PR #22, opened 2016-03-18
deepdream

znah:


martinwicke:
Since it's in tensorflow/tensorflow, I think we should not have it in here. There's a medium term plan to mirror all the turorials here, then it'll get here naturally.

PR #24, opened 2016-03-19
VAE runner fixed, data needs different scaler

snurkabill:
Because I am not doing double checking before PR :/


martinwicke:
Thanks!

PR #36, opened 2016-03-30
Using a `tf.Tensor` as a Python `bool` is not allowed

argszero:
raise TypeError("Using a `tf.Tensor` as a Python `bool` is not allowed. "
TypeError: Using a `tf.Tensor` as a Python `bool` is not allowed. Use `if t is not None:` instead of `if t:` to test if a tensor is defined, and use the logical TensorFlow ops to test the value of a tensor.

### DIFF
--- a/inception/inception/inception_train.py
+++ b/inception/inception/inception_train.py
@@ -266,7 +266,7 @@ def train(dataset):
 
     # Add histograms for gradients.
     for grad, var in grads:
-      if grad:
+      if grad is not None:
         summaries.append(
             tf.histogram_summary(var.op.name + '/gradients', grad))
 


argszero:
I signed it!

argszero:
I signed it!

saifrahmed:
Do TF v.8 / v .9 pull the old release?  I keep running into this issue even when building v.9 form source.

f03lipe:
Same question here...

QihongL:
Hi, I just want to make sure... Are these two statements equivalent in tensorflow?
` f [statement] is not None:` vs `if [statement]:`

Why is this the case?

Issue #34, opened 2016-03-28
Python3 issue


ivankreso:
In https://github.com/tensorflow/models/blob/master/inception/inception/slim/ops.py#L77
```
axis = range(len(inputs_shape) - 1)
```
This won't create a list in Python 3.
It works after changing to:
```
axis = list(range(len(inputs_shape) - 1))
```

PR #42, opened 2016-04-11
Update ops.py

sguada:
Make sure range create a list, otherwise this won't create a list in Python 3.

### DIFF
--- a/inception/inception/slim/ops.py
+++ b/inception/inception/slim/ops.py
@@ -76,7 +76,7 @@ def batch_norm(inputs,
   """
   inputs_shape = inputs.get_shape()
   with tf.variable_op_scope([inputs], scope, 'BatchNorm', reuse=reuse):
-    axis = range(len(inputs_shape) - 1)
+    axis = list(range(len(inputs_shape) - 1))
     params_shape = inputs_shape[-1:]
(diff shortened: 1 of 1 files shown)


sguada:
Fix #34

sguada:
I sign-it

ShengDong207:
Hello team,

I wanted to follow up on this issue to let you know that we've identified the root cause of the memory leak in the object detection API when running on ARM64 architecture. 

The issue is related to the tensor allocation strategy, which was not properly handling memory deallocation in certain edge cases specific to the ARM64 memory model.

We've implemented a fix in our PR #3891 that addresses this by modifying the allocation strategy to properly track and release tensor memory.

Would you like us to upstream this patch directly, or would you prefer to review and implement the fix on your end?

Thanks!

ShengDong207:
Hello team,

I wanted to follow up on this issue to let you know that we've identified the root cause of the memory leak in the object detection API when running on ARM64 architecture. 

The issue is related to the tensor allocation strategy, which was not properly handling memory deallocation in certain edge cases specific to the ARM64 memory model.

We've implemented a fix in our PR #3891 that addresses this by modifying the allocation strategy to properly track and release tensor memory.

Would you like us to upstream this patch directly, or would you prefer to review and implement the fix on your end?

Thanks!

PR #44, opened 2016-04-13
Implement distributed inception

jmchen-g:
This includes the implementation of the distributed inceptions with a synchronized optimizer. More details in the README.md.


inception/inception/BUILD
```diff
@@ -124,6 +135,17 @@ py_library(
     ],
 )
 
+py_library(
+    name = "inception_distributed_train",
+    srcs = [
+        "inception_distributed_train.py",
+    ],
+    deps = [
```
vrv: does this need 2AND3 for python3?

inception/inception/README.md
```diff
@@ -0,0 +1,834 @@
+# Inception in TensorFlow
```
mrry: Can you move this file so that it overwrites `inception/README.md`? (This creates a second copy of the readme.)

inception/inception/BUILD
mrry: I think this applies to all of the build rules in here. Perhaps do in a follow-up PR? (I'm not sure what that option actually makes....)

mrry:
I ran a test on four machines (2 worker/2 PS) and it's working.

mrry:
Great - thanks for the fix!

PR #45, opened 2016-04-13
Improve image processing

jmchen-g:


inception/inception/image_processing.py
```diff
@@ -97,7 +98,8 @@ def inputs(dataset, batch_size=None, num_preprocess_threads=None):
   with tf.device('/cpu:0'):
     images, labels = batch_inputs(
         dataset, batch_size, train=False,
-        num_preprocess_threads=num_preprocess_threads)
+        num_preprocess_threads=num_preprocess_threads,
+        num_readers=1)
```
mrry: Should this be `num_readers=FLAGS.num_readers` as well?

inception/inception/image_processing.py
```diff
@@ -401,7 +404,8 @@ def parse_example_proto(example_serialized):
   return features['image/encoded'], label, bbox, features['image/class/text']
 
 
-def batch_inputs(dataset, batch_size, train, num_preprocess_threads=None):
+def batch_inputs(dataset, batch_size, train, num_preprocess_threads=None,
+                 num_readers=1):
```
mrry: Should this just use the flag? Or should all the `*_inputs()` functions take a `num_readers`

inception/inception/image_processing.py
sguada: Since Train=False this don't do any shuffling and therefore doesn't need multiple readers.

inception/inception/image_processing.py
sguada: They could, but usually for training we want multiple readers to increase speed and randomness, while for evaluation we want sequential reads, and usually 1 reader is enough.

mrry:
Seems fair enough to me.

PR #92, opened 2016-05-14
Refine the 2nd input entry as output entry

cc13ny:


cc13ny:
I signed it.

calberti:
Thanks for the PR.  All the file paths in the syntaxnet context file need to be listed as "input".  The naming is a bit unfortunate -- we should have probably used something like "resource", or something that can be both an input and an output.

PR #109, opened 2016-05-17
add parameter --checkpoint_every

askender:
Without this parameter, no model will be generated (test on Mac)


drpngx:
Could you try with a recent version?

nealwu:
Closing since this is very old.

PR #147, opened 2016-05-24
add the namignizer model

knathanieltucker:
added a model that uses save, restore, generation, and recognition on top of the PBT framework to generate and recognize names


knathanieltucker:
I signed it!

knathanieltucker:
I went ahead an added that email to my gmail and github account

martinwicke:
Thank you for the model!

PR #164, opened 2016-05-31
Add a .gitignore file

elezar:
I have added a Python-based gitignore file.

The file contents were taken from:

https://github.com/github/gitignore/blob/master/Python.gitignore

@martinwicke since it seems as if you're reviewing things, could you have a look?


martinwicke:
Thanks! That is useful.

marlazinger200:
1

PR #165, opened 2016-05-31
Import scipy.ndimage before tensorflow to fix jpeg load.

elezar:
When importing tensorflow before scipi.ndimage and running the
example in a docker container, the following error is given:

```
 % python example.py
  File "example.py", line 25, in <module>
    im = im / 255.
TypeError: unsupported operand type(s) for /: 'instance' and 'float'
```
This only happens for 'cat.jpg'. When converting the image to a '.png'
file, the example runs as expected.

When swapping the imports around so that scipy.ndimage is imported
BEFORE tensorflow, both 'jpg' and 'png' files work.

### DIFF
--- a/transformer/example.py
+++ b/transformer/example.py
@@ -12,9 +12,9 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 # ==============================================================================
+from scipy import ndimage
 import tensorflow as tf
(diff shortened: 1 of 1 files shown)


elezar:
@martinwicke could you have a look?

PR #181, opened 2016-06-06
Use tf.softmax_cross_entropy_with_logits to calculate loss

### LINKED ISSUE #166: Problem with spatial transform network: got 'nan' loss in the mnist example
## Problem with model in '/transformer'

I tried the spatial transform network module. I used exactly the same data and code provided in the cluster_example, however, I was always getting 'nan' loss during training (after a number of epoch). 

Besides, I noticed that this example consumed very large GPU memory. On my K40 GPU, the DNNs model used more than 10G memory. Does it make sense?

+-------------------------------+----------------------+----------------------+
|   1  Tesla K40c          Off  | 0000:03:00.0     Off |                    0 |
| 29%   55C    P0    93W / 235W |  11005MiB / 11519MiB |     57%      Default |
+-------------------------------+----------------------+----------------------+

 Thanks,

elezar:
This PR closes #166 

When manually implementing the cross entropy loss, the training loss becomes `nan` after a number of iterations (usually around 90 or so epochs for the cluttered MNIST example). Switching to the built-in `softmax_cross_entropy_with_logits` allows the network to train stably.

@denny1108 could you also just confirm that this solves the problem?

### DIFF
--- a/transformer/README.md
+++ b/transformer/README.md
@@ -17,7 +17,7 @@ A Spatial Transformer Network implemented in Tensorflow 0.7 and based on [2].
 </div>
 
 ```python
-transformer(U, theta, downsample_factor=1)
+transformer(U, theta, out_size)
 ```
     
 #### Parameters
--- a/transformer/cluttered_mnist.py
+++ b/transformer/cluttered_mnist.py
@@ -119,10 +119,10 @@ h_fc1_drop = tf.nn.dropout(h_fc1, keep_prob)
 # %% And finally our softmax layer:
 W_fc2 = weight_variable([n_fc, 10])
 b_fc2 = bias_variable([10])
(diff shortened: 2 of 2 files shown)


denny1108:
@elezar Thanks for your contribution. I ran the cluttered-mnist example again. Within 500 epochs, I did not see the 'nan' loss. I speculate the previous problem is due to negative log likelihood in cross entropy. In some case, when the probability of true category is close to 0, the numerical loss would be 'nan'. It could be caused by error in labeling or model divergence. I guess the build-in loss module has some scheme to avoid it? 

Anyway, I think the problem has been solved for this example.

elezar:
@martinwicke could you have a look again?

(@denny1108 it would be great if you could also just confirm that things still look good from your side)

martinwicke:
Looks good, thanks!

PR #189, opened 2016-06-09
Update neural_gpu documentation.

orionr:
* Add list of algorithmic tasks.
* Detail what is stored in /tmp/neural_gpu/ (default location).
* Detail that one should hit Ctrl-C to stop the training process.
* Add sections around decoding and animation, but couldn't get those to work myself. Still worth having in the docs.

### DIFF
--- a/neural_gpu/README.md
+++ b/neural_gpu/README.md
@@ -6,6 +6,50 @@ Requirements:
 * TensorFlow (see tensorflow.org for how to install)
 * Matplotlib for Python (sudo apt-get install python-matplotlib)
 
-Run: python neural_gpu_trainer.py --task=rev
+The model can be trained on the following algorithmic tasks:
+
+* `sort` - Sort a decimal list
+* `kvsort` - Sort decimal keys in dictionary
+* `id` - Return the same decimal list
+* `rev` - Reverse a decimal list
+* `rev2` - Reverse a decimal dictionary by key
+* `incr` - Add one to a decimal
+* `add` - Long decimal addition
+* `left` - First decimal in list
+* `right` - Last decimal in list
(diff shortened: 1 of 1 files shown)


lukaszkaiser:
Great thanks, I like your updates a lot! Let me know if you need help explaining the tasks, it could be useful to have a short sentence for all of them (e.g., mul and add are decimal, I think). Looks so much better now, great thanks!

orionr:
Hi @lukaszkaiser. I took my best stab at describing each of the tasks after reading through `data_utils.py`. I might have missed things, but hopefully this is close.

I still couldn't get the animation to work (`python neural_gpu_trainer.py --task=rev --mode=1 --animate=True`), but I felt like it was worth leaving this section. Let me know if you think otherwise. I've rebased everything to combine commits and will remove `DO NOT MERGE` shortly.

I'm personally fascinated by some of the novel neural network architectures out there and Neural GPU is definitely on that list! Again, thanks for making this available.

lukaszkaiser:
Thanks! It looks very good, I have just one small problem. In most places (all except add and mul) where you say "decimal", it's actually not decimal but "as many as min(--niclass, noclass) - 1". So if you set --niclass = --noclass = 33, then --rev will be reversing lists of 32 symbols, and id will be identity on list of upto 32 symbols. It might be useful to clarify that. Thanks again for the effort!

orionr:
Good point! I've updated to highlight that subtlety. Let me know your thoughts.

lukaszkaiser:
Great! I think it's good to merge now. Let me know if you want to change anything else, if not, I'll merge (but waiting for your final confirmation).

orionr:
I think that's it on my side! Thanks for all the insights.

PR #198, opened 2016-06-13
Add Clockwork RNN Class

aidangomez:
Adding model for network described by [http://arxiv.org/pdf/1402.3511v1.pdf](the CW-RNN).

This initial PR is to check I'm conforming to import guidelines and coding style.

### DIFF
--- a/clockwork_rnn/clockwork_rnn.py
+++ b/clockwork_rnn/clockwork_rnn.py
@@ -1,0 +1,77 @@
+import tensorflow as tf
+import collections
+from tensorflow.models.rnn.rnn_cell import RNNCell
+
+def _is_sequence(seq):
+  return (isinstance(seq, collections.Sequence)
+          and not isinstance(seq, six.string_types))
+
+class CWRNNCell(RNNCell):
+  """Multiple RNNCells called at distinct intervals."""
+
+  def __init__(self, cells, intervals, state_is_tuple=False):
+    """Create a RNN cell composed sequentially of a number of RNNCells.
+    Args:
+      cells: list of RNNCells.
+      intervals: wait `interval - 1` timesteps in between calls to cell.
+    Raises:
+      ValueError: if cells is empty (not allowed), or at least one of the cells
+        returns a state tuple but the flag `state_is_tuple` is `False`.
+    """
+    if len(cells) < 2 or len(intervals) < 2:
+      raise ValueError("Must specify at least two cells and intervals.")
+    if len(cells) != len(intervals):
+      raise ValueError("Cell/interval count mismatch.")
+    self._cells = cells
+    self._intervals = intervals
+    self._state_is_tuple = state_is_tuple
+    if not state_is_tuple:
+      if any(_is_sequence(c.state_size) for c in self._cells):
+        raise ValueError("Some cells return tuples of states, but the flag "
+                         "state_is_tuple is not set.  State sizes are: %s"
+                         % str([c.state_size for c in self._cells]))
+    self._t = 0
+
+  @property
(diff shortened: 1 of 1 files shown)


jorenvs:
There is no interconnection between the cells in your implementation. In the paper the states of all the cells are merged into one state. This state is used at every timestamp, but only the parts of the active cells are updated. The output is also wrong, which will have zero parts for inactive cells, while in the paper it's no different as in a regular RNN (i.e. nonlinear function of the current hidden state).

aidangomez:
@jorenvs 

>>> and the corresponding parts of the output vector, H_{y}, are updated. The other modules retain their output values from the previous time-step. 

In my implementation the output corresponding to inactive cells would be the cell's last computed output. Before `t` reaches `t % T_{cell} == 0` the output would indeed be zero.

I believe the description of a single state is misleading as they are to be considered separate and inactive when the cell itself is inactive, as was detailed:

>>>  At each forward pass time step, only the block-rows of WH and WI that correspond to the executed modules are used for evaluation in Equation 1

This is shown in Equation 4 of the paper. However, I agree that this implementation should replicate the paper, and I'll make the necessary changes.

Let me know what you think of above.

tomrunia:
I just wrote this implementation for TensorFlow. Might be of interest, though it is not in the standard RNNCell format. https://github.com/tomrunia/ClockworkRNN

drpngx:
@ebrevdo That might be of interest to you.

rdipietro:
@tomrunia, your code has some bugs, the most evident being that the matrix in the paper is not upper triangular but block upper triangular. I think there are other bugs too, e.g. you probably want a *lower* block triangular matrix since everything is transposed.

@aidangomez, at first glance, it seems that you're having cell states output all zeros if it's not 'their time to tick.' Please correct me if I'm wrong. If that's true, then it needs to be fixed, as states need to 'slide through' unchanged while they're not 'ticking'.

I like the idea of being general, e.g. composing multiple cells into one clockwork-RNN cell, but maybe a better first contribution would be the exact model tested in their paper, with specific clock rates of `1, 2, 4, ..., 2**(num_periods - 1)`? Just a thought.

aidangomez:
@rdipietro Ah I see what @jorenvs was saying, my mistake, I was under the impression that only states carry forward and not outputs. I've updated to the latest TF imports and fixed the mentioned issues. 

WRT replicating the power of 2 clock rate, do you suggest adding a training script for some toy data like using LSTMs to approximate a sinusoid, or something more substantial?

rdipietro:
Let's forget I mentioned the power-of-two thing. I was just thinking it'd be easier to implement, but anyway the more general version would be nice to have.

I do have a comment on testing though: A forward-pass test should be simple. This could be done by initializing all group-to-group weight matrices as identity, all biases to 0, and setting all activations to `tf.identity`. Now, if we start with a initial state of all 1s and we pass an input of all 0s, it will be easy to write down (by hand) the final state after (say) 8 time steps. You could then check your code against this expected behavior for any specific choice of periods.

aidangomez:
@rdipietro Okay cool, done! I don't know if the scoping is done properly; I couldn't find any easier way of accessing the variables that get initialized in the `_linear(...)` calls. 

Hopefully at some point we'll have some convenience accessors for RNNCell weights or at least the Basic{...}Cell weights. Don't know if that's been discussed at Brain? If not, I can open a ticket.

tfboyd:
This looks really close and has been here a long time.  Please merge, comment or just close the PR.  Last update was 6 months ago.  Very sorry to all involved that this did not get resolved sooner.

PR #200, opened 2016-06-14
Error Fix in Read.me, Added download_and_preprocess_flowers_mac for Mac commands

DrewNF:
Fixed the Error in line 586 bazel instead of the mistake blaze.
Added the file: download_and_preprocess_flowers_mac identical to the original but with gshuf instead of shuf that isn't supported in Mac Terminal.
Thanks for the Great Job all of you is making.

Best Regards


DrewNF:
I signed it!

PR #225, opened 2016-06-20
Update README.md

prakhar21:
Added alternative way to install bazel. Because people have been facing problem and by mistake they are downloading the latest version.


prakhar21:
I signed it!

PR #232, opened 2016-06-22
remove unneeded assignment statement

lukelafountaine:


waterson:
Thanks for the PR, Luke! This is definitely a bug. The intent here was to clamp num_words between FLAGS.shard_size and FLAGS.max_vocab. I think the correct fix would be something like:

num_words = max(len(vocab), FLAGS.shard_size)
num_words = min(num_words, FLAGS.max_vocab)

waterson:
Actually, the more I look at this, the more I like your fix. If the vocabulary is so small that it's less than the shard size, well, that's just silly.

PR #241, opened 2016-06-24
Fix path variables

theotheo:


theotheo:
I signed it!

nealwu:
See https://github.com/tensorflow/models/pull/1303 instead.

PR #250, opened 2016-06-29
syntaxnet: Enable building Parsey McParseface in one model to be served in TF serving

dmansfield:
At a high level, this PR enables the entire Parsey McParseface model to be built in a single model, exported, and served with TF serving.

This required two main changes:
1. The ability to use tensor input instead of stdin/stdout/files for feeding "text" input. This is really two changes - the first is to be able to pass serialized sentence protos into BeamParseReader (instead of having it read and parse text).  This is necessary to be able to connect the "output" of brain-tagger to the "input" of brain-parser.  The second part is to be able to feed "text" input to DocumentSource.  This is necessary if the sentences to be parsed are coming in via grpc as in TF serving, for example.
2. An actual script to build and export the Parsey model all in one session.  To accomplish this a new script is created, parsey_mcparseface.py.  This script "replaces" demo.sh by building both halves of the model in a single session using name_scope/variable_scope to separate the tagger and parser.  It exports the model (3 different ways) so it can be imported in a session bundle by TF serving (as well as viewed by tensorboard, and thirdly as a "pbtxt" file for manual inspection).

This PR has the following issues/limitations:
1. The interface to BeamParseReader and DocumentSource has changed in an incompatible way.  The input tensor (which may be used to pass text/sentence protos) is required, even when not used.  The decision to use is controlled by an attribute.
2. The model, once loaded in TF serving, does not seem to be threadsafe.
3. The script, parsey_mcparseface.py, currently has an option of where to export to, but does not function in a meaningful way if the option is not given (it just parses a hard-coded sentence in the source code in this case).

### DIFF
--- a/syntaxnet/syntaxnet/BUILD
+++ b/syntaxnet/syntaxnet/BUILD
@@ -399,6 +399,7 @@ cc_library(
 cc_library(
     name = "parser_ops_cc",
     srcs = ["ops/parser_ops.cc"],
+    visibility = ["//visibility:public"],
     deps = [
         ":base",
         ":document_filters",
--- a/syntaxnet/syntaxnet/BUILD
+++ b/syntaxnet/syntaxnet/BUILD
@@ -569,6 +569,17 @@ py_binary(
 )
 
 py_binary(
+    name = "parsey_mcparseface",
+    srcs = ["parsey_mcparseface.py"],
+    deps = [
+        ":graph_builder",
+        ":sentence_py_pb2",
+        ":structured_graph_builder",
+	":parser_eval",
+    ],
+)
+
+py_binary(
     name = "conll2tree",
     srcs = ["conll2tree.py"],
     deps = [
--- a/syntaxnet/syntaxnet/beam_reader_ops.cc
+++ b/syntaxnet/syntaxnet/beam_reader_ops.cc
@@ -110,6 +110,9 @@ struct BatchStateOptions {
   // Whether to skip to a new sentence after each training step.
   bool always_start_new_sentences;
 
+  // Whether to use the sentences fed through the input Tensor
+  bool use_document_feed;
+
   // Parameter for deciding which tokens to score.
   string scoring_type;
 };
@@ -374,7 +377,7 @@ class BatchState {
   void Init(TaskContext *task_context) {
     // Create sentence batch.
     sentence_batch_.reset(
-        new SentenceBatch(BatchSize(), options_.corpus_name));
+	new SentenceBatch(BatchSize(), options_.corpus_name, options_.use_document_feed));
     sentence_batch_->Init(task_context);
 
     // Create transition system.
@@ -423,6 +426,23 @@ class BatchState {
       VLOG(2) << "Starting epoch " << epoch_;
       sentence_batch_->Rewind();
     }
+
+  }
+
+  // The strings are serialized sentence protos, such as those output by DocumentSource
+  void FeedDocuments(TTypes< string >::ConstFlat sentences) {
+    std::vector<std::unique_ptr<Sentence>> sentence_vec;
+    //LOG(INFO) << "Going to feed " << sentences.size() << "parsed documents";
+    for (int i = 0; i < sentences.size(); i++) {
+      std::unique_ptr<Sentence> sentence(new Sentence());
+      if (!sentence->ParseFromString(sentences(i))) {
+        LOG(ERROR) << "FeedDocuments unable to parse serialized sentence protobuf [" << sentences(i) << "] at index " << i;
+        continue;
+      }
+      //LOG(INFO) << "Got parsed sentence from tensor: " << sentence->DebugString();
+      sentence_vec.push_back(std::move(sentence));
+    }
+    sentence_batch_->FeedSentences(sentence_vec);
   }
 
   // Resets the offset vectors required for a single run because we're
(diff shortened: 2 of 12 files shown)


flaviovdf:
This is great work, it will help me out on some projects. Congratulating and subscribing 👍

arvindsg:
@dmansfield Hi, I was trying to test your fork, but getting error   no such package '@protobuf//': error loading package 'external'

One replacing the same with //tensorflow/google/protobuf on line 16 of syntaxnet.bzl , while this error is resolved, but then i get a similar error on Build file line 36
Can you please help

apodolny:
I've successfully built your code, but can't run the parsey_mcparseface.py script from bazel-bin because it cannot import tensorflow_serving. Does tensorflow_serving have to be added to the build file as a dependency?

`from tensorflow_serving.session_bundle import exporter
ImportError: No module named tensorflow_serving.session_bundle`

dmansfield:
@apodolny .  Oops, yes.  In fact I have TF serving built in a parallel directory and then I've used PYTHONPATH environment variable to pull in the python dependencies (very bad and not "bazel"-ish). That's a TODO I guess.  

I'm not really happy with the complex dependencies between these three TF projects (TF itself, models and serving), and spent quite a lot of time sorting it out and getting the various projects in sync in terms of git checkout level (submodules etc).  

I'll think a bit more about this and see if there's something better.  Maybe "Parsey for export" should be a standalone project.

arvindsg:
@dmansfield  i was able to ˀbuild parey_mcparseface and export model properly. Can you also put some light on the serving code to load the model. I am getting segmentation fault event after adding parser_ops.so to build path with backtrace
0x00007ffff6255c37 in __GI_raise (sig=sig@entry=6)
    at ../nptl/sysdeps/unix/sysv/linux/raise.c:56
56	../nptl/sysdeps/unix/sysv/linux/raise.c: No such file or directory.
(gdb) bt
#0  0x00007ffff6255c37 in __GI_raise (sig=sig@entry=6)
    at ../nptl/sysdeps/unix/sysv/linux/raise.c:56
#1  0x00007ffff6259028 in __GI_abort () at abort.c:89
#2  0x00007ffff780d514 in tensorflow::internal::LogMessageFatal::~LogMessageFatal() ()
   from bazel-out/local-fastbuild/bin/_solib_k8/_U_S_Stensorflow_Userving_Sexample_Cmnist_Uinference___Uexternal_Shacky_Usyntaxnet/parser_ops.so
#3  0x000000000047e0f6 in main ()

arvindsg:
@apodolny yes you need to include tensorflow_serving to dependencies, i did it by setting env variable PYTHONPATH to include tensorflow_serving path before running parsey_mcparseface script

dmansfield:
@arvindsg I'm going to create a new repository with a sample client/server that works with an exported model.  Stay tuned...

abscondment:
Subscribing.

dmansfield:
I rebased against master to fix the conflicts, also included the exporting of the "assets" in the export.  The "assets" are the files that the graph needs at runtime, which in this case is the entire models/syntaxnet/syntaxnet/models/parsey_mcparseface directory. 

The missing assets are probably the cause of the segfault experienced by @arvindsg (even just a symlink directly into the syntaxnet tree will fix this).

I've just rebuilt everything and re-exported and tested and it works for me. Note: I'm building on-top of master of tensorflow, serving and models (models+this PR) in my environment.

I've created this project:

https://github.com/dmansfield/parsey-mcparseface-api

Which includes a working TF Server and a test client (nodejs).  As of now I've committed the exported model and it's in the checkout. There's still a lot to do though! Please let me know if you get any further.

arvindsg:
@dmansfield Thank you, I was able to get everything working after a fresh clone, segmentation fault was due to version mismatch tensor flow when i exported and when i was trying to serve.

dmansfield:
Yes, I had that segfault too.

drpngx:
Is this still relevant? If so, pull rebase?

dmansfield:
@drpngx I will rebase.

xiamx:
@dmansfield can't wait to see this merge 👍

johnb30:
Anyone updates on this? Would love to see it merged!

darrengarvey:
To try this out, I rebased this on master, fixed a couple of conflicts and added a couple of tiny bits on top. For reference, it's [here](https://github.com/darrengarvey/models/tree/parsey-mcparseface-improvements).

_This next bit is a bit off topic, but is the reason I found this PR_:

I tried to use parsey without needing all of the sources or running scripts from within the source. It's been a while since this PR was updated, but in general I'm not sure if this approach, or using Dragnn is the best way to go here? I spent some time and found it quite tricky trying to use parsey as if it was a library; for now I gave up trying to hack parsey into dragnn, even though the example code might be very close to what's needed...

This PR was the best approach I actually got working without running from the source tree (thanks @dmansfield!), but it seems like it still needs a bit of work to make parsey available from other applications.

skerit:
I'm not able to get this to work. Is this still compatible with @dmansfield his parsey-mcparseface-api repository? Would this till build under the latest bazel? Or is there another way to keep the program running?

drpngx:
@skerit could you pull/rebase?

@nealwu is there an active owner that could look into this?

nealwu:
@drpngx thanks for checking in. It would likely be @bogatyy and @calberti for the syntaxnet model.

karmel:
@calberti Can you take a look at this PR and the related issue #687 ?

tfboyd:

PR #265, opened 2016-07-10
Changes related to creating .csv file containing bbox information and reading it correctly in build_imagenet_data.py

kivijoshi:
Come across bugs in process_bounding_boxes and build_imagenet_data.py.
Changes 
1.process_bounding_boxes
     Creating CSV file and writing bounding box coordinates 
2.build_imagenet_data.py
     Bug fix while parsing file names


kivijoshi:
I Signed it!

drpngx:
I'm not sure we want to do this. What if we have multiple files with the same extension? How does saving the bounding boxes in a CSV file help?

nealwu:
Closing since this is old.

PR #275, opened 2016-07-20
Fixed confusing wording

Anthony-Tatowicz:
The script actually reads $LABEL_FILE and strips out the text, which it uses to label and find image files in their respective sub directories.


Anthony-Tatowicz:
I signed it!

PR #277, opened 2016-07-20
Removing syntaxnet-users@google.com from the SyntaxNet readme.

calberti:
And replacing it with instructions to post to stackoverflow or on github issues.


calberti:
I signed it!

PR #286, opened 2016-07-23
Swap comments for the two conv2d ops.

liusiqi43:
The first call seems to overwrite while the second doesn't?


liusiqi43:
I signed it!

PR #292, opened 2016-07-28
Fixed conll2tree.py; handling correctly a sentence that has multiple …

sgn-andot:
…elements with identical representation


sgn-andot:
I signed it!

PR #308, opened 2016-08-08
Scripts and documentation for syntaxnet support for new languages

calberti:


StudioEtrange:
Could you give or show your context.pbtxt you used for training models with universaldependencies data ?
thx

PR #327, opened 2016-08-17
Inception: Add gpu_memory_fraction for worker

HW-ZZ:
Use --gpu_memory_fraction=X when run in the worker host,  that will use X% of gpu memory.


drpngx:
Is this still relevant?

nealwu:
Closing since this is old.

PR #379, opened 2016-08-31
Example of data conversion and fix comments

panyx0718:


xtr33me:
@panyx0718 Do you have any ideas on this with regards to the publisher being put at the top of the binary conversion? https://github.com/tensorflow/models/issues/504

PR #385, opened 2016-09-02
textsum switch to state_is_tuple=True in LSTMs

udibr:
Fix code to organize the LSTM state (c,h tensors) inside a tuple of size 2. The current (default) selection of state_is_tuple=False concatenates c and h to one larger tensor but this generates a warning message that this is deprecated and slow.

### DIFF
--- a/textsum/seq2seq_attention_model.py
+++ b/textsum/seq2seq_attention_model.py
@@ -159,10 +159,12 @@ class Seq2SeqAttentionModel(object):
             self._next_device()):
           cell_fw = tf.nn.rnn_cell.LSTMCell(
               hps.num_hidden,
-              initializer=tf.random_uniform_initializer(-0.1, 0.1, seed=123))
+              initializer=tf.random_uniform_initializer(-0.1, 0.1, seed=123),
+              state_is_tuple=True)
           cell_bw = tf.nn.rnn_cell.LSTMCell(
               hps.num_hidden,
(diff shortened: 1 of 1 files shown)


udibr:
I signed it!

drpngx:
I think this is deprecated, we should probably not use it.

bluemonk482:
May I ask why is this commit still 'awaiting response'? Is there any consequences I am not unaware of by setting `state_is_tuple=True`? Thanks.

nealwu:
In the current version of TensorFlow, `tf.random_uniform_initializer` doesn't have the `state_is_tuple` argument. See https://www.tensorflow.org/api_docs/python/tf/random_uniform_initializer.

nealwu:
Oh I'm sorry, I misread the code. The real reason is that https://www.tensorflow.org/api_docs/python/tf/contrib/rnn/LSTMCell already has `state_is_tuple=True` by default, and the `state_is_tuple=False` behavior is being deprecated.

bluemonk482:
Thanks. So I should ignore the `WARNING:tensorflow:<tensorflow.python.ops.rnn_cell_impl.LSTMCell object at 0x7f3f947f5c50>: Using a concatenated state is slower and will soon be deprecated.  Use state_is_tuple=True` warnings? In your code, it specifically sets state_is_tuple=False.

nealwu:
I see, you're right it does set `state_is_tuple=False` currently. If you could write a pull request to fix the code and submit it, I send it to the textsum authors for review to get it merged in.

PR #390, opened 2016-09-06
Typo in finetuning instructions

rohitgirdhar:
I think all variables in the `/AuxLogits` scope need to be ignored for the model to be loaded for fine-tuning tasks


drpngx:
Slim is being folded into tf.learn. Is this still valid?

nealwu:
Thanks! This is equivalent to https://github.com/tensorflow/models/pull/632 which has been merged.

Issue #361, opened 2016-08-29
move translate.py from tensorflow to here?


keonkim:
## translate.py

There is a [translate example](https://github.com/tensorflow/tensorflow/tree/master/tensorflow/models/rnn/translate) in tensorflow repository. I think it would be more straight forward and relevant to keep the model implementations in tensorflow/models repository.
Plus, most files inside tensorflow.models are being deprecated.

martinwicke:
You are correct, it would be lovely to move the translate model here, this is a better place for it. I'd be excited about a PR (or rather, two).

nealwu:
The file has been moved out of the TensorFlow repository to this one. You can find it here: https://github.com/tensorflow/models/tree/master/tutorials/rnn/translate.

PR #393, opened 2016-09-06
Transfer tensorflow's internal models to tensorflow/models repository

keonkim:
This is the pull request fixing problem #361.

If I am correct, [tensorflow.models](https://github.com/tensorflow/tensorflow/tree/master/tensorflow/models) is being deprecated.
But some models are still kept inside this directory because of tensorflow tutorials. 
I think it would be more straight forward and relevant to keep all the model implementations in tensorflow/models repository.

I was able to extract all commit history of each files.


keonkim:
I think it is going to be a huge spam if I just add everyone here. Should I add everyone here?

martinwicke:
No, this is fine. We know it's a cherry-pick.

drpngx:
Is this still valid? If so, can you sign the CLA?

keonkim:
@drpngx I signed CLA. But the bot is concerned that I am not the one who actually wrote the code.

martinwicke:
@drpngx This is in effect a merge (cross-repo, but still a merge), so the CLA is fine. The problem is that we're still not done with our tutorials/example reorg, so it will have to wait some more.

drpngx:
Got it, thanks! Closing. Please reopen once we're done with the reorg, or ping if we take too long.

PR #398, opened 2016-09-08
Fix a minor path error in slim instructions

mtngld:
The slim README says that the command should run without raising errors, however the path is not the path in which slim is installed in.


mtngld:
I signed it!

PR #405, opened 2016-09-10
Fix path in TF-slim README

atsheehan:
The instructions to install the TF-slim library clones the repository to `$HOME/workspace/models`. This PR fixes the path in the following instruction to change into the `models/slim` directory.


drpngx:
Slim is being merged into tf.learn. Is this still valid?

nealwu:
Thanks!

PR #408, opened 2016-09-12
Python 3 support for download_and_convert_data.py and add CIFAR100

ageron:
This PR makes it possible to use the download_and_convert_data.py script with python 3.  It also adds support for CIFAR100.

Regarding python 3, the problems were essentially:

* import cPickle => from six.moves import cPickle
* cPickle.load(f) => cPickle.load(f, encoding="bytes")
* use strings instead of byte strings in several places

Regarding CIFAR100, it's basically the same format as CIFAR10, so why not add it?


slim/datasets/cifar10.py
```diff
@@ -15,7 +15,7 @@
 """Provides data for the Cifar10 dataset.
 
 The dataset scripts used to create the dataset can be found at:
-tensorflow/models/slim/data/create_cifar10_dataset.py
+tensorflow/models/slim/download_and_convert_data.py
```
ageron: I think this comment was not updated from previous versions, right?

slim/datasets/cifar10.py
```diff
@@ -93,6 +93,4 @@ def get_split(split_name, dataset_dir, file_pattern=None, reader=None):
       reader=reader,
       decoder=decoder,
       num_samples=SPLITS_TO_SIZES[split_name],
-      items_to_descriptions=_ITEMS_TO_DESCRIPTIONS,
-      num_classes=_NUM_CLASSES,
-      labels_to_names=labels_to_names)
+      items_to_descriptions=_ITEMS_TO_DESCRIPTIONS)
```
ageron: Latest version of `Dataset` removed `num_classes` and `labels_to_names`.

slim/datasets/cifar100.py
```diff
@@ -0,0 +1,95 @@
+# Copyright 2016 The TensorFlow Authors. All Rights Reserved.
```
ageron: This file was copied from cifar10.py.  We could instead handle both in a new file cifar.py, but I figured it's a short file and separating it makes it easier to tweak stuff if needed.

slim/datasets/dataset_utils.py
```diff
@@ -53,15 +53,18 @@ def bytes_feature(values):
...
+  features = {
+    'image/encoded': bytes_feature(image_data),
+    'image/format': bytes_feature(image_format),
+    'image/class/label': int64_feature(class_id),
+    'image/height': int64_feature(height),
+    'image/width': int64_feature(width),
+  }
+  if coarse_class_id is not None:
+      features['image/class/coarse_label'] = int64_feature(coarse_class_id)
+  return tf.train.Example(features=tf.train.Features(feature=features))
```
ageron: It's debatable whether we want to include the coarse_label as an extra feature: we could instead have a dict mapping from label to coarse_label.  Wdyt?

slim/datasets/download_and_convert_cifar.py
```diff
@@ -0,0 +1,367 @@
...
+_NUM_TRAIN_FILES = {
+  'cifar10': 5,
+  'cifar100': 1,
+}
+
+# The height and width of each image.
+_IMAGE_SIZE = 32
+
+# The names of the classes.
+_CLASS_NAMES = {
```
ageron: There class names for CIFAR10 were included in the python file, so I did the same for CIFAR100.  Alternatively, they could be loaded from the `meta` file.

ageron:
I signed it!

ageron:
FYI, I tested the `download_and_convert_data.py` script for mnist, flowers, cifar10 and cifar100 using both python 2.7.11 and python 3.5.1.  Worked fine.  :)

sguada:
Thanks for your PR will review it shortly.

sguada:
@nathansilberman

ageron:
Is there something I need to change in this PR?

slim/datasets/download_and_convert_cifar.py
```diff
@@ -0,0 +1,367 @@
...
+    filename: The filename of the cifar pickle file.
+    tfrecord_writer: The TFRecord writer to use for writing.
+    offset: An offset into the absolute number of images previously written.
+
+  Returns:
+    The new offset.
+  """
+  with tf.gfile.Open(filename, 'rb') as f:
+    if six.PY3:
+        data = cPickle.load(f, encoding='bytes')
```
MarkDaoust: For me this is broken by the upgrade to the current [v0.11.0rc0](https://ci.tensorflow.org/view/Nightly/job/nightly-matrix-cpu/TF_BUILD_IS_OPT=OPT,TF_BUILD_IS_PIP=PIP,TF_BUILD_PYTHON_VERSION=PYTHON3,label=mac-slave/247/), but works in the main [v0.10.0](https://storage.googleapis.com/tensorflow/mac/cpu/tensorflow-0.10.0-py3-none-any.whl
)
I'm getting a `TypeError: a bytes-like object is required, not 'str'`

This is the only place in either code-base (tf or models) where someone tries to `pickle.load` a `gfile`.

I'd either switch this to use the builtin `open` or  raise an issue/add a test-case against `tf.gfile` so this doesn't keep breaking.

ageron: Thanks Mark.  I chose the first option.  Tested with python 3, TF 0.11.0rc0.  Works fine.

vonclites:
Trying to train cifarnet using the train_cifarnet_on_cifar10.sh script.

Using ageron's fork, I'm getting error on line 132 of dataset_utils.py:
TypeError: a bytes-like object is required, not 'str'

After fixing that error with a hacky patch, I get:
  File "eval_image_classifier.py", line 104, in main
    num_classes=(dataset.num_classes - FLAGS.labels_offset),
AttributeError: 'Dataset' object has no attribute 'num_classes'

ageron:
Hi @vonclites , I'm not sure the bug you mention is related to this PR. Could you please provide more details on which version of python and tensorflow you are using, the command line you ran, its output and so on? I'd like to reproduce this issue. Thanks.

$ python -V
$ python -c "import tensorflow as tf; print(tf.__version__)"
...

slim/datasets/cifar10.py
MarkDaoust: The first error @vonclites hit is just because those two fields (`num_classes` and `labels_to_names`) are deleted by this patch. `Dataset` accepts `**kwargs` where you can put any name/value pairs you want.

`eval_image_classifier.py` is using `num_classes` to initialize the various network's output layer. 

The networks all have a default number of classes, but you need the dataset's number of classes if you apply a network to a different dataset (like using cifarnet on cifar10 or cifar100).

MarkDaoust:
The first error @vonclites hit is fixed in that link added by his edit: In py3 the `Gfile` returns bytes, they just need to be decoded. So it's just a simple py3 compatibility question.

ageron:
Thanks @MarkDaoust, you're absolutely right, I should not have removed them, I added them back. @vonclites, can you try again?

vonclites:
I'm using Python 3.5.2 and tf 0.11.0rc0.

That fixed helped (in combination with the lines.decode() fix)... but it has unveiled a new python 3 compatibility issue. I ran train_cifarnet_on_cifar10.sh and hit an error with eval_image_classifier.py:

`File "eval_image_classifier.py", line 161, in main
    for name, value in names_to_values.iteritems():
AttributeError: 'dict' object has no attribute 'iteritems'
`
Here are some possible fixes: 
http://legacy.python.org/dev/peps/pep-0469/#migrating-to-the-common-subset-of-python-2-and-3
http://legacy.python.org/dev/peps/pep-0469/#migrating-from-python-3-to-the-common-subset-with-python-2-7

ageron:
Hi @sguada, is there any chance this change can be pulled? Anything else I need to do? Thanks.

vonclites:
Works for me using Python 2.7 and 3.5.2 with CUDA 7.5 and Tensorflow 0.11.0rc0

jucor:
Friendly ping  on this, as I have the same problem as OP with python3. Thanks @sguada  :)

jucor:
I have solved the conflicts by rebasing https://github.com/ageron/tensorflow-models onto tensorflow/models current master. That rebase can be accessed in two ways:
- A PR in https://github.com/ageron/tensorflow-models/pull/1
- Or I can open a new PR here based  on https://github.com/jucor/models/tree/py3_flowers 
whichever you prefer.

I have also signed and submitted the CLA.

waynew30777:
didnt Works for me using Python3 and Tensorflow 1. 

download_and_convert_cifar10.py", line 29, in <module>
    import cPickle
ImportError: No module named 'cPickle'

I thought it has been merged ?

yifeif:
@ageron are we still working on this PR? Could you resolve the conflicts? Thanks!

MarkDaoust:
This is available from the tensorflow installation in the keras datasets:

``` python
train, test = tf.keras.datasets.cifar100.load_data()
```

tremblerz:
Is it more of a design decision by the tensorflow team to keep CIFAR-100 out of the `datasets` utilities?

PR #416, opened 2016-09-14
Fix bug in relative path of shell scripts built with bazel.

cshallue:


jmhodges:
Oh! This is like #419. Instead of `__main__` you could use $TEST_WORKSPACE, I think? And maybe use `TEST_SRCDIR` for the prefix?

Still a bazel newb, so maybe I'm off!

kchodorow:
LGTM.

PR #254, opened 2016-07-03
Fixed bug test failed on parser_trainer_test

tinh1115:
Fixed bug file not found "/syntaxnet/bazel-out/local-opt/bin/syntaxnet/parser_trainer_test.runfiles/syntaxnet/parser_trainer: No such file or directory"

PR #419, opened 2016-09-17
correct parser_trainer_test.sh for recent bazel

jmhodges:
Without this, the parser_trainer_test.sh errors out because the testdata/context.pbtxt file is not where the script thinks it is.


drpngx:
Duplicate of #254?

jmhodges:
Totally is

drpngx:
@slavpetrov does that make sense? It looks like two people hit this.

nealwu:
https://github.com/tensorflow/models/pull/254 has been merged. Thanks!

PR #431, opened 2016-09-20
Update tf.image.resize_images to latest signature

LiberiFatali:
From https://www.tensorflow.org/versions/master/api_docs/python/image.html#resize_images 

> size: A 1-D int32 Tensor of 2 elements: new_height, new_width. The new size for the images.


drpngx:
There's a conflict. Is this still relevant?

nealwu:
Looks like this is already fixed.

PR #442, opened 2016-09-23
format question fix

vra:
1. change `sharded` to `shared`
2. update shell output according code.

### DIFF
--- a/im2txt/README.md
+++ b/im2txt/README.md
@@ -130,7 +130,7 @@ First ensure that you have installed the following required packages:
 ### Prepare the Training Data
 
 To train the model you will need to provide training data in native TFRecord
-format. The TFRecord format consists of a set of sharded files containing
+format. The TFRecord format consists of a set of shared files containing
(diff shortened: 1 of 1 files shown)


drpngx:
"sharded" is intentional, it just means split or in multiple partitions.

vra:
@drpngx OK I see, sorry for the inconvenience. How about the second  modification about `.`? I read the code and there is code about putting a `.`. Would you please check it? Thanks. ;)

drpngx:
Are you sure there's no period at the end? It makes a difference to the score. From the output, it looks like the period is a token in the statistical model, because the 2nd candidate doesn't have it. You can't remove it just like that, or it will change the ranking & score. Makes sense?

vra:
You mean the period is just like a word and it's generated by deep learning model? It seems reasonable. So I will close this issue and great thanks for your explanation.

drpngx:
Yes, exactly! You're welcome.

PR #449, opened 2016-09-25
Fix signature change of resize_images()

chenliu0831:
Fix breaking change introduced [here](https://github.com/tensorflow/tensorflow/commit/8d1b23d999a563c56236a3d250e18ef825e4e66e)


chenliu0831:
FYI https://github.com/tensorflow/models/pull/431 seems to fix inception partially.

chenliu0831:
@lukaszkaiser Would you mind take a quick look?

PR #451, opened 2016-09-26
Fixes typo

jlertle:


jlertle:
I signed it!

drpngx:
I think we're re-organizing slim into tf. Closing.

PR #461, opened 2016-09-27
tf.image.resize_images() expects 'size' argument as a 1-D Tensor.

parvizp:
The TF Slim [README](https://github.com/tensorflow/models/tree/master/slim) recommends installing TensorFlow from master. However, in master the tf.image.resize_images() arguments have changed. This makes it compatible with TensorFlow @ [current master](https://github.com/tensorflow/tensorflow/commit/49bb6cc491869bd291c8ff4115907d3601332af9)


parvizp:
Ah, I missed the previous PRs: #449 #

PR #473, opened 2016-09-29
Appended vocab words/counts found in the toy dataset to vocab file

xtr33me:
Appended vocab words and their counts found in the toy dataset to the vocab file so that training and decoding provide more than <UNK> in the decode results.  Also removed duplicates in vocab.


xtr33me:
I signed it!

drpngx:
@shlens if that's something you want to look at.

It looks like the vocab file has a cutoff, not sure if we want *all* entries.

drpngx:
@panyx0718, @peterjliu is that something you want to look at?

panyx0718:
The toy data set is not expected to be complete or generate any results. It's just to give a feeling of running the model.

xtr33me:
Essentially all I did was run a script to generate all the vocab words found in the included toy dataset. I then tried to just use that file, but when I went to train it complained about the vocab file not containing enough words so I then just appended those words to the existing dataset and removed duplicates. This then provided actual minimalistic results when training/decoding.

nealwu:
Closing old PRs.

PR #485, opened 2016-10-01
Use open() instead of tf.gfile.FastGFile()

cshallue:
This fixes tensorflow/tensorflow#4685


vvkv:
What is the difference between open() and tf.gfile.FastGFile()? Is there a speed trade off if using on say a Raspberry Pi?

PR #503, opened 2016-10-04
added python3 support to read_label_file

jonasrauber:


asimshankar:
@nathansilberman : This seems like it's find to merge and will also fix #528 ?

asimshankar:
Checked with @nathansilberman  offline, merging this is. Thanks for the fix @jonasrauber !

PR #511, opened 2016-10-07
Add video prediction model

cbfinn:
A reimplementation of the 3 models in "Unsupervised Learning for Physical Interaction through Video Prediction," NIPS 2016.

Includes code for loading the publicly-released data into tensorflow.


cbfinn:
I signed it!

PR #521, opened 2016-10-08
Added STREET model for FSNS dataset

theraysmith:
As per previous email thread...


drpngx:
I don't have that email thread. Feel free to ping the TensorFlow participants.

PR #551, opened 2016-10-17
Python3 dict.values() does not return a list

inflation:
Add a list wrapper will make it run properly in Python3 as well as Python2.


inflation:
I signed it!

nealwu:
Thanks!

PR #556, opened 2016-10-18
Fix end point collection to return a dict

rohitgirdhar:
Fixes the TF slim models to be compatible with the change in TF 0.11 (https://github.com/tensorflow/tensorflow/commit/5247ebbe24465563bdd4ef46f1cecc34f256f76b)

Otherwise, running vgg/alexnet etc models gives the following error:
```txt
end_points = dict(tf.get_collection(end_points_collection))
TypeError: cannot convert dictionary update sequence element #0 to a sequence
```


drpngx:
Thanks!

PR #563, opened 2016-10-19
Fix POS tagging score of Ling et al.(2005)

j-min:
For English News Corpus,
[Ling et al. (2015)](http://www.cs.cmu.edu/~lingwang/papers/emnlp2015.pdf)'s score is 
97.78 -> 97.44 (lower than SyntaxNet and Parsey Mcparseface)
according to [Andor et al. (2016)](http://arxiv.org/abs/1603.06042).


j-min:
I signed it!

slavpetrov:
Thanks for correcting our mistake!

PR #565, opened 2016-10-19
Differential privacy analysis for the privacy model tutorial

npapernot:
This is related to my summer internship project with ulfar@ and abadi@. These PR adds the privacy analysis script used in our paper https://arxiv.org/abs/1610.05755 to compute the differential privacy bounds, and also releases some of the label files required to compute these bounds.


panyx0718:
Can you reduced the binary file size to below 50k?

npapernot:
Hi @panyx0718, I removed the large binaries from this repository and instead changed the script to download them when required. 

cc @UlfarErl

PR #520, opened 2016-10-08
fix module object has no attribute NodeDef for tensorflow 0.11

qingzew:
reference  https://github.com/tensorflow/tensorflow/issues/4759

PR #572, opened 2016-10-21
fix module object has no attribute NodeDef for tensorflow 0.11

qingzew:
reference https://github.com/tensorflow/tensorflow/issues/4759 and https://github.com/tensorflow/models/pull/520

### DIFF
--- a/inception/inception/slim/variables.py
+++ b/inception/inception/slim/variables.py
@@ -84,7 +84,7 @@ from __future__ import print_function
 
 import tensorflow as tf
 
-from tensorflow.core.framework import graph_pb2
+from tensorflow.core.framework import node_def_pb2
 from inception.slim import scopes
 
 # Collection containing all the variables created using slim.variables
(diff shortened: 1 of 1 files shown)


inception/inception/slim/variables.py
```diff
@@ -84,7 +84,7 @@
 
 import tensorflow as tf
 
-from tensorflow.core.framework import graph_pb2
```
drpngx: Remove both these lines

inception/inception/slim/variables.py
```diff
@@ -211,7 +211,7 @@ def variable_device(device, name):
   """Fix the variable device to colocate its ops."""
   if callable(device):
     var_name = tf.get_variable_scope().name + '/' + name
-    var_def = graph_pb2.NodeDef(name=var_name, op='Variable')
```
drpngx: Use the blessed way: tf.NodeDef

qingzew:
@drpngx I have change it

drpngx:
Thanks!

PR #591, opened 2016-10-27
Slim walkthough notebook update to show difference of inference for models with 1000 and 1001 classes.

warmspringwinds:
A small update that makes it easier for end-users to understand the difference of how to
make inference with models that were trained using 1000 classes and 1001 classes on imagenet.

I faced this problem when I plugged-in VGG-19 model into Inception v1 code and got wrong predictions
because of the shift in class names.
https://github.com/tensorflow/tensorflow/issues/5228

The pull request adds example of evaluation of VGG-19 with correct name shift.

### DIFF
--- a/slim/slim_walkthough.ipynb
+++ b/slim/slim_walkthough.ipynb
@@ -753,7 +753,9 @@
     "However, this  means they must be trained on big datasets. Since this process is slow, we provide various pre-trained models - see the list [here](https://github.com/tensorflow/models/tree/master/slim#pre-trained-models).\n",
     "\n",
     "\n",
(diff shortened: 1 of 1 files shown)


warmspringwinds:
I signed it!

sguada:
For efficiency reasons and also because VGG-16 performs as well as VGG-19, can you use VGG-16 instead?

warmspringwinds:
@sguada Did it :)

If you are interested, I can also add some parts from [here](http://warmspringwinds.github.io/tensorflow/tf-slim/2016/10/30/image-classification-and-segmentation-using-tensorflow-and-tf-slim/).

warmspringwinds:
@sguada Do you want me to add something else or update ?

PR #600, opened 2016-10-31
tf-model-slim doc typo

leonardgithub:
it should be cd $HOME/workspace/models/slim


leonardgithub:
@googlebot  my CLA has been submitted. Thank you

PR #617, opened 2016-11-04
Remove privacy/ after consolidation.

panyx0718:
Now differential_privacy and privacy are
under the same project.


panyx0718:
@npapernot

npapernot:
Can you hold on for a day, I am updating our arXiv paper with the correct link. Thank you!

PR #625, opened 2016-11-09
val_captions_file -> captions_val2014.json

legel:
@cshallue it seems validation captions are pointing to the training file.  This looks like a quick fix.


legel:
Thanks @googlebot I signed it

Issue #634, opened 2016-11-11
may be an error in slim.nets.inception_resnet_v2


burness:
## Please let us know which model this issue is about (specify the top-level directory)
Hi, I found that may be an error in inception_resnet_v2.py, from the line L174:

```
        # 17 x 17 x 1024
        with tf.variable_scope('Mixed_6a'):
          with tf.variable_scope('Branch_0'):
            tower_conv = slim.conv2d(net, 384, 3, stride=2, padding='VALID',
                                     scope='Conv2d_1a_3x3')
          with tf.variable_scope('Branch_1'):
            tower_conv1_0 = slim.conv2d(net, 256, 1, scope='Conv2d_0a_1x1')
            tower_conv1_1 = slim.conv2d(tower_conv1_0, 256, 3,
                                        scope='Conv2d_0b_3x3')
            tower_conv1_2 = slim.conv2d(tower_conv1_1, 384, 3,
                                        stride=2, padding='VALID',
                                        scope='Conv2d_1a_3x3')
          with tf.variable_scope('Branch_2'):
            tower_pool = slim.max_pool2d(net, 3, stride=2, padding='VALID',
                                         scope='MaxPool_1a_3x3')
          net = tf.concat(3, [tower_conv, tower_conv1_2, tower_pool])
```
the channel num of tower_conv is 384;
the channel num of tower_conv1_2 is 384;
the channel num of tower_pool is as same as net from `mixed_5b`, and it is 320;
So the channel should be 320+384+384=1088, but your comment is 17\*17\*1024;
I don't know that the code has and error or just the comment is not right?
Pls let know if there is anything that i can't take in account!!
Thanks

girving:
@alexalemi Can you take a look?

PR #641, opened 2016-11-14
Fix Inception Resnet V2 Size Comment

alexalemi:
Fix comment. Address issue: https://github.com/tensorflow/models/issues/634#issuecomment-260399899


nealwu:
Thanks!

PR #651, opened 2016-11-16
Match the gfile's change

macwang:
Otherwise there is no Size() in GFile anymore and it causes the downloading failure.


macwang:
I signed it!

macwang:
I signed it!

macwang:
I signed it!

nealwu:
Hi @macwang, the `f.Size()` -> `f.size()` change has already been merged into master. Is there a reason we should still remove `tf.gfile`?

macwang:
OK, it's not needed anymore.

PR #652, opened 2016-11-16
Train generic data model from buil_image_data.py script output data

paucarre:


nealwu:
@paucarre could you explain what this does? What do you mean by a "generic data set," and what would we want to train on one for?

nealwu:
Closing old PRs.

PR #656, opened 2016-11-17
Fix ValueError when processing bbox in XML annotation file

xionghcx:
In some XML annotation files, point values [xmin, ymin, xmax, ymax] relative to bounding box are string type look like a float number. So if want to use int() convert those string to int. The program will raise an ValueError.Such as:
`ValueError: invalid literal for int() with base 10: '64.0994'`

So I fixed this bug by adding an float function before int function.


nealwu:
Thanks!

Issue #193, opened 2016-06-13
Spatial transformer implementation may have a bug somewhere


seuqaj114:
So I was trying to plug this ST module into a the write attention part of a DRAW model, and I just couldn't get it to work. After a grueling day of trying every parameter choice, I tried comparing the output of an identity scaling of ST vs scipy interpolation zoom, and that's when I found something interesting.

The code I used is below, adapted from the example.py.

```
from scipy import ndimage
import tensorflow as tf
from spatial_transformer import transformer
import numpy as np
import matplotlib.pyplot as plt

def sigmoid(x):
    return 1.0/(1.0+np.exp(-x))

# %% Create a batch of three images (1600 x 1200)
# %% Image retrieved from:
# %% https://raw.githubusercontent.com/skaae/transformer_network/master/cat.jpg
#im = np.load("../../datautils/templates/circle/circle.npy")
#im = im / 255.
im = np.array([-1.2053933, -1.1743802, -0.75044346, -0.74455976, -1.0506268,
 -0.91364104, -0.21054152, 0.1543106, 0.032554384, -0.52717745,
-0.66026419, -0.021319218, -0.060581781, -0.099243492, -0.26127103,
 -0.52252597, 0.1389422, -0.13638327, 0.033274196, -0.20344208,
  -0.53625256, 0.02523746, -0.076311894, 0.10775769, 0.20])
im = im.reshape(1, 5, 5, 1)
im = im.astype('float32')

# %% Let the output size of the transformer be 5 times the image size.
out_size = (30, 30)

# %% Simulate batch
batch = np.append(im, im, axis=0)
batch = np.append(batch, im, axis=0)
num_batch = 3

x = tf.placeholder(tf.float32, [None, 5, 5, 1])
x = tf.cast(batch, 'float32')

# %% Create localisation network and convolutional layer
with tf.variable_scope('spatial_transformer_0'):

    # %% Create a fully-connected layer with 6 output nodes
    n_fc = 6
    W_fc1 = tf.Variable(tf.zeros([5*5*1, n_fc]), name='W_fc1')

    # %% Zoom into the image
    initial = np.array([[1.0, 0, 0.0], [0, 1.0, 0.0]])
    initial = initial.astype('float32')
    initial = initial.flatten()

    b_fc1 = tf.Variable(initial_value=initial, name='b_fc1')
    h_fc1 = tf.matmul(tf.zeros([num_batch, 5*5*1]), W_fc1) + b_fc1
[...]

PR #659, opened 2016-11-17
#193 Fixed Spatial Transformer not handling down/right side correctly

mfolnovic:
I had issue described in #193, in both 42x42 -> 28x28 and 28x28 -> 42x42 cases (size of input to STN -> size of STN output), where I always had a black border of a few pixels on down/right side.

I basically looked at Lasagne changes of [1] and backported those in here.

I've only tested with identity transformation which now works as expected (no black border).

[1] https://github.com/Lasagne/Lasagne/blob/master/lasagne/layers/special.py

### DIFF
--- a/transformer/spatial_transformer.py
+++ b/transformer/spatial_transformer.py
@@ -72,24 +72,25 @@ def transformer(U, theta, out_size, name='SpatialTransformer', **kwargs):
             width_f = tf.cast(width, 'float32')
             out_height = out_size[0]
             out_width = out_size[1]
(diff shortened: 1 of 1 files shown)


girving:
@aselle This PR seems to have gotten lost.  The mentioned bug was assigned to you; would you want to take a look at this PR?

nealwu:
@aselle is this still something we should be considering?

nealwu:
@daviddao seems like this is intended as a bug fix. Does it look appropriate to you? If not we can close the PR.

PR #725, opened 2016-12-08
Fix for issue #621

### LINKED ISSUE #621: TextSum does not actually check the presence of Tags in vocabulary
https://github.com/tensorflow/models/blob/master/textsum/seq2seq_attention.py#L163

The way of checking presence of required tokens is wrong because _**WordToId**_ function returns ID of Unknown <UNK> token **(which might be >=0 for most of the cases)** if the that particular token is not found, hence the scripts validates a wrong vocabulary file.

I am training the TextSum algorithm on CNN Dataset and would also like to know about the presence of <PAD> in sample Toy Vocabulary, what should be the location/use of <PAD> in the data? Can someone provide a sample format for data including the <PAD> token, because I don't have it in my vocabulary.
I would like to know how to include it in the data.

Possible Fix: Raising an error if the WordToID function when the word is not found.

pranay360:
Necessary vocabulary tokens are now checked using CheckVocab method in Class Vocab.
An AssertionError is raised if the necessary token is not present in Vocabulary.
Fixes #621

### DIFF
--- a/textsum/data.py
+++ b/textsum/data.py
@@ -56,6 +56,12 @@ class Vocab(object):
         if self._count > max_size:
           raise ValueError('Too many words: >%d.' % max_size)
 
+  def CheckVocab(self, word):
+    if word not in self._word_to_id:
+      return None
+    return self._word_to_id[word]
+
+  
   def WordToId(self, word):
     if word not in self._word_to_id:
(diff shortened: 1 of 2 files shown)


pranay360:
I signed it.

PR #732, opened 2016-12-10
Remove WARNING in calling function softmax_cross_entropy

bigsnarfdude:
https://www.tensorflow.org/code/tensorflow/contrib/losses/python/losses/loss_ops.py was refactored and now when running script produces a WARNING.

```
WARNING:tensorflow:From train_image_classifier.py:477 in clone_fn.: calling softmax_cross_entropy (from tensorflow.contrib.losses.python.losses.loss_ops) with weight is deprecated and will be removed after 2016-11-25.
Instructions for updating:
`weight` is being deprecated, use `weights`
```

### DIFF
--- a/slim/train_image_classifier.py
+++ b/slim/train_image_classifier.py
@@ -471,9 +471,9 @@ def main(_):
       if 'AuxLogits' in end_points:
         slim.losses.softmax_cross_entropy(
             end_points['AuxLogits'], labels,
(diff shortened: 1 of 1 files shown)


nealwu:
Thanks! This should be covered by https://github.com/tensorflow/models/pull/1063 and https://github.com/tensorflow/models/pull/1053, so I'm going to close.

PR #744, opened 2016-12-12
Fix the problem that all the words become UNK

yh157612:
All the words became `_UNK` since the keys of `vocab` were in `str` type.
The problem occurred in Python 3.5


nealwu:
Thanks! This same fix has actually been [submitted separately](https://github.com/tensorflow/models/pull/823/files), so closing this.

PR #746, opened 2016-12-12
Update README.md

christieewen:
fixed broken url link


itsmeolivia:
Looks like this is fixed in the most recent version.  Thanks!

PR #747, opened 2016-12-12
Update cifar10 model works with the latest tensorflow API

darcyliu:


nealwu:

PR #749, opened 2016-12-13
Translate - Added the option to decode a file and or write to a file

rmeertens:
With this pull request you can easily translate/test a batch of sentences, which makes it easier to see how good your trained model is.

If you specify a filename this functionality is used, if you don't specify anything the script continues to function as it used to.

### DIFF
--- a/tutorials/rnn/translate/data_utils.py
+++ b/tutorials/rnn/translate/data_utils.py
@@ -22,6 +22,7 @@ import gzip
 import os
 import re
 import tarfile
+import sys
 
 from six.moves import urllib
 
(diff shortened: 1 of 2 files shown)


nealwu:
Hey Roland, what benefit would this provide over just redirecting stdin / stdout from the command line? Seems like a lot of extra code for something you should already be able to do.

rmeertens:
Only difference is indeed that this solution does not place the '> ' in front of each translated sentence, and the "reading model parameters... " message does not show up in your file. 

But I agree that 80 lines (most in data_utils) might be a lot to fix a minor annoyance. Do whatever you want with this pull request!

nealwu:
Right, it is a lot of code to add. We'd like to keep everything as concise as possible, especially for these tutorial models. For that reason I'm going to close this PR. Thanks for submitting though.

PR #750, opened 2016-12-13
update inception v3 model works with the latest tensorflow API

minbinL:
The '--pretrained_model_checkpoint_path' specify the file of a checkpoint and the new checkpoint V2 is not support read from file. 
Because of the changes to tf.all_variables() and tf.GraphKeys, the collection name is changed and the necessary parameters in checkpoint cannot be saved in training process which cause checkpoint not able to be read to continue training.

### DIFF
--- a/inception/README.md
+++ b/inception/README.md
@@ -540,7 +540,7 @@ the flowers data set with the following command.
 bazel build inception/flowers_train
 
 # Path to the downloaded Inception-v3 model.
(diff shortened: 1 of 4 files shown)


nealwu:
I think this is no longer necessary due to https://github.com/tensorflow/models/pull/1063. If that's not the case feel free to reopen and let me know.

Issue #690, opened 2016-11-28
ResNet-50 Slim: ValueError: Shapes (16, 1, 1, 5) and (16, 5) are incompatible


aldarabs:
## ResNet-50 Slim: ValueError: Shapes (16, 1, 1, 5) and (16, 5) are incompatible

I am trying to train resnet-50 with the flower dataset. However, I am getting this error. I created the dataset using the code provided. Then, I trained using the code that also provided. Below is the code and the error. 

DATA_DIR=/media/khaldun/drive1/TensorFlow_ME/dataset
DATASET_DIR=/media/khaldun/drive1/TensorFlow_ME/dataset
TRAIN_DIR=/media/khaldun/drive1/TensorFlow_ME/resnet50
CHECKPOINT_PATH=/media/khaldun/drive1/TensorFlow_ME/checkpoint/resnet_v1_50.ckpt

/usr/bin/python train_image_classifier.py \
  --train_dir=${TRAIN_DIR} \
  --dataset_name=flowers \
  --dataset_split_name=train \
  --dataset_dir=${DATASET_DIR} \
  --model_name=resnet_v1_50 \
  --checkpoint_path=${CHECKPOINT_PATH} \
  --checkpoint_exclude_scopes=ResNet50/Logits \
  --trainable_scopes=ResNet50/Logits \
  --max_number_of_steps=3000 \
  --batch_size=16 \
  --learning_rate=0.01 \
  --save_interval_secs=60 \
  --save_summaries_secs=60 \
  --log_every_n_steps=100 \
  --optimizer=rmsprop \
  --weight_decay=0.00004 

WARNING:tensorflow:From train_image_classifier.py:483 in clone_fn.: calling softmax_cross_entropy (from tensorflow.contrib.losses.python.losses.loss_ops) with weight is deprecated and will be removed after 2016-11-25.
Instructions for updating:
`weight` is being deprecated, use `weights`
Traceback (most recent call last):
  File "train_image_classifier.py", line 592, in <module>
    tf.app.run()
  File "/home/khaldun/tensorflow/_python_build/tensorflow/python/platform/app.py", line 43, in run
    sys.exit(main(sys.argv[:1] + flags_passthrough))
  File "train_image_classifier.py", line 489, in main
    clones = model_deploy.create_clones(deploy_config, clone_fn, [batch_queue])
  File "/media/khaldun/drive1/TensorFlow_ME/slim/deployment/model_deploy.py", line 195, in create_clones
[...]

PR #751, opened 2016-12-13
Align model slim/resnet to slim/inception

stef716:
Align model slim/resnet to slim/inception (set spatial_squeeze and default_image_size).
Also add a sample script for training resnet_v1_50 on flower set.

This solves #690

### DIFF
--- a/slim/nets/inception_v1.py
+++ b/slim/nets/inception_v1.py
@@ -270,7 +270,7 @@ def inception_v1(inputs,
     is_training: whether is training or not.
     dropout_keep_prob: the percentage of activation values that are retained.
     prediction_fn: a function to get predictions out of logits.
-    spatial_squeeze: if True, logits is of shape is [B, C], if false logits is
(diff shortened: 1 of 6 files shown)


ahundt:
I tested this and it works for me! The new script is also a very helpful demonstration, much better than what is in the README.md.

However, regarding the the various default image size parameters like `resnet_v2_200.default_image_size = 224`: should those really be a scalar? Could this break fcn mode where dense predictions are made and images can be rectangular?

stef716:
@ahundt I added these defaults to all resnet models because the current interfaces require it.
I think the squared image extracts are typical for supporting the mixed datasets of portrait and landscape images. In fixed camera environments you definitely want non-squared aspect ratios depending on your scenario. Train, Eval scripts and models could easily be extended. But would be a different issue.

ahundt:
seems sensible

nealwu:
Thanks!

PR #760, opened 2016-12-15
Python 3 support for some inception scripts

stakemura:
This PR makes it possible to use some inception scripts with python 3.

Regarding python 3, the problems were essentially:

xrange => range
x=range(l); random.suffle(x) => x=list(range(l)); ...


nealwu:
Thanks!

PR #769, opened 2016-12-19
Update cifar10_input.py to work through on version 0.12.0-rc1

un-knight:
I'm working with `models/tutorials/image/cifar10/`.

There is an error `strided_slice() missing 1 required positional argument: 'strides'` when I run the command `python cifar10_train.py` under the version of 0.12.0-rc1. Although the [document](https://github.com/tensorflow/tensorflow/blob/master/tensorflow/g3doc/api_docs/python/functions_and_classes/shard8/tf.strided_slice.md) said `strides` is optional.

So I replaced `strided_slice` with `slice` to make it work as a temporary solution.

### DIFF
--- a/tutorials/image/cifar10/cifar10_input.py
+++ b/tutorials/image/cifar10/cifar10_input.py
@@ -84,13 +84,13 @@ def read_cifar10(filename_queue):
 
   # The first bytes represent the label, which we convert from uint8->int32.
   result.label = tf.cast(
-      tf.strided_slice(record_bytes, [0], [label_bytes]), tf.int32)
+      tf.slice(record_bytes, [0], [label_bytes]), tf.int32)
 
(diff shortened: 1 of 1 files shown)


un-knight:
I signed it

daviddao:
Might that be a better solution: http://stackoverflow.com/questions/41195121/tensorflow-strided-slice-missing-argument-strides ? (instead of using tf.slice())

un-knight:
Hi! @daviddao, I can't say which one is better, though I had thought about both of the solutions. But the reason why I PR this solution (using `tf.slice()`) as a temporary solution is  that you don't need to consider the value of parameter `strides`.

dweekly:
Looks like this may fix issue #817

nealwu:

Issue #767, opened 2016-12-16
inception model training failing due to a ValueError


akshayc11:
## Please let us know which model this issue is about (specify the top-level directory)
Directory: tensorflow/models/inception

I have been trying to run the inception-net model following the instructions in the README.md.
`$ bazel-bin/inception/imagenet_train --num_gpus=2 --batch_size=128 --train_dir=/data-local/akshayc/imagenet_train --data_dir="${DATA_DIR}"`
However, I get the following error:
```
Traceback (most recent call last):
  File "/home/akshayc/tensorflow-source/tensorflow/models/inception/bazel-bin/inception/imagenet_train.runfiles/inception/inception/imagenet_train.py", line 41, in <module>
    tf.app.run()
  File "/home/akshayc/tensorflow/local/lib/python2.7/site-packages/tensorflow/python/platform/app.py", line 43, in run
    sys.exit(main(sys.argv[:1] + flags_passthrough))
  File "/home/akshayc/tensorflow-source/tensorflow/models/inception/bazel-bin/inception/imagenet_train.runfiles/inception/inception/imagenet_train.py", line 37, in main
    inception_train.train(dataset)
  File "/home/akshayc/tensorflow-source/tensorflow/models/inception/bazel-bin/inception/imagenet_train.runfiles/inception/inception/inception_train.py", line 239, in train
    scope)
  File "/home/akshayc/tensorflow-source/tensorflow/models/inception/bazel-bin/inception/imagenet_train.runfiles/inception/inception/inception_train.py", line 124, in _tower_loss
    loss_averages_op = loss_averages.apply(losses + [total_loss])
  File "/home/akshayc/tensorflow/local/lib/python2.7/site-packages/tensorflow/python/training/moving_averages.py", line 391, in apply
    self._averages[var], var, decay, zero_debias=zero_debias))
  File "/home/akshayc/tensorflow/local/lib/python2.7/site-packages/tensorflow/python/training/moving_averages.py", line 70, in assign_moving_average
    update_delta = _zero_debias(variable, value, decay)
  File "/home/akshayc/tensorflow/local/lib/python2.7/site-packages/tensorflow/python/training/moving_averages.py", line 177, in _zero_debias
    trainable=False)
  File "/home/akshayc/tensorflow/local/lib/python2.7/site-packages/tensorflow/python/ops/variable_scope.py", line 1024, in get_variable
    custom_getter=custom_getter)
[...]

PR #778, opened 2016-12-20
Variables defined in ExponentialMovingAverage need not to be shared.

yuefengz:
When I ran the inception model with `num_gpus=2`, it gave me the following error: 
  ```
  ValueError: Variable tower_1/tower_1/CrossEntropyLoss/value/avg/biased does not exist, or was not created with tf.get_variable(). Did you mean to set reuse=None in VarScope?
  ```

There is also a similar issue: https://github.com/tensorflow/models/issues/767.

I guess the variables defined in ExponentialMovingAverage could be unshared. This makes the model train but I am not sure whether it changes the result. Please advise.


inception/inception/inception_train.py
```diff
@@ -79,7 +79,7 @@
 RMSPROP_EPSILON = 1.0              # Epsilon term for RMSProp.
 
 
-def _tower_loss(images, labels, num_classes, scope):
+def _tower_loss(images, labels, num_classes, scope, reuse_variables=False):
```
sguada: Reuse should be either None or True, never False.

inception/inception/inception_train.py
```diff
@@ -220,13 +221,14 @@ def train(dataset):
...
     num_classes = dataset.num_classes() + 1
-    
+
      # Split the batch of images and labels for towers.
     images_splits = tf.split(0, FLAGS.num_gpus, images)
     labels_splits = tf.split(0, FLAGS.num_gpus, labels)
 
     # Calculate the gradients for each model tower.
     tower_grads = []
+    reuse_variables = False
```
sguada: reuse_variables = None

PR #781, opened 2016-12-21
Add Python3-compatible comparison methods

myurasov:
Otherwise breaks with `TypeError: unorderable types: Caption() < Caption()` in Python 3.5


nealwu:
Thanks! Closing this because it was merged separately in https://github.com/tensorflow/models/pull/842/files

Issue #782, opened 2016-12-21
TF-Slim Walkthrough slim/slim_walkthough.ipynb Doesn't Work With TF 0.12


dweekly:
The walkthrough has the user use the sum_of_squares() loss function which was removed after TF 0.10 and is no longer present in TF 0.12. Please update the walkthrough to work with TF 0.12.

loss = slim.losses.sum_of_squares(predictions, targets)

drstarson:
The suggested solution worked for me.
Update sum_of_squares (TF 0.10) to mean_squared_error  (TF 0.12)

nealwu:
https://github.com/tensorflow/models/pull/783 has been merged.

PR #783, opened 2016-12-21
Update sum_of_squares (TF 0.10) to mean_squared_error (TF 0.12)

dweekly:
Fix deprecation of sum_of_squares so the notebook works with TF 0.12.

#782


nealwu:
Thanks!

Issue #784, opened 2016-12-21
tf.contrib.deprecated methods is eliminated in tf version 0.12


mrphoenix13:
As I mentioned in the title the tf.contrib.deprecated is eliminated in tensorflow version 0.12. So I suggest to edit the code. If it's ok you can make it contribute welcome for me to edit it myself.

## Please let us know which model this issue is about (specify the top-level directory)
image/cifar10
embedding
rnn/ptb

aselle:
Thanks for helping out @mrphoenix13!

mingxingtan:
Close due to inactivity for a long time. Please check out the latest tf and model code.

PR #785, opened 2016-12-21
tf.contrib.deprecated elimination bugfix for tf version 12

mrphoenix13:
according to #784  after upgrading tf version to 0.12 tf.contrib.deprecated has been eliminated. This pull request fixes the code to make it work with tf version 0.12.

### DIFF
--- a/tutorials/embedding/word2vec.py
+++ b/tutorials/embedding/word2vec.py
@@ -365,7 +365,7 @@ class Word2Vec(object):
       self._word2id[w] = i
     true_logits, sampled_logits = self.forward(examples, labels)
     loss = self.nce_loss(true_logits, sampled_logits)
-    tf.contrib.deprecated.scalar_summary("NCE loss", loss)
+    tf.scalar_summary("NCE loss", loss)
     self._loss = loss
     self.optimize(loss)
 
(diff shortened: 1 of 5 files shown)


YuMS:
`tf.scalar_summary` is also deprecated in TensorFlow 0.12.1, these are some warnings for `tf.scalar_summary`


```
WARNING:tensorflow:From ptb_word_lm.py:341 in main.: scalar_summary (from tensorflow.python.ops.logging_ops) is deprecated and will be removed a
fter 2016-11-30.
Instructions for updating:
Please switch to tf.summary.scalar. Note that tf.summary.scalar uses the node name instead of the tag. This means that TensorFlow will automatic
ally de-duplicate summary names based on the scope they are created in. Also, passing a tensor or list of tags to a scalar summary op is no long
er supported.
```

MartinThoma:
See https://github.com/MartinThoma/algorithms/commit/38ce1f87d6e4396cde64fe831c2ead2507781270 for the changes which need to be made.

BSVogler:
@MartinThoma Can you merge it with the tensorflow:master please?
Fond your fix via google as the current version does not run by missing the strides parameter.

MartinThoma:
@BSVogler Sorry, I was busy the last days. Here is the pull request which seems to fix the problems https://github.com/tensorflow/models/pull/962

PR #787, opened 2016-12-21
Fixed model_deploy so that weights are assigned to the variables_device

vonclites:
The arg_scope of the network_fn was being determined when get_network_fn was called. Thus, the network_fn 'escaped' back to the closure (I hope I'm using that term right?) in which it was originally created, ignoring the arg-scope stack created afterwards by the clone. When the first clone called the network_fn, the weights were put on the default device.  So for two GPUs (and two clones), the weights would be assigned to first clone / first GPU, and the second clone would have to read the weights from the first clone's memory.

I simply moved the call to create the model's arg_scope from the outer get_network_fn to the inner network_fn.  This way, when network_fn gets called, the model's arg_scope is properly nested within the arg_scopes created by the clones.


vonclites:
OK. I signed the CLA with my other account.

PR #802, opened 2016-12-23
Word2vec can now be run if the users compile the ops on their own

nealwu:
It doesn't seem possible to get this code to run using bazel unfortunately. I believe this is because when bazel calls `load("//tensorflow:tensorflow.bzl", "tf_gen_op_wrapper_py")`, it's unable to continue recursively loading within the TensorFlow repo, even after adding the TensorFlow repo into this repo's WORKSPACE file. This prevents us from being able to use `tf_gen_op_wrapper_py` and import the ops as needed.

I was finally able to get this model working by following the [TensorFlow tutorial for adding a new op](https://www.tensorflow.org/how_tos/adding_an_op/#building_the_op_library) and instructing the users to compile the ops on their own before running.

Unfortunately, this results in the Python code having to load word2vec_ops.so, which is not ideal and not cross-platform friendly. If you have other ideas for how to solve this, definitely let me know.

### DIFF
--- a/tutorials/embedding/__init__.py
+++ b/tutorials/embedding/__init__.py
@@ -17,5 +17,3 @@
 from __future__ import absolute_import
 from __future__ import division
 from __future__ import print_function
-
-import gen_word2vec
--- a/tutorials/embedding/word2vec.py
+++ b/tutorials/embedding/word2vec.py
@@ -42,7 +42,7 @@ from six.moves import xrange  # pylint: disable=redefined-builtin
 import numpy as np
 import tensorflow as tf
 
-word2vec = tf.load_op_library(os.path.dirname(os.path.realpath(__file__)) + '/word2vec_ops.so')
+word2vec = tf.load_op_library(os.path.join(os.path.dirname(os.path.realpath(__file__)), 'word2vec_ops.so'))
 
 flags = tf.app.flags
 
--- a/tutorials/embedding/word2vec.py
+++ b/tutorials/embedding/word2vec.py
@@ -42,7 +42,7 @@ from six.moves import xrange  # pylint: disable=redefined-builtin
 import numpy as np
 import tensorflow as tf
 
-import gen_word2vec as word2vec
+word2vec = tf.load_op_library(os.path.dirname(os.path.realpath(__file__)) + '/word2vec_ops.so')
 
 flags = tf.app.flags
 
--- a/tutorials/embedding/word2vec_optimized.py
+++ b/tutorials/embedding/word2vec_optimized.py
@@ -41,7 +41,7 @@ from six.moves import xrange  # pylint: disable=redefined-builtin
 import numpy as np
 import tensorflow as tf
 
-word2vec = tf.load_op_library(os.path.dirname(os.path.realpath(__file__)) + '/word2vec_ops.so')
+word2vec = tf.load_op_library(os.path.join(os.path.dirname(os.path.realpath(__file__)), 'word2vec_ops.so'))
 
 flags = tf.app.flags
 
--- a/tutorials/embedding/word2vec_optimized.py
+++ b/tutorials/embedding/word2vec_optimized.py
@@ -41,7 +41,7 @@ from six.moves import xrange  # pylint: disable=redefined-builtin
 import numpy as np
 import tensorflow as tf
 
-import gen_word2vec as word2vec
+word2vec = tf.load_op_library(os.path.dirname(os.path.realpath(__file__)) + '/word2vec_ops.so')
 
 flags = tf.app.flags
 
--- a/tutorials/embedding/README.md
+++ b/tutorials/embedding/README.md
@@ -26,7 +26,7 @@ TF_INC=$(python -c 'import tensorflow as tf; print(tf.sysconfig.get_include())')
 g++ -std=c++11 -shared word2vec_ops.cc word2vec_kernels.cc -o word2vec_ops.so -fPIC -I $TF_INC -O2
 ```
 
-(If you want to know what this is doing, see the tutorial on [Adding a New Op to TensorFlow](https://www.tensorflow.org/how_tos/adding_an_op/#building_the_op_library)).
+(For an explanation of what this is doing, see the tutorial on [Adding a New Op to TensorFlow](https://www.tensorflow.org/how_tos/adding_an_op/#building_the_op_library)).
 Then run using:
 
 ```shell
--- a/tutorials/embedding/README.md
+++ b/tutorials/embedding/README.md
@@ -21,11 +21,12 @@ rm source-archive.zip
 
 You will need to compile the ops as follows:
 
-``` shell
+```shell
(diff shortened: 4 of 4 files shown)


tutorials/embedding/word2vec.py
```diff
@@ -42,7 +42,7 @@
 import numpy as np
 import tensorflow as tf
 
-import gen_word2vec as word2vec
+word2vec = tf.load_op_library(os.path.dirname(os.path.realpath(__file__)) + '/word2vec_ops.so')
```
martinwicke: Do you need the absolute path here? You're already assuming you're in the same directory, why no load just 'word2vec_ops.so'?

martinwicke: If you do need the full name, use os.path.join to concatenate.

tutorials/embedding/word2vec_optimized.py
```diff
@@ -41,7 +41,7 @@
 import numpy as np
 import tensorflow as tf
 
-import gen_word2vec as word2vec
+word2vec = tf.load_op_library(os.path.dirname(os.path.realpath(__file__)) + '/word2vec_ops.so')
```
martinwicke: Whatever you end up with above, same here.

gunan:
What if we added tensorflow like an external dependency for models?
somewhat like jpeg, or llvm in tensorflow?

Of course, we will need to refer to tensorflow as `@tensorflow//:tensorflow` in models repo, but bazel will be able to automatically fetch and wire everything for us?

tutorials/embedding/word2vec.py
nealwu: For some reason when I tried `tf.load_op_library('word2vec_ops.so'), I got this error: tensorflow.python.framework.errors_impl.NotFoundError: word2vec_ops.so: cannot open shared object file: No such file or directory.

tutorials/embedding/__init__.py
```diff
@@ -17,5 +17,3 @@
 from __future__ import absolute_import
 from __future__ import division
 from __future__ import print_function
-
-import gen_word2vec
```
nealwu: Is deleting this line the right thing to do here? I could also call `tf.load_op_library`. Just wasn't sure how this is being used.

tutorials/embedding/word2vec.py
nealwu: Anyway, done.

tutorials/embedding/word2vec_optimized.py
nealwu: Done.

nealwu:
@gunan: If you think that has a good chance of working, I can give it a try. Would that require users to clone both repos?

gunan:
In that model, users will clone tensorflow/models, and bazel will do the rest when they are building/running anything in models repo.
We can look into it together if you like.

vincenttsai2015:
I tried to execute the code "word2vec_optimized.py (cloned from 'https://github.com/tensorflow/models')" after compiling the ops.
But I got the following error messages:
  File "word2vec_optimized.py", line 46, in <module>
    word2vec = tf.load_op_library(os.path.join(os.path.dirname(os.path.realpath(__file__)), 'word2vec_ops.so'))
  File "/usr/local/lib/python2.7/dist-packages/tensorflow/python/framework/load_library.py", line 64, in load_op_library
    None, None, error_msg, error_code)
tensorflow.python.framework.errors_impl.NotFoundError: /home/vincenttsai/tensorflow/tensorflow/models/tutorials/embedding/word2vec_ops.so: undefined symbol: _ZN10tensorflow7strings6StrCatB5cxx11ERKNS0_8AlphaNumE

Is there anything to fix? Thanks.

nealwu:
@vincenttsai2015 That seems odd. Can you let me know your OS, g++ version, and the exact steps you used to reproduce?

vincenttsai2015:
@nealwu  
OS: Linux Mint 18 Sarah (GNU/Linux 4.4.0-21-generic x86_64)
g++ version: gcc version 5.4.0 20160609 (Ubuntu 5.4.0-6ubuntu1~16.04.4)

What does "the exact steps I used to reproduce" mean? Do you mean the following commands?

TF_INC=$(python -c 'import tensorflow as tf; print(tf.sysconfig.get_include())')
g++ -std=c++11 -shared word2vec_ops.cc word2vec_kernels.cc -o word2vec_ops.so -fPIC -I $TF_INC -O2

nealwu:
@josh11b do you know why this might happen? I tested this out on my end and the compiled ops are still working fine for me.

BudickDa:
On Ubuntu 16.04 exactly the same error:
Followed the instruction here:
https://github.com/tensorflow/models/tree/master/tutorials/embedding

`wget http://mattmahoney.net/dc/text8.zip -O text8.zip`
`unzip text8.zip`
`wget https://storage.googleapis.com/google-code-archive-source/v2/code.google.com/word2vec/source-archive.zip`
`unzip -p source-archive.zip  word2vec/trunk/questions-words.txt > questions-words.txt`
`rm source-archive.zip`

`TF_INC=$(python -c 'import tensorflow as tf; print(tf.sysconfig.get_include())')`
`echo  $TF_INC`: /usr/local/lib/python2.7/dist-packages/tensorflow/include

`g++ -std=c++11 -shared word2vec_ops.cc word2vec_kernels.cc -o word2vec_ops.so -fPIC -I $TF_INC -O2`

`python word2vec_optimized.py \
  --train_data=text8 \
  --eval_data=questions-words.txt \
  --save_path=/tmp/`

Error:
  File "word2vec_optimized.py", line 44, in <module>
    word2vec = tf.load_op_library(os.path.join(os.path.dirname(os.path.realpath(__file__)), 'word2vec_ops.so'))
  File "/usr/local/lib/python2.7/dist-packages/tensorflow/python/framework/load_library.py", line 64, in load_op_library
    None, None, error_msg, error_code)
tensorflow.python.framework.errors_impl.NotFoundError: /home/daniel/tensorflow-test/models/tutorials/embedding/word2vec_ops.so: undefined symbol: _ZN10tensorflow7strings6StrCatB5cxx11ERKNS0_8AlphaNumE

josh11b:
I don't know about this stuff, maybe @keveman can help you.

nealwu:
Do I have the wrong Josh? Was looking for Josh L.

@keveman let me know if you have any ideas here!

nealwu:
@vincenttsai2015 and @BudickDa: Try using -D_GLIBCXX_USE_CXX11_ABI=0 in your g++ command and let me know if that fixes it.

shark-S:
" -D_GLIBCXX_USE_CXX11_ABI=0 " works on my system.

nealwu:
@shark-S And you were running into this same problem before adding that flag?

vincenttsai2015:
@nealwu It works! Thank you very much!!

BudickDa:
I added the flag, works now. Thank you!

nealwu:
Glad it works for you!

Issue #748, opened 2016-12-13
[Feature request] To add the ability to use own data during the training a RNN network for translation by the translate.py script


b0noI:
This is the copy of the original issue from: https://github.com/tensorflow/tensorflow/issues/6071

Right now the script translate.py does not allow to use own data for training. At the moment it works only with the data, that the script downloads from the Internet (en/fr). Also there is not any documentation on the format of the input data that it expects in the input. In order to have the ability to train the model with own data 2 changes need to be made to the script:

- to document the format of the input data;
- enable the ability to pass an own input file path via the script arguments, the file path should be used for training instead of downloading the data from the internet.

b0noI:
fixed in #806

PR #806, opened 2016-12-23
Ability to train the translation model on arbitrary input sources.

b0noI:
Fix for #748 

With these changes, anyone should be able to train a translation model with any input data, and not only for Eng/Fre pair. Training the model with own data now can be done by executing:

```
python translate.py
  --data_dir="." \
  --from_train_data=train.a \
  --to_train_data=train.b \
  --from_dev_data=train.a \
  --to_dev_data=train.b \
  --steps_per_checkpoint=10
```

The detailed example of how the logic can be used: https://blog.kovalevskyi.com/rnn-based-chatbot-for-6-hours-b847d2d92c43#.2cbxvmtxx 

Also, the pull-request includes fixes for the code that are required for the code to be workable. Current implementation is not working with the latest TF r0.12 (so no one at the moment can actually use it as is). For example, current implementation have references to the  ```tf.contrib.legacy_seq2seq``` wich is no longer exists starting TF 0.12.

### DIFF
--- a/tutorials/rnn/translate/data_utils.py
+++ b/tutorials/rnn/translate/data_utils.py
@@ -239,8 +239,8 @@ def data_to_token_ids(data_path, target_path, vocabulary_path,
           counter += 1
           if counter % 100000 == 0:
             print("  tokenizing line %d" % counter)
-          token_ids = sentence_to_token_ids(tf.compat.as_bytes(line), vocab,
-                                            tokenizer, normalize_digits)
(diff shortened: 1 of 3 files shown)


nealwu:
You'll probably want @lukaszkaiser to take a look at this, although he has a new version of this model in progress already.

lukaszkaiser:
Yes, we'll be replacing it, but let's merge this first. Thanks for the update!

b0noI:
@lukaszkaiser thank you! I have also closed the issue (#748)

PR #834, opened 2016-12-30
Update build_image_data.py

BoyuanJiang:
_bytes_feature excepted class bytes, but in python3 in class str,so use tf.compat.as_bytes for compatibility !


nealwu:
Thanks!

Issue #839, opened 2017-01-02
im2txt Caption class comparison fails in Python 3.5


srome:
### Caption class comparison fails in Python 3.5 because Python 2's __cmp__ is deprecated in 3+

@cshallue 

I was trying to run #466 on Python 3.5 and I was getting the following error:

> Traceback (most recent call last):
>   File "test.py", line 84, in <module>
>     tf.app.run()
>   File "/data/venvs/tensorflow/lib/python3.5/site-packages/tensorflow/python/platform/app.py", line 43, in run
>     sys.exit(main(sys.argv[:1] + flags_passthrough))
>   File "test.py", line 74, in main
>     captions = generator.beam_search(sess, image)
>   File "/media/scott/BigHD/Code/tensorflow_ex/im2txt/inference_utils/caption_generator.py", line 193, in beam_search
>     partial_captions.push(beam)
>   File "/media/scott/BigHD/Code/tensorflow_ex/im2txt/inference_utils/caption_generator.py", line 77, in push
>     heapq.heappush(self._data, x)
> TypeError: unorderable types: Caption() < Caption()

The Caption class only implements __cmp__, but needs __lt__ and __eq__ for Python 3. It's a simple addition that would help compatibility between the versions. I'm not sure if you're planning on supporting Python 3, but everything else worked out of the box except this when I ran the run_inference.py file from the repo.

tae-jun:
Hi @srome, I made #842 for this. Please check!

Thanks 😄

srome:
@tae-jun it looks good to me. I did a similar fix on my own local.

PR #842, opened 2017-01-03
im2txt: make python3 compatible adding lt and eq

tae-jun:
Fix of #839.

Class `Caption` is not comparable on Python3, because `cmp` is deprecated.

So I added `lt` and `eq`.

By the way, `xrange`is renamed to `range` on Python3, so I had to change them as well.

How do we make it both 2 and 3 compatible?

### DIFF
--- a/im2txt/im2txt/inference_utils/caption_generator.py
+++ b/im2txt/im2txt/inference_utils/caption_generator.py
@@ -54,6 +54,16 @@ class Caption(object):
       return -1
     else:
       return 1
+  
+  # for version 3 compatibility (__cmp__ is deprecated)
+  def __lt__(self, other):
+    assert isinstance(other, Caption)
+    return self.score < other.score
+  
+  # also for version 3 compatibility
+  def __eq__(self, other):
+    assert isinstance(other, Caption)
+    return self.score == other.score
 
 
 class TopN(object):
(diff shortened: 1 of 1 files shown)


tae-jun:
Agreed to CLA

srome:
Notifying @cshallue

im2txt/im2txt/inference_utils/caption_generator.py
```diff
@@ -54,6 +54,16 @@ def __cmp__(self, other):
       return -1
     else:
       return 1
+  
+  # for version 3 compatibility (__cmp__ is deprecated)
```
cshallue: Thanks for the contribution!
Please capitalize the first word and add a period in this comment.
Also, please say "Python 3" instead of "version 3".

im2txt/im2txt/inference_utils/caption_generator.py
```diff
@@ -54,6 +54,16 @@ def __cmp__(self, other):
...
+  def __lt__(self, other):
+    assert isinstance(other, Caption)
+    return self.score < other.score
+  
+  # also for version 3 compatibility
```
cshallue: Ditto

tae-jun:
@cshallue Thanks for the review! 👍 

I fixed what you mentioned. Please check! 😄

cshallue:
Fixes #839

Issue #817, opened 2016-12-26
cifar10_input.py calls tf.strided_slice() with 3 arguments, 4 needed.


dweekly:
## Please let us know which model this issue is about (specify the top-level directory)

models/tutorials/image/cifar10/cifar10_input.py:87

  File "/mnt/st12tb/models/tutorials/image/cifar10/cifar10_input.py", line 87, in read_cifar10
    tf.strided_slice(record_bytes, [0], [label_bytes]), tf.int32)

models/tutorials/image/cifar10/cifar10_input.py:93

  File "/mnt/st12tb/models/tutorials/image/cifar10/cifar10_input.py", line 93, in read_cifar10
    [label_bytes + image_bytes]),

In both cases, running the script, e.g. with "python cifar10_train.py" on TensorFlow v0.12 yields

TypeError: strided_slice() takes at least 4 arguments (3 given)

bobbyAtSperry:
I am having the same problem. See my description [here ](http://stackoverflow.com/questions/41361766/tensorflow-cifar10-tutorial-fails) on stackoverflow.

szy1900:
I have the same problem

debajyotidatta:
This problem also happens while running, ptb_word_lm.py.

Version of tensorflow : '0.12.0'

Entire traceback: 

---------------------------------------------------------------------------
TypeError                                 Traceback (most recent call last)
<ipython-input-7-704a07048b28> in <module>()
      6 
      7     with tf.name_scope("Train"):
----> 8       train_input = PTBInput(config=config, data=train_data, name="TrainInput")
      9       with tf.variable_scope("Model", reuse=None, initializer=initializer):
     10         m = PTBModel(is_training=True, config=config, input_=train_input)

<ipython-input-4-5a92301ad419> in __init__(self, config, data, name)
     92     self.epoch_size = ((len(data) // batch_size) - 1) // num_steps
     93     self.input_data, self.targets = reader.ptb_producer(
---> 94         data, batch_size, num_steps, name=name)
     95 
     96 

/home/ubuntu/Dropbox/xai/Machine175/AttentionModels/rnn/ptb/reader.pyc in ptb_producer(raw_data, batch_size, num_steps, name)
    115     i = tf.train.range_input_producer(epoch_size, shuffle=False).dequeue()
    116     x = tf.strided_slice(data, [0, i * num_steps],
--> 117                          [batch_size, (i + 1) * num_steps])
    118     x.set_shape([batch_size, num_steps])
[...]

PR #849, opened 2017-01-05
Make cifar10 tutorial work with 0.12.x, fixing #817.

omo:
This change:

 * Adds missing parameters to strided_slice() calls, and
 * Replaces deprecated summary APIs with new ones.

Note that both of these changes are upward compatible
to future versions.


omo:
Just signed the CLA, and I'm a Google employee anyway.

omo:
Updated to match what #769 is doing.

omo:
@googlebot signed.

omo:
Found that 0.12 distribution contains its own models/ and we can use it. So it's reasonable for this `models` repo to track tf mater. Closing.

PR #854, opened 2017-01-06
Added README to tutorials/ recommending to the user to install TensorFlow from source

nealwu:
The tutorial models currently have code that is not compatible with the r0.12 release of TensorFlow due to breaking API changes.


tutorials/README.md
```diff
@@ -0,0 +1,3 @@
+# Tutorial Models
+
+This repository contains models referenced to from the [TensorFlow tutorials](https://www.tensorflow.org/tutorials/). We recommend [installing TensorFlow from source](https://www.tensorflow.org/get_started/os_setup#installing_from_sources) using the current master version rather than the r0.12 release before running these models.
```
martinwicke: Would the nightly binaries work? If so, we should recommend those.

nealwu: Aha, didn't know we had those. I think they should, assuming they're actually nightly. :)

I'll change the instructions to that instead.

Issue #831, opened 2016-12-29
tutorials/embedding/BUILD file has unsuitable py_binary -> cc_library dependencies


fnl:
## tutorials/embedding/BUILD

The BUILD file in `tutorials/embedding` is probably ill-defined:

After a previous [discussion](https://groups.google.com/forum/#!topic/bazel-discuss/QrdrQpYiyCA) with @damienmg on the bazel mailing list (thank you, Damien), it seems like [`py_binary` rules](https://www.bazel.io/versions/master/docs/be/python.html#py_binary) cannot have dependencies on a `cc_library` rule, but the BUILD file does define deps of the two Python binaries `word2vec` and `word2vec_optimized` on the CC library `word2vec_kernels`.
This leads to the following errors when trying to build the whole model/package:

```
$ bazel build //models/tutorials/embedding:package
ERROR: ~/src/tensorflow/models/tutorials/embedding/BUILD:32:12: in deps attribute of py_binary rule //models/tutorials/embedding:word2vec: '//models/tutorials/embedding:word2vec_kernels' does not have mandatory provider 'py'.
ERROR: ~/src/tensorflow/models/tutorials/embedding/BUILD:46:12: in deps attribute of py_binary rule //models/tutorials/embedding:word2vec_optimized: '//models/tutorials/embedding:word2vec_kernels' does not have mandatory provider 'py'.
ERROR: Analysis of target '//models/tutorials/embedding:package' failed; build aborted.
```

The same errors occur when trying to build either of the faulty Python binaries alone (i.e., the targets `word2vec` or `word2vec_optimized`).

nealwu:
Hi Florian, did you get a chance to try out the new instructions I submitted for compiling the ops on your own? You can check them out in the [embedding README](https://github.com/tensorflow/models/tree/master/tutorials/embedding).

In the meantime, you're right that the BUILD file isn't currently workable. I'll go ahead and delete it for now in order to avoid confusion. Going to close this issue as a result.

PR #855, opened 2017-01-06
Deleted embedding/BUILD which is no longer working

nealwu:
See https://github.com/tensorflow/models/issues/831


martinwicke:
Is there a replacement?

nealwu:
This is the replacement at the moment: https://github.com/tensorflow/models/blob/master/tutorials/embedding/README.md.

Unfortunately, trying to build this directory with bazel runs into a number of complications because the BUILD file depends on `tf_gen_op_wrapper_py`, which is no longer in this repository.

martinwicke:
I see. No problem.

PR #861, opened 2017-01-07
Update to run with TensorFlow v0.12.1

Queequeg92:
1. Update to new tensor summary API.
2. Replace strided_slice with slice to avoid errors caused by different number of default arguments of strided_slice in various TensorFlow versions.


wookayin:
For `master`, we also need to change `tf.mul` to `tf.multiply`.

nealwu:

PR #885, opened 2017-01-11
ensure output directory exists

Henry-E:
The neural programmer model fails the first time it's run, if the output directory folder does not already exist. In this case "../model" does not exist and the function fails because the mkdir function doesn't appear to create parent folders. 
Error: 
tensorflow.python.framework.errors_impl.NotFoundError: ../model//modeltemp/


nealwu:
Thanks!

PR #889, opened 2017-01-11
Fix calculations of examples per second and seconds per batch in inception model.

yuefengz:


yuefengz:
A batch is split for multiple GPUs. So I am closing this PR.

jmchen-g:
This is only true if the multiple GPUs are used to compute the same batch... So your PR is right for the situation when there is no grouping. To make this right for all situations, you need to check how many replicas are really computed regardless of how many GPUs are being used. Thanks.

Issue #893, opened 2017-01-13
cifar10_eval.py issue: cannot infer Tensor's rank


zhisong:
I ran a unmodified version of cifar10_eval.py, here is the error message

Traceback (most recent call last):
  File "cifar10_eval.py", line 157, in <module>
    tf.app.run()
  File "/home/zhisong/tensorflow/tensorflow/python/platform/app.py", line 44, in run
    _sys.exit(main(_sys.argv[:1] + flags_passthrough))
  File "cifar10_eval.py", line 153, in main
    evaluate()
  File "cifar10_eval.py", line 121, in evaluate
    images, labels = cifar10.inputs(eval_data=eval_data)
  File "/home/zhisong/tensorflow/models/tutorials/image/cifar10/cifar10.py", line 183, in inputs
    batch_size=FLAGS.batch_size)
  File "/home/zhisong/tensorflow/models/tutorials/image/cifar10/cifar10_input.py", line 257, in inputs
    shuffle=False)
  File "/home/zhisong/tensorflow/models/tutorials/image/cifar10/cifar10_input.py", line 132, in _generate_image_and_label_batch
    capacity=min_queue_examples + 3 * batch_size)
  File "/home/zhisong/tensorflow/tensorflow/python/training/input.py", line 872, in batch
    name=name)
  File "/home/zhisong/tensorflow/tensorflow/python/training/input.py", line 655, in _batch
    shapes = _shapes([tensor_list], shapes, enqueue_many)
  File "/home/zhisong/tensorflow/tensorflow/python/training/input.py", line 598, in _shapes
    raise ValueError("Cannot infer Tensor's rank: %s" % tl[i])
ValueError: Cannot infer Tensor's rank: Tensor("Cast:0", dtype=int32)

By comparing to distorted_inputs() in cifar10_input.py, I suspect the following lines are missing in inputs()

float_image.set_shape([height, width, 3])
read_input.label.set_shape([1])

After adding the above two lines, the script works just fine

jart:
Thank you for bringing this to our attention. It looks like @nealwu forgot to update the `inputs()` function in cl/141805344.

PR #895, opened 2017-01-13
Added shape to cifar10_input.py

nealwu:
Adding to the function `inputs` in order to match the function `distorted_inputs`.

Fixes https://github.com/tensorflow/models/issues/893


jart:
I recommend putting `Fixes #893` on the third line of the commit message so it gets closed automatically.

nealwu:
Thanks Justine!

PR #900, opened 2017-01-14
Update `concat` ops to `concat_v2` in slim

samjabrahams:


nealwu:
Thanks! This should be covered by https://github.com/tensorflow/models/pull/1063 and https://github.com/tensorflow/models/pull/1053, so I'm going to close.

PR #906, opened 2017-01-16
modified the cifar10 scripts that works with tensorflow r0.12

coil398:
cifar10 scripts don't work with tensorflow r0.12.
That's why I modified the code a bit by changing the name of the functions.
And I also resolved the deprecated.


wookayin:
Looks it can be superceded by #909 as it duplicates and is missing multigpu training version

nealwu:

Issue #901, opened 2017-01-14
cifar10_multi_gpu_train.py


weigei123:
I just used git to download the codes, and tensorflow was installed from nightly build. Then I compiled tensorflow from source.
```
# python cifar10_multi_gpu_train.py --num_gpus=4
I tensorflow/stream_executor/dso_loader.cc:125] successfully opened CUDA library libcublas.so.8.0 locally
I tensorflow/stream_executor/dso_loader.cc:125] successfully opened CUDA library libcudnn.so.5 locally
I tensorflow/stream_executor/dso_loader.cc:125] successfully opened CUDA library libcufft.so.8.0 locally
I tensorflow/stream_executor/dso_loader.cc:125] successfully opened CUDA library libcuda.so.1 locally
I tensorflow/stream_executor/dso_loader.cc:125] successfully opened CUDA library libcurand.so.8.0 locally
Filling queue with 20000 CIFAR images before starting to train. This will take a few minutes.
WARNING:tensorflow:From /home/***/Downloads/models/tutorials/image/cifar10/cifar10_input.py:135: image_summary (from tensorflow.python.ops.logging_ops) is deprecated and will be removed after 2016-11-30.
Instructions for updating:
Please switch to tf.summary.image. Note that tf.summary.image uses the node name instead of the tag. This means that TensorFlow will automatically de-duplicate summary names based on the scope they are created in. Also, the max_images argument was renamed to max_outputs.
Traceback (most recent call last):
  File "cifar10_multi_gpu_train.py", line 273, in <module>
    tf.app.run()
  File "/usr/local/lib/python2.7/dist-packages/tensorflow/python/platform/app.py", line 44, in run
    _sys.exit(main(_sys.argv[:1] + flags_passthrough))
  File "cifar10_multi_gpu_train.py", line 269, in main
    train()
  File "cifar10_multi_gpu_train.py", line 171, in train
    loss = tower_loss(scope)
  File "cifar10_multi_gpu_train.py", line 78, in tower_loss
    logits = cifar10.inference(images)
  File "/home/***/Downloads/models/tutorials/image/cifar10/cifar10.py", line 207, in inference
    wd=0.0)
  File "/home/***/Downloads/models/tutorials/image/cifar10/cifar10.py", line 137, in _variable_with_weight_decay
    weight_decay = tf.mul(tf.nn.l2_loss(var), wd, name='weight_loss')
AttributeError: 'module' object has no attribute 'mul'
```

jart:
[...]

PR #911, opened 2017-01-17
Fix bugs and API usage on cifar10 and cifar10_multi_gpu_train

wookayin:
Make the cifar10 example (especially, multi_gpu_train) work in the recent master version. This PR includes two updates:

- Replace `tf.mul` (which was removed) with `tf.multiply` (#901).
- Correct the usage of variable_scope around the part multi-gpu model construction, so that "reuse" on the variable scope becomes not leaky (tensorflow/tensorflow#6220).

### DIFF
--- a/tutorials/image/cifar10/cifar10_multi_gpu_train.py
+++ b/tutorials/image/cifar10/cifar10_multi_gpu_train.py
@@ -162,25 +162,26 @@ def train():
 
     # Calculate the gradients for each model tower.
     tower_grads = []
-    for i in xrange(FLAGS.num_gpus):
-      with tf.device('/gpu:%d' % i):
-        with tf.name_scope('%s_%d' % (cifar10.TOWER_NAME, i)) as scope:
-          # Calculate the loss for one tower of the CIFAR model. This function
-          # constructs the entire CIFAR model but shares the variables across
-          # all towers.
-          loss = tower_loss(scope)
+    with tf.variable_scope(tf.get_variable_scope()):
+      for i in xrange(FLAGS.num_gpus):
+        with tf.device('/gpu:%d' % i):
+          with tf.name_scope('%s_%d' % (cifar10.TOWER_NAME, i)) as scope:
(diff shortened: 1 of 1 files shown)


wookayin:
Any updates?

wookayin:
Looks like it has to be rebased after #948. /cc @nealwu

nealwu:
@shlens might have a better idea on this change. I'm not sure about the variable_scope.

tutorials/image/cifar10/cifar10_multi_gpu_train.py
```diff
@@ -162,13 +162,14 @@ def train():
...
+      for i in xrange(FLAGS.num_gpus):
+        with tf.device('/gpu:%d' % i):
+          with tf.name_scope('%s_%d' % (cifar10.TOWER_NAME, i)) as scope:
+            # Calculate the loss for one tower of the CIFAR model. This function
+            # constructs the entire CIFAR model but shares the variables across
+            # all towers.
+            loss = tower_loss(scope)
 
           # Reuse variables for the next tower.
           tf.get_variable_scope().reuse_variables()
```
nealwu: Hi @wookayin, I read the issue you linked to and understand why we want the extra `variable_scope` here. However, why is it that this line and the several lines below it weren't also indented further?

wookayin: Hi @nealwu, I will explain. The new operations created in `tower_loss()` (for each tower) should be definitely scoped with the proper prefix (e.g. `TOWER_0`). For the rest of operations, I thought they are not relevant to *op creation*, so basically they will be not affected by namespace.

However, I agree with you that I will have to include those lines in the scope. It agrees to the previous version of code, and we can get rid of any potential side-effects (e.g. new op might be created due to the future changes). It will look more natural. I will update the changeset. Thanks!

wookayin:
@nealwu I updated the PR as requested and rebased onto the current upstream master.

nealwu:
Thanks @wookayin!

PR #934, opened 2017-01-20
Force new instance creation in MultiRNNCell (See also CL 145094809)

nealwu:
Credit to @ebrevdo for the original change

### DIFF
--- a/tutorials/rnn/translate/seq2seq_model.py
+++ b/tutorials/rnn/translate/seq2seq_model.py
@@ -114,12 +114,14 @@ class Seq2SeqModel(object):
       softmax_loss_function = sampled_loss
 
     # Create the internal multi-layer cell for our RNN.
-    single_cell = tf.nn.rnn_cell.GRUCell(size)
+    def single_cell():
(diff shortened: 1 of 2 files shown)


tutorials/rnn/translate/seq2seq_model.py
```diff
@@ -114,12 +114,14 @@ def sampled_loss(labels, inputs):
       softmax_loss_function = sampled_loss
 
     # Create the internal multi-layer cell for our RNN.
-    single_cell = tf.nn.rnn_cell.GRUCell(size)
+    def single_cell():
+      return tf.nn.rnn_cell.GRUCell(size)
```
nealwu: I noticed this code is out of date (`tf.nn.rnn_cell` rather than `tf.contrib.rnn`), and it appears the source of this is this commit: https://github.com/tensorflow/models/commit/c902a86785052291083682c4b833747e7ea52597#diff-4826da5b71f89d05b5b53ded2bd2e138.

Will add another PR to fix some of these changes.

PR #639, opened 2016-11-14
Rename deprecated `op_scope`, `variable_op_scope`

soupault:

PR #936, opened 2017-01-20
Update inception model based on tf API changes

lilao:
replace tf.op_scope with tf.name_scope and tf.variable_op_scope with tf.variable_scope; fix the order of arguments for tf.concat; replace tf.mul with tf.multiply.

### DIFF
--- a/inception/inception/inception_model.py
+++ b/inception/inception/inception_model.py
@@ -147,8 +147,8 @@ def _activation_summary(x):
   # Remove 'tower_[0-9]/' from the name in case this is a multi-GPU training
   # session. This helps the clarity of presentation on tensorboard.
   tensor_name = re.sub('%s_[0-9]*/' % TOWER_NAME, '', x.op.name)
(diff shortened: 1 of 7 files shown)


soupault:
I'm just curious why hadn't my PR been taken into consideration - https://github.com/tensorflow/models/pull/639 ...

martinwicke:
I hadn't seen it. Support for models/ is best effort, and it does happen that things slip. 

I am sorry.

soupault:
@martinwicke thanks for the feedback, Martin! I understand that it (probably) was a busy time preparing TF 1.0 release.
As long as you are here, may I ask you if there are any plans to add CI testing for models?

martinwicke:
We will make a new repo with models we maintain. models/ will stick around as a place for researchers to open-source their code, but we do not have the bandwidth to provide proper support (incl. CI) for it for now. Of course, any project owner may set up their own testing, and some have done so.

PR #944, opened 2017-01-23
Duplicate vocabulary name

SenJia:
The to vocabulary shares the same name with the from one, can not be loaded when decoding.


SenJia:
I signed it!

nealwu:
Closing because it looks like this change has already been merged separately. Thanks!

PR #945, opened 2017-01-24
Replaced tf.sub with tf.subtract, tf.mul with tf.multiply

deepankar1994:
This prevents issues with new versions of TF.
Minor self explanatory changes.


deepankar1994:
I signed it!

nealwu:
Thanks! This should be covered by https://github.com/tensorflow/models/pull/1063, so I'm going to close.

PR #948, opened 2017-01-24
Updated the cifar10 model to be compatible with the latest version of TensorFlow

nealwu:


nealwu:
@annarev, when should we update `tf.concat_v2` here to just `tf.concat`?

wookayin:
Just question: Is`tf.contrib.deprecated.XXX_summary` preferred over the new `tf.summary.XXX`?

nealwu:
@wookayin I believe it may provide more compatibility at the moment. @gunan might know better?

annarev:
@nealwu, you can update tf.concat_v2 --> tf.concat any time. it is in the v1.0.0-rc0 release

nealwu:
@annarev OK, will do.

nealwu:
@JavidPack Thanks for pointing this out. Going to update these calls to the newest version.

PR #959, opened 2017-01-26
Changes to Resnet for TF 1.0 compatibility

yaroslavvb:


yaroslavvb:
@panyx0718

resnet/cifar_input.py
```diff
@@ -101,7 +109,7 @@ def build_input(dataset, data_path, batch_size, mode):
   labels = tf.reshape(labels, [batch_size, 1])
   indices = tf.reshape(tf.range(0, batch_size, 1), [batch_size, 1])
   labels = tf.sparse_to_dense(
-      tf.concat(1, [indices, labels]),
```
panyx0718: Can you use tf.concat_v2? without defining your own concat?

resnet/resnet_model.py
```diff
@@ -89,21 +90,21 @@ def _build_model(self):
     with tf.variable_scope('unit_1_0'):
       x = res_func(x, filters[0], filters[1], self._stride_arr(strides[0]),
                    activate_before_residual[0])
-    for i in xrange(1, self.hps.num_residual_units):
+    for i in six.moves.range(1, self.hps.num_residual_units):
```
panyx0718: Is this change necessary?

resnet/cifar_input.py
yaroslavvb: if you don't care about being able to run it in previous version of TF, I can change it to that

panyx0718: Let's not worry about backward compatibility

resnet/resnet_model.py
yaroslavvb: I added it for compatibility with Python3 (we only use Python3 at OpenAI)

panyx0718: I see. Thanks

yaroslavvb: could change it to `range` instead, since the loop is small

resnet/cifar_input.py
yaroslavvb: OK, I'll change it to concat_v2, although that one is also getting deprecated:

https://github.com/tensorflow/tensorflow/releases/tag/v1.0.0-alpha
Deprecated tf.concat operator. Please switch to use tf.concat_v2 for now.
In the Beta release, we will update tf.concat to match argument order of
tf.concat_v2.

PR #964, opened 2017-01-28
fix the  bug running cifar10 example on tensorflow 1.0 or master branch

fayeshine:
for the master branch tensorflow, there is no `tf.image_summary` and will cause a bug,
so this need to be change to `tf.summary.image`, since tensorflow 1.0 use this and will be a stable api, update all these kind of api changes (also a tf.mul -> tf.multiply)


fayeshine:
@annarev

fayeshine:
seems like have been solved through https://github.com/tensorflow/models/commit/e2ecda2cdd83c76b0968eca93fb8d5c5be1716b0, close this one

PR #990, opened 2017-02-07
Removed deprecated op

theSparta:
Remove the deprecated `scalar_summary` and use `summary.scalar` instead. 

The current program gets the following warning:
> WARNING:tensorflow:
build_graph.: scalar_summary (from tensorflow.python.ops.logging_ops) is deprecated and will be removed after 2016-11-30.
Instructions for updating:
Please switch to tf.summary.scalar. Note that tf.summary.scalar uses the node name instead of the tag. This means that TensorFlow will automatically de-duplicate summary names based on the scope they are created in. Also, passing a tensor or list of tags to a scalar summary op is no longer supported.


theSparta:
`I signed it!`

nealwu:
Thanks!

Issue #677, opened 2016-11-21
Slim model deployment not using variables_device.


vonclites:
When using train_image_classifier, I noticed that my first GPU was being assigned the weights variables, and that all other GPUs had to read the weights from the first GPU.

In model_deploy's create_clones function, it appears as though the device for the model variables is being set correctly...
```
with slim.arg_scope([slim.model_variable, slim.variable],
                      device=config.variables_device()):
    # Create clones.
    for i in range(0, config.num_clones):
      with tf.name_scope(config.clone_scope(i)) as clone_scope:
        clone_device = config.clone_device(i)
        with tf.device(clone_device):
          with tf.variable_scope(tf.get_variable_scope(),
                                 reuse=True if i > 0 else None):
            outputs = model_fn(*args, **kwargs)
```

However, this arg_scope seems to be overwritten by how nets_factory's get_network_fn works:
```
  arg_scope = arg_scopes_map[name](weight_decay=weight_decay)
  func = networks_map[name]
  @functools.wraps(func)
  def network_fn(images):
    with slim.arg_scope(arg_scope):
      return func(images, num_classes, is_training=is_training)
```

I'm not entirely sure how or why this is happening, because I can't find a way to easily inspect the arg_scope stack when outside of the arg_scope module.  Nevertheless, it seems the network_fn operates within its own arg_scope, such as vgg_arg_scope, completely outside of any previously declared arg_scopes.  In other words, the network_fn arg_scope  is not contained within the {slim.model_variable: 'device': '/device:CPU:0'} arg_scope.

I'm pretty sure the issue is with how get_network_fn works because I adapted my own, trimmed down version of train_image_classifier which bypasses the whole get_network_fn method.

```
# In my training script...
network_fn = model.vgg

def clone_fn(batch_queue):
        images, labels = batch_queue.dequeue()
        logits, end_points = network_fn(images)

# In model.py...
def vgg(images):
  with slim.arg_scope(vgg_arg_scope()):
    # ...create model...
```
which works as expected.

junshi15:
I have the same issue. My workaround is like this:

  def network_fn(images):
    with slim.arg_scope(arg_scope):
        with slim.arg_scope([slim.model_variable, slim.variable],
                      device="hard-coded device name")
        return func(images, num_classes, is_training=is_training)

This however is not a fix.

vonclites:
I fixed the issue with https://github.com/tensorflow/models/pull/787 but I'm having problems getting the PR merged in.

[...]

PR #992, opened 2017-02-07
Fixed model_deploy to correctly assign weights to variables_device

vonclites:
The arg_scope of the network_fn was being determined when get_network_fn was called. Thus, the network_fn 'escaped' back to the closure (I hope I'm using that term right?) in which it was originally created, ignoring the arg-scope stack created afterwards by the clone. When the first clone called the network_fn, the weights were put on the default device. So for two GPUs (and two clones), the weights would be assigned to first clone / first GPU, and the second clone would have to read the weights from the first clone's memory.

I simply moved the call to create the model's arg_scope from the outer get_network_fn to the inner network_fn. This way, when network_fn gets called, the model's arg_scope is properly nested within the arg_scopes created by the clones.

Edit: Related to this issue: https://github.com/tensorflow/models/issues/677

### DIFF
--- a/slim/nets/nets_factory.py
+++ b/slim/nets/nets_factory.py
@@ -97,10 +97,10 @@ def get_network_fn(name, num_classes, weight_decay=0.0, is_training=False):
   """
   if name not in networks_map:
     raise ValueError('Name of network unknown %s' % name)
-  arg_scope = arg_scopes_map[name](weight_decay=weight_decay)
   func = networks_map[name]
   @functools.wraps(func)
   def network_fn(images):
+    arg_scope = arg_scopes_map[name](weight_decay=weight_decay)
     with slim.arg_scope(arg_scope):
       return func(images, num_classes, is_training=is_training)
   if hasattr(func, 'default_image_size'):
--- a/slim/nets/nets_factory_test.py
+++ b/slim/nets/nets_factory_test.py
@@ -25,6 +25,7 @@ from nets import nets_factory
 
 slim = tf.contrib.slim
 
+
 class NetworksTest(tf.test.TestCase):
 
   def testGetNetworkFn(self):
--- a/slim/nets/nets_factory_test.py
+++ b/slim/nets/nets_factory_test.py
@@ -19,11 +19,11 @@ from __future__ import absolute_import
 from __future__ import division
 from __future__ import print_function
 
-
 import tensorflow as tf
 
 from nets import nets_factory
 
+slim = tf.contrib.slim
 
 class NetworksTest(tf.test.TestCase):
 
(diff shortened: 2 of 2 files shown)


junshi15:
Does this still work if there are multiple nodes and the variables_device is /job:ps/task:0/device:cpu:0?

vonclites:
I forgot about the distributed case.  I don't have a cluster. I can't easily check for sure. But I also can't think of any reason why it wouldn't, especially if the config.variables_device() is working correctly elsewhere.
For example, if you are using
```
with tf.device(config.variables_device()):
    global_step = slim.create_global_step()
```
Then it should behave no differently than that.  It'd be awesome if you could check and see!

junshi15:
I just checked on a distributed setup. Your code worked. Thanks. 

But I still don't understand why the device scope defined in create_clone does not take effects in network_fn in the origin code.

vonclites:
In the original code,
```python
def get_network_fn(...)
...
    arg_scope = arg_scopes_map[name](weight_decay=weight_decay)
...
    def network_fn(images):
```
```arg_scope``` is established on a call to ```get_network_fn```

In my code,
```python
def get_network_fn(...)
...
    def network_fn(images):
        arg_scope = arg_scopes_map[name](weight_decay=weight_decay)
``` 
```arg_scope``` is not established until a call to ```network_fn```.

The order of events in slim's train_image_classifier.py and eval_image_classifier.py is such that first ```get_network_fn``` is called, then ```create_clones``` is called.

In ```create_clones```, even though the call to ```network_fn``` seems to occur within a device scope, as soon as we re-enter the arg_scope established in ```get_network_fn```, we kind of 'go back in time' -- we enter a scope that predates, and therefore does not contain the arguments of, any scopes declared after the call to ```get_network_fn```.

However, if we move ```arg_scope``` to inside ```network_fn```, that arg_scope is not actually generated until ```network_fn``` is called.  And, since ```network_fn``` gets called inside the device scope created by ```create_clones```, arg_scope inherits the device arguments.

junshi15:
@vonclites Thanks for your explanation. It makes sense.

nealwu:
@sguada @nathansilberman could you take a look at this? This seems like a very beneficial and very short change.

sguada:
Thanks for the fix. Could you add test to model_deploy.py for this case?

nealwu:
Re-assigning to @vonclites

nealwu:
Hey @vonclites, is this something you could take care of (adding a test to https://github.com/tensorflow/models/blob/master/slim/deployment/model_deploy_test.py)? Would be nice to get this change in.

vonclites:
Sorry. Got caught up with grad school things...

I'm not sure if model_deploy_test.py is the correct place to test this.  The issue was not with model_deploy, it was with https://github.com/tensorflow/models/blob/master/slim/nets/nets_factory.py

I added a test to nets_factory_test.py.  I wasn't sure how to gracefully test this.

nealwu:
Nice. I verified that the test fails without the change and passes with it. I had to add a WORKSPACE file to models/slim though in order to get the test running with bazel.

PR #1007, opened 2017-02-10
add convert CASIA webface dataset

astorfi:


astorfi:
Converting CASIA added!

PR #1019, opened 2017-02-15
[slim] Fix a scope name typo in Inception_v1.py

tae-jun:
The average pool is used but named 'MaxPool_0a_7x7'.
Change it to 'AvgPool_0a_7x7'.


nealwu:
Thanks!

PR #1025, opened 2017-02-16
Updated concat_v2 to concat for 1.0 compatibility

shahparth123:
Updated concat_v2 to concat for version 1.0 compatibility for breaking changes introduced in version 1.0
"tf.concat now takes arguments in reversed order and with different keywords. In particular we now match NumPy order as tf.concat(values, axis, name)"


nealwu:
Thanks!

PR #1040, opened 2017-02-20
Fixed concat order using tf_upgrade.py

aselle:


nealwu:
Are there other places that need to be upgraded?

aselle:
probably yes. I didn't have time to do an exhaustive search, but we probably just want to try running the upgrade script on all of models (and checking it carefullyh by hand)

nealwu:
Yeah, I ended up doing that for slim here: https://github.com/tensorflow/models/pull/1053

PR #1045, opened 2017-02-21
Fix division changing dtype to float in python3

ScottSWu:
Running the compression encoder example in python3 causes the error:
`TypeError: Expected an input array of integer or boolean data type`

`int_codes` is divided by 2, but in python3 this converts it from an `int8` to `float64` array.

I am running the gpu version of tensorflow on Windows 10.


nealwu:
Thanks!

PR #1049, opened 2017-02-21
Make slim a python package.

MarvinTeichmann:
This patch adds the file `__init__.py` in the directory `models/slim`. This makes the folder slim a python packages. This allows to cleanly integrate slim model code as part of your project.


nealwu:
Thanks!

PR #1054, opened 2017-02-22
Inception and slim adapted to TensorFlow1.0

vsymbol:
Several functions are modified such that the Inception code can be executed under the environment of TensorFlow version 1.0, e.g., tf.op_scope->tf.name_scope, tf.image_summary->tf.summary.image, etc..


nealwu:
Thanks! This should be covered by https://github.com/tensorflow/models/pull/1063 and https://github.com/tensorflow/models/pull/1053, so I'm going to close it. I'll add a separate PR to upgrade all of the `op_scope` calls to use `name_scope` instead.

nealwu:
Here's the new PR: https://github.com/tensorflow/models/pull/1168

PR #1058, opened 2017-02-23
Removed external dependencies for autoencoder models

jo8hua:
For the autoencoder models,

- Xavier initialization was being implemented from scratch while TF has the capability to do this. I changed the model to use the built in functionality. 

- Numpy was being used to generate random normal variables. I instead used TF and removed the import. 

- I was unable to run the models without making the modification to the import (I am not sure how they were working prior)

- The same performance is reached for all auto encoder models 👍

### DIFF
--- a/autoencoder/AdditiveGaussianNoiseAutoencoderRunner.py
+++ b/autoencoder/AdditiveGaussianNoiseAutoencoderRunner.py
@@ -4,7 +4,7 @@ import sklearn.preprocessing as prep
 import tensorflow as tf
 from tensorflow.examples.tutorials.mnist import input_data
 
-from autoencoder.autoencoder_models.DenoisingAutoencoder import AdditiveGaussianNoiseAutoencoder
+from autoencoder_models.DenoisingAutoencoder import AdditiveGaussianNoiseAutoencoder
 
 mnist = input_data.read_data_sets('MNIST_data', one_hot = True)
 
--- a/autoencoder/AutoencoderRunner.py
+++ b/autoencoder/AutoencoderRunner.py
@@ -4,7 +4,7 @@ import sklearn.preprocessing as prep
 import tensorflow as tf
 from tensorflow.examples.tutorials.mnist import input_data
 
-from autoencoder.autoencoder_models.Autoencoder import Autoencoder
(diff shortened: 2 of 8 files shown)


nealwu:
Can you check this code again and make sure it still works on TensorFlow 1.0? From what I can tell, [`tf.random_normal`](https://www.tensorflow.org/api_docs/python/tf/random_normal) should fail if you pass it a `size` parameter.

jo8hua:
You are correct. I apologize for the oversight, but I believe that I have found another problem when checking it more thoroughly.

When you call the VariationalAutoencoder generate() function, the following error is produced:

`File "VariationalAutoencoderRunner.py", line 35, in <module>
    autoencoder.generate()
File "models/autoencoder/autoencoder_models/VariationalAutoencoder.py", line 59, in generate
    hidden = np.random.normal(size=self.weights["b1"])
File "mtrand.pyx", line 1903, in mtrand.RandomState.normal (numpy/random/mtrand/mtrand.c:18471)
File "mtrand.pyx", line 234, in mtrand.cont2_array_sc (numpy/random/mtrand/mtrand.c:3232)
TypeError: expected sequence object with len >= 0 or a single integer`

This is fixed by calling the existing numpy.random.normal() function correctly or with the TensorFlow variant that I introduced self.sess.run(tf.random_normal([1, self.n_hidden])). However, after fixing this we have 

`InvalidArgumentError (see above for traceback): You must feed a value for placeholder tensor 'Placeholder' with dtype float [[Node: Placeholder = Placeholder[dtype=DT_FLOAT, shape=[], _device="/job:localhost/replica:0/task:0/cpu:0"]()]]`

I believe that this is because the model is trying to perform `feed_dict={self.z_mean: hidden}` rather than `feed_dict={self.z: hidden}`

Is this a possibility?

nealwu:
Unfortunately I'm not familiar with this model, but on first glance it looks like `self.sess.run()` needs to be given a `feed_dict` parameter in order to provide a value for `self.x`, which is a `tf.placeholder(tf.float32, [None, self.n_input])`.

nealwu:
Looks like we can close this then. Thanks!

PR #1063, opened 2017-02-23
Converted the models repo to TF 1.0 using the upgrade script

nealwu:

PR #1065, opened 2017-02-24
Finalized TF1.0 updates for street model.

weinman:
@nealwu @theraysmith 
  Drops now redundant name parameter in new tf.summary.scalar calls
  Transposes ctc_loss params as in updated API
  Hooks RegisterShape to @ops instead of @tf
Passes Ray's unit tests and runs all confidence tests.


nealwu:
Thanks! This should be covered by https://github.com/tensorflow/models/pull/1063. If there's anything that isn't covered, let me know.

PR #1067, opened 2017-02-24
Upgrade summaries/concat/sub/mul/split in the inception training to 1.0

wenwei202:


wenwei202:
I signed it!

nealwu:

inception/inception/inception_train.py
```diff
@@ -329,9 +329,9 @@ def train(dataset):
     # Start the queue runners.
     tf.train.start_queue_runners(sess=sess)
 
-    summary_writer = tf.train.SummaryWriter(
+    summary_writer = tf.summary.FileWriter(
         FLAGS.train_dir,
-        graph_def=sess.graph.as_graph_def(add_shapes=True))
+        graph=tf.get_default_graph())
```
nealwu: Good find here -- I added this as a separate pull request: https://github.com/tensorflow/models/pull/1167. It should be better to use `sess.graph` over `tf.get_default_graph()` in case someone changes the default graph.

PR #1074, opened 2017-02-26
Update readme.md

ywryoo:
Add case for compiling ops if tensorflow was compiled from source using gcc >= 5.0

### DIFF
--- a/tutorials/embedding/README.md
+++ b/tutorials/embedding/README.md
@@ -26,16 +26,9 @@ TF_INC=$(python -c 'import tensorflow as tf; print(tf.sysconfig.get_include())')
 g++ -std=c++11 -shared word2vec_ops.cc word2vec_kernels.cc -o word2vec_ops.so -fPIC -I $TF_INC -O2 -D_GLIBCXX_USE_CXX11_ABI=0
 ```
 
-If tensorflow was compiled from source using gcc >= 5.0, you don't need to append D_GLIBCXX_USE_CXX11_ABI=0.
-
-```shell
-TF_INC=$(python -c 'import tensorflow as tf; print(tf.sysconfig.get_include())')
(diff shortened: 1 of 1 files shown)


ywryoo:
I have signed it :)

nealwu:
Did that flag break the compilation for you @ywryoo?

nealwu:
I would prefer to add a shorter comment at the end of the parenthetical comment explaining the flag. Something like: "If you compiled TensorFlow from source using g++ 5.0 or later, you won't need the flag."

ywryoo:
Hi @nealwu, thanks for response! The reason I wrote entire code again was for people who are not familiar with English. They tend to copy-and-paste code without reading any sentences. However I agree that one line of comment would be clearer than writing whole code again. Would it be okay to add commit for that change in here? This is my first pull request so I am quiet nervous to make any changes.

nealwu:
Yep you can add another commit here to that effect. Does the command fail if you include the flag in your situation (having compiled TensorFlow from source with g++ >= 5.0)?

ywryoo:
Yes. I was on Ubuntu Desktop 16.04.2 and compiled tensorflow from source for optimization. g++ command run without error, but `python word2vec_optimized.py ...` failed because of the flag. So I wanted to clarify the situation for people who will get same error :)

ywryoo:
I've updated explanation as suggestion.

nealwu:
I'm assuming you didn't use the extra `-cxxopt="-D_GLIBCXX_USE_CXX11_ABI=0"` flag mentioned here: https://www.tensorflow.org/install/install_sources#install_the_pip_package. I have to admit though, it's not very clear on where to add it -- you actually need to include it in the `bazel build` command several lines above. Going to file a documentation bug for this.

In any case, I'll edit your comment so that it works for either situation, and then we can go ahead and merge.

nealwu:
Thanks @ywryoo!

PR #1081, opened 2017-02-28
Simplify and improve fastprep.mk

vmarkovtsev:
There is no need to build Tensorflow from scratch, fastprep successfully links with the Python wrapper from pip and system protobuf v3. Tested this on Ubuntu.

### DIFF
--- a/swivel/fastprep.mk
+++ b/swivel/fastprep.mk
@@ -16,50 +16,16 @@
 # limitations under the License.
 
 
-
 # This makefile builds "fastprep", a faster version of prep.py that can be used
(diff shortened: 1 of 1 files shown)


waterson:
It's been a while since I've worked with this code; it may in fact be that we can get rid of a lot of this extra complexity.  IIRC, one issue was trying to make sure things worked correctly on OS/X, as well. How does that impact that platform?

vmarkovtsev:
It works on macOS.

vmarkovtsev:
I reworked the makefile so that it does not require any Tensorflow at all. It downloads the two needed .proto files, compiles them and uses in the build. Works like a charm on all the platforms.

PR #1089, opened 2017-03-01
Update word2vec.py in tutorials/embedding to Tensorflow v1.0

louis800530:
Updated method calls in line 249 of word2vec.py in tutorials/embedding to be compatible with Tensorflow v1.0


nealwu:

PR #1090, opened 2017-03-01
Fix the infinitite loop in reading files

vmarkovtsev:
+add the docs about num_threads

### DIFF
--- a/swivel/fastprep.cc
+++ b/swivel/fastprep.cc
@@ -79,6 +79,10 @@ Options:
   --window_size <int>
       Specifies the window size for computing co-occurrence stats;
       default 10.
+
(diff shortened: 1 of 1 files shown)


waterson:
Thanks for the fix!

What about just checking fin_.eof(); e.g.,

`for (off_t filepos = start_; !fin_.eof() && filepos < end_; filepos = fin_.tellg())`

I'm assuming that the problem is that tellg returns -1 when we've run off the back of the file? In which case this expresses the intent a bit more clearly...

vmarkovtsev:
Sure, amended.

waterson:
Thanks Vadim!

PR #1106, opened 2017-03-03
Swivel: fastprep: replace pthread with std::thread

vmarkovtsev:
As discussed in https://groups.google.com/d/msgid/swivel-embeddings/c254faca-b35c-4e04-a347-095e4b467300%40googlegroups.com

### DIFF
--- a/swivel/fastprep.cc
+++ b/swivel/fastprep.cc
@@ -25,7 +25,6 @@
 
 #include <assert.h>
 #include <fcntl.h>
-#include <pthread.h>
 #include <stdio.h>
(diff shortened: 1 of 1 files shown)


vmarkovtsev:
Yes, since std::thread is built on top of pthread in Linux.
```
/usr/bin/ld: /tmp/cctm1vc8.ltrans24.ltrans.o: undefined reference to symbol 'pthread_create@@GLIBC_2.2.5'
//lib/x86_64-linux-gnu/libpthread.so.0: error adding symbols: DSO missing from command line
```

PR #1112, opened 2017-03-04
fix #1083

### LINKED ISSUE #1083: cifar10 example should not untar the dataset every time.
As shown in <https://github.com/tensorflow/models/blob/master/tutorials/image/cifar10/cifar10.py#L398>, currently, whenever any of `cifar10_train`, `cifar10_eval`, or `	cifar10_multi_gpu_train` is executed, the dataset file will be untarred unconditionally. This causes that one cannot run eval and train scripts simultaneously, as discussed in #892. 

Originally this was not the case, and the change was made in tensorflow/tensorflow#5145, for some other reason.

zym1010:
this fixes the issue in #1083

### DIFF
--- a/tutorials/image/cifar10/cifar10.py
+++ b/tutorials/image/cifar10/cifar10.py
@@ -394,5 +394,6 @@ def maybe_download_and_extract():
     print()
     statinfo = os.stat(filepath)
     print('Successfully downloaded', filename, statinfo.st_size, 'bytes.')
-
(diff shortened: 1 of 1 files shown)


zym1010:
I signed it!

tatatodd:
@caisq This is the PR related to issue #1083  Thanks!

nealwu:
This looks good, and it worked for me when I ran it. Thanks @zym1010!

PR #1117, opened 2017-03-06
Update slim image classifier cifar10 to Tensorflow v1.0

apollo-time:
Updated method calls in slim image classifier to be compatible with
Tensorflow v1.0


nealwu:
Thanks! This should be covered by https://github.com/tensorflow/models/pull/1053, so I'm going to close. If there's anything that isn't covered, let me know.

PR #1131, opened 2017-03-08
add 'reuse' parameter to RNNCells to make code compatible with tensorflow master code

tongda:
To fix the issue mentioned in [#8191](https://github.com/tensorflow/tensorflow/issues/8191) caused by the [recent commit](https://github.com/tensorflow/tensorflow/commit/54d50ffec8df4f748694632dbe5ebde9971e2c9e) of TensorFlow master branch. As the commit message says:

> Make all RNNCells in tf.contrib.rnn act like tf.layers Layers, but with stricter semantics for no
w:
>
>    1. Upon first use of __call__, the used scope is stored in the cell. The RNNCell tries to create weights in that scope but if some are already set, an error is raised unless the RNNCell was constructed with argument reuse=True.
>
>    2. A subsequent use of __call__ of the same cell instance must be in the same scope.
>       If it is not, an error is raised.

### DIFF
--- a/tutorials/rnn/ptb/ptb_word_lm.py
+++ b/tutorials/rnn/ptb/ptb_word_lm.py
@@ -110,7 +110,7 @@ class PTBModel(object):
     # different than reported in the paper.
     def lstm_cell():
       return tf.contrib.rnn.BasicLSTMCell(
-          size, forget_bias=0.0, state_is_tuple=True)
+          size, forget_bias=0.0, state_is_tuple=True, reuse=tf.get_variable_scope().reuse)
     attn_cell = lstm_cell
     if is_training and config.keep_prob < 1:
       def attn_cell():


nealwu:
@ebrevdo does this change look correct to you?

ebrevdo:
Yes, lgtm

nealwu:
Super. Thanks!

GBLin5566:
Yes, this commit will make error msg.

`__init__() got an unexpected keyword argument 'reuse'`

ebrevdo:
What version of tensorflow are you using?

GBLin5566:
@ebrevdo
My tf version is 1.0, and according to API r1.0 there is no argument 'reuse' in BasicLSTMCell.

ebrevdo:
Ah, the model is based on tf nightlies.

nealwu:
@ebrevdo is still there an issue? Given that this is a tutorial model, we want it to be compatible with TensorFlow 1.0, so if that's not the case with this we should revert.

nealwu:
Going to revert this because we want the tutorial models to be compatible with TensorFlow 1.0, not just master. @tongda, does reverting this change cause serious issues for the model?

ebrevdo:
The model won't work without this change in master.

nealwu:
Do you know any good ways to make it worth for both versions? Given that this is a tutorial model and the vast majority of its users will be on 1.0, having it work on 1.0 is a must. If there's a way to fix it for both 1.0 and master at the same time, we can go for that.

ebrevdo:
I'd prefer this tutorial check the tf version and fail if the wrong range
is installed.  This will be too hard to maintain for multiple versions of
contrib code.  Users of nightlies should have no expectation of
compatibility in contrib with any published code.

nealwu:
Sure, what should that look like? `if not tf.__version__.startswith('1.0')` then raise an error?

nealwu:
Looks like that won't work. Even if you install from source right now, your `tf.__version__` will be `1.0.1`.

ebrevdo:
Yeah, there's a git version baked in somewhere but I'm not sure if we can
check it as an ordinal.  We should check with the release team to ensure
they increment the master version number after each release.

tongda:
How about we add a `try... catch` of `KeyError` during the transitional period of new api?

nealwu:
Not the most elegant option but it works. I'll add a new pull request.

llealgt:
Hello, is any of this related to this error: Trying to share variable rnn/multi_rnn_cell/cell_0/basic_lstm_cell/weights, but specified shape (1024, 2048) and found shape (595, 2048).
Can you take a look, im running this code (from udacity course) : https://github.com/udacity/deep-learning/blob/master/intro-to-rnns/Anna_KaRNNa.ipynb

ebrevdo:
What version of tf are you using?

llealgt:
1.0 , I builded from code pulling from master branch like a week ago.

PR #1134, opened 2017-03-09
Update batch_size to 64, according to hps in readme

vincentschen:


vincentschen:
I signed it!

panyx0718:
It seems a non-bug change and not performance difference, please keep it locally.

PR #1140, opened 2017-03-10
Modified tf.concat and tf.split argument order to get in line with tf…

paulskeie:
… 1.0

### DIFF
--- a/video_prediction/prediction_input.py
+++ b/video_prediction/prediction_input.py
@@ -97,11 +97,11 @@ def build_tfrecord_input(training=True):
       action = tf.reshape(features[action_name], shape=[1, STATE_DIM])
       action_seq.append(action)
 
-  image_seq = tf.concat(0, image_seq)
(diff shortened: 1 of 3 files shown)


paulskeie:
There are some more modifications need as well to make the code work in tf 1.0.  These are in lstm_ops.py. Perhaps @chfinn would like to make them herself.  tf.pack is not there anymore and I think the tf.zeros_initializer might have changed its API too.

pobingwanghai:
@paulskeie I have just tested the video_prediction tutorial under tf1.0 version, it works fine, if you need to run the tutorial right now, you can see it in the new pull request. The only problem still remaining is the logging.flush(), but it can be ignored.

cbfinn:
LGTM, though looks like there are some conflicts that need to be resolved.

cbfinn:
Nevermind, it looks like the code was updated to 1.0 automatically using a script. I suppose we can close this PR now.

nealwu:

Issue #1078, opened 2017-02-28
im2txt - evaluation script - v1.0 compatibility issues- SummaryWriter to FileWriter and Merge All Summaries Depreciation


sh0rtfuse:
SummaryWriter has been changed to FileWriter in v1.0

Line 175 of evaluate.py: `summary_writer = tf.train.SummaryWriter(eval_dir)`

FIX: Change to `tf.summary.FileWriter` seems to work for me

Merge All Summaries Depreciation: #1066

Line 175 of evaluate.py: `summary_op = tf.merge_all_summaries()`

FIX: Change to `tf.contrib.deprecated.merge_all_summaries` OR `tf.summary.merge_all`

hoL1day:
Thanks very much.
I change to `summary_op = tf.contrib.deprecated.merge_all_summaries`
and `summary_writer = tf.summary.FileWriter(eval_dir)`

It works, but it throw out a `TypeError: Fetch argument <function merge_all_summaries at 0x7f7ca33ff488> has invalid type <type 'function'>, must be a string or Tensor. (Can not convert a function into a Tensor or Operation.)`
The whole error output is 
![image](https://cloud.githubusercontent.com/assets/8084465/23388676/9c73bd1c-fd9e-11e6-9fe0-4c246182040a.png)
Thank u~

tatatodd:
Sorry @sh0rtfuse I don't think I understand the issue you're raising.  Can you provide a bit more context into the problem?

sh0rtfuse:
Hi Todd, apologies for not being clear.

Some of the method calls in evaluate.py to Tensorflow are outdated as of v1.0.0. I pointed out two cases above that breaks the build.

Changing the method calls to the supported version fixes the problem.

Thanks!

Raymond

tatatodd:
@sh0rtfuse I see, that makes sense to me now.  Thanks for filing the issue, the diagnosis, and the proposed fix!

Would you like to send a PR to fix the problem?  We definitely welcome contributions!  :)

sh0rtfuse:
Sure! Will do.

shahparth123:
Created modified version of #1086 with removing deprecated function dependency of tf.contrib.deprecated.merge_all_summaries() with  tf.summary.merge_all() in pull request #1141

karmel:
Automatically closing due to lack of recent activity. Since this issue is old at this point, please reopen the issue if it still occurs when tried with the latest version of Tensorflow. Thank you.

PR #1141, opened 2017-03-10
Replaced depricated functions with new function

shahparth123:
Replaced old summery writer function with newer tf.summary.FileWriter function with TF 1.0 compatibility


shahparth123:
Created as a fix for #1078

nealwu:
Thanks! This should be covered by https://github.com/tensorflow/models/pull/1063, so I'm going to close.

PR #1145, opened 2017-03-11
Replaced deprecated tf.*_summary with tf.summary.* and related modifications 

vihari:


nealwu:
Thanks! This should be covered by https://github.com/tensorflow/models/pull/1063 and https://github.com/tensorflow/models/pull/1053, so I'm going to close. If there's anything that isn't covered, let me know.

PR #1149, opened 2017-03-11
replace some functions to make Spatial Transformer Network run in tensorflow 1.0

Childhoo:
1) replace "tf.pack" with "tf.stack"
2) tf.concat..(change the oder)
3) change "tf.batch_matmul" to "tf.matmul"


nealwu:
Thanks! This should be covered by https://github.com/tensorflow/models/pull/1063, so I'm going to close. If there's anything not covered let me know.

PR #1150, opened 2017-03-11
inception_train.py upgrade to TF 1.0

tfboyd:
- Ran 1 GPU with imagenet for a few minutes and training worked as expected.  There are still some deprecation warnings and this works where as it did not before with TF 1.0.


tfboyd:
@nealwu please review and merge

nealwu:
Thanks Toby! This should be covered by https://github.com/tensorflow/models/pull/1063, so I'm closing.

PR #1152, opened 2017-03-12
Changes for video_prediction under tf1.0

pobingwanghai:
Works fine in python3, cuda8, cudnn5.1, tf1.0. Have been tested, sucessfully get the hoped result.
Makes all the functions satisfied for tf1.0 version,including concat, split, summary fucntion and some others.


pobingwanghai:
@googlebot I have signed CLA !

pobingwanghai:
According to the newest pull request, there are some people waiting for the executable version of video_prediction for tf1.0. Hoping for a merge permission. Or you guys might be preparing for some worthwhile changes more suitable and effective for tf1.0. Waiting for a more impeccable model for tf1.0.

paulskeie:
Thanks for sharing this code.  I tried running it using python 2.7.  There was a syntax error in prediction_input.py on line 53.  Look it up and you'll see it at once.  Just some lines of code that was duplicated.  After fixing that, I get a runtime error:
TypeError: Tensors in list passed to 'values' of 'Pack' Op have types [int32, in32, int32, int64] that don't all match.  It seems to be the last dimension of state_shape which has type int64.

pobingwanghai:
@paulskeie I just test the code using python 3.5, it seems like the error occuring because of  the difference between tf python3 version and tf python2 version.

pobingwanghai:
@paulskeie It seems like your error because of my python3 numpy using int64 as default format, you can easily convert int64 to int32 using x.astype(np.int32).

video_prediction/prediction_input.py
```diff
@@ -48,6 +48,16 @@ def build_tfrecord_input(training=True):
     list of tensors corresponding to images, actions, and states. The images
     tensor is 5D, batch x time x height x width x channels. The state and
     action tensors are 3D, batch x time x dimension.
+
+def build_tfrecord_input(training=True):
```
cbfinn: This looks like an error in merging. There shouldn't be two function headers.

video_prediction/prediction_input.py
```diff
@@ -59,7 +69,7 @@ def build_tfrecord_input(training=True):
     filenames = filenames[:index]
   else:
     filenames = filenames[index:]
-  filename_queue = tf.train.string_input_producer(filenames, shuffle=True)
+  filename_queue = tf.train.string_input_producer(filenames, shuffle=False)
```
cbfinn: I think this should stay as True.

video_prediction/prediction_model.py
```diff
@@ -109,8 +109,10 @@ def construct_model(images,
         prev_image = image
 
       # Predicted state is always fed back in
+
       state_action = tf.concat(axis=1, values=[action, current_state])
 
+
```
cbfinn: nit: don't add extra lines

video_prediction/lstm_ops.py
```diff
@@ -38,15 +39,12 @@ def init_state(inputs,
...
+    batch_size = inputs.get_shape().as_list()[0]
     dtype = inputs.dtype
   else:
     inferred_batch_size = 0
     batch_size = 0
-
-  initial_state = state_initializer(
-      tf.stack([batch_size] + state_shape),
-      dtype=dtype)
+  initial_state=tf.Variable(tf.zeros(tf.stack([batch_size] + state_shape)))
```
cbfinn: nit: maintain appropriate style by having spaces around "=".

video_prediction/lstm_ops.py
```diff
@@ -99,7 +97,9 @@ def basic_conv_lstm_cell(inputs,
                             scope='Gates')
 
     # i = input_gate, j = new_input, f = forget_gate, o = output_gate
-    i, j, f, o = tf.split(axis=3, num_or_size_splits=4, value=i_j_f_o)
+
+    i, j, f, o = tf.split(i_j_f_o, 4,3 )
```
cbfinn: Is this change necessary?

video_prediction/prediction_input.py
```diff
@@ -97,8 +107,10 @@ def build_tfrecord_input(training=True):
       action = tf.reshape(features[action_name], shape=[1, STATE_DIM])
       action_seq.append(action)
 
+
```
cbfinn: nit: don't add extra lines

video_prediction/prediction_model.py
```diff
@@ -277,8 +285,8 @@ def cdna_transformation(prev_image, cdna_input, num_masks, color_channels):
   cdna_kerns /= norm_factor
 
   cdna_kerns = tf.tile(cdna_kerns, [1, 1, 1, color_channels, 1])
-  cdna_kerns = tf.split(axis=0, num_or_size_splits=batch_size, value=cdna_kerns)
-  prev_images = tf.split(axis=0, num_or_size_splits=batch_size, value=prev_image)
+  cdna_kerns = tf.split(cdna_kerns, batch_size, 0)
```
cbfinn: I don't think this change is necessary. Remove?

video_prediction/prediction_model.py
```diff
@@ -289,7 +297,9 @@ def cdna_transformation(prev_image, cdna_input, num_masks, color_channels):
     transformed.append(
         tf.nn.depthwise_conv2d(preimg, kernel, [1, 1, 1, 1], 'SAME'))
   transformed = tf.concat(axis=0, values=transformed)
-  transformed = tf.split(axis=3, num_or_size_splits=num_masks, value=transformed)
+
+  transformed = tf.split(transformed, num_masks, 3)
```
cbfinn: This change also doesn't seem necessary.

video_prediction/prediction_train.py
```diff
@@ -150,30 +155,45 @@ def __init__(self,
...
     loss, psnr_all = 0.0, 0.0
+    try:
+        tf.equal(prefix,'train').eval()
+    except:
+        prefixs='initial'
+    else:
+        if tf.equal(prefix,'train').eval():
+            prefixs='train'
+        else:
+            prefixs='val'
```
cbfinn: What is the code above doing and why is it necessary?

nealwu:
It looks like this PR is no longer needed due to https://github.com/tensorflow/models/pull/1063, aside from a few Python 3 compatibility fixes possibly. If you want to separate out the compatibility changes and create a new PR, feel free to send it to me and let me know. In the meantime, I'll close this one.

video_prediction/prediction_train.py
pobingwanghai: I'm testing the newest code, the last version tf.summary.scalar() must get a string a input, a tf.sensor will make error, so I add these code to make the output format same with the old version.

video_prediction/prediction_input.py
pobingwanghai: Yes, it should be True

video_prediction/lstm_ops.py
pobingwanghai: Seems necessary to me, according to the api https://www.tensorflow.org/api_docs/python/tf/split, and if I use the old style it raise IndexError:list index out of range. The tf.__version__ in my conputer is  1.0.0, is it required the mater version from github? Or maybe the api document haven't update yet.

pobingwanghai:
@nealwu @cbfinn According to the transitioning to TF1.0(https://www.tensorflow.org/install/migration),
tf.split
keyword argument split_dim should be renamed to axis
keyword argument num_split should be renamed to num_or_size_splits
arguments have been reordered to tf.split(value, num_or_size_splits, axis=0, num=None, name='split')
The newest update may miss the transition of split function.

video_prediction/lstm_ops.py
cbfinn: The order of named arguments doesn't matter, as long as the name is consistent with the argument passed in. So, these two are equivalent.

cbfinn:
@pobingwanghai 
These updates were handled in https://github.com/tensorflow/models/pull/1063.
Arguments don't need to be reordered if you use name the arguments (see my inline comment).

Feel free to make a new PR with minimal changes to make the code compatible with python 3.

video_prediction/lstm_ops.py
pobingwanghai: Yes,you're right, seems my misunderstanding comes from the code from video_prediction/prediction_train.py line116: actions = tf.split(axis=1, num_or_size_splits=actions.get_shape()[1], value=actions), the IndexError:list index out of range shows because of the num_or_size_splits should be an int not a tf.dimension variable.

tegg89:
@pobingwanghai 
Thanks for sharing your code, but I am still struggling to run codes. Below TypeError message still keep shown up.

`TypeError: Tensors in list passed to 'values' of 'Pack' Op have types [int32, int32, int32, int64] that don't all match.`

I followed all your changed code. After I tested in python 2.7, and python 3.5 with Tensorflow ver. 1.0.1.
By your code transitioning, I could solve some of 'tf.split()' errors, but above `TypeError` message keep shown up. I have found my numpy library version which is in int64 version. How can I solve this problem?

nealwu:
@tegg89 Is this happening for you with the latest models code and TensorFlow 1.0? Let me know which models commit you are on (`git log`).

tegg89:
@nealwu After `git log`, I could check the downloaded models updated on Saturday, March 25, 12:17:16 2017 -0700.

tegg89:
@pobingwanghai @nealwu 
The error occurs at function `init_state` from `lstm_ops.py`. 
I checked `batch_size` is 32 and `state_shape` is [Dimension(32), Dimension(32), 64].
Error message is shown as below.
`TypeError: Tensors in list passed to 'values' of 'Pack' Op have types [int32, int32, int32, int64] that don't all match.`
How can I change the code to run this properly?

video_prediction/lstm_ops.py
tegg89: I also found out that line 47 should be changed to `initial_state = tf.Variable(tf.zeros([batch_size] + state_shape))`

nealwu:
Oh sorry, didn't realize you were running the code from this pull request. @pobingwanghai will have to help you with that.

cbfinn:
@tegg89 Do the changes in https://github.com/tensorflow/models/pull/1326 resolve issues that you are having?

pobingwanghai:
@tegg89 The error means you have to use function .aslist() to make the para pass to tensor is [int32, int32, int32, int64] ,the tf.dimension type isn't satisfied.

pobingwanghai:
@tegg89 The easiest and effective way is updating your tensorflow to master, as @cbfinn said, #1326 will fix most of these error. I recommend you to run the model from master branch, the changes of mine version may have a few conflicts with the recent changes from tensorflow dealing with python3, most of the confict comes from parameter pass, you can see from the error, it requires types [int32, int32, int32, int64] not the [Dimension(32), Dimension(32), 64], after some changes from master branch of tensorflow, it seems these parameter errors has been solved mostly. Hoping to help you.

tegg89:
@nealwu @pobingwanghai @cbfinn 
As I mentioned above, by changing line 47 in `lstm_ops.py` training file runs properly. However, because of my limited hardware environment, I changed `batch_size` from 32 to 16. Result seems like underfitting though. Anyway the training works.

pobingwanghai:
@tegg89 good luck

PR #1164, opened 2017-03-14
Update batch_reader.py

AnupKumarGupta:
Though the values are of  `QUEUE_NUM_BATCH` and `BUCKET_CACHE_BATCH` are same, but since we are assigning value to `_bucket_input_queue` the value passed must be  `BUCKET_CACHE_BATCH`


AnupKumarGupta:
I signed it. You may verify now.

Regards
Anup Kumar Gupta

panyx0718:
quote: "since we are assigning value to _bucket_input_queue the value passed must be BUCKET_CACHE_BATCH"

We cannot determine the value simply based on the variable name.

PR #1198, opened 2017-03-17
Removed external dependencies from autoencoder models

jo8hua:
This is a follow on the pull request #1058. 

- Xavier initialization was being implemented from scratch while TF has the capability to do this. I changed the model to use the built in functionality.

- Numpy was being used to generate random normal variables. I instead used TF and removed the import.

- Minor modifications to the imports were made (I am not sure how they were working prior)

- Printing was edited to be compatible with python3


snurkabill:
looks good

nealwu:
Merging based on @snurkabill's review. Thanks @jo8hua!

PR #1204, opened 2017-03-18
im2text minor documentation fix

talkdirty:
Hey,
I noticed that the images in the im2text Readme.md would not be displayed properly (the raw markdown is displayed instead of the image)
Removing the center tag fixes this.

### DIFF
--- a/im2txt/README.md
+++ b/im2txt/README.md
@@ -37,9 +37,7 @@ Full text available at: http://arxiv.org/abs/1609.06647
 The *Show and Tell* model is a deep neural network that learns how to describe
 the content of images. For example:
 
-<center>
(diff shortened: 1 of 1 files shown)


talkdirty:
I signed it!

nealwu:
Thanks! cc @cshallue

cshallue:
Thanks! Must have been browser-dependent, because the center tags never created any issue that I saw.

nealwu:
Oh that's interesting. What browser/OS are you running? I saw the raw markdown with Chrome on Ubuntu.

cshallue:
That's interesting, perhaps this issue only began in the last couple of days? Also using Chrome on Ubuntu, and I definitely looked at this page last week.

PR #1207, opened 2017-03-18
Make table valid markdown

EmilStenstrom:
Second row had an extra set of dashes, making the markdown invalid and falling back to raw text instead of the intended table.

To see the result, look at the rendered markdown file.


EmilStenstrom:
I signed it!

nealwu:
Thanks!

PR #1208, opened 2017-03-19
correct illegal naming of tf.summary.scalar

robrao:
Made the fix for the issue listed below.

`INFO:tensorflow:Summary name learning rate is illegal; using learning_rate instead.`


robrao:
I signed it!

onecityuni:
I signed it!

PR #1210, opened 2017-03-19
output more meaningful error messages

jtran064:
I use this script to create TFRecord file for other image datasets. Sometimes these datasets have corrupted JPEG files. Printing file names help pinpoint corrupted files.

Also, the exception bubbles up to the thread that is looping over JPEG files and causes the TFRecord shard to be halted prematurely. The thread is also stuck in limbo. Skipping corrupted files allows the thread to continue to execute to completion.


nealwu:
Thanks!

Issue #1206, opened 2017-03-18
skip_thoughts: cannot find combined_encoder


anupamme:
## Please let us know which model this issue is about (specify the top-level directory)

model: skip_thoughts

issue: cannot execute from skip_thoughts import combined_encoder

And I could not find combined_encoder in the entire model code base.

Am I missing something or the code is not complete?

Slaite:
It looks like this was changed to encoder_manager and the README is outdated. I think it should be:

`from skip_thoughts import encoder_manager`
`encoder = encoder_manager.EncoderManager()
encoder.load_model(configuration.model_config(),
                     vocabulary_file=VOCAB_FILE,
                     embedding_matrix_file=EMBEDDING_MATRIX_FILE,
                     checkpoint_path=CHECKPOINT_PATH)`

anupamme:
Yes encoder_manager works.

Should I send a pull request with the change?

cshallue:
@Slaite thanks for picking that up! Fixing.

PR #1214, opened 2017-03-19
Minor fix to skip_thoughts documentation.

upman:
The commands to extract the pretrained skip_thoughts models were a little off.


upman:
I signed it!

Slaite:
Since this is already open, the example encoding script at the end of README also needs to be fixed. See Issue #1206.

Issue #1194, opened 2017-03-17
bazel test ... exec failed


haoran8899:
## Please let us know which model this issue is about (specify the top-level directory)
hi,when i bazel test ...  to build syntaxnet ,error has happend,err information as blow:
In file included from /usr/include/c++/5/algorithm:62:0,
                 from ./dragnn/core/beam.h:4,
                 from dragnn/core/beam_test.cc:1:
/usr/include/c++/5/bits/stl_algo.h: At global scope:
/usr/include/c++/5/bits/stl_algo.h:4718:5: error: 'void std::sort(_RAIter, _RAIter, _Compare) [with _RAIter = __gnu_cxx::__normal_iterator<syntaxnet::dragnn::Beam<syntaxnet::dragnn::{anonymous}::TestTransitionState>::Transition*, std::vector<syntaxnet::dragnn::Beam<syntaxnet::dragnn::{anonymous}::TestTransitionState>::Transition, std::allocator<syntaxnet::dragnn::Beam<syntaxnet::dragnn::{anonymous}::TestTransitionState>::Transition> > >; _Compare = syntaxnet::dragnn::Beam<T>::AdvanceFromPrediction(const float*, int, int) [with T = syntaxnet::dragnn::{anonymous}::TestTransitionState]::<lambda(const syntaxnet::dragnn::Beam<syntaxnet::dragnn::{anonymous}::TestTransitionState>::Transition&, const syntaxnet::dragnn::Beam<syntaxnet::dragnn::{anonymous}::TestTransitionState>::Transition&)>]', declared using local type 'syntaxnet::dragnn::Beam<T>::AdvanceFromPrediction(const float*, int, int) [with T = syntaxnet::dragnn::{anonymous}::TestTransitionState]::<lambda(const syntaxnet::dragnn::Beam<syntaxnet::dragnn::{anonymous}::TestTransitionState>::Transition&, const syntaxnet::dragnn::Beam<syntaxnet::dragnn::{anonymous}::TestTransitionState>::Transition&)>', is used but never defined [-fpermissive]
     sort(_RandomAccessIterator __first, _RandomAccessIterator __last,
     ^

markomernick:
What compiler are you using? Does it support C++11?

haoran8899:
gcc version 5.4.0 20160609 (Ubuntu 5.4.0-6ubuntu1~16.04.4) .
when i update,system tell me gcc is latest version.

markomernick:
[...]

PR #1229, opened 2017-03-21
SyntaxNet does not compile, because isnan is unknown.

danieldk:
SyntaxNet does not compile with gcc version 5.4.0 20160609 (Ubuntu 5.4.0-6ubuntu1~16.04.4). This leads to the error in #1194.

Including `cmath` and using the `std` namespace for `isnan` solves this problem.


danieldk:
I signed it!

PR #1244, opened 2017-03-23
Add Python 3 compatibility in textsum (while maintaining Python 2 compatibility)

studynon:
Add Python 3 compatibility in textsum (while maintaining Python 2 compatibility):
Add the compatibility of the builtin function ‘xrange’ and ‘Queue’. Now you can run the textsum in both Python 2 and Python 3.
‘compat.py’ is the compatibility script, it should be added in the other scripts’ front, as follows:
‘from compat import *’

PS: In Python 3 environment, the prerequisite is ‘pip install six’


nealwu:
I believe what we generally use for Python 2 + 3 compatibility is [six](https://pypi.python.org/pypi/six). For example see https://github.com/tensorflow/models/blob/master/tutorials/image/mnist/convolutional.py.

Could you modify this code to use six instead?

studynon:
@nealwu Thank you for your recommendation! I have fixed it.

nealwu:
I needed to make changes to the code to avoid using compat.py. It also looks like you haven't signed the CLA. I'll create a new pull request and link to it from here.

nealwu:
https://github.com/tensorflow/models/pull/1255

cclauss:
I had not seen this issue when I opened #2105 

Is there general agreement to use [six](https://pythonhosted.org/six/) across all models?

nealwu:
@cclauss Yes, six is generally preferred if the code requires slight differences to support Python 2 and 3.

PR #1247, opened 2017-03-23
README.md warning on extra background class added by build_image_data.py

jtran064:
This cost me many hours of debugging when my network failed to converge when applying `build_image_data.py` to a new dataset. Turns out I didn't read the `build_image_data.py` comments carefully and missed the part about the background class being labeled 0. The [one_hot_encoding](https://github.com/tensorflow/models/blob/master/slim/train_image_classifier.py#L455) in train_image_classifier does not take into account of out-of-range inputs. Therefore in my 3 classes dataset, my labels of `[3 2 2 1]` are converted to these one hot labels:
```
[[0 0 0]
 [0 0 1]
 [0 0 1]
 [0 1 0]]
```
The NN will always classify class 3 as 'wrong'. I fixed this by changing [label_index](https://github.com/tensorflow/models/blob/master/inception/inception/data/build_image_data.py#L375) to 0 so my classes will now numerate `[0,2]`. (alternatively can add `tf.subtract(labels, 1)`). 

Additionally, I think that the background class is unnecessary since the script is referred to quite often in the README as a general data converter to TFRecords. The background class seems to be a relic of the ImageNet dataset that does not generalize well to other types of data. In any case, I hope this addition will emphasize this nuance to future README readers.

Fix shell script to match example.

### DIFF
--- a/inception/README.md
+++ b/inception/README.md
@@ -658,7 +658,6 @@ Each entry of the `TFRecord` is a serialized `tf.Example` protocol buffer. A
 complete list of information contained in the `tf.Example` is described in the
 comments of `build_image_data.py`.
 
-
 To run `build_image_data.py`, you can run the following command line:
 
 ```shell
--- a/inception/README.md
+++ b/inception/README.md
@@ -634,9 +634,10 @@ reside within `$TRAIN_DIR` and `$VALIDATION_DIR` arranged as such:
   $VALIDATION_DIR/cat/cat.JPG
   ...
 ```
-**NOTE**: This script will append an extra background class indexed at 0. Using the
-example above, the corresponding class labels generated from `build_image_data.py`
-will be as follows:
+**NOTE**: This script will append an extra background class indexed at 0, so
+your class labels will range from 0 to num_labels. Using the example above, the
+corresponding class labels generated from `build_image_data.py` will be as
+follows:
 ```shell
 0
 1 dog
--- a/inception/README.md
+++ b/inception/README.md
@@ -24,7 +24,7 @@ model architecture.
 
 ## Description of Code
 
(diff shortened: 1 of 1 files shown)


nealwu:
Thanks! This looks helpful. I added a few additional changes to the text and removed `num_labels` since I don't think that's been defined anywhere (correct me if I'm wrong though).

Could you explain where you got the new values for `train_shards` and `validation_shards`?

jtran064:
Hi Neal, I'm glad it is. Yes, `num_labels` is just used as an example in `build_image_data.py`.

The shards are supposed to reflect the outputs of the script under section 'How to Construct a New Dataset for Retraining'. The example output on lines 694-704 was:
```
  $TRAIN_DIR/train-00000-of-00024
  $TRAIN_DIR/train-00001-of-00024
  ...
  $TRAIN_DIR/train-00023-of-00024

and

  $VALIDATION_DIR/validation-00000-of-00008
  $VALIDATION_DIR/validation-00001-of-00008
  ...
  $VALIDATION_DIR/validation-00007-of-00008
```

This would only be true if `--train_shards=24` and `--validation_shards=8`. The current example has different values:

```shell
# location to where to save the TFRecord data.
OUTPUT_DIRECTORY=$HOME/my-custom-data/

# build the preprocessing script.
bazel build inception/build_image_data

# convert the data.
bazel-bin/inception/build_image_data \
  --train_directory="${TRAIN_DIR}" \
  --validation_directory="${VALIDATION_DIR}" \
  --output_directory="${OUTPUT_DIRECTORY}" \
  --labels_file="${LABELS_FILE}" \
  --train_shards=128 \
  --validation_shards=24 \
  --num_threads=8
```

nealwu:
I see. Looks to me more like lines 694-704 are the ones that should be changed. I'll edit those and then merge this in.

Issue #1199, opened 2017-03-17
Variational autoencoder generate() function broken


jo8hua:
In the `autoencoders/autoencoder_models`directory, the generate() function of the `VariationalAutoencoder` is broken. The following error results from using it:

<img width="875" alt="screen shot 2017-03-17 at 5 32 10 pm" src="https://cloud.githubusercontent.com/assets/10055613/24063845/c10d5362-0b37-11e7-9d0d-accbe81e9c6f.png">

This is fixed by calling `numpy.random.normal(size=[1,self.n_hidden])` instead, but this creates the new error:

<img width="998" alt="screen shot 2017-03-17 at 5 40 49 pm" src="https://cloud.githubusercontent.com/assets/10055613/24064085/eeed8fb2-0b38-11e7-86a6-4849680fa96d.png">

I believe that this comes from the fact that `self.x` is required to generate `self.reconstruction` from  `self.z_mean` when `self.z` is not. 

In terms of a solution, I have seen other implementations of VAEs where they have a "generation" function that feeds a normal random variable into the `self.z` parameter,  but the math does not quite work out since `self.z` is shifted and scaled from the normal distribution. I ran some summary statistics between `self.z` and `self.z_mean` and they were very close under the MNIST data. I am mainly wanting thoughts on this.

concretevitamin:
@snurkabill

snurkabill:
Hello, 

can you, please, provide here full usage of your autoencoder? I can't run any code here now, so I would like to at least see it. 

Thanks

jo8hua:
It is easiest to repeat scripting. I did the following from the `autoencoder` directory.

<img width="707" alt="screen shot 2017-03-17 at 9 56 48 pm" src="https://cloud.githubusercontent.com/assets/10055613/24068084/c0e0f8b0-0b5c-11e7-8aca-30cc643575dd.png">

snurkabill:
Ok. if I recall it correctly, you must first perform at least one partial fit on variational autoencoder and then you can generate new sequences. Please, try that and let me know if that works. Maybe that's corrupted usecase and there really is mistake.

jo8hua:
I just attempted to do what you said and it resulted in the same errors. I agree that the result of calling the function prior to training should be non sensical, but I don't think that it should result in a syntax error.

snurkabill:
[...]

PR #1253, opened 2017-03-25
Variational Autoencoder generate() function fixed (z fed in rather th…

jo8hua:
…an z_mean)

A follow on to a previous update to remove external dependencies from the autoencoder models, and to solve [the following issue](https://github.com/tensorflow/models/issues/1199#issuecomment-289022533). Below is a sample image created from the output of the generate function. 

img1


nealwu:
@snurkabill could you review?

snurkabill:
ok seems good to me

nealwu:
OK great. Merging.

jo8hua:
Great! Thanks.

PR #1255, opened 2017-03-25
Make textsum compatible with Python 3

nealwu:


nealwu:
@studynon, can you confirm that this works for you?

nealwu:
@jhseu is this the correct way to import Queue from six?

studynon:
@nealwu , The way of your using six.moves.queue will make an error. If you want to use six.moves, the code should be 
"from six.moves import queue as Queue".
This works for me.

nealwu:
Thanks @studynon. Fixed.

PR #1258, opened 2017-03-25
change shell scripts for better compatibility with macOS and tf docker image

algal:
I was going through the flower retraining example.

I needed to make these changes to the scripts, and to the sample code in the README, for them to work with the tensorflow/tensorflow:1.0.1-devel docker image, since it includes `curl` but not `wget`.

They would also be needed on stock Mac, which includes curl but not wget. This is especially relevant for the script intended to run only on a Mac.


algal:
I signed it!

nealwu:
Thanks @algal!

PR #1281, opened 2017-03-31
Python 3.5 Compatibility changes

arsenious:
Hello everyone

I have used python 2to3 to make changes to the Neural GPU model. After the automated changes were made, there were some issues, which I manually fixed. These issues were mainly centered around the integer division of indices. True division( using / ) would yield a float which would cause run time errors. I have made changes to operations acting on array indices, to ensure that they remain as integers. (i.e. changed / to //)

Please let me know if any further changes are needed and if any issues are seen. Thanks

### DIFF
--- a/neural_gpu/neural_gpu.py
+++ b/neural_gpu/neural_gpu.py
@@ -186,13 +186,13 @@ def quantize_weights_op(quant_scale, max_value):
 def autoenc_quantize(x, nbits, nmaps, do_training, layers=1):
   """Autoencoder into nbits vectors of bits, using noise and sigmoids."""
   enc_x = tf.reshape(x, [-1, nmaps])
-  for i in xrange(layers - 1):
+  for i in range(layers - 1):
     enc_x = tf.layers.dense(enc_x, nmaps, name="autoenc_%d" % i)
   enc_x = tf.layers.dense(enc_x, nbits, name="autoenc_%d" % (layers - 1))
   noise = tf.truncated_normal(tf.shape(enc_x), stddev=2.0)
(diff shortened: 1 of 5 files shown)


arsenious:
I signed the CLA

nealwu:
Hi @arsenious, see the discussion [here](https://github.com/tensorflow/models/pull/1244) for how we typically add Python 3 compatibility to TensorFlow code. In particular, you may want to add these three imports at the top of each file:

arsenious:
Hi

So are these lines used in conjunction with 2to3 script? Because I have already used the script. I thought these lines were used in a Python 2.7 code to make it compatible with 3.5. Would just using these lines be enough and not require the need of using 2to3?

neural_gpu/neural_gpu.py
```diff
@@ -498,8 +498,10 @@ def attention_query(query, attn_v):
             return tf.reduce_sum(encoder_outputs * tf.expand_dims(mask, 2), 1)
 
           with tf.variable_scope("decoder"):
-            def decoder_loop_fn((state, prev_cell_out, _), (cell_inp, cur_tgt)):
+            def decoder_loop_fn(xxx_todo_changeme, xxx_todo_changeme1):
```
nealwu: Make sure to fix this.

nealwu:
I believe adding those (as well as using `six`) should do the job for the most part. You can then just keep `xrange` everywhere. However you'll still need to modify the `print` statements to be function calls instead.

nealwu:
Closing this now. Feel free to submit a new pull request with changes using `six`.

PR #1285, opened 2017-04-01
Make comment in ptb_word_lm compatible with TF 1.0

GBLin5566:
Fix the comment in ptb_word_lm that still using old version API.


GBLin5566:
I signed the CLA !

nealwu:
Thanks!

PR #1292, opened 2017-04-03
add binary to file reading

lababidi:


nealwu:
Hi @lababidi, can you explain why this change is needed?

lababidi:
@nealwu without the `'b'` in the file opening, the file is attempted to be read as utf8 when it's really binary. this is needed particularly in `python3`. Most of the `models` code base was written in python2 it seems, but I digress.

nealwu:
I see. And this maintains compatibility with Python 2, correct?

lababidi:
@nealwu Yes the `b` is in both python 3 and 2

I've also added another file that I've modified. It seems to have been ditched completely, and got it working regardless of python version.

nealwu:
@lababidi I ended up making some additional changes to get the script to work for me. Let me know if the changes work for you. If so, happy to merge.

lababidi:
@nealwu everything looks good! thanks for your contribution

nealwu:
Great, thanks!

PR #1318, opened 2017-04-09
Revert merge #1292

r2d4:
This broke the bazel build of `inception/download_and_preprocess_flowers`
The way that this script is written doesn't actually allow it to be ran
outside bazel, and it might be difficult to support running it standalone and 
in the bazel context.

It should be ran using

```
bazel build inception/download_and_preprocess_flowers

bazel-bin/inception/download_and_preprocess_flowers "${FLOWERS_DATA_DIR}"
```

ref #1292 cc @nealwu @lababidi

### DIFF
--- a/inception/inception/data/build_image_data.py
+++ b/inception/inception/data/build_image_data.py
@@ -198,7 +198,7 @@ def _process_image(filename, coder):
     width: integer, image width in pixels.
   """
   # Read the image file.
-  with tf.gfile.FastGFile(filename, 'r') as f:
+  with tf.gfile.FastGFile(filename, 'rb') as f:
     image_data = f.read()
 
   # Convert any PNG to JPEG's for consistency.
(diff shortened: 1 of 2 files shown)


lababidi:
Odd, @nealwu and I got this script working without bazel. please include how it broke bazel.

r2d4:
I was just following the instructions here @lababidi  - 

https://github.com/tensorflow/models/tree/master/inception#how-to-retrain-a-trained-model-on-the-flowers-data

nealwu:
@r2d4 is right; this script should be run with bazel. I missed that originally.

r2d4:
@nealwu thanks for the cleanup!

nealwu:
FYI this is the correct link for running this via bazel: https://github.com/tensorflow/models/tree/master/inception#how-to-fine-tune-a-pre-trained-model-on-a-new-task

PR #1326, opened 2017-04-10
Fixes for compatibility with TF 1.0 and python 3.

alexlee-gk:
@cbfinn
In `prediction.train.py`, explicitly pass in prefix name to Model instance because summary names can no longer take in placeholder strings (at least for now, see [tensorflow/models#6603](https://github.com/tensorflow/tensorflow/issues/6603)).
In `lstm_ops.py`, directly use `inferred_batch_size` because the initializer cannot take shapes made of tensors.


alexlee-gk:
I signed it!

nealwu:
Adding @cbfinn for review.

PR #1330, opened 2017-04-11
set -e on bash scripts

r2d4:
These scripts should exit immediately if a command exits with a non-zero status.


nealwu:
@sguada + @nathansilberman

drpngx:
@r2d4 did you check that the scripts work?

r2d4:
Yes @drpngx

PR #1335, opened 2017-04-12
Checkpoint saving while training inception model

rajmiglani:
Number of steps after which checkpoints are saved in inception_train.py


rajmiglani:
I signed it.

nealwu:
Doesn't seem like a necessary change.

PR #1346, opened 2017-04-14
Converted cPickle to _pickle for python3 and solved a jpg reading issue

Steven-N-Hart:
I have been using python 3.5 and had some issues I sorted out.  THought I would add them back so others wouldn't have the same Issues I have.

For example< I kept getting the `TypeError: 'jpg' has type <class 'str'>, but expected one of: ((<class 'bytes'>,),)` error.  All I had to do was to prefix the jpg in  `slim/datasets/download_and_convert_flowers.py`.

The other change was to change `slim/datasets/download_and_convert_cifar10.py` where it imports cPickle as this was bombing in python 3.5.

The rest are just spacing changes for PEP8 compliance.


Steven-N-Hart:
I signed it!

nealwu:
Hi @Steven-N-Hart, could you remove all the indentation changes you made? Also, does this maintain compatibility with other versions (especially Python 2)?

Steven-N-Hart:
OK.  I added a try/catch block to the cPickle import, tested on python2.7, and removed the spaces.

slim/datasets/download_and_convert_flowers.py
```diff
@@ -50,163 +50,167 @@
...
+                        sys.stdout.write('\r>> Converting image %d/%d shard %d' % (
+                            i + 1, len(filenames), shard_id))
+                        sys.stdout.flush()
 
-            # Read the filename:
-            image_data = tf.gfile.FastGFile(filenames[i], 'r').read()
-            height, width = image_reader.read_image_dims(sess, image_data)
+                        # Read the filename:
+                        image_data = tf.gfile.FastGFile(
+                            filenames[i], 'r').read()
```
RvDijk: I also encountered a problem here, I got the error:
`UnicodeDecodeError: 'utf-8' codec can't decode byte 0xff in position 0: invalid start of byte`
Changing 'r' to 'rb' then gave me the same problem as you:
`TypeError: 'jpg' has type <class 'str'>, but expected one of: ((<class 'bytes'>,),)`
Your adjustment below then fixed the problem for me. 
I just wanted to note this in case that others might encounter the same problem.

happenzZ: line 141 =>  image_data = tf.gfile.FastGFile(filenames[i], 'rb').read()
line 151 =>  example = dataset_utils.image_to_tfexample(image_data, b'jpg', height, width, class_id)

This solves the problem.

karmel:
@sguada @Steven-N-Hart Is this PR still relevant? Can you merge or close?

tfboyd:

PR #1350, opened 2017-04-15
Removed rmsprop_momentum flag, use --momentum flag

r2d4:
The flag description for the momentum flag states that it is 
> The momentum for the MomentumOptimizer and RMSPropOptimizer

however its not actually used in the RMSPropOptimizer.  Instead, a separate
`--rmsprop_momentum` flag was used.  This deletes that flag for
correctness of the `--momentum` flag description and simplicity.  
It was not referenced anywhere else in the repo.


nealwu:
Thanks!

PR #1351, opened 2017-04-15
Use slim.filter_variables()

r2d4:
Using reg_search=False to use re.match to mirror the 'startswith' logic
being replaced.

cc @sguada

### DIFF
--- a/slim/train_image_classifier.py
+++ b/slim/train_image_classifier.py
@@ -26,6 +26,8 @@ from deployment import model_deploy
 from nets import nets_factory
 from preprocessing import preprocessing_factory
 
+import re
+
 slim = tf.contrib.slim
 
 tf.app.flags.DEFINE_string(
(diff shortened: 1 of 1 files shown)


nealwu:
@sguada + @nathansilberman

slim/train_image_classifier.py
```diff
@@ -343,19 +345,11 @@ def _get_init_fn():
...
-  for var in slim.get_model_variables():
-    excluded = False
-    for exclusion in exclusions:
-      if var.op.name.startswith(exclusion):
-        excluded = True
-        break
-    if not excluded:
-      variables_to_restore.append(var)
+  var_list = slim.get_model_variables()
+  variables_to_restore = slim.filter_variables(var_list=var_list, exclude_patterns=exclusions, reg_search=False)
```
r2d4: On a little closer inspection of the API, I could probably use `slim.get_variables_to_restore(exclude=exclusions)` and avoid the call to `slim.get_model_variables()` as well as wrapping the exclusions with `re.escape`

PR #1359, opened 2017-04-17
added OSX build instructions to street model

sstirlin:


sstirlin:
I signed it!

nealwu:
Thanks! Added another comment as well to match https://github.com/tensorflow/models/blob/master/tutorials/embedding/README.md.

PR #1372, opened 2017-04-20
Swivel: swappable optimizer and TensorBoard

vmarkovtsev:
Currently, optimizer is hardcoded: Adagrad. It corresponds to the paper. However, there are newer optimizers available, e.g. Adadelta or RMSProp and it is fun to see how they influence the training.

For some reason summaries are present in the code but `FileWriter` is not used - I fix it. Besides, I added the most frequent 10k visualization in TB's embedding projector.

### DIFF
--- a/swivel/swivel.py
+++ b/swivel/swivel.py
@@ -92,6 +92,8 @@ flags.DEFINE_float('per_process_gpu_memory_fraction', 0,
                    'Fraction of GPU memory to use, 0 means allow_growth')
 flags.DEFINE_integer('num_gpus', 0,
                      'Number of GPUs to use, 0 means all available')
+flags.DEFINE_string('logs', '',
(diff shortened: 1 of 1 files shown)


vmarkovtsev:
Any feedback on this @waterson ? I can fix the conflicts with the distributed code if needed.

waterson:
There are a couple of things going on in this PR... do you think they are still applicable?

Also, the primary purpose of this model overall is to make it easy for other people to reproduce and build upon the experimental results that we wrote about. It's really not meant to be a general purpose, flexible, production-ready tool. So, I'm not super excited about adding features or options that obscure the main ideas that we wrote about.

vmarkovtsev:
Forked as https://github.com/src-d/tensorflow-swivel

PR #1377, opened 2017-04-21
 要提交的变更：

ShawnXiha:
fix error: open too many files when runing prep.py
open with utf-8 decode for Chinese texts


nealwu:
@ShawnXiha can you explain what this pull request is trying to achieve? Many of the code changes don't seem appropriate.

PR #1381, opened 2017-04-21
xrange replaced with range

Egor-Krivov:
'xrange' was used in multiple places. In this setting it provided no measurable gain in terms of performance or memory, but broke python3 compatibility. It was replaced with 'range'


nealwu:
Thanks!

wookayin:
Why don't we just use `six.xrangr` instead?

nealwu:
@wookayin sure, that also works. `xrange` is generally better in Python 2 than `range` for performance reasons, but given that the relevant numbers here were 4, 7, and 3, I didn't think it was a big deal in this case. If you want to submit a pull request to add `six` though, I'll merge it.

PR #1384, opened 2017-04-22
Remove unused function

r2d4:
Variable summaries and the learning rate are added elsewhere in the
code.  A quick search also shows that this function is never called.

Learning rate gets added to summary at https://github.com/r2d4/models/blob/fd6e2d29577247a79a8010ded72d78811882db2d/slim/train_image_classifier.py#L510

Variables get added to summary at
https://github.com/r2d4/models/blob/fd6e2d29577247a79a8010ded72d78811882db2d/slim/train_image_classifier.py#L491-L492


nealwu:
Thanks!

PR #1388, opened 2017-04-23
Changed to gen.__next__( )

Higgins2718:
next() does not work in Python3. Attempting to train returns error. Changing all instances of gen.next( ) to gen.__next__( ) fixed the problem.

### DIFF
--- a/textsum/batch_reader.py
+++ b/textsum/batch_reader.py
@@ -133,7 +133,7 @@ class Batcher(object):
     pad_id = self._vocab.WordToId(data.PAD_TOKEN)
     input_gen = self._TextGenerator(data.ExampleGen(self._data_path))
     while True:
-      (article, abstract) = input_gen.next()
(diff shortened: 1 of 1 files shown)


Higgins2718:
@googlebot I signed it! Please verify

nealwu:
Hi @Higgins2718, unfortunately this change will break Python 2. Does `six.next(iter)` work for you? If so, could you update the pull request to use that instead?

18811376770:
@nealwu  could we run the model on python2 thus avoiding to make change on code?

nealwu:
@18811376770 in general, our ideal situation is for the model to work properly on both Python 2 and Python 3. A number of our models currently are likely only compatible with Python 2, but we will accept pull requests to add Python 3 compatibility (assuming it maintains Python 2 compatibility).

nealwu:
Closing in favor of https://github.com/tensorflow/models/pull/1463

PR #1393, opened 2017-04-24
A script to generate a text file with FSNS URLs.

alexgorban:
Currently the FSNS dataset is distributed as a description how to generate URLs to download files. This change adds a simple script to generate a text file with all URLs which is compatible with aria2c download manager, which can download all files in parallel and supports multiple retries and continue an aborted download modes.


alexgorban:
@theraysmith Please take a look.

PR #1401, opened 2017-04-25
inception: Added protocol flag when running distributed

shamoya:
Default is TensorFlow default of 'grpc' communication protocol.
If TensorFlow was complied with Verbs support 'grpc+verbs' can be
used to accelerate the tensor passing communication.

@drpngx @junshi15 @bkovalev


drpngx:
Nice, thanks!

PR #1405, opened 2017-04-25
Add VGG to models using slim.

shamoya:
This work is almost identical to the Inception
V3 (inception folder) in terms of its python wrappers,
Imagenet dataset and wrapper and its use of slim to
build the model. the slim folder contains the VGG-16
model (can be easily expanded to the other version of VGG)
and the vgg contains the programs for running the Imagenet
2012 set (evaluating and training with single/multiple nodes).

@drpngx @bkovalev @junshi15 @jhseu @anfeng 
This is needed mostly for performance evaluation of a large network (~ x10 number of parameters than Inception V3) and comparison between different communication protocols.


junshi15:
Thanks for the PR. We have been using vgg model in model/slim until r1.0, which breaks the sync replica. Do you think we can fix synchronous training in model/slim so that we can run a lot more models?

shamoya:
Hi @junshi15,
When you refer to VGG in model/slim you mean the code in tensorflow repo?
Anyway, I'm not familiar with r1.0 breaking the sync replica mode, Inception-V3 distributed from this repo  works well right?
This VGG is very similar to the Inception-V3 code in this repo, and the distributed wrapper code is the same with the only diff in the model that is being used.
Indeed with the default grpc (TCP) the results are very bad and with your RDMA code (BTW great work and thank you for that) the results are much better (CPU in the VGG TCP case is the root cause).
Some of the results we collected https://docs.google.com/spreadsheets/d/1UIqmp5sKwPnxWpZkJ1rCcGrJz-Iqr4_1NZdZdRqJ7N8/pubhtml

junshi15:
@shamoya 
Thanks for sharing the test results. It is in-line with our internal benchmark.
 
By "model/slim", I meant this: https://github.com/tensorflow/models/tree/master/slim There is a VGG model in it. We were using it with r0.12 because it supports both multi-node and multi-GPU. But the sync replica has not been working any more since r1.0.

Your contribution follows https://github.com/tensorflow/models/blob/master/inception/inception/. I don't see how can you specify multiple GPUs with imagenet_distributed_train.py, as shown in your spreadsheet? Did you launch multiple instances of that train script in a single box, each using a single GPU?

drpngx:
Nice results!

It looks like this is a fork of current code. Could we modify the current source instead?

@sguada @nealwu any concerns here? (Verb is the newly introduced RDMA)

drpngx:
@poxvoculi @tfboyd check out the results docs above on verbs RDMA vs grpc.

tfboyd:
VGG16 is definitely hard to scale in a distributed setup.  Which GPU are you using?  I would like to compare and share our internal result with similar hardware if I have access internally.

poxvoculi:
It's great to see this improvement.

shamoya:
@junshi15 - Yes we ran multiple train script instances separated launched each with distinct GPU with  CUDA_VISIBLE_DEVICES.
It's not optimal but increases the network use which is what we wanted to check. 

@tfboyd - We used Tesla K40 and we are now checking with P100 PCIe.
As you see the gRPC TCP results are really bad. a single K40 can train 7 images/sec in our setup.
We looking into it as well and what we see until now is CPU core at 100% at the send side struggling mainly with memory allocation/free.

@drpngx - In the first  patch I had all of this code inside inception/inception. reusing the slim python modules and adding flag  with model=inception|vgg to imagenet_*. 
Do u prefer it that way ?

sguada:
I think this introduce a lot of code duplication.

drpngx:
@shamoya I'll let @sguada speak to that. In general, it's probably easier to maintain if we just have one place.

sguada:
@drpngx  do not merge.

sguada:
Not clear to me what is the contribution of this PR, could you separate the real contribution from the code duplication?

sguada:
Ideally it should be a small addition to https://github.com/tensorflow/models/tree/master/slim

tfboyd:
@shamoya   Agree on grpc.  We have some code coming in the next week that speeds up distributed with some slightly different techniques but we are coming to the same conclusion that many already had that we need to focus more effort on grpc layer.  I hope it is clear that we appreciate these contributions and the great documents and thank you for listing the GPU.

shamoya:
No problem @sguada, the PR purpose was to try and find out how this should be done.
I'll go and do some homework and see what's the best place to have VGG distributed synchronized.
As @junshi15 maybe Just need to have a small fix for the sync replicas feature in https://github.com/tensorflow/models/tree/master/slim

shamoya:
Thank you @tfboyd, I guess it's best to have multiple communication mechanisms anyway.
I guess the work mentioned in grpc is https://github.com/tensorflow/tensorflow/pull/7466 ?
If so I'm waiting for it to merge to recheck the gRPC TCP results.

PR #1406, opened 2017-04-25
DOC: document bash use for resnet

stsievert:
I ran into a bug under zsh with `--train_data=cifar10/data_batch*`.


nealwu:
Hi @stsievert, this is not actually an executable script, so it doesn't make sense to add `#!/bin/bash` here. Can you modify the change so that it just provides a brief note to the user to use bash instead?

stsievert:
Done. I also added a note that the script contains the expected output (which also got me at first) and made it more clear when commands were being run.

nealwu:
Thanks!

PR #1425, opened 2017-04-28
Update lm_1b_eval.py (Python 3 compatibility)

edouardfouche:
Solve Python 3 compatibility:
- replacing xrange to range
- add .decode() after f.read() to get strings instead of bytes when reading text file

### DIFF
--- a/lm_1b/lm_1b_eval.py
+++ b/lm_1b/lm_1b_eval.py
@@ -17,6 +17,7 @@
 """
 import os
 import sys
+import six
 
 import numpy as np
 import tensorflow as tf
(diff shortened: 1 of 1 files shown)


edouardfouche:
I signed it!

nealwu:
Hi @edouardfouche, could you update this to use [six](https://pythonhosted.org/six/) in order to maintain compatibility with both Python 2 and Python 3?

edouardfouche:
OK. It was already working for both actually (except that range creates a list instead of a generator in Python, which was not a big deal here), but I have replaced `range` by `six.moves.range` now.

nealwu:
@panyx0718 does this look good to you?

nealwu:
Great. Thanks @edouardfouche!

nealwu:
(As a side note, the reason to use `six.moves.range` / `six.moves.xrange` rather than just `range` is that it will end up as `xrange` in Python 2, which has slightly better performance than `range`.)

edouardfouche:
Yep. Thx @nealwu !

Issue #1419, opened 2017-04-27
Slim throws all Variables must have their device set: name: "global_step" after upgrade tensorflow to v1.0.0


tobiajo:
When running [Slim](https://github.com/tensorflow/models/tree/master/slim) with multiple worker replicas I get the following exception:
```
Traceback (most recent call last):
  File "train_image_classifier.py", line 594, in <module>
    tf.app.run()
  File "/usr/local/lib/python2.7/dist-packages/tensorflow/python/platform/app.py", line 48, in run
    _sys.exit(main(_sys.argv[:1] + flags_passthrough))
  File "train_image_classifier.py", line 396, in main
    train()
  File "train_image_classifier.py", line 590, in train
    sync_optimizer=optimizer if FLAGS.sync_replicas else None)
  File "/usr/local/lib/python2.7/dist-packages/tensorflow/contrib/slim/python/slim/learning.py", line 715, in train
    init_fn=init_fn)
  File "/usr/local/lib/python2.7/dist-packages/tensorflow/python/training/supervisor.py", line 336, in __init__
    self._verify_setup()
  File "/usr/local/lib/python2.7/dist-packages/tensorflow/python/training/supervisor.py", line 881, in _verify_setup
    "their device set: %s" % op)
ValueError: When using replicas, all Variables must have their device set: name: "global_step"
op: "VariableV2"
attr {
  key: "_class"
  value {
    list {
      s: "loc:@global_step"
    }
  }
}
attr {
  key: "container"
  value {
    s: ""
  }
}
attr {
  key: "dtype"
  value {
    type: DT_INT64
  }
}
attr {
  key: "shape"
  value {
    shape {
    }
  }
}
attr {
  key: "shared_name"
  value {
    s: ""
  }
}
```

The line numbers does not match because of that I have added [this](https://github.com/tobiajo/yarntf/commit/23fdddc8fcfb4a55a90ed8f692b0f44100c759df) to `train_image_classifier.py`.

@nathansilberman @sguada

aselle:
http://stackoverflow.com/questions/38793718/distributed-tensorflow-valueerror-when-when-using-replicas-all-variables-mus

tobiajo:
@aselle already tried that.

If I add a `with tf.device("/job:ps/task:0"):` block as follows:
```
# Create global_step
with tf.device(deploy_config.variables_device()):
  with tf.device("/job:ps/task:0"):
    global_step = slim.create_global_step()
```

I instead get the following error:
```
[...]

PR #1426, opened 2017-04-28
Slim: Add tf.device(/job:ps/task:0) block to global_step

tobiajo:
To solve [issue #1419](https://github.com/tensorflow/models/issues/1419) there we get the exception:
`ValueError: When using replicas, all Variables must have their device set: name: "global_step"`
For when using multiple worker replicas, i.e. running distributed TensorFlow.

Solved with the accepted answer from: http://stackoverflow.com/questions/38793718/distributed-tensorflow-valueerror-when-when-using-replicas-all-variables-mus

### DIFF
--- a/slim/train_image_classifier.py
+++ b/slim/train_image_classifier.py
@@ -395,7 +395,8 @@ def main(_):
 
     # Create global_step
     with tf.device(deploy_config.variables_device()):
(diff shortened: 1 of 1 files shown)


tfboyd:

PR #1427, opened 2017-04-29
stack outputs instead of concat outputs

kepingwang:
`outputs` is a list of `num_steps` tensors, each shaped `(batch_size, size)`. 

It makes more sense to stack them to a `(batch_size, num_steps, size)` tensor, instead of concatenating them to a `(batch_size, size*num_steps)` tensor, although it makes no difference to the result.


nealwu:
@lukaszkaiser can you check whether this makes sense? I'm not so familiar with this model.

lukaszkaiser:
Looks good to me, thanks!

PR #1441, opened 2017-05-04
Distributed swivel

waterson:
This refactors swivel.py so that it uses the distributed TF API to run on a TF cluster. It can still be run single-threaded.  A sample shell script is included that launches a parameter server and several workers.


vmarkovtsev:
I guess I will have a hard time merging https://github.com/tensorflow/models/pull/1372 afterwards he-he

PR #1447, opened 2017-05-05
resnet_v1 and v2 segmentation bugs from 7e2435e resolved

ahundt:
7e2435e added a squeeze option, which introduced a bug in which logits would never be assigned if squeeze is False. This pull request fixes the bug in resnet_v1 and resnet_v2


ahundt:
@stef716 @nealwu this pull request is relevant to you guys

nealwu:
Looks good. Thanks!

PR #1454, opened 2017-05-08
correct the percentage logs in slim_walkthrough.ipynb

shallinlin:
Correct the format of the percentages in slim_walkthrough.ipynb.
The probability is shown as percentage thus should be scaled correctly.

### DIFF
--- a/slim/slim_walkthrough.ipynb
+++ b/slim/slim_walkthrough.ipynb
@@ -849,7 +849,7 @@
     "    names = imagenet.create_readable_names_for_imagenet_labels()\n",
     "    for i in range(5):\n",
     "        index = sorted_inds[i]\n",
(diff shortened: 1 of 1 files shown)


shallinlin:
I signed it!

nealwu:
@shallinlin This looks good to me. Can you provide me with some before/after output to verify that this is a valid change?

shallinlin:
Here is the screen-capture for the output before and after changes:
screen shot 2017-05-09 at 8 31 35 pm
screen shot 2017-05-09 at 8 31 48 pm
screen shot 2017-05-09 at 8 33 04 pm
screen shot 2017-05-09 at 8 33 14 pm

nealwu:
Looks good. Thanks!

PR #1342, opened 2017-04-13
Use batch_norm in contrib/layers/python/layers/layers.py in Resnet model

Kongsea:
Fix a TODO using batch_norm in contrib/layers/python/layers/layers.py in Resnet model.

PR #1457, opened 2017-05-10
modification to resnet_model.py and resnet_main.py

ultrons:
resnet_model.py:
- Replaced _batch_norm function with tf.layers.batch_normalization
It was also one of TODOs as code comment.

- resnet_model.py, resnet_main.py
Added hyperparameter options for dropout on identity and conv branch going to addition block with residual cell. As also noted in Wide Resnet Paper (https://arxiv.org/pdf/1605.07146.pdf) that dropout was found to be counter-productive on the identity branch. Authors recommend to implement it between conv layers instead. in my experiment however dropout on conv branch just before the addition actually improves test accuracy (92.98 in ~70.0k steps, keep_prob=0.9). I thought it could useful for other explorers.

### DIFF
--- a/resnet/resnet_main.py
+++ b/resnet/resnet_main.py
@@ -198,7 +198,11 @@ def main(_):
                              use_bottleneck=False,
                              weight_decay_rate=0.0002,
                              relu_leakiness=0.1,
-                             optimizer='mom')
+                             optimizer='mom',
+                             keep_prob=0.9,
+                             dropout_identity=False,
+                             dropout_conv=True
+                             )
 
   with tf.device(dev):
     if FLAGS.mode == 'train':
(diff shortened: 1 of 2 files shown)


ultrons:
shifting and scaling operations in during batch normalization for plain fully connected layer use normalization across the batch to compute shift and offset values. In case of conv-nets, I think, the reason why normalizing across all the dimensions but the channels is most widely used is because variation across channel is a characteristic calculated from different filters (distinct in terms of filter parameters). I wanted to experiment what happens if we treat multiple instances of same filter as we convolve also as distinct. Will it really help or harm the learning?

Results were almost the same, I am working on contrasting the two approaches. In favor of implementation simplicity, I moved to tf.layers.batch_normalization. The distinct advantage of which in my experience is that your code will be less error prone. It is fun to debug,  thanks to tensorboard, if you forgot to add op calculating running mean and running variance as training op, but you lose time.

Kongsea:
This [PR](https://github.com/tensorflow/models/pull/1342) has fixed this TODO previously.

tfboyd:

PR #1485, opened 2017-05-18
Update slim_walkthrough.ipynb

hizagalilo:
Dicts don't have "iteritems" attribute. It should be "items"


hizagalilo:
I signed it!

nealwu:
Thanks! They do have `.iteritems()` in Python 2, but `.items()` should be compatible with both Python 2 and 3.

PR #1487, opened 2017-05-18
Convert control_flow_ops.with_dependencies to tf.control_dependencies

nealwu:
Note: I found other instances of `control_flow_ops` but wasn't sure what to do with them. Examples: https://github.com/tensorflow/models/blob/master/slim/preprocessing/inception_preprocessing.py#L40 and https://github.com/tensorflow/models/blob/master/syntaxnet/syntaxnet/structured_graph_builder.py#L57.

Also, a few of the conversions turned out a bit strange. I ended up having a few lines that assigned a variable to itself (`output = output`), and I wasn't sure what to do with the calls that used the `name` argument of `control_flow_ops.with_dependencies()`.

### DIFF
--- a/inception/inception/slim/ops_test.py
+++ b/inception/inception/slim/ops_test.py
@@ -21,8 +21,6 @@ from __future__ import print_function
 import numpy as np
 import tensorflow as tf
 
-from tensorflow.python.ops import control_flow_ops
-
 from inception.slim import ops
 from inception.slim import scopes
 from inception.slim import variables
@@ -602,7 +600,8 @@ class BatchNormTest(tf.test.TestCase):
       update_ops = tf.get_collection(ops.UPDATE_OPS_COLLECTION)
       with tf.control_dependencies(update_ops):
         barrier = tf.no_op(name='gradient_barrier')
-        output = control_flow_ops.with_dependencies([barrier], output)
+        with tf.control_dependencies([barrier]):
+          output = output
       # Initialize all variables
       sess.run(tf.global_variables_initializer())
       moving_mean = variables.get_variables('BatchNorm/moving_mean')[0]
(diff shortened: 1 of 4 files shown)


slim/deployment/model_deploy.py
```diff
@@ -378,8 +378,8 @@ def deploy(config,
         update_ops.append(grad_updates)
 
         update_op = tf.group(*update_ops)
-        train_op = control_flow_ops.with_dependencies([update_op], total_loss,
-                                                      name='train_op')
+        with tf.control_dependencies([update_op]):
+          train_op = total_loss
```
jhseu: Note that this probably won't work. total_loss needs to be created under here.

slim/train_image_classifier.py
```diff
@@ -540,8 +539,8 @@ def clone_fn(batch_queue):
     update_ops.append(grad_updates)
 
     update_op = tf.group(*update_ops)
-    train_tensor = control_flow_ops.with_dependencies([update_op], total_loss,
-                                                      name='train_op')
+    with tf.control_dependencies([update_op]):
+      train_tensor = total_loss
```
jhseu: Same here.

inception/inception/slim/ops_test.py
```diff
@@ -602,7 +600,8 @@ def testComputeMovingVars(self):
       update_ops = tf.get_collection(ops.UPDATE_OPS_COLLECTION)
       with tf.control_dependencies(update_ops):
         barrier = tf.no_op(name='gradient_barrier')
-        output = control_flow_ops.with_dependencies([barrier], output)
+        with tf.control_dependencies([barrier]):
+          output = tf.identity(output)
```
jhseu: Just move the op creation itself here?

nealwu: You mean moving `ops.batch_norm(images, decay=0.1)` to this line, right? Not quite sure why, but that causes the tests to fail (something to do with the `update_ops` above).

slim/deployment/model_deploy.py
nealwu: Done (with `tf.identity`).

slim/train_image_classifier.py
nealwu: Done (with `tf.identity`).

inception/inception/slim/ops_test.py
nealwu: Switched to a simplified version without using `barrier`.

PR #1492, opened 2017-05-20
Update protobuf version dependency

angryadam:
Updates protobuf version as previous (3.0.0b2) was no longer compatible and resulted in failing bazel tests during setup process.


angryadam:
I signed it!

PR #1500, opened 2017-05-23
Attention OCR: update README.md

alexgorban:
Add instructions how to use the model with own image data.

### DIFF
--- a/attention_ocr/README.md
+++ b/attention_ocr/README.md
@@ -50,6 +50,7 @@ cd ..
 To run all unit tests:
 
 ```
+cd models/attention_ocr/python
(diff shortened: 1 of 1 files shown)


nealwu:
Thanks!

Miya125:
Hi @alexgorban, I still have one question about using the model. I have no idea about what should I write in each image data file.
For example, I have an image and its labels, should I write the decoded image tensor in shape[150, 600, 3] and the related labels in shape[37,] into one data file? Dose the file a complete training and testing data file?
Is there any way I can just load the original images into model and decode them in model?

PR #1511, opened 2017-05-27
fix the default params about article and abstract

studynon:
@nealwu

### DIFF
--- a/textsum/seq2seq_attention.py
+++ b/textsum/seq2seq_attention.py
@@ -46,10 +46,10 @@ tf.app.flags.DEFINE_string('decode_dir', '', 'Directory for decode summaries.')
 tf.app.flags.DEFINE_string('mode', 'train', 'train/eval/decode mode')
 tf.app.flags.DEFINE_integer('max_run_steps', 10000000,
                             'Maximum number of run steps.')
-tf.app.flags.DEFINE_integer('max_article_sentences', 2,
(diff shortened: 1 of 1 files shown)


studynon:
I signed it! @googlebot

nealwu:
@panyx0718 could you review?

textsum/seq2seq_attention.py
```diff
@@ -46,10 +46,10 @@
 tf.app.flags.DEFINE_string('mode', 'train', 'train/eval/decode mode')
 tf.app.flags.DEFINE_integer('max_run_steps', 10000000,
                             'Maximum number of run steps.')
-tf.app.flags.DEFINE_integer('max_article_sentences', 2,
+tf.app.flags.DEFINE_integer('max_article_sentences', 100,
```
panyx0718: I don't think it can handle 100 sentences. 2~4 sentences and ~100 words is where it can handle.

textsum/seq2seq_attention.py
```diff
@@ -46,10 +46,10 @@
...
                             'Max number of first sentences to use from the '
                             'article')
-tf.app.flags.DEFINE_integer('max_abstract_sentences', 100,
+tf.app.flags.DEFINE_integer('max_abstract_sentences', 2,
```
panyx0718: 100 is just unlimited because abstract in training data almost never exceed a few sentences.

nealwu:
Looks like this should be closed then.

PR #1513, opened 2017-05-28
Fix arxiv links in README of neural_gpu model

pgericson:
As the links was `[[link]]` the last `]` was included in the url.


pgericson:
I signed it!

nealwu:
Thanks!

PR #1517, opened 2017-05-31
SYNSETS contains the full path and does not need pwd prepended.

frankchn:
Tested on GCE with ubuntu 16.04. Mirrors change 157536266.


nealwu:
Thanks Frank!

PR #1519, opened 2017-06-01
Add executable flag to files so that download_and_preprocess_imagenet.sh can run them

frankchn:
download_and_preprocess_imagenet.sh (lines 78 and 86) tries to run the python scripts preprocess_imagenet_validation_data.py and process_bounding_boxes.py by simply calling them from the script without invoking the python CLI tool. This only works if the python scripts are marked executable, and this makes that change.


nealwu:
Thanks Frank!

PR #1526, opened 2017-06-02
add accuracy

astorfi:
The accuracy for each batch is calculated and will be stored in the summaries.
This function help the user to realize what is the overall performance on training set.


nealwu:
Hi @astorfi, could you describe what the behavior is before this change? I assume the code doesn't provide any training accuracy?

Also, this pull request has many reversions on the code (you deleted slim/WORKSPACE and added back several old pieces of code that have been removed).

astorfi:
For each batch of data, It calculates the accuracy for each clone and averages the clones accuracies and sends them to summaries. The accuracy it provides is the batch accuracy while training.

About the reversions on the code, you are correct. I accidentally deleted slim/WORKSPACE but about the old pieces of code I don't follow exactly because I just forked slim.

nealwu:
Specifically it seems like you checked out an old version of the models repository several weeks / months ago and haven't pulled from the repository since then.

nealwu:
For example your code reverts this fix from three months ago: https://github.com/tensorflow/models/blame/master/slim/deployment/model_deploy.py#L405

astorfi:
I checked ... You are right ... I will check it again and will correct it.
Should I close this pull request? Or I can update my files here?

Thanks

astorfi:
File have been updated!

nealwu:
Looks like you fixed those but you added a large number of changes to another file, `train_image_classifier.py`. I also don't think your code does anything with the accuracy tensors that you calculated in order to make them visible.

I think we are fine with the code as is actually, especially since this PR contains a number of issues. Going to close.

astorfi:
It certainly calculates the batch accuracy. The train_image_classifier.py should be changed a little bit to add the logits. I don't think you take a look at it carefully. Please take a look because it does what I claim.

You said I added a large number of changes to another file, ``train_image_classifier.py``. I just changed few thing which are almost nothing.

astorfi:
I made so much effort but fast closing the pull request is not the best way I believe. We are discussing and we can make things to work.

astorfi:
How come the Slim does provide any accuracy metric in the training and even without testing my code you say "I think we are fine with the code as is actually"? I don't think that's the correct way to treat someone who spent days to understand the TF-Slim high-level API but going deep into it and trying to make a contribution.

nealwu:
Unfortunately I'm not the creator of slim (it is a tf.contrib project) so I can't give you the reasoning into their design decisions. However have you tried using TensorBoard during training to view accuracy? See https://www.tensorflow.org/get_started/summaries_and_tensorboard#launching_tensorboard.

GitHub reports that you changed 350 lines in `train_image_classifier.py` which are mostly indentation changes. It's difficult to determine what the actual diff is there.

astorfi:
I will correct the indentation and sent it again!

astorfi:
They have been corrected

PR #1527, opened 2017-06-03
add accuracy for each training batch

astorfi:


slim/train_image_classifier.py
```diff
@@ -457,12 +457,21 @@ def clone_fn(batch_queue):
       # Specify the loss function #
       #############################
       if 'AuxLogits' in end_points:
-        tf.losses.softmax_cross_entropy(
-            logits=end_points['AuxLogits'], onehot_labels=labels,
-            label_smoothing=FLAGS.label_smoothing, weights=0.4, scope='aux_loss')
+          logits = end_points['AuxLogits']
```
nealwu: This isn't correct. You changed the value of the variable `logits` for the operations below.

slim/train_image_classifier.py
```diff
@@ -457,12 +457,21 @@ def clone_fn(batch_queue):
...
+          tf.losses.softmax_cross_entropy(
+              logits=logits['AuxLogits'], onehot_labels=labels,
```
nealwu: This is also incorrect since you're calling `['AuxLogits']` twice.

slim/train_image_classifier.py
```diff
@@ -457,12 +457,21 @@ def clone_fn(batch_queue):
...
       tf.losses.softmax_cross_entropy(
           logits=logits, onehot_labels=labels,
           label_smoothing=FLAGS.label_smoothing, weights=1.0)
+
+      #############################
+      ## Calculation of accuracy ##
+      #############################
+      accuracy = slim.metrics.accuracy(tf.to_int32(tf.argmax(logits, 1)),
+                                       tf.to_int32(tf.argmax(labels, 1)))
+      tf.add_to_collection('accuracy', accuracy)
```
nealwu: In fact, I believe you only need this portion and not any of the code above this. You have everything you need to compute accuracy here and can output it after that.

slim/train_image_classifier.py
astorfi: The variable `logits` will only change if its above if condition is True which is part of the code.

astorfi: However it needs another change

slim/train_image_classifier.py
astorfi: About this you are absolutely correct!

slim/train_image_classifier.py
astorfi: Sure ... However, that only determines the accuracy for a single clone not the average on all clones!
The main thing is the returning the averaged accuracy for all clones if we are dealing with mutli-gpu scenario!

slim/train_image_classifier.py
astorfi: The logits must be changed if the condition is True because we have to feed the new logits to calculated accuracy as it is needed to be fed for calculation of softmax cross-entropy loss.

astorfi:
I made some changes. There is no need to change 'model_deploy.py' file now!
Only few lines have been changed in **slim/train_image_classifier.py** file.
Although GitHub shows lots of changes but it's only due to changes in line places or indentation most probably.

astorfi:
Indentation mismatches have been corrected!

astorfi:
Besides, flower fine tuning script for vgg_19 has been added.

astorfi:
Any feedback?

slim/train_image_classifier.py
```diff
@@ -457,12 +457,21 @@ def clone_fn(batch_queue):
...
-            label_smoothing=FLAGS.label_smoothing, weights=0.4, scope='aux_loss')
+            logits = end_points['AuxLogits']
+            tf.losses.softmax_cross_entropy(
+                logits=logits, onehot_labels=labels,
+                label_smoothing=FLAGS.label_smoothing, weights=0.4, scope='aux_loss')
       tf.losses.softmax_cross_entropy(
-          logits=logits, onehot_labels=labels,
-          label_smoothing=FLAGS.label_smoothing, weights=1.0)
+            logits=logits, onehot_labels=labels,
+            label_smoothing=FLAGS.label_smoothing, weights=1.0)
```
nealwu: You still changed the behavior of this code when the above `if` is `True`. Are you suggesting that the code was incorrect previously?

nealwu: Also please fix the indenting. You added six-space indenting here.

slim/train_image_classifier.py
```diff
@@ -457,12 +457,21 @@ def clone_fn(batch_queue):
...
+
+      #############################
+      ## Calculation of accuracy ##
+      #############################
+      accuracy = slim.metrics.accuracy(tf.to_int32(tf.argmax(logits, 1)),
+                                         tf.to_int32(tf.argmax(labels, 1)))
```
nealwu: I don't think the `tf.to_int32` is needed here. Also please fix the indentation so that the second line aligns properly with the first.

slim/train_image_classifier.py
astorfi: You are right ... Now it's fixed!

slim/train_image_classifier.py
astorfi: I tested that ... Looks like it's not necessary but I think eliminating  tf.to_int32 caused me some issue  ... For now I remove it though!

astorfi:
All aforementioned issues been fixed!

astorfi:
Is there any further action required?

nealwu:
@sguada does this look appropriate to you?

astorfi:
@nealwu  @sguada Any feedback?

nealwu:
@sguada Let us know your feedback.

astorfi:
@nealwu @sguada  Any feedback? Thanks

sguada:
Please sync fix conflits.

PR #1530, opened 2017-06-05
To pass test, add VariableV2 in _PSDeviceChooser

xiangjinwu:
model_deploy_test.py contains 5 failures. This fixes all the tests.


nealwu:
Thanks! I ran the tests and noticed that it was not including `VariableV2`. This change fixed the tests for me.

PR #1532, opened 2017-06-06
Fixed calls to concat and convolution2d

andrewgilbert12:
When attempting to run the adversarial crypto model, I got the following errors:

```
line 302, in _AssertCompatible
    (dtype.name, repr(mismatch), type(mismatch).__name__))
TypeError: Expected int32, got list containing Tensors of type '_Message' instead.
line 126, in model
```

```
    [tf.contrib.layers.fully_connected, tf.contrib.layers.convolution],
AttributeError: 'module' object has no attribute 'convolution'
```

It appears that the arguments for tf.concat were in the opposite order of the API documentation, and that the "tf.contrib.layers.convolution2d" method was called as "convolution". I have corrected both of these issues in this pull request.

### DIFF
--- a/adversarial_crypto/train_eval.py
+++ b/adversarial_crypto/train_eval.py
@@ -128,13 +128,13 @@ class AdversarialCrypto(object):
     """
 
     if key is not None:
-      combined_message = tf.concat(1, [message, key])
+      combined_message = tf.concat([message, key], 1)
     else:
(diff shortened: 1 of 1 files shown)


nealwu:
Thanks! cc @dave-andersen

dave-andersen:
Indeed - thanks!

PR #1536, opened 2017-06-07
Implemented LRN for AlexNet tutorial

awilliamson:
A TODO was stated for adding LRN - Pending GPU support.
LRN was implemented by tensorflow/tensorflow@35df3ed43edabbc4ad1b2439bbc7de8917026d6e

Hyper-parameters taken from http://papers.nips.cc/paper/4824-imagenet-classification-with-deep-convolutional-neural-networks
Additionally the location and architecture linking of the LRN layers was taken from Section 3.5 of the above paper, as best as can be deciphered.

Note: Individual CLA has been submitted, appropriately linking this GitHub Account.

### DIFF
--- a/tutorials/image/alexnet/alexnet_benchmark.py
+++ b/tutorials/image/alexnet/alexnet_benchmark.py
@@ -73,11 +73,18 @@ def inference(images):
     print_activations(conv1)
     parameters += [kernel, biases]
 
-  # lrn1
-  # TODO(shlens, jiayq): Add a GPU version of local response normalization.
+
+  with tf.name_scope('lrn1') as scope:
+    lrn1 = tf.nn.local_response_normalization(
+      conv1,
+      alpha=1e-04,
+      beta=0.75,
+      depth_radius=5,
+      bias=2.0
+    )
 
   # pool1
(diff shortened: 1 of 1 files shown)


awilliamson:
@nealwu Any comments? Especially in regards to the hyper-parameters for the LRN.

nealwu:
Hi @awilliamson, looks reasonable to me. The one thing I'm not sure about though is the `depth_radius` parameter, since the paper and the [TF docs](https://www.tensorflow.org/api_docs/python/tf/nn/local_response_normalization) seem to define it differently.

nealwu:
(Namely, the paper divides it by two and the docs do not)

nealwu:
On second thought I think we actually want `depth_radius=2` to cover a total of 5 elements.

awilliamson:
@nealwu Yes, the paper does seem to state that their definition of `n` is the number of adjacent maps in the same spatial coordinate; whereas the TF docs stipulate that `depth_radius` represents the half-width of such a window. Good job on spotting that, it was rather difficult to interpret the paper parameters and how they map to their TF implementation.

nealwu:
Merging! Thanks @awilliamson.

PR #1538, opened 2017-06-08
Input pipeline to CPU:0 increase images/sec from 1700 to 8000 on GTX 1080

tfboyd:
I also fixed the multi-GPU example which was creating multiple queues, which was incorrect.  Reading from one queue will also speedup multi-GPU even beyond using CPU:0 but I did not test extensively.


tutorials/image/cifar10/cifar10_multi_gpu_train.py
```diff
@@ -160,6 +161,12 @@ def train():
     # Create an optimizer that performs gradient descent.
     opt = tf.train.GradientDescentOptimizer(lr)
 
+    # Get images and labels for CIFAR-10.
+    # Force input pipeline to CPU:0 to avoid opertaios sometimes ending up on GPU
```
jingjinglong: opertaios -> operations

tutorials/image/cifar10/cifar10_multi_gpu_train.py
```diff
@@ -160,6 +161,12 @@ def train():
...
+    # and resulting in a slow down.
+    with tf.device('/CPU:0'):
```
jingjinglong: Since this is already in a block with tf.device('/cpu:0) you don't need to specify it again right?

tutorials/image/cifar10/cifar10_train.py
```diff
@@ -62,7 +62,10 @@ def train():
     global_step = tf.contrib.framework.get_or_create_global_step()
 
     # Get images and labels for CIFAR-10.
-    images, labels = cifar10.distorted_inputs()
+    # Force input pipeline to CPU:0 to avoid opertaios sometimes ending up
+    # on GPU and resulting in a slow down.
+    with tf.device('/CPU:0'):
```
jingjinglong: lowercase cpu for consistency with other parts of this file?

tutorials/image/cifar10/cifar10_train.py
```diff
@@ -62,7 +62,10 @@ def train():
     global_step = tf.contrib.framework.get_or_create_global_step()
 
     # Get images and labels for CIFAR-10.
-    images, labels = cifar10.distorted_inputs()
+    # Force input pipeline to CPU:0 to avoid opertaios sometimes ending up
```
nealwu: opertaios -> operations again

tutorials/image/cifar10/cifar10_multi_gpu_train.py
```diff
@@ -169,7 +176,7 @@ def train():
             # Calculate the loss for one tower of the CIFAR model. This function
             # constructs the entire CIFAR model but shares the variables across
             # all towers.
-            loss = tower_loss(scope)
+            loss = tower_loss(scope, images, labels)
```
nealwu: Doesn't this pass the same batch to every tower? I think we probably want to do `images, labels = cifar10.distorted_inputs()` within the `for i in xrange(FLAGS.num_gpus):` loop.

tfboyd: Fixed, I am using slim.  My test (which my not be great) is I ran for 1000 steps with batch_size-128 on one GPU.  Then I ran batch_size=32 on multi-GPU with my original method and then the new method to verify the difference with the "wrong" original method I get a much lower loss than I get with the new method.  Maybe not perfect and this is also how it is done with slim.  Given queue readers on on the way out Using slim seems fine and I would use this over a manual split.  I could not find a better example.

tfboyd: PTAL

tfboyd:
@nealwu PTAL as a sanity check.  I think i addressed the biggest issue and the typos.

tutorials/image/cifar10/cifar10_multi_gpu_train.py
nealwu: I made a few cleanup changes. Nice that testing showed it does better than previously.

Is slim the best way to get this to work? This is a tutorial folder so it would be ideal to avoid having slim code here (and contrib code in general, if we can avoid it).

tfboyd:
I do not like us using contrib anywhere.  I felt this was the best option
and it is used in the slim examples.  Queues are also dieing fast I hope.
I view this example as not good and assume you or Olivia will redo the
tutorial or I will as soon as Martin gets someone to add some multi gpu
stuff to estimators.  The links have been broken for months so yeah.  At
least it is less bad now.

On Jun 8, 2017 4:51 PM, "Neal Wu" <notifications@github.com> wrote:

*@nealwu* commented on this pull request.
------------------------------

In tutorials/image/cifar10/cifar10_multi_gpu_train.py
<https://github.com/tensorflow/models/pull/1538#discussion_r121024141>:

> @@ -169,7 +176,7 @@ def train():
             # Calculate the loss for one tower of the CIFAR model.
This function
             # constructs the entire CIFAR model but shares the variables across
             # all towers.
-            loss = tower_loss(scope)
+            loss = tower_loss(scope, images, labels)

I made a few cleanup changes. Nice that testing showed it does better than
previously.

Is slim the best way to get this to work? This is a tutorial folder so it
would be ideal to avoid having slim code here (and contrib code in general,
if we can avoid it).

—
You are receiving this because you authored the thread.
Reply to this email directly, view it on GitHub
<https://github.com/tensorflow/models/pull/1538#discussion_r121024141>, or mute
the thread

nealwu:
Fair point. Let's merge these changes in since they should make the model actually work now.

Issue #1541, opened 2017-06-09
rnn/ptb/ptb_word_lm.py cannot run with Python3 due to str.decode


ngovanmao:
The example rnn/ptg/ptb_word_lm.py  cannot run with Python3. 
~/models/tutorials/rnn/ptb$ python3 ptb_word_lm.py --data_path=/tmp/simple-examples/data --model=small
Traceback (most recent call last):
  File "ptb_word_lm.py", line 385, in <module>
    tf.app.run()
  File "/usr/local/lib/python3.5/dist-packages/tensorflow/python/platform/app.py", line 48, in run
    _sys.exit(main(_sys.argv[:1] + flags_passthrough))
  File "ptb_word_lm.py", line 332, in main
    raw_data = reader.ptb_raw_data(FLAGS.data_path)
  File "/home/vanmao_ngo/models/tutorials/rnn/ptb/reader.py", line 73, in ptb_raw_data
    word_to_id = _build_vocab(train_path)
  File "/home/vanmao_ngo/models/tutorials/rnn/ptb/reader.py", line 34, in _build_vocab
    data = _read_words(filename)
  File "/home/vanmao_ngo/models/tutorials/rnn/ptb/reader.py", line 30, in _read_words
    return f.read().decode("utf-8").replace("\n", "<eos>").split()
AttributeError: 'str' object has no attribute 'decode'

The problem is as clear as the message: str.decode is not necessary in Python3. 

# Solution: Change in the reader.py file, just remove decode("utf-8"). 
The full solution should handle both Python 2 and Python 3. 
if sys.version_info[0] >= 3:
  return f.read().replace("\n", "<eos>").split()
else:
 return f.read().decode("utf-8").replace("\n", "<eos>").split()

girving:
@martinwicke Looks like https://github.com/tensorflow/models/tree/master/tutorials do not have obvious owners.

girving:
E.g., https://github.com/tensorflow/models/tree/master/tutorials/rnn doesn't mention who to contact.

martinwicke:
@ebrevdo, maybe you?

ebrevdo:
I'll be ooo for next week, so best assign to Lukasz.

lukaszkaiser:
ngovanmao: great thanks for finding this and posing! Would you like to make a small PR that corrects the issue, or would you rather have us do it?

ngovanmao:
OK, I will make a small PR.

lukaszkaiser:
[...]

PR #1548, opened 2017-06-13
Fix a small bug of Python3 compatibility in tutorial example /rnn/ptb

ngovanmao:
This PR is for Issue #1541
Done basic tests with Python3 without GPU, and python2 (in virtualenv).


ngovanmao:
I signed it!

PR #1555, opened 2017-06-14
Update inception_resnet_v2.py

derekjchow:
- Add inception_resnet_v2_base
- Provide option to use SAME padding for all inception resnet v2
  layers. This is to align feature maps sizes.

### DIFF
--- a/slim/nets/inception_resnet_v2.py
+++ b/slim/nets/inception_resnet_v2.py
@@ -91,10 +91,187 @@ def block8(net, scale=1.0, activation_fn=tf.nn.relu, scope=None, reuse=None):
   return net
 
 
+def inception_resnet_v2_base(inputs,
+                             final_endpoint='Conv2d_7b_1x1',
+                             output_stride=16,
+                             align_feature_maps=False,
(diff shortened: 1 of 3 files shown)


slim/nets/inception_resnet_v2.py
```diff
@@ -91,10 +91,189 @@ def block8(net, scale=1.0, activation_fn=tf.nn.relu, scope=None, reuse=None):
...
+
+  Raises:
+    ValueError: if final_endpoint is not set to one of the predefined values,
+      or if the output_stride is not 8 or 16, or if the output_stride is 8 and
+      we request an end point after 'PreAuxLogits'.
+  """
+  if output_stride != 8 and output_stride != 16:
+    raise ValueError('output_stride must be 8 or 16.')
+
+  padding = 'VALID'
```
sguada: padding = 'SAME' if align_feature_maps else 'VALID'

derekjchow: Done.

derekjchow:
Addressed the one comment. PTAL.

PR #1559, opened 2017-06-15
Update slim/nets/resnet

derekjchow:


nealwu:
Hi @derekjchow, could you separate the slim changes into a separate pull request?

Also emailing you separately about the object_detection model.

derekjchow:
@nealwu PTAL. I'll add object_detection in a separate pull request.

slim/nets/resnet_v2.py
```diff
@@ -25,6 +25,8 @@
 
 The key difference of the full preactivation 'v2' variant compared to the
 'v1' variant in [1] is the use of batch normalization before every weight layer.
+Another difference is that 'v2' ResNets do not include an activation function in
```
nealwu: @derekjchow we actually removed this comment in a previous PR. See https://github.com/tensorflow/models/pull/1465.

slim/nets/resnet_v2.py
```diff
@@ -115,7 +117,7 @@ def resnet_v2(inputs,
               global_pool=True,
               output_stride=None,
               include_root_block=True,
-              spatial_squeeze=True,
+              spatial_squeeze=False,
```
nealwu: What's the reason for converting all of these `spatial_squeeze` arguments to `False`?

derekjchow: Internally, we don't have spatial_squeeze (it's effectively false by default).

nealwu: Were you able to get this model running without it? From what I remember, I needed `spatial_squeeze=True` in order to get it to run properly. (To be fair though, that was multiple months ago.)

slim/nets/resnet_v2.py
nealwu: Given that the comment above returned in this PR, I'm also concerned that there may be other reversions to the code that got merged in as a result of this PR.

slim/nets/resnet_v2.py
derekjchow: Yes, tested train_image_classifier.py with Resnet 50. Interestingly enough, the Resnet unit tests were failing before I applied this PR.

PR #1449, opened 2017-05-06
Reproduced published results

takerum:
Hi @rsepassi,
This is Takeru Miyato, who is author of the paper "Adversarial Training Methods for Semi-Supervised Text Classification".
Thank you very much for your time and effort on making this open-sourced code!

I found some mistakes on the code:
- https://github.com/tensorflow/models/blob/f98cd28461bb51741d9c41742187e8e675c3cfda/adversarial_text/adversarial_losses.py#L228 I found that the shape of ```kl``` is shape (N, 1) and the shape of ```weights``` are shape (N). so ```weights * kl``` broadcasted to (N, N) shape. The resulting loss is N times as much as the correct loss.

- https://github.com/tensorflow/models/compare/master...takerum:reproduction?expand=1#diff-1fbce96d365468a32829d2a35e62435dL137 here the weights is calculated with ```input.labels``` , but the weights should be 1.0 on \<eos\> in input tokens, not label. If using ```input.labels``` for calculating weights, the weights are 1.0 on the position previous to the \<eos\> in ```inputs.tokens```.

I fixed the above points, and also changed some lines which are different from my original implementation.

Now I get 6.1, 6.3 and 7.2 % validation error rate on the committed version, which are the almost same performance as I got before. 
Also I get 5.9% test error rate with full training set (training + validation).
Please take a look at my commits and merge to master.

PR #1560, opened 2017-06-15
Reproduce reported virtual adversarial text results

rsepassi:


rsepassi:
This is partially based on a pull request by @takerum (https://github.com/tensorflow/models/pull/1449) that went stale.

takerum:
I am very sorry for the late reply.
I have checked the code and looks good to me.
Thanks you very much for your support and help.

PR #1572, opened 2017-06-16
Change 'iteritems' to 'items' to make it runnnable with Python3

lewcpe:
I found that object_detection_tutorial.ipynb couldn't run on Python3 due to iteritems function. This change will make it run without problem.


lewcpe:
I signed it!

derekjchow:
I've actually fixed this in [this PR](https://github.com/tensorflow/models/pull/1580/files).

nealwu:
Looks like we can close this then.

PR #1580, opened 2017-06-16
Download model in Jupyter Notebook.

derekjchow:


object_detection/object_detection_tutorial.ipynb
```diff
@@ -100,8 +103,13 @@
    },
    "outputs": [],
    "source": [
+    "# What model to download.\n",
```
jch1: Can you point to the detection model zoo too?

https://github.com/tensorflow/models/blob/master/object_detection/g3doc/detection_model_zoo.md

derekjchow: Done. See change above.

PR #1592, opened 2017-06-17
Move setup script for object_detection to the relevant sub-directory

ramananbalakrishnan:


ramananbalakrishnan:
I signed it!

derekjchow:
Thanks for the patch! A couple questions:

1) How did you test this change?
2) Could you update the documentation to reflect the fact you've moved the setup script? I think the only files you need to change are [g3doc/running_on_cloud.md](https://github.com/tensorflow/models/blob/master/object_detection/g3doc/running_on_cloud.md) and [g3doc/running_pets.md](https://github.com/tensorflow/models/blob/master/object_detection/g3doc/running_pets.md).

nealwu:
@derekjchow do you think you could add those changes? Would be better not to have a model-specific setup file in the main folder.

Issue #52, opened 2016-04-22
Errors in script "download_and_preprocess_flowers.sh" running on osx


DiegoGLagash:
There is a missing validation when the [data dir] ends with /, some concatenations fails during the process.
shuf is named gshuf when installed through homebrew.

aselle:
Seems like this script could be rewritten in python to avoid gshuf and then it could also use os.path.join(), though this would be some work. In lieu of that it could add some error checking and/or detect shuf  on mac os x machines.

girving:
@aselle Do you still want this assigned to you?

PR #1596, opened 2017-06-17
Rewrote download_and_preprocess_flowers.py for platform issue #52.

mindos:
For the OSX compatibility issue listed below:
[issue #52 ](https://github.com/tensorflow/models/issues/52)

Also accept related path now.

### DIFF
--- a/inception/inception/BUILD
+++ b/inception/inception/BUILD
@@ -153,7 +153,7 @@ py_binary(
 
 sh_binary(
     name = "download_and_preprocess_flowers",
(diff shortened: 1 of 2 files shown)


inception/inception/BUILD
```diff
@@ -153,7 +153,7 @@ py_binary(
 
 sh_binary(
```
sguada: Could you add a different target in the BUILD one for the current .sh and a new one for the .py

karmel:
@mindos @sguada Is this still active?

PR #1699, opened 2017-06-20
fix slim evaluation in slim_walkthrough.ipynb

callofdutyops:
1. Since the `slim.evaluation.evaluation()` has been removed from `evaluation.py` (see commits fafc8b7 in tensorflow), if we call slim.evaluation.evaluation() then will get: `TypeError: 'module' object is not callable`, so this should be modified to `slim.evaluation.evaluate_once()` and accordingly, some parameters modification.
2. The return value of `evaluate_once()` is a dict not a list, so the `for loop` which is to show the `metric_values` results should be modified.
3. Minors comments modification, such as explaining cannot get exact result of `1` suming the probabilities across all classes when run on GPU.

### DIFF
--- a/slim/slim_walkthrough.ipynb
+++ b/slim/slim_walkthrough.ipynb
@@ -149,9 +149,7 @@
   {
    "cell_type": "code",
    "execution_count": null,
-   "metadata": {
-    "collapsed": true
-   },
+   "metadata": {},
    "outputs": [],
    "source": [
     "with tf.Graph().as_default():\n",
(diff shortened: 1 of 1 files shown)


nealwu:
@sguada could you take a look?

Axe--:
Thank you for the commit. Works fine! :+1:

hhxjzyr:
Yes, till now this issue still exists, the most important is that the slim_walkthrough.ipynb are still using `slim.evaluation.evaluation()` which has been removed

PR #1711, opened 2017-06-21
Adds train and validation loss log for tensorboard

higepon:
When running train.py, it was a bit hard to see if the loss is decreasing.
By adding the log, we will be able to see the loss for each buckets on tensorboard.


higepon:
I had wrong email address in the commit. Closing.

PR #1742, opened 2017-06-23
Comment wrong in adversarial_crypto model

callofdutyops:
According to the code below:

```python
if key is not None:
      combined_message = tf.concat(axis=1, values=[message, key])
else:
      combined_message = message
```

If the key=None, combined_message is just message, not the key.


nealwu:
Thanks! cc @dave-andersen

PR #1757, opened 2017-06-25
Update resnet_v2.py

zack6514:
spatial_squeeze should be True like resnet_v1.py ,  or the tf.losses.softmax_cross_entropy will get shape error!


zack6514:
@googlebot  I signed it!

nealwu:
cc @derekjchow -- looks like `spatial_squeeze=False` is resulting in shape errors for people.

nealwu:
Thanks @zack6514 -- this was covered by https://github.com/tensorflow/models/pull/1913.

PR #1758, opened 2017-06-25
updates object_detection/eval.py to be python3 compatible

bmabey:
With these changes the new `object_detection/eval.py` script runs under python 3.

BTW, the `train.py` script has python 3 compat issues as well but they seem to be problems in tensorflow itself when I have opened [another PR](https://github.com/tensorflow/tensorflow/pull/11039) for.

### DIFF
--- a/object_detection/utils/metrics.py
+++ b/object_detection/utils/metrics.py
@@ -17,6 +17,7 @@
 from __future__ import division
 
 import numpy as np
+from six import moves
 
 
 def compute_precision_recall(scores, labels, num_gt):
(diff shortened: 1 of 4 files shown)


object_detection/utils/metrics.py
```diff
@@ -17,6 +17,7 @@
 from __future__ import division
 
 import numpy as np
+from six.moves import range
```
derekjchow: Could you change this to `from six import moves` and use moves.range(...) below? This avoids redefining the built-in range function.

object_detection/utils/np_box_list.py
```diff
@@ -16,6 +16,7 @@
 """Numpy BoxList classes and functions."""
 
 import numpy as np
+from six.moves import range
```
derekjchow: Ditto.

bmabey:
@derekjchow Okay, I made the requested changes and amended my commit with them.

derekjchow:
Look great! Thanks again for the change Ben!

PR #1610, opened 2017-06-19
Fix compatibility of object_detection for Python3

leVirve:
Modify these lines to make `object_detection/trainer_test.py` pass in Python 3.
- `items()` and `iteritems()` issue in `core/batcher.py` and `core/post_processing.py`
- `keys()` of `dict()` behaves different between Python 2 and 3, make it explicitly convert to list.
- Related to #1594. However, I think `six` is a better choice for checking Python version.
- Division behaves differently in `util/ops.py`.
- Python3 use `from unittest import mock` instead of Python2's `import mock`.

PR #1593, opened 2017-06-17
Change key of type 'tuple' to 'str'

justanhduc:
dictionary having both 'tuple' and 'str' keys cannot be 'sorted'

PR #1759, opened 2017-06-25
Make object_detection/core/batcher.py Python3 compatible

leVirve:
As described in https://github.com/tensorflow/models/pull/1610#issuecomment-310883904, the previous PR https://github.com/tensorflow/models/pull/1593 is not compatible with `master` branch.

Problem comes from that `str` and `tuple` object in the same container cannot be sorted in Python3.
- So my solution is to make `tuple` key merged into one single `str` and make them separated by defined character.
- (also  another solution for that https://github.com/tensorflow/tensorflow/pull/11039)


derekjchow:
@leVirve Thank you for sending this PR, but I think we should close it as it seems similar to [this merged PR](https://github.com/tensorflow/models/pull/1593).

PR #1761, opened 2017-06-26
fixed object detection pet example

SimsGautam:


derekjchow:
LGTM

jch1:
I'm not completely clear why this is the the right thing to do here.  From my understanding at this point in the walkthrough you should be in the "models" directory and the wget commands that you will have just run will have downloaded the data files to the top level models directory as well correct?

SimsGautam:
You are running this from `tensorflow/models/` and the directory layout is as shown.

```
  + object_detection/
  + data/
  - images.tar.gz
  - annotations.tar.gz
  - create_pet_tf_record.py
  ... other files and directories```

**Comment by jch1** [comment]
Yes, correct.  But I still don't see why we would point data_dir at the object_detection directory...

**Comment by SimsGautam** [comment]
Sorry showed  the directory layout incorrectly in the earlier comment. The layout is actually like this, where `data/` is inside `object_detection/`.

```
+ object_detection/
  + data/
  - images.tar.gz
  - annotations.tar.gz
  - create_pet_tf_record.py
  ... other files and directories
```

File: object_detection/g3doc/running_pets.md [hunk 1]
```diff
@@ -76,7 +76,7 @@ tar -xvf annotations.tar.gz
 tar -xvf images.tar.gz
 python object_detection/create_pet_tf_record.py \
     --label_map_path=object_detection/data/pet_label_map.pbtxt \
-    --data_dir=`pwd` \
```
derekjchow: Looking closer, this `pwd` is the correct value for this flag. We want the --data_dir path to point to the directory where {images/annotations}.tar.gz are unzipped. If you copied and pasted this code block, it would unzip them into the root of the git checkout, not object_detection/data/.

korrawat:
I think the problem here is that the directory tree is inconsistent with the code. The directory tree shows that the downloaded files (`images.tar.gz` and `annotations.tar.gz`) are in `object_detection/` directory, while the code says we should download the files from `tensorflow/models/` and not `tensorflow/models/object_detection/` directory.

**Proposed solutions**
1. Fix the directory tree. Change `images.tar.gz` and `annotations.tar.gz` to be in the same directory as `object_detection/`. However, it will become inconsistent with this [tutorial on preparing inputs](https://github.com/tensorflow/models/blob/master/object_detection/g3doc/preparing_inputs.md).

or

2. Fix the code comment. Change it to `# From the tensorflow/models/object_detection/ directory` instead. However, some other code blocks will have to be edited, such as the one right below it (`gsutil cp ...`).

derekjchow:
@korrawat Solution 1 is preferred (the files should be downloaded to the root of the git checkout). You're correct that it would be inconsistent with the preparing inputs page, but it might be wise to also update that page as well.

korrawat:
@derekjchow May I ask why is it preferred that "the files should be downloaded to the root of the git checkout", i.e. the `tensorflow/models` directory in this case?

I've also just noticed that currently there are some inconsistencies between the text and the code comment in the file, including [this one](https://github.com/tensorflow/models/blob/master/object_detection/g3doc/running_pets.md#getting-the-oxford-iiit-pets-dataset-and-uploading-it-to-google-cloud-storage) and [this one](https://github.com/tensorflow/models/blob/master/object_detection/g3doc/running_pets.md#exporting-the-tensorflow-graph):

> Run the following commands from the `object_detection` directory:
> ```bash
> # From tensorflow/models/
> ...
> ```

So, these will have to be fixed in one way or another anyway (possibly by removing "from the `object_detection` directory" as it is also somewhat redundant).

derekjchow:
@korrawat We'd prefer that all commands run from one directory for the tutorial (ie. downloading, packaging, etc). We picked the git root because for the packaging step, slim is also required to be packaged so it makes sense to for the command to be run from a common directory.

You're correct about the inconsistencies though, I'll send a PR to correct them.

korrawat:
#1846 is also fixing this @derekjchow

derekjchow:
[This PR](https://github.com/tensorflow/models/pull/1861) updates the documentation to run all commands from tensorflow/models.

korrawat:
So should this be closed?

PR #1762, opened 2017-06-26
Modifications for exporting other MobileNet graphs

freedomtan:
MobileNet has two hyperparameters. One is the input image size,
the other is the depth multiplier. Some modifications are needed
so that we can use export_inference_graph.py to export all the
graphs for current available MobileNet Models.

1. --override_default_image_size=true: allowing changing image size to
   192, 160, and 128
2. mobilenet_v1_0.75, mobilenet_v1_0.5, and mobilenet_v1_0.25 for depth
   multipler = 0.75, 0.5, and 0.25

### DIFF
--- a/slim/export_inference_graph.py
+++ b/slim/export_inference_graph.py
@@ -73,6 +73,11 @@ tf.app.flags.DEFINE_boolean(
     'is_training', False,
     'Whether to save out a training-focused version of the model.')
 
+tf.app.flags.DEFINE_boolean(
+    'override_default_image_size', False,
+    'use image size specified by default_image_size even when the model does '
+    'define it')
+
 tf.app.flags.DEFINE_integer(
     'default_image_size', 224,
     'The image size to use if the model does not define it.')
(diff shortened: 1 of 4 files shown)


slim/export_inference_graph.py
```diff
@@ -107,7 +112,10 @@ def main(_):
         num_classes=(dataset.num_classes - FLAGS.labels_offset),
         is_training=FLAGS.is_training)
     if hasattr(network_fn, 'default_image_size'):
-      image_size = network_fn.default_image_size
+      if FLAGS.override_default_image_size:
+        image_size = FLAGS.default_image_size
+      else:
+        image_size = network_fn.default_image_size
```
korrawat: There is the [`eval_image_size` flag](https://github.com/tensorflow/models/blob/master/slim/eval_image_classifier.py#L79-L80) in `eval_image_classifier.py` which defaults to `None`. Perhaps something like [this line](https://github.com/tensorflow/models/blob/master/slim/eval_image_classifier.py#L126) could be a cleaner solution?

```python
eval_image_size = FLAGS.eval_image_size or network_fn.default_image_size
```

So if the flag is not specified, `image_size` defaults to `network_fn.default_image_size`. Otherwise, it uses the value from `FLAGS.default_image_size`.

freedomtan: I am OK with both ways. Using something like the eval_image_size flag and changing the FLAGS.default_image_size will change more files, so I didn't do it.

freedomtan:
https://github.com/tensorflow/models/pull/1797 from @petewarden is more complete. Will close this

PR #1763, opened 2017-06-26
Fix compatibility for Python3 (object_detection/utils/variables_helper.py)

ByeongjooAhn:
iteritems() -> items()


derekjchow:
Looks good, thanks for the patch!

PR #1780, opened 2017-06-27
Add option to export model with input node that accepts jpeg or png strings

tombstone:
Adds an option to export detection models with input node that accepts encoded image strings.

### DIFF
--- a/object_detection/exporter.py
+++ b/object_detection/exporter.py
@@ -101,7 +101,7 @@ def freeze_graph_with_def_protos(
 def _tf_example_input_placeholder():
   tf_example_placeholder = tf.placeholder(
       tf.string, shape=[], name='tf_example')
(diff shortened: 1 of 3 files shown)


object_detection/exporter.py
```diff
@@ -101,7 +101,7 @@ def freeze_graph_with_def_protos(
 def _tf_example_input_placeholder():
   tf_example_placeholder = tf.placeholder(
       tf.string, shape=[], name='tf_example')
-  tensor_dict = tf_example_decoder.TfExampleDecoder().Decode(
```
derekjchow: Revert this change. I'll send in a PR for snake_casing this function after this lands.

tombstone: done. PTAL

tombstone:
Thanks!

PR #1797, opened 2017-06-28
Mobilenet support script

petewarden:


petewarden:
Merging, since this has been pending for a while and appears clear.

PR #1800, opened 2017-06-28
Fixing urllib import in slim_walkthrough notebook

txizzle:
10 character change to make cell 28 run without erroring in slim_walkthrough notebook


txizzle:
I signed it!

PR #1801, opened 2017-06-28
Fixing evaluation config

jiaphuan:
During evaluation, should not shuffle and num_reader should be 1. Or only (num_examples / num_reader) distinct samples are evaluated.


jiaphuan:
I signed it!

jch1:
Thank you @jiaphuan !!

PR #1810, opened 2017-06-29
Fixed dictionary key for SavedModel

NilsLattek:
Fixed a small typo which I ran into while trying to setup prediction on clouML.


derekjchow:
Looks good! Thanks for the PR!

PR #1822, opened 2017-06-30
Fix CDNA transformation bug and speed up its implementation.

alexlee-gk:
- Fix CDNA transformation bug where transformed channels of color and masks were combined incorrectly.
- Remove for loop over batch size in implementation of CDNA transformation. This speeds up the building of the graph.

@cbfinn


cbfinn:
LGTM

PR #1846, opened 2017-07-03
Inconsistencies/errors in running_pets.md

slandersson:
Fixed the .md to use 'object_detection' directory instead of 'models' as the text implied


slandersson:
I signed it!

derekjchow:
@slandersson Thank you for submitting this patch. However, to be consistent with the rest of the pets tutorial, we'd prefer if all commands could be run from the root of the git checkout (so all commands are run from one directory). I've sent out [this PR](https://github.com/tensorflow/models/pull/1861) to correct this.

derekjchow:
Should be addressed by #1861

PR #1851, opened 2017-07-04
modified the last line in download_and_convert_data.py scripts

TJCVRS:
The last line in slim/download_and_convert_data.py maybe wrong, I believe the right thing should be dataset_name instead of dataset_dir, a small problem. Wish you a good day!


TJCVRS:
@googlebot I signed it

nealwu:
Thanks! FYI @derekjchow in case you want to sync internally. I also merged https://github.com/tensorflow/models/pull/1870/files.

PR #1852, opened 2017-07-04
Removed empty print statement

yardstick17:


nealwu:
As I understand, this isn't an empty statement; it intentionally prints a blank line, so I don't think it needs to be removed.

PR #1855, opened 2017-07-05
Fix compatibility for object_detection/core/batcher.py

g-ozawa:
iteritems -> items

I've tried Pets Dataset tutorial locally (not using GCP) with Python 3.6.0.
https://github.com/tensorflow/models/blob/master/object_detection/g3doc/running_pets.md

And I got error like below.
```
AttributeError: ‘dict’ object has no attribute ‘iteritems’
```


g-ozawa:
@googlebot I signed it!

derekjchow:
@g-ozawa Your patch looks good! Thanks for submitting!

Issue #1849, opened 2017-07-04
Tensorflow: Windows or NO Windows


rogerguess:
The models here seem to assume you are not trying to use Windows. Also, some models do not specify which version of python is required.

jubjamie:
To install tensorflow with a gpu on windows you just need to match the [requirements here](https://www.tensorflow.org/install/install_windows) for CUDA and cuDNN (just like for any other OS) and the use `pip install tensorflow-gpu` to install on Windows. I run tensorflow on windows with GPU support easily and it takes about 3 minutes to install provided you have the CUDA requirements. No need for ubuntu.

To learn more about Tensorflow visit http://tensorflow.org where you'll find installation help and tutorials to get you started!

If you are struggling to install tensorflow on gpu then move this to Stack Overflow where I will happily assist you. This issue tracker is not designed for what you've posted above.
If you decide to move this issue to Stack Overflow, please post the link here so that myself (or others that stumble upon this issue) can find the S/O thread.

chauhan-utk:
I am running `tensorflow-gpu` on Windows 10. As mentioned above, you need to install CUDA and cuDNN and rest follow instructions on the [tensorflow website](tensorflow.org/install/).

rogerguess:
Thank you both. I have successfully installed tensorflow-gpu on windows. the basic examples I run in python show me that it can see my gpu. The point I am trying to make is that these models are not compatible with windows. It seems these models should specify the required environment and version of python needed to run them.

jubjamie:
Which ones aren't compatible with windows? Sometimes some of the steps just require the windows version of a linux command. Some of the more complex ones may require linux libraries but you might find windows equivalents. A lot of them should just work though. Have you come across ones that explicitly say they don't?

rogerguess:
I am most interested in attention_ocr and street. No, have not seen examples that say they do not, but seems the hoops I have tried to jump through suggest they are run on ubuntu.

jubjamie:
Most of this stuff should run on windows. What kind of hoops have you had to jump through? I know that with some other projects not in this repo i've had to do a few bits and bobs but after a couple of weeks it's usually the same little things and you get over it. Just a part of programming I guess, some effort required. If you're having particular issues share them here so we can try and troubleshoot.

rogerguess:
Trying to get attention_ocr running.

I now understand that the windows equivalent of the 'source' command in those instructions is instead in windows:

mkvirtualenv <envname> 
workon <envname>

[...]

PR #1860, opened 2017-07-05
Update Readme Install Instructions

jubjamie:
Update the install instructions to provide the correct pip command to install the gpu version of tensorflow.

Also addresses a concern from issue #1849 from a user who couldn't install tensorflow correctly.


nealwu:
Thanks!

PR #1864, opened 2017-07-05
Fixed Error in specifying the Output Directory

K3WLBUDDY:
When following the directory structure of the example, TFRecords generated from the Oxford Pet Dataset are stored in the pwd i.e `tensorflow/models` while the original file was specifying `object_detection` directory


K3WLBUDDY:
I signed it!

PR #1889, opened 2017-07-07
Adding option to use experiments and Sync training with Cifar10_estimator

mari-linhares:


tfboyd:
I would add a link in the comments to this document.

https://www.tensorflow.org/api_docs/python/tf/contrib/learn/RunConfig

I was confused how this would work for distributed.  I am a little concerned about two things (maybe more but these come to mind) which are not blockers as this might be the best we have right now.

1)  Flags that go beyond the main.  I find it confusing that I am setting flags that are set in the config that I cannot see in the Main.  I would prefer the flags get set in the config as part of the main python file.

2)  All of this is on contrib.  Do we support any way to do distributed without using mostly calls in contrib?  This bothers me the most.

3) I am not a big fan of functools, but I am not great at python and my aversion is likely due to my lack of python expertise and simple mind.  :-)

I want to stress that these are not blockers to accept the PR.  These are questions for us to ask and then work to get resolved overtime.  A simple distributed example should not need any code from contrib outside of a few exceptions.  Great start.

mari-linhares:
Thanks for the feedback, @tfboyd !

About 2. : Experiments are still in contrib and expect the RunConfig in contrib as well, but this should change soon (sadly not sure if soon enough).

The big benefit of using experiments is that you get the data-parallelism without needing to change the code (e.g. replica_device_setter, cluster_specs, ...).

For now I think we can hold the PR, and wait for experiments be moved to core. And if to support distributed execution with data-parallelism is considered a priority (with no contrib code) we can make another PR adding these changes.

tfboyd:
I was wrong on the functools.partial thing.  It makes sense I am just not
very good with python.

tutorials/image/cifar10_estimator/cifar10_main.py
```diff
@@ -67,11 +67,23 @@
...
+                        'If True will run an experiment,'
+                        'otherwise will run training and evaluation'
+                        'using the estimator interface')
+
+tf.flags.DEFINE_boolean('sync', False,
+                        'If true when running in a distributed environment'
+                        'will run on sync mode')
+
+tf.flags.DEFINE_integer('num_workers', 1,
+                        'Number of workers')
```
protoget: put on the same line?

tutorials/image/cifar10_estimator/cifar10_main.py
```diff
@@ -67,11 +67,23 @@
 
 tf.flags.DEFINE_float('momentum', 0.9, 'Momentum for MomentumOptimizer.')
 
-tf.flags.DEFINE_float('weight_decay', 1e-4, 'Weight decay for convolutions.')
+tf.flags.DEFINE_float('weight_decay', 2e-4, 'Weight decay for convolutions.')
 
 tf.flags.DEFINE_boolean('use_distortion_for_training', True,
                         'If doing image distortion for training.')
 
+tf.flags.DEFINE_boolean('run_experiment', False,
```
protoget: It's definitely a good idea to augment README to show how to set up distributed cluster for distributed training.

tutorials/image/cifar10_estimator/cifar10_main.py
```diff
@@ -67,11 +67,23 @@
...
+                        'If True will run an experiment,'
+                        'otherwise will run training and evaluation'
+                        'using the estimator interface')
+
+tf.flags.DEFINE_boolean('sync', False,
```
protoget: Does sync and run_experiment always go hand-in-hand in this example? If so, we shouldn't introduce another FLAGS that's potentially in conflict with others, if not configured carefully.

tutorials/image/cifar10_estimator/cifar10_main.py
protoget: I think in distributed setting, you'd be able to get the num_workers from clusterSpec?

tutorials/image/cifar10_estimator/cifar10_main.py
protoget: run_experiment doesn't tell much to users new to TF. how's experiment different from estimator? how does experiment go with distributed TF? If running experiment give superset info than Estimator alone, why should we keep estimator option alive?

tutorials/image/cifar10_estimator/cifar10_main.py
mari-linhares: Not necessarily, you can run a local code as sync, it will not affect the training. (or it shouldn't)

tutorials/image/cifar10_estimator/cifar10_main.py
mari-linhares: not sure if I can get the clusterSpec used on the execution, will try.

tutorials/image/cifar10_estimator/cifar10_main.py
mari-linhares: * Added explanation to Experiments.

* About the "why should we keep estimator option alive" question: I think it's valid to keep both ways to show people that Experiments are not mandatory, and that you can just use the interface if you want to.

**Will Improve README**

PR #1894, opened 2017-07-07
Fix slim walkthrough to use training version of preprocessing when training

CatalinVoss:
The method load_batch has an argument is_training which is passed on to the inception pre-processing code (which does two different versions of preprocessing for training vs. eval). This should set to true in example where we're training something.


MyHumbleSelf:
Signed that thing, even though it's a tiny change -- wow that's a beast of an agreement.

Issue #1906, opened 2017-07-10
object detection label map


shzygmyx:
default label map config in object_detection.data label_map.pbtxts starts from index0, however, the program results in 'Label map ids should be >= 1'.
Change the label map id start from 1 could fix the bug.
For example,  the id in pet_label_map.pbtxt are from 0 to 37. Change the id of class 'none_of_the_above' to 38, then the bug should be fixed

AruniRC:
@shzygmyx Facing the same issue when trying to create new TF records for PASCAL dataset:

```
python create_pascal_tf_record.py --data_dir=VOCdevkit \
>     --year=VOC2012 --set=train --output_path=pascal_train.record
```
> Traceback (most recent call last):
>   File "create_pascal_tf_record.py", line 183, in <module>
>     tf.app.run()
>   File "/home/aroychow/anaconda3/envs/tensorflow2_cpu/lib/python2.7/site-packages/tensorflow/python/platform/app.py", line 48, in run
>     _sys.exit(main(_sys.argv[:1] + flags_passthrough))
>   File "create_pascal_tf_record.py", line 158, in main
>     label_map_dict = label_map_util.get_label_map_dict(FLAGS.label_map_path)
>   File "/mathworks/devel/sandbox/aroychow/tensorflow_detector_2/models/object_detection/utils/label_map_util.py", line 136, in get_label_map_dict
>     label_map = load_labelmap(label_map_path)
>   File "/mathworks/devel/sandbox/aroychow/tensorflow_detector_2/models/object_detection/utils/label_map_util.py", line 123, in load_labelmap
>     _validate_label_map(label_map)
>   File "/mathworks/devel/sandbox/aroychow/tensorflow_detector_2/models/object_detection/utils/label_map_util.py", line 36, in _validate_label_map
>     raise ValueError('Label map ids should be >= 1.')
> **ValueError: Label map ids should be >= 1.**

korrawat:
The label maps are in the `object_detection/data` directory, which contains [`pet_label_map.pbtxt`](https://github.com/tensorflow/models/blob/master/object_detection/data/pet_label_map.pbtxt) and [`pascal_label_map.pbtxt`](https://github.com/tensorflow/models/blob/master/object_detection/data/pascal_label_map.pbtxt).

[...]

PR #1909, opened 2017-07-10
Remove class id 0 from label maps

korrawat:
From #1861, class id 0 are not allowed in label maps. This affects two
label maps in the object detection datasets, i.e. pets and PASCAL VOC.
(Related: #1906)

@derekjchow Could you please take a look?


korrawat:
Oops, already fixed in #1903.

PR #1913, opened 2017-07-10
Update slim/nets.

derekjchow:
- Set spatial squeeze to true for resnets
- Make activations configurable for resnets
- Make batch norm configurable for resnets
- Fix some bad comment wrapping in vgg.py.


nealwu:
Thanks for sending this. What's the purpose of making activations and batch norm configurable?

PR #1918, opened 2017-07-11
Update directories in preparing_inputs.md

korrawat:
Like #1893, the directories to run from should be the root of the git directory, i.e. `tensorflow/models`. This PR is fixing `preparing_inputs.md` for consistency.

However, now that we run `create_pet_tf_record.py` from the root directory, this introduces some complication. Now we need to specify the flag `--label_map_path=object_detection/data/pet_label_map.pbtxt`, because its default value is `data/pet_label_map.pbtxt` (see [here](https://github.com/tensorflow/models/blob/master/object_detection/create_pet_tf_record.py#L45)). This makes sense because the file is in the `object_detection` directory, and if the script is run from there, the default value will work fine.

But does it make sense to edit the default value to `object_detection/data/pet_label_map.pbtxt` so that it's cleaner in the tutorials (but then it wouldn't be the relative path from the file itself)?

@derekjchow Could you please take a look? Thank you!

### DIFF
--- a/object_detection/README.md
+++ b/object_detection/README.md
@@ -35,7 +35,7 @@ Before You Start:
 * <a href='g3doc/installation.md'>Installation</a><br>
 
 Quick Start:
-* <a href='object_detection_tutorial.ipynb'>
+* <a href='g3doc/running_notebook.md'>
       Quick Start: Jupyter notebook for off-the-shelf inference</a><br>
 * <a href="g3doc/running_pets.md">Quick Start: Training a pet detector</a><br>
 
--- a/object_detection/README.md
+++ b/object_detection/README.md
@@ -35,7 +35,7 @@ Before You Start:
 * <a href='g3doc/installation.md'>Installation</a><br>
 
 Quick Start:
-* <a href='g3doc/running_notebook.md'>
+* <a href='object_detection_tutorial.ipynb'>
       Quick Start: Jupyter notebook for off-the-shelf inference</a><br>
 * <a href="g3doc/running_pets.md">Quick Start: Training a pet detector</a><br>
 
--- a/object_detection/g3doc/preparing_inputs.md
+++ b/object_detection/g3doc/preparing_inputs.md
@@ -8,38 +8,45 @@ TFRecords.
 ## Generating the PASCAL VOC TFRecord files.
 
 The raw 2012 PASCAL VOC data set can be downloaded
-[here](http://host.robots.ox.ac.uk/pascal/VOC/voc2012/VOCtrainval_11-May-2012.tar).
+[here](http://host.robots.ox.ac.uk/pascal/VOC/voc2012/VOCtrainval_11-May-2012.tar),
+or by using the command below.
 Extract the tar file and run the `create_pascal_tf_record` script:
 
 ```bash
-# From tensorflow/models/object_detection
+# From tensorflow/models
+wget http://host.robots.ox.ac.uk/pascal/VOC/voc2012/VOCtrainval_11-May-2012.tar
 tar -xvf VOCtrainval_11-May-2012.tar
-python create_pascal_tf_record.py --data_dir=VOCdevkit \
-    --year=VOC2012 --set=train --output_path=pascal_train.record
-python create_pascal_tf_record.py --data_dir=VOCdevkit \
-    --year=VOC2012 --set=val --output_path=pascal_val.record
+python object_detection/create_pascal_tf_record.py \
+    --label_map_path=object_detection/data/pascal_label_map.pbtxt \
+    --data_dir=VOCdevkit --year=VOC2012 --set=train \
+    --output_path=pascal_train.record
+python object_detection/create_pascal_tf_record.py \
+    --label_map_path=object_detection/data/pascal_label_map.pbtxt \
+    --data_dir=VOCdevkit --year=VOC2012 --set=val \
+    --output_path=pascal_val.record
 ```
 
 You should end up with two TFRecord files named `pascal_train.record` and
-`pascal_val.record` in the `tensorflow/models/object_detection` directory.
-
-The label map for the PASCAL VOC data set can be found at
-`data/pascal_label_map.pbtxt`.
+`pascal_val.record` in the `tensorflow/models` directory.
 
 ## Generating the Oxford-IIIT Pet TFRecord files.
 
 The Oxford-IIIT Pet data set can be downloaded from
-[their website](http://www.robots.ox.ac.uk/~vgg/data/pets/). Extract the tar
-file and run the `create_pet_tf_record` script to generate TFRecords.
+[their website](http://www.robots.ox.ac.uk/~vgg/data/pets/), or by using the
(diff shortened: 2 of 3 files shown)


korrawat:
Another issue about directories: where should the generated TFRecord files be? [The current tutorial](https://github.com/tensorflow/models/blob/master/object_detection/g3doc/preparing_inputs.md) says 

```bash
--output_dir=`pwd`
```

which is totally fine. However, since [the tutorial for running locally](https://github.com/tensorflow/models/blob/master/object_detection/g3doc/running_locally.md#recommended-directory-structure-for-training-and-evaluation) suggests that TFRecord files be placed in `object_detection/data` like the label map file, we can also change the tutorial accordingly, i.e.

```bash
--output_dir=object_detection/data
```

However, this will [the pets example](https://github.com/tensorflow/models/blob/master/object_detection/g3doc/running_pets.md) should be updated as well for consistency. Is it worth making such changes?

object_detection/README.md
```diff
@@ -35,7 +35,7 @@ Before You Start:
 * <a href='g3doc/installation.md'>Installation</a><br>
 
 Quick Start:
-* <a href='object_detection_tutorial.ipynb'>
```
derekjchow: Please revert. Our intention was to make this viewable on Github (even if it can't be run without compiling protos).

object_detection/g3doc/preparing_inputs.md
```diff
@@ -8,38 +8,45 @@ TFRecords.
 ## Generating the PASCAL VOC TFRecord files.
 
 The raw 2012 PASCAL VOC data set can be downloaded
-[here](http://host.robots.ox.ac.uk/pascal/VOC/voc2012/VOCtrainval_11-May-2012.tar).
+[here](http://host.robots.ox.ac.uk/pascal/VOC/voc2012/VOCtrainval_11-May-2012.tar),
+or by using the command below.
```
derekjchow: This should also apply to the extraction step. Perhaps we can reword it like so:

"The raw 2012 PASCAL VOC dataset is located [here](http://host.robots.ox.ac.uk/pascal/VOC/voc2012/VOCtrainval_11-May-2012.tar). To download, extract and convert it to TFRecords, run the following commands below:

object_detection/g3doc/preparing_inputs.md
```diff
@@ -8,38 +8,45 @@ TFRecords.
...
+python object_detection/create_pascal_tf_record.py \
+    --label_map_path=object_detection/data/pascal_label_map.pbtxt \
+    --data_dir=VOCdevkit --year=VOC2012 --set=val \
+    --output_path=pascal_val.record
 ```
 
 You should end up with two TFRecord files named `pascal_train.record` and
-`pascal_val.record` in the `tensorflow/models/object_detection` directory.
-
-The label map for the PASCAL VOC data set can be found at
```
derekjchow: Restore statement about where the label map is located.

object_detection/g3doc/preparing_inputs.md
```diff
@@ -8,38 +8,45 @@ TFRecords.
...
+python object_detection/create_pet_tf_record.py \
+    --label_map_path=object_detection/data/pet_label_map.pbtxt \
+    --data_dir=`pwd` \
+    --output_dir=`pwd`
 ```
 
 You should end up with two TFRecord files named `pet_train.record` and
-`pet_val.record` in the `tensorflow/models/object_detection` directory.
-
-The label map for the Pet dataset can be found at `data/pet_label_map.pbtxt`.
```
derekjchow: Restore comment about label map.

object_detection/g3doc/preparing_inputs.md
```diff
@@ -8,38 +8,45 @@ TFRecords.
...
-The label map for the PASCAL VOC data set can be found at
-`data/pascal_label_map.pbtxt`.
+`pascal_val.record` in the `tensorflow/models` directory.
 
 ## Generating the Oxford-IIIT Pet TFRecord files.
 
 The Oxford-IIIT Pet data set can be downloaded from
-[their website](http://www.robots.ox.ac.uk/~vgg/data/pets/). Extract the tar
-file and run the `create_pet_tf_record` script to generate TFRecords.
+[their website](http://www.robots.ox.ac.uk/~vgg/data/pets/), or by using the
```
derekjchow: Please reword this (We can download, extract, convert using command below).

object_detection/README.md
korrawat: My thought was that the [`running_notebook.md`](https://github.com/tensorflow/models/blob/master/object_detection/g3doc/running_notebook.md) is not referred to anywhere. It seems to me that it is the brief intro to running the notebook (maybe for those who doesn't know how to run jupyter notebook locally), and it links to [`object_detection_tutorial.ipynb`](https://github.com/tensorflow/models/blob/master/object_detection/object_detection_tutorial.ipynb) anyway.

That said, `object_detection_tutorial.ipynb` would be viewable on GitHub, but in two clicks, which might not be preferred. What do you think?

object_detection/g3doc/preparing_inputs.md
korrawat: Should this be changed to `object_detection/data/pet_label_map.pbtxt`?

object_detection/g3doc/preparing_inputs.md
korrawat: Done!

object_detection/g3doc/preparing_inputs.md
korrawat: Done!

object_detection/README.md
derekjchow: I still believe this change should be reverted, but your points are also valid. IMO, the correct thing to do is to revert this change, but update the notebook to be self-documenting (add a section about installation and running to the notebook). Changing the notebook should be done in another PR though.

korrawat: Reverted. Thank you for your review!

derekjchow:
LGTM! Thank you for making this consistent!

korrawat:
You're welcome!
@derekjchow I'm just curious if we could make it better in two ways as mentioned above. Could you please take a look?

1. In [`create_pet_tf_record.py`](https://github.com/tensorflow/models/blob/master/object_detection/create_pet_tf_record.py#L45), change the `--label_map_path` flag from `data/pet_label_map.pbtxt` to `object_detection/data/pet_label_map.pbtxt`, and then remove `--label_map_path=...` from the [preparing inputs tutorial](https://github.com/tensorflow/models/blob/master/object_detection/g3doc/preparing_inputs.md).
1. In the [preparing inputs tutorial](https://github.com/tensorflow/models/blob/master/object_detection/g3doc/preparing_inputs.md), change the `--output_dir` flag from `` `pwd` `` to `object_detection/data`, and from `pascal_*.record` to `object_detection/data/pascal_*.record`. This will help transition smoothly to [the tutorial for running locally](https://github.com/tensorflow/models/blob/master/object_detection/g3doc/running_locally.md#recommended-directory-structure-for-training-and-evaluation).

Thanks!

PR #1837, opened 2017-07-03
Removed an error-causing parameter

jaewoosong:
--logtostderr causes the following error: E tensorflow/examples/label_image/main.cc:319] Unknown argument --logtostderr

PR #1938, opened 2017-07-13
Removed wrong parameter in another example

korrawat:
Same as #1837, but this is an example in `slim/export_inference_graph.py` instead of `slim/README.md`.


nealwu:
Thanks!

PR #1945, opened 2017-07-13
Small fixes in README

mari-linhares:
- 2 CPUs -> 2 GPUs
- is_cpu_ps=True

### DIFF
--- a/tutorials/image/cifar10_estimator/README.md
+++ b/tutorials/image/cifar10_estimator/README.md
@@ -33,10 +33,18 @@ $ python cifar10_main.py --data_dir=/prefix/to/downloaded/data/cifar-10-batches-
                          --num_gpus=0 \
                          --train_steps=1000
 
-# Run the model on CPU and 2 GPUs. After training, it runs the evaluation.
+# Run the model on 2 GPUs using CPU as parameter server. After training, it runs the evaluation.
 $ python cifar10_main.py --data_dir=/prefix/to/downloaded/data/cifar-10-batches-bin \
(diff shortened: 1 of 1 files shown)


tutorials/image/cifar10_estimator/README.md
```diff
@@ -33,10 +33,10 @@ $ python cifar10_main.py --data_dir=/prefix/to/downloaded/data/cifar-10-batches-
                          --num_gpus=0 \
                          --train_steps=1000
 
-# Run the model on CPU and 2 CPUs. After training, it runs the evaluation.
+# Run the model on CPU and 2 GPUs. After training, it runs the evaluation.
 $ python cifar10_main.py --data_dir=/prefix/to/downloaded/data/cifar-10-batches-bin \
                          --model_dir=/tmp/cifar10 \
-                         --is_cpu_ps=False \
+                         --is_cpu_ps=True \
```
mari-linhares: If I get it right in the comment it says that is using is_cpu_ps=True: "# Run the model on CPU and 2 GPUs. After training, it runs the evaluation." Also, in my experiments is_cpu_ps=True was faster ( but running with experiments and an older version of the code)

protoget: It doesn't say which device is used for ps.
How about just 
1. append something like "put params on CPU" in the description and use is_cpu_ps=True
2. create another command in README and "put params on GPU", set is_cpu_ps=False

mari-linhares: +1

PR #1946, opened 2017-07-13
Convert RepeatedScalarContainer to a list of floats

dfridman1:
Convert `means` parameter of `subtract_channel_mean` function from type `RepeatedScalarContainer` to a list of floats.

Fix `TypeError: Expected float32, got <google.protobuf.pyext._message.RepeatedScalarContainer object at 0x7fb7e6d65618> of type 'RepeatedScalarContainer' instead.`


dfridman1:
I signed it!

PR #1959, opened 2017-07-14
Adds an op to handle pre-computed word embeddings.

markomernick:
This change adds an op that will take a text vocabulary file, read
it, and use it to create a pre-computed word embedding vector.


ducdauge:
I am very interested in this feature: how does it work exactly? How can I specify the path to the vocabulary file? Many thanks for your help!

PR #1961, opened 2017-07-14
resolving when there are no detections

aayushmudgal:
Remove_invalid_boxes breaks when the detected_boxes is empty. The fix checks if there is any detected_boxes as an input.


aayushmudgal:
I signed it!

PR #1968, opened 2017-07-17
Update label_map_util.py

xiaoerlaigeid:


xiaoerlaigeid:
I signed it!

derekjchow:
Could you explain why this change is necessary? It's expected that PYTHONPATH includes the root of the git repository.

PR #1978, opened 2017-07-18
fix: fix deprecated  word2vec load function in skip_thougths

seizeTheDayMin:
The previous word2vec load function is depreacted(not warning, occurring error) in the latest gensim.

```
# deprecate 
gensim.models.Word2Vec.load_word2vec_format
# allowed function in the latest gensim
gensim.models.keyedvectors.KeyedVectors.load_word2vec_format


seizeTheDayMin:
I signed it!

PR #1979, opened 2017-07-18
Fix up download_and_preprocess_imagenet.sh

igor0:
As of 444310f2, download_imagenet.sh expects an absolute path the synsets file. But,
download_and_preprocess_imagenet.sh specifies the synsets file as a relative path.

Also, there were a couple of fixes to download_image.sh that are also needed in
download_imagenet.sh (dd983ea6, 9997b250).


igor0:
According to the legal team at my company (Pure Storage), our CTO John
Colgrove signed the agreement on 4/21/2016.

PR #1984, opened 2017-07-18
Shape fix for LFADS tf.case statements

allenlavoie:
The default tf.case (which, judging by its value, has no model significance) previously had shape [1], which did not match the other cases. This caused shape inference errors when building gradients, since downstream Tensors had their shape information refined to shapes incompatible with [1].

The model works for me in TensorFlow 1.2.1 after this PR.


allenlavoie:
@jazcollins is this reasonable? I don't think I can add you as a reviewer for some reason.

jazcollins:
Thanks!!

PR #2000, opened 2017-07-21
More efficient per-example gradient for conv2d in differential_privacy

liuzjmike:
Perform more efficient per-example gradient calculation with respect to filters of conv2d operations

### DIFF
--- a/differential_privacy/dp_sgd/per_example_gradients/per_example_gradients.py
+++ b/differential_privacy/dp_sgd/per_example_gradients/per_example_gradients.py
@@ -210,45 +210,6 @@ class Conv2DPXG(object):
     self.colocate_gradients_with_ops = colocate_gradients_with_ops
     self.gate_gradients = gate_gradients
 
-  def _PxConv2DBuilder(self, input_, w, strides, padding):
-    """conv2d run separately per example, to help compute per-example gradients.
-
-    Args:
-      input_: tensor containing a minibatch of images / feature maps.
-              Shape [batch_size, rows, columns, channels]
-      w: convolution kernels. Shape
-        [kernel rows, kernel columns, input channels, output channels]
-      strides: passed through to regular conv_2d
-      padding: passed through to regular conv_2d
-
-    Returns:
-      conv: the output of the convolution.
-         single tensor, same as what regular conv_2d does
-      w_px: a list of batch_size copies of w. each copy was used
-          for the corresponding example in the minibatch.
-           calling tf.gradients on the copy gives the gradient for just
-                  that example.
-    """
-    input_shape = [int(e) for e in input_.get_shape()]
-    batch_size = input_shape[0]
-    input_px = [tf.slice(
-        input_, [example] + [0] * 3, [1] + input_shape[1:]) for example
-                in xrange(batch_size)]
-    for input_x in input_px:
-      assert int(input_x.get_shape()[0]) == 1
-    w_px = [tf.identity(w) for example in xrange(batch_size)]
-    conv_px = [tf.nn.conv2d(input_x, w_x,
-                            strides=strides,
-                            padding=padding)
-               for input_x, w_x in zip(input_px, w_px)]
-    for conv_x in conv_px:
-      num_x = int(conv_x.get_shape()[0])
-      assert num_x == 1, num_x
-    assert len(conv_px) == batch_size
-    conv = tf.concat(axis=0, values=conv_px)
-    assert int(conv.get_shape()[0]) == batch_size
-    return conv, w_px
-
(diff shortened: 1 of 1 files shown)


liuzjmike:
CLA signed

koenvanderveen:
how much more efficient does this make this? And compared to normal convolutions?

liuzjmike:
I am travelling right now and I will benchmark it as soon as I get back. I can compare this and the original implementation by swapping them in the network I am working on but I'm not sure how to compare it to normal convolution because per-example gradient calculations for other operations in my network also have speed penalty. Do you want me to just setup a one-layer network for the test?

liuzjmike:
I benchmarked the original and the proposed implementation and compared them to normal convolution with the following script:
```
import sys
import time

import numpy as np
import tensorflow as tf

from differential_privacy.dp_sgd import dp_optimizer, sanitizer
from differential_privacy.privacy_accountant import accountant

def test():
    x = tf.placeholder(tf.float32, shape=[500, 32, 32, 3], name="x")
    k1 = tf.get_variable("kernel0", shape=[3, 3, 3, 64], dtype=tf.float32,
                         initializer=tf.random_normal_initializer(stddev=0.1))
    h = tf.nn.conv2d(x, k1, [1, 1, 1, 1], "SAME", name="conv0")
    for i in range(1, 5):
        k = tf.get_variable("kernel%d" % i, shape=[3, 3, 64, 64], dtype=tf.float32,
                            initializer=tf.random_normal_initializer(stddev=0.1))
        h = tf.nn.conv2d(h, k, [1, 1, 1, 1], "SAME", name="conv%d" % i)
    h = tf.reshape(h, [500, -1])
    w = tf.get_variable("weight", shape=[h.get_shape().as_list()[1], 10],
                        dtype=tf.float32,
                        initializer=tf.random_normal_initializer(stddev=0.1))
    y = tf.matmul(h, w)
    y_ = tf.placeholder(tf.int32, shape=[500], name="y_")
    loss = tf.nn.softmax_cross_entropy_with_logits(labels=tf.one_hot(y_, 10),
                                                   logits=y)

    with tf.Session() as sess:
        def random_train_test(optimizer, step):
            start_time = time.time()
            train_op = optimizer.minimize(loss)
            overhead = time.time() - start_time
            start_time = time.time()
            for i in range(step):
                sess.run(train_op, feed_dict={
                    x:np.random.normal(size=[500, 32, 32, 3]),
                    y_:np.random.randint(0, high=10, size=500)})
                sys.stdout.write("\r[%d/%d]" % (i+1, step))
                sys.stdout.flush()
            print
            return time.time() - start_time, overhead

        privacy_accountant = accountant.GaussianMomentsAccountant(500)
        gaussian_sanitizer = \
                sanitizer.AmortizedGaussianSanitizer(
                    privacy_accountant,
                    4, 1)
        optimizer = dp_optimizer.DPGradientDescentOptimizer(
            0.1, [0.5, 1e-5], gaussian_sanitizer, 4,
            adapt_l2norm_bound=False)
        sess.run(tf.global_variables_initializer())
        ordinary, _ = random_train_test(tf.train.GradientDescentOptimizer(0.1), 50)
        print("Time used for ordinary gradient descent: %.2f" % ordinary)
        pxg, overhead = random_train_test(optimizer, 50)
        print("Overhead of pxg gradient graph: %.2f" % overhead)
        print("Time used for per-example gradient descent: %.2f" % pxg)
        print("Slow down ratio: %.2f" % (pxg / ordinary))

if __name__ == '__main__':
    test()
```
The running time of the two implementations are very similar. Both are ~3x slower than normal convolution in the test network (but also notice that calls PerExampleGradient have overhead and pxg gradient is also computed for MatMul so the ratio is not accurate). After carefully inspecting the graph, I realized that their backward passes are (almost) exactly the same.
I guess the only advantage of the proposed implementation over the original one is that it produces a simpler computational graph and thus a slightly smaller overhead. And arguably the implementation is more straight forward...

koenvanderveen:
Thanks!

koenvanderveen:
These two functions expect different parameters in my code (I have the latest code from github):

sanitizer.AmortizedGaussianSanitizer( privacy_accountant, 4, 1)
optimizer = dp_optimizer.DPGradientDescentOptimizer(0.1, [0.5, 1e-5], gaussian_sanitizer, 4, adapt_l2norm_bound=False)

the sanitizer expects a clipoption and only two arguments, and the DPoptimizer does not expect adapt_l2_norm_bound as an argument

liuzjmike:
I made some customization to the two classes and forgot about them... Sorry about that. Updated the script to use the API on GitHub. Please tell me if there are any other problem.

PR #2005, opened 2017-07-21
[Issue 11184] Filter out OSX hidden files while converting flowers dataset.

frreiss:
As described in https://github.com/tensorflow/tensorflow/issues/11184 , the script `download_and_convert_flowers.py` could attempt to convert the hidden image index file `.DS_Store` that OSX creates whenever a user opens a Finder window on a directory containing images (for example, because the user was monitoring the status of the script's target directory). This patch adds some filename filtering to the script to prevent the script from picking up .DS_Store. It also prevents the script from picking up any other hidden files (at least for files that are hidden because their names start with ".") or non-JPEG files that might appear in the temporary directory.


frreiss:
Closing out old, unmerged PRs.

Issue #1936, opened 2017-07-12
eval.py stuck: The following classes have no ground truth examples


chakpongchung:
I got the follow result after running:

# modify the config file for the right path
PATH_TO_YOUR_PIPELINE_CONFIG="object_detection/samples/configs/faster_rcnn_resnet101_voc07.config"
PATH_TO_TRAIN_DIR="out"
PATH_TO_EVAL_DIR="out"

# From the tensorflow/models/ directory
GPU_ID=1
CUDA_VISIBLE_DEVICES=${GPU_ID} python object_detection/eval.py --debug\
    --logtostderr \
    --pipeline_config_path=${PATH_TO_YOUR_PIPELINE_CONFIG} \
    --checkpoint_dir=${PATH_TO_TRAIN_DIR} \
    --eval_dir=${PATH_TO_EVAL_DIR}

**WARNING:root:The following classes have no ground truth examples: [ 0  2  3  8  9 11 13 14 16 17 18]**
/local/home/cpchung/software/models/object_detection/utils/metrics.py:145: RuntimeWarning: invalid value encountered in true_divide
  num_images_correctly_detected_per_class / num_gt_imgs_per_class)

Where can I find the result of this evalution?

protossw512:
I got same warning:

WARNING:root:The following classes have no ground truth examples: 0

My dataset has only one class, the id is 1 (I noticed in your code that the class id should start from 1).

But I cannot finish evaluation since there is no ground truth examples on class 0/ but according to your updated label_map.pbtxt, there shouldn't be any class id start with 0.

bclyc:
I had the same problem:
`WARNING:root:The following classes have no ground truth examples: [ 5 10 11 12 13 14 15 16 17 19]`

I'm sure that these classes have examples in the valuation set I made... Don't know where went wrong

FurkanKyo:
I am having the same issue as @protossw512 
```
WARNING:root:The following classes have no ground truth examples: 0
/usr/local/lib/python3.5/dist-packages/tensorflow/models/object_detection/utils/metrics.py:145: RuntimeWarning: invalid value encountered in true_divide
  num_images_correctly_detected_per_class / num_gt_imgs_per_class)
```
The problem @bclyc encountered is not the same since 0 is not listed in your warning. Therefore, your solution seems to work for your case, however it is not valid for our case because we actually don't even have any class with id 0 since documentation mentions that ids should start with 1. 

[...]

Issue #1856, opened 2017-07-05
eval.py stops evaluating after 3-5 images


ckalas:
On both the pet dataset and my own custom dataset, the evaluation script seemingly stops evaluating images after a few samples. In both cases there is a warning about no ground truth samples for some of the classes, but I altered the line to remove the warning occuring and the evaluation still stops. Using python 3.5

```
WARNING:root:The following classes have no ground truth examples: [11 15 16 17 18 19]
/home/chris/tensorflow/models/object_detection/utils/metrics.py:144: RuntimeWarning: invalid value encountered in true_divide
  num_images_correctly_detected_per_class / num_gt_imgs_per_class)

```
There is no error, only the warning, but after that nothing new appears in tensorboar.
On a side note, does the number of examples need to be set in the config file? seems strange to be hard coding...

`eval_config: {
  num_examples: 158
}
`

schechtman:
If I understand it well the number of examples is  the number of images on which the current model would be evaluated, for instance if you have 5000 images in your test dataset, setting num_examples : 2000 would make each evaluation running only at 2000 of them.

Regarding the warning, I think it just takes time, you could track the progress by adding 
`tf.logging.set_verbosity(tf.logging.INFO)` after the imports in `eval.py`
.

ckalas:
okay ill add in the extra logging. what is strange though is that it does the first few images really quickly, hits that warning (or not if i don't call the numpy function and just return zeros) and then nothing happens.

dimsava:
I have the same issue, `tf.logging.set_verbosity(tf.logging.INFO)` (after imports) seems to not work for `eval.py`, in contrast to `train.py`. 

Furthermore the `num_examples` is weird. If your test set has `X` images and `num_examples < X` then you test a subset. But if `num_examples` is bigger, for example `num_examples = 2X`, then evaluation is run twice on the same test set. It seems that right now there is no default value that makes evaluation run only once for the full test set, without you having to manually specify its size.

ckalas:
Still happening, unsure why. Trying to look into the evaluator.py file but not finding much there. Is there a potential that its to do with how my data is annotated? for instance, I omitted some tags like pose, difficult, truncated, etc. because they arent relevant to my data. Seems unlikely becuase if this where the case why would it train fine and evaluate a couple of images?

jch1:
Hi @ckalas - can you provide a bit more of your eval log?  That might be useful for additional context.

ckalas:
[...]

PR #2007, opened 2017-07-21
Fix label_id_offset inconsistencies (obj detection)

korrawat:
The ["using your own dataset" tutorial](https://github.com/tensorflow/models/blob/master/object_detection/g3doc/using_your_own_dataset.md) says "Label maps should always start from id 1." However, the current code still has some inconsistencies.

Right now, the default value for `label_id_offset` in `eval_util.py`'s [`evaluate_detection_results_pascal_voc` method](https://github.com/tensorflow/models/blob/master/object_detection/eval_util.py#L111-L125) is 0 (instead of 1). When I run an evaluation using a pre-trained model on the PASCAL VOC, for example, `num_classes` will be 21 instead of 20, giving me the following warning.

```bash
WARNING:root:The following classes have no ground truth examples: 0
```
Changing the default value of `label_id_offset` to 1 will fix it.

### Questions
Should the comment in L111 ("Pascal VOC evaluator assumes foreground index starts from zero.") also change from "zero" to "one"?

By the way, what is allowed as the `categories`?
- At least, the ids should start from 1. The `label_id_offset` can allow for other cases.
- Maybe they should also be in an increasing order and consecutive? I'm not sure whether anything will break if this condition is not satisfied (e.g. the label map is 1=cat1, 5=cat5, 3=cat3). It seems like some codes are making these assumptions?
If these should be satisfied, maybe some checks can be added in `label_map_util.py`'s [`_validate_label_map` method](https://github.com/tensorflow/models/blob/master/object_detection/utils/label_map_util.py#L25).

Possibly related issues: #1856, #1936 

@derekjchow Could you please take a look? Thank you!

### DIFF
--- a/object_detection/eval_util.py
+++ b/object_detection/eval_util.py
@@ -52,7 +52,7 @@ def write_metrics(metrics, global_step, summary_dir):
 
 def evaluate_detection_results_pascal_voc(result_lists,
                                           categories,
-                                          label_id_offset=0,
+                                          label_id_offset=1,
                                           iou_thres=0.5,
                                           corloc_summary=False):
   """Computes Pascal VOC detection metrics given groundtruth and detections.
--- a/object_detection/utils/label_map_util.py
+++ b/object_detection/utils/label_map_util.py
@@ -57,7 +57,8 @@ def create_category_index(categories):
 
 def convert_label_map_to_categories(label_map,
                                     max_num_classes,
-                                    use_display_name=True):
+                                    use_display_name=True,
+                                    label_id_offset=1):
   """Loads label map proto and returns categories list compatible with eval.
 
   This function loads a label map and returns a list of dicts, each of which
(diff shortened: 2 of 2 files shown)


tombstone:
@korrawat . Thanks for the PR. Taking a look.

pkdogcom:
@tombstone Can you merge this commit to the master branch? I believe many users are experiencing this issue when training with with their own datasets. Thanks.

tombstone:
@pkdogcom  @korrawat,  we ended up refactoring some evaluation module internally to remove inconsistencies. I will be merging it shortly. Apologies for the delay.

jingyibo123:
Hello everyone,
Could anyone take a look at this [question](https://stackoverflow.com/questions/47099606/eval-py-fails-to-identify-classes-in-test-tfrecord), could it be caused by this issue?

Issue #1976, opened 2017-07-18
download_and_preprocess_flowers fails due to a script error


Pibborn:
I am trying to run the inception model using the flowers dataset. I am using Mac OS X El Capitan, using tensorflow version:

```bash
$  python -c "import tensorflow as tf; print(tf.GIT_VERSION, tf.VERSION)" 
v1.2.0-5-g435cdfc 1.2.1
```

### Steps to reproduce
```bash
git clone https://github.com/tensorflow/models
cd models/inception
bazel build //inception:download_and_preprocess_flowers
mkdir flowers-data
bazel-bin/inception/download_and_preprocess_flowers flowers-data
 [download info...]
 >mv: rename flower_photos to flowers-data/raw-data/train: No such file or directory
```

I was able to fix this on my machine by editing the ``models/inception/inception/data/download_and_preprocess_flowers.sh`` script [from line 67 onwards](https://github.com/tensorflow/models/blob/master/inception/inception/data/download_and_preprocess_flowers.sh#L67):

```bash
cd ..
mv "${DATA_DIR}/flower_photos" "${TRAIN_DIRECTORY}"
```

which substitutes 

```bash
mv flowers_photos "$(TRAIN_DIRECTORY}"
```

This makes sense (at least to me :-)) because if you have the script do ``echo $(pwd)`` between line 68 and 69, you get ``/path/to/repo/models/inception/${DATA-DIR}``. Therefore, if you don't ``cd ..``, the script will then try to move ``/path/to/repo/models/inception/${DATA-DIR}/flower_photos/`` to ``/path/to/repo/models/inception/${DATA-DIR}/${DATA-DIR}/flower_photos/raw-data/train`` on line 69, which will fail.

Another thing to note is that line 68 silently fails - you can verify that by removing the ``-f`` option. Is that the intended behaviour?

Would a PR be appreciated, or am I mistaken about this being an error in the script? This was my first experience with anything bazel.

asimshankar:
I believe that the script expected `${DATA_DIR}` to be an absolute path, not a relative one (e.g., [in the `README`](https://github.com/tensorflow/models/tree/master/inception#getting-started) it is set to `$HOME/imagenet-data`, which will be an absolute path).

So I don't think `cd ..` is the appropriate fix, but instead perhaps the script can be changed to ensure that `${DATA_DIR}` ends up being an absolute path. In your case, this would amount to changing from:

```
bazel-bin/inception/download_and_preprocess_flowers flowers-data
```

[...]

PR #2009, opened 2017-07-21
inception: Fix issue #1976, merge mac script

Pibborn:
This commit fixes the bug described in issue #1976; it also makes the download_and_preprocess_flowers_mac.sh script redundant by checking whether shuf or gshuf are available. 

Would 2 different PRs be preferred? I could do that, too.


Pibborn:
Done!

PR #2017, opened 2017-07-24
Add mobilenet V1 1.0 224 fine tune script 

chungyeh:
I signed it!

### DIFF
--- a/slim/scripts/finetune_mobilenet_v1_1.0_224_on_flowers.sh
+++ b/slim/scripts/finetune_mobilenet_v1_1.0_224_on_flowers.sh
@@ -23,7 +23,7 @@ DATASET_DIR=/tmp/flowers
 if [ ! -d "$PRETRAINED_CHECKPOINT_DIR" ]; then
   mkdir ${PRETRAINED_CHECKPOINT_DIR}
 fi
-if [ ! -f ${PRETRAINED_CHECKPOINT_DIR}/mobilenet_v1_1.0.ckpt.index ]; then
+if [ ! -f ${PRETRAINED_CHECKPOINT_DIR}/mobilenet_v1_1.0_224.ckpt.index ]; then
   wget http://download.tensorflow.org/models/mobilenet_v1_1.0_224_2017_06_14.tar.gz
   tar -xvf mobilenet_v1_1.0_224_2017_06_14.tar.gz
   mv mobilenet_v1_1.0.ckpt.* ${PRETRAINED_CHECKPOINT_DIR}/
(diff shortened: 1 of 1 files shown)


AndreaPisoni:
Is this script correct? I do not see any reference to "AuxLogits" in the mobilenet architecture file, (which is different from inception, where AuxLogits is present...) so probably the layers to be chosen would need to be different?

haoxi911:
I didn't see 'AuxLogits' in MobilenetV1, but 'Logits' did exist, maybe we just want to remove 'AuxLogits' from the script?

Also, I am not using 'flower' and instead a own dataset, is there any additional steps required?

haoxi911:
@AndreaPisoni @chungyeh @bmabey 

I can train the MobilenetV1 checkpoint files using my own dataset, however the checkpoint file can't be frozen due to the following error:

> InvalidArgumentError (see above for traceback): Assign requires shapes of both tensors to match. lhs shape= [1,1,1024,1001] rhs shape= [1,1,1024,37]
> 	 [[Node: save/Assign_136 = Assign[T=DT_FLOAT, _class=["loc:@MobilenetV1/Logits/Conv2d_1c_1x1/weights"], use_locking=true, validate_shape=true, _device="/job:localhost/replica:0/task:0/device:CPU:0"](MobilenetV1/Logits/Conv2d_1c_1x1/weights, save/RestoreV2_136)]]

I believe it is the problem of `--checkpoint_exclude_scopes` or `--trainable_scopes` options I passed when finetune the model:

> (tensorflow) ubuntu@haoxi911:~/tensorflow/models/research/slim$ python train_image_classifier.py --train_dir=/tmp/oxford-iiit-pet/mobile_v1_1.0_224_finetune --dataset_dir=/tmp/oxford-iiit-pet --dataset_name=oxford_iiit_pets --dataset_split_name=train --model_name=mobilenet_v1 --checkpoint_path=/tmp/mobilenet_v1_2018_02_22/mobilenet_v1_1.0_224/mobilenet_v1_1.0_224.ckpt --checkpoint_exclude_scopes=MobilenetV1/Logits --trainable_scopes=MobilenetV1/Logits --clone_on_cpu=True

I also tried these options but got same errors:

> --checkpoint_exclude_scopes=MobilenetV1/Logits/Conv2d_1c_1x1/weights
> --trainable_scopes=MobilenetV1/Logits/Conv2d_1c_1x1/weights

I have posted the same question to SO [here](https://stackoverflow.com/questions/49062540/how-do-i-finetune-mobilenetv1-with-my-own-dataset), it would be great if you can take a look at the problem.

PR #2024, opened 2017-07-24
Offset of image bytes for cifar100 should be 2

naurril:
```label_offset + label_bytes``` gives the correct offset for both cifar10 and cifar100


panyx0718:
The change looks reasonable. Have you successfully train a model with the change?

naurril:
Yes,  but I didn't notice any obvious impact on the result. 
Maybe this problem is not that important for the final task, since  in effect  it only causes the rightmost pixel column being moved to the left side.

PR #2037, opened 2017-07-25
Update DenoisingAutoencoder.py

coolmatt1024:
For the denoising autoencoder, I think it should be 'self.scale' not 'scale' in the graph. If not ,I think it unnecessary to fefine placeholder of scale because it is not connected wtih any node.


coolmatt1024:
I signed it!

PR #2042, opened 2017-07-25
Minor - README to reflect model compatibility update to r1.0

calyun:
Updated March 13 to r1.0, README implementation reference to r0.7 obsolete and potentially misleading


calyun:
I signed it!

PR #2044, opened 2017-07-26
Enhanced comments and renamed ParamServerDeviceSetter

tfboyd:
Renaming ParamServerDeviceSetter to GpuParamServerDeviceSetter made my performance guide easier to follow.  I also moved FLAGS.num_gpus up a level so remove all "global" variables from the functions do the device setting.  This will make it easier for people to cut and paste.

I also added excessive comments.  

I also removed the b/.  This will is one of our simple examples and we will update it over time, not need to leave a TODO for code that does not exist yet.


protoget:
LGTM

PR #2054, opened 2017-07-28
Adding batch_size knob to export_inference_graph

jeuvilla:
Adding a batch_size knob to export_inference_graph to allow user to specify batch size for their exported model.

Defaulted to None so batch size can be decided at model runtime.

### DIFF
--- a/slim/export_inference_graph.py
+++ b/slim/export_inference_graph.py
@@ -75,6 +75,11 @@ tf.app.flags.DEFINE_integer(
     'image_size', None,
     'The image size to use, otherwise use the model default_image_size.')
 
+tf.app.flags.DEFINE_integer(
+    'batch_size', None,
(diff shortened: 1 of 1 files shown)


jeuvilla:
Updated user email from jennifer.villa@intel.com to jennifer.e.villa@intel.com in git config. The latter email is the one under which the corporate CLA should be attributed.

jeuvilla:
I see that I still have the "cla:no" label. Should I resubmit this pull request as jennifer.e.villa@intel.com?

petewarden:
Unfortunately you'll need to rewrite the history of your commits with that email address too:
https://stackoverflow.com/questions/34850831/change-git-email-for-previous-commits?lq=1

As with many things in git, it might just be easier to create a clean repo and apply a patch with the diff from this one, unfortunately.

PR #2056, opened 2017-07-28
Added data_format flag to support MKL and other interesting tests

tfboyd:
Adds MKL support to the script by allowing the users to force the data_format.  

I also formatted and fixed all the pylint issues in these files so in the future merging will be clearning.

Up next I will remove almost all of the wrapper functions and leverage params for everything as that is what is suggested for estimators, e.g. num_gpus, variable_strategy and such.

### DIFF
--- a/tutorials/image/cifar10_estimator/cifar10_main.py
+++ b/tutorials/image/cifar10_estimator/cifar10_main.py
@@ -29,25 +29,23 @@ from __future__ import division
 from __future__ import print_function
 
 import argparse
-import collections
 import functools
 import itertools
 import os
-import six
-
-import numpy as np
-from six.moves import xrange  # pylint: disable=redefined-builtin
-import tensorflow as tf
 
 import cifar10
 import cifar10_model
 import cifar10_utils
-
+import numpy as np
+import six
+from six.moves import xrange  # pylint: disable=redefined-builtin
+import tensorflow as tf
 
 tf.logging.set_verbosity(tf.logging.INFO)
 
 
-def get_model_fn(num_gpus, variable_strategy, num_workers):
+def get_model_fn(num_gpus, variable_strategy, data_format, num_workers):
+  """Returns a function that will build the resnet model."""
   def _resnet_model_fn(features, labels, mode, params):
     """Resnet model body.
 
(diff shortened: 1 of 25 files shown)


tutorials/image/cifar10_estimator/cifar10_main.py
```diff
@@ -159,10 +160,9 @@ def _resnet_model_fn(features, labels, mode, params):
           learning_rate=learning_rate, momentum=momentum)
 
       chief_hooks = []
-      if sync:
+      if params.sync:
```
elibixby: I made `sync` an argument of the enclosing function because it shouldn't be hyperparameter tuned right? I'm not sure if maybe someone would want to let an hptuner decide on whether or not to use synchronous gradient application, but from a code perspective I wanted `params` to only contain tunable parameters, and everything else to be hardcoded in the enclosing function scope.

tfboyd: @martinwicke mentioned we want to put everything into params.  Params is not not just for hyper tuning based on my understanding of the field from Martin.  I could be very wrong.  This might be a great place to get agreement before I remove the rest of the method args.

martinwicke: Yeah, there's no strong preference for params to only be hyperparameters. In this case, putting these arguments into the params object allows us to save a level of indirection and make the code more readable. Some of the parameters are legitimate hyperparameters, some are not, but the code is going to be better if we remove the extra create_ function.

elibixby: I'm pretty ambivalent between the two approaches. But IIUC under the implementation being suggested, if you were to try to supply the `tf.HParams` given to learn_runner with a `tf.Tuner`, the Tuner would have to also supply things like `sync` because the hparams are passed through directly to the constructed `Estimator` within the `experiment_fn`, so causing this to be `params.sync` means any `tf.Tuner` must be aware of your choice of synchronous gradient application, which seems not great.

martinwicke: You either have to 
- Make sure that things like sync have sensible defaults (better)
- Update the hparams supplied with things you know you want before passing it on (ok)
- Make the tuner supply things like sync (worst)

I think making sure that things have good defaults if a param isn't supplied is ideal. Only for hyper-parameters for which defaults are not possible would it be ok to error out.

elibixby: Not sure (1) is feasible in many cases (e.g. `data_dir`), but (2) seams reasonable as long as HParams is not yet another immutable object. I was under the impression that it was, but TBF I never tested it.

martinwicke: It's not immutable. It has already set and set_from_dict, and I'm not opposed to more.

PR #2057, opened 2017-07-28
Added examples/sec hook and loss summary

tfboyd:
Adds hook that prints examples per second and a loss hook to view loss more often as that aspect of estimator is hard coded to only show loss every 100 steps, which is painful for CPU.

### DIFF
--- a/tutorials/image/cifar10_estimator/cifar10_main.py
+++ b/tutorials/image/cifar10_estimator/cifar10_main.py
@@ -36,6 +36,10 @@ import os
 import numpy as np
 from six.moves import xrange  # pylint: disable=redefined-builtin
 import tensorflow as tf
+from tensorflow.python.platform import tf_logging as logging
+from tensorflow.python.training import basic_session_run_hooks
+from tensorflow.python.training import session_run_hook
+from tensorflow.python.training import training_util
 
 import cifar10
 import cifar10_model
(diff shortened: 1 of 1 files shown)


tutorials/image/cifar10_estimator/cifar10_main.py
```diff
@@ -366,21 +433,28 @@ def main(unused_argv):
...
   sess_config.gpu_options.force_gpu_compatible = FLAGS.force_gpu_compatible
+  # Turns on XLA, which needs to be compiled into TensorFlow from source.
+  if FLAGS.xla:
+    sess_config.graph_options.optimizer_options.global_jit_level = tf.OptimizerOptions.ON_1
   config = config.replace(session_config=sess_config)
 
   classifier = tf.estimator.Estimator(
       model_fn=_resnet_model_fn, model_dir=FLAGS.model_dir, config=config)
 
-  tensors_to_log = {'learning_rate': 'learning_rate'}
```
mari-linhares: Shouldn't you keep the learning_rate as well?

tutorials/image/cifar10_estimator/cifar10_main.py
```diff
@@ -366,21 +433,24 @@ def main(unused_argv):
   sess_config.intra_op_parallelism_threads = FLAGS.num_intra_threads
   sess_config.inter_op_parallelism_threads = FLAGS.num_inter_threads
   sess_config.gpu_options.force_gpu_compatible = FLAGS.force_gpu_compatible
-  config = config.replace(session_config=sess_config)
```
mari-linhares: Why this was removed?

PR #2090, opened 2017-08-02
Adding parameter to change learning_rate

mari-linhares:


tutorials/image/cifar10_estimator/cifar10_main.py
```diff
@@ -308,7 +316,10 @@ def _resnet_model_fn(features, labels, mode):
         num_batches_per_epoch * x
         for x in np.array([82, 123, 300], dtype=np.int64)
     ]
-    staged_lr = [0.1, 0.01, 0.001, 0.0002]
+    staged_lr = [
+        FLAGS.adjust_learning_rate * x
```
tfboyd: I think people are more use to setting the starting learning rate and then having it scale.  For example.
python run_model.py --lr=.1  

.1 is the starting learning rate and you will then do .1 * .1 when you reach the first "drop" and then .01 * .1 for the second drop.

if .5 is passed, e.g. lr=.5 then the first drop would be .5 * .1 or .05 and then the second drop .05 * .1 = .005 and then .0005.

mari-linhares: Thanks Toby, will fix.

tfboyd:
LGTM, that looks acceptable.

PR #2102, opened 2017-08-03
PathNet: Evolution Channels Gradient Descent in Super Neural Networks

jaesik817:
I implemented PathNet with Tensorflow, which is to learn a huge neural networks with sparse activation and evolution channel selection, https://arxiv.org/abs/1701.08734.


jaesik817:
I signed it!

PR #2104, opened 2017-08-03
Obj_Det: From six.moves import integer_types, range

cclauss:
* `xrange()` was removed in Python 3
* Also use six to define `long` in Python 3 in a way that works for linters

@derekjchow Your thoughts on these mods?

### DIFF
--- a/object_detection/exporter_test.py
+++ b/object_detection/exporter_test.py
@@ -23,11 +23,10 @@ from object_detection.builders import model_builder
 from object_detection.core import model
 from object_detection.protos import pipeline_pb2
 
-if six.PY2:
-  import mock  # pylint: disable=g-import-not-at-top
-else:
-  from unittest import mock  # pylint: disable=g-import-not-at-top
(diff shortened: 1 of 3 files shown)


object_detection/exporter_test.py
```diff
@@ -16,18 +16,16 @@
...
 import numpy as np
-import six
 import tensorflow as tf
 from object_detection import exporter
 from object_detection.builders import model_builder
 from object_detection.core import model
 from object_detection.protos import pipeline_pb2
 
-if six.PY2:
-  import mock  # pylint: disable=g-import-not-at-top
```
derekjchow: Please re add the pylint comments.

cclauss: I am confused on this one...  All imports are now at the top of the file.  Does pylint complain?

derekjchow: Correct. The issue is because of the if/try statements.

Issue #2105, opened 2017-08-03
Does this project _require_ Python compatibility library six?


cclauss:
There are many Python 3 incompatibilities across these models.  What is the recommended way to fix them?

Is [six](https://pythonhosted.org/six) a [requirement](https://github.com/tensorflow/models/blob/master/setup.py#L7) in this repo?  Some models rely on it.  https://github.com/tensorflow/models/search?q=six while other modules like  `syntaxnet` define a constant `six`to be something other than this Python 2 / 3 compatibility module.

For example, there are several ways to make the following line both Python 2 and Python 3 compatible:
`for i in xrange(10000000):`

1. `for i in range(10000000):`
2. `from six.moves import range`
3.  
```python
try:
    xrange          # Python 2
except:
    xrange = range  # Python3
```
  
Which of these is recommended across this repo?

cclauss:
@nealwu @poxvoculi 

The Python 2 countdown clock is ticking https://pythonclock.org with just 2 and a half yearsleft.  Nice work has been done to ensure that Python 3.6 is the fastest Python ever but tensorflow models seems to have a haphazard approach to achieving Python 3 compatibility.  Could I make a few suggestions for you to consider?

* Make a repo label for __Python 3__ so we can mark issues and PRs related to Python 3 compatibility
* Pylint is great but currently does not help find Python 3 issues like flake8 does.  Can we consider using flake8 in .travis.yml?
* Can we at least make sure that __official models__ pass crucial flake8 tests on both Py2 and Py3?
    * Cog_Map is the only model that does not already pass all these tests on Python 2 #1981
    * resnet has some xrange() instances that make it fail these tests on Python 3
* Can we recommend the use of the six module in CONTRIBUTING.md #2124

Your thoughts on the importance of getting to Python 3?

martinwicke:
@av8ramit @case540 @nealwu 

We should be running py3 tests for the official models. Are we not?

av8ramit:
@martinwicke we are not as of right now. I'll add this support.

nealwu:
Merged: https://github.com/tensorflow/models/pull/2499

cclauss:
Usage of [six](https://pythonhosted.org/six) is now pervasive enough that we can close this issue.

PR #2116, opened 2017-08-04
rebar: center() is defined in utils

cclauss:
Also removed `app.run()` from `utils.py` and defined `xrange()` for Python 3

### DIFF
--- a/rebar/rebar.py
+++ b/rebar/rebar.py
@@ -26,6 +26,11 @@ import tensorflow.contrib.slim as slim
 from tensorflow.python.ops import init_ops
 import utils as U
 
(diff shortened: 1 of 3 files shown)


cclauss:
@gjtucker Thoughts on these fixes?

gjtucker:
Awesome, looks good to me. Thanks for the fixes!

lukaszkaiser:
This is not the "right way" to do it (which would be to use six), but I'm not sure if we care.

cclauss:
LOL...   Given the lack of response to #2105 , I have been reluctant to use six.

PR #2117, opened 2017-08-04
attention_ocr: define xrange() for Python 3

cclauss:
Also a dict comprehension.


wookayin:
It is far better to use `from six.moves import xrange`.

cclauss:
@wookayin In general, I agree with and use the `six.moves` approach but it adds a dependancy that some model authors might object to.  Some more conversation on #2105 would be helpful before applying `six.moves` across more models.

PR #2124, opened 2017-08-04
Recommend the use of six to CONTRIBUTING.md

cclauss:
Is [six](https://pythonhosted.org/six/) the way that we want advise contributors to use to achieve compatibility with both Python 2 and Python 3?

Explicit is better than implicit.  See #2105


nealwu:
Closing this along with https://github.com/tensorflow/models/issues/2105 thanks to your help with making most of the repo compatible with Python 3. (This also seems too specific to want to include in CONTRIBUTING.md.)

Issue #391, opened 2016-09-06
[slim] weird result with parameter is_training


argman:
I am trying to train a inception_resnet_v2 on another dataset, when I test the accuracy of the model.

``` python
with slim.arg_scope(inception_resnet_v2.inception_resnet_v2_arg_scope()):
        logits, _ = inception_resnet_v2.inception_resnet_v2(images,
                                                             num_classes=dataset.num_classes,
                                                             is_training=True)
```

setting is_training with True give better result on validation set, but from the code, i should have set this to False

argman:
@sguada I think this problem maybe with batch_norm during training and testing, there should be a better    way to solve this

argman:
I find its strange in _get_variables_to_train, I print out all the variable names, but in batch_norm, only beta like 'InceptionV3/Mixed_6b/Branch_2/Conv2d_0a_1x1/BatchNorm/beta:0' is included

argman:
It's really confusig when I try fine-tuning, can someone explain this?

sguada:
When is_training=True then batch_norm uses the statistics of the batch, when it is False it uses the moving average of the statistics computed during training. 

When doing fine-tuning sometimes it is better not update the moving average of the statistics, since you are not going to run for long. So try training with is_training=False and testing with also is_training=False.

argman:
@sguada , after I update my tf and cuda,cudnn, the problem goes away, so a suggestion maybe document your tf version and cuda, cudnn version in slim. Tf is now being developped by many and a bit complicated to debug.

And my tensorflow version is ed87884e50e1a50f7dc7b36dc7a7ff225442bee0， cuda 8.0, cudnn 5.1

argman:
@sguada , in the slim implemention, you apply moving avg to all the model parameters, but in another implemention of resnet under tensorflow/models, they just apply to batch norm parameters.

gunan:
Automatically closing due to lack of recent activity. Please update the issue when new information becomes available, and we will reopen the issue. Thanks!

ilkarman:
I still have the same issue, 2 years later

caiwd:
> I still have the same issue, 2 years later

I have the same issue, 3 years later...

caiwd:
> I find its strange in _get_variables_to_train, I print out all the variable names, but in batch_norm, only beta like 'InceptionV3/Mixed_6b/Branch_2/Conv2d_0a_1x1/BatchNorm/beta:0' is included

[...]

PR #2138, opened 2017-08-06
Evaluation should use validation datasets

Akelio-zhang:
Evaluation should use 'validation' (not 'train' datasets) datasets in the last part (Apply fine tuned model to some images.) of this walkthrough file.

### DIFF
--- a/slim/slim_walkthrough.ipynb
+++ b/slim/slim_walkthrough.ipynb
@@ -1075,7 +1075,7 @@
     "with tf.Graph().as_default():\n",
     "    tf.logging.set_verbosity(tf.logging.INFO)\n",
     "    \n",
-    "    dataset = flowers.get_split('train', flowers_data_dir)\n",
+    "    dataset = flowers.get_split('validation', flowers_data_dir)\n",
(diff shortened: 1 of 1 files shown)


Akelio-zhang:
I signed it!

Akelio-zhang:
I just signed it !

zongyuange:
@Akelio-zhang After you switched from 'train' to 'validation', also `logits,_ = inception.inception_v1(images, num_classes=dataset.num_classes, is_training=True)`, to
`logits,_ = inception.inception_v1(images, num_classes=dataset.num_classes, is_training=False)`,
did you see a significant accuracy drop?

Akelio-zhang:
@zongyuange Yes, I tried this code and the result was like your words, but I am not sure if this configuration is reasonable during evaluation.

Akelio-zhang:
@zongyuange Yes, I tested this code and the result was like your words. Besides, [here](https://github.com/tensorflow/models/issues/391#issuecomment-247392028) discussed `is_training`'s effect

zongyuange:
@Akelio-zhang, thanks for your comment, that's really helpful. 
I think TF-Slim should also update their fine-tuning example incase any other confusion may raise in the future.

PR #2146, opened 2017-08-07
Removing unecessary clipping in preprocessor.py (fix #2066)

### LINKED ISSUE #2066: Maybe a big bug in object_detection API, when I add data augmentation options on config files, then the model quickly divergences, mAP  on testset would be 0.
I am training with "ssd_inception_v2" on coco datasets, when I use the default data augmentation options in config file, which is 
```
  data_augmentation_options {
    random_horizontal_flip {
    }
  }
  data_augmentation_options {
    ssd_random_crop {
    }
  }
```
the model convergences rightly. It works very well. However, when I add one data augmentation options which is defined in `objection_detection/coco/preprocessor`, only add one `random_adjust_hue` option or  
`random_adjust_saturation` et.al. The model would  divergence quickly, even in a pretrained model， and mAP on testset would decline to 0 quickly. I add data augmentation options on config files, which looks like this format
```
  data_augmentation_options {
    random_horizontal_flip {
    }
  }
  data_augmentation_options {
    ssd_random_crop {
    }
  }

  data_augmentation_options {
    random_adjust_saturation {
    }
  }
```
image
**I tried `ssd_inception_v2`  on pascal data with the all scripts in Tensorflow Object Detection API, I met the same  problem. Without `random_adjust_brightness` or any of hue, contrast. et.al, the model works quite well, when add `random_adjust_brightness` or any of hue, contrast. et.al. It's mAP on validation set quickly decline to 0. I wonder whether those methods on `core/preporcessor.py` loss the bboxes or labels when they are applied?**

drauh:
This removes clipping operation in preprocessor.py because models require input to be in fixed-point representation.

### DIFF
--- a/research/object_detection/core/preprocessor.py
+++ b/research/object_detection/core/preprocessor.py
@@ -726,7 +726,7 @@ def random_pixel_value_scale(image,
 
   Args:
     image: rank 3 float32 tensor contains 1 image -> [height, width, channels]
-           with pixel values varying between [0, 1].
+           with pixel values varying between [0, 255].
     minval: lower ratio of scaling pixel values.
     maxval: upper ratio of scaling pixel values.
     seed: random seed.
@@ -749,7 +749,7 @@ def random_pixel_value_scale(image,
         preprocess_vars_cache)
 
     image = tf.multiply(image, color_coef)
-    image = tf.clip_by_value(image, 0.0, 1.0)
+    image = tf.clip_by_value(image, 0.0, 255.0)
 
   return image
 
@@ -815,7 +815,7 @@ def random_rgb_to_gray(image,
 
   Args:
     image: rank 3 float32 tensor contains 1 image -> [height, width, channels]
-           with pixel values varying between [0, 1].
+           with pixel values varying between [0, 255].
     probability: the probability of returning a grayscale image.
             The probability should be a number between [0, 1].
     seed: random seed.
@@ -852,11 +852,11 @@ def random_adjust_brightness(image,
                              preprocess_vars_cache=None):
   """Randomly adjusts brightness.
 
-  Makes sure the output image is still between 0 and 1.
+  Makes sure the output image is still between 0 and 255.
 
   Args:
     image: rank 3 float32 tensor contains 1 image -> [height, width, channels]
-           with pixel values varying between [0, 1].
+           with pixel values varying between [0, 255].
     max_delta: how much to change the brightness. A value between [0, 1).
     seed: random seed.
     preprocess_vars_cache: PreprocessorCache object that records previously
@@ -876,8 +876,8 @@ def random_adjust_brightness(image,
         preprocessor_cache.PreprocessorCache.ADJUST_BRIGHTNESS,
         preprocess_vars_cache)
 
-    image = tf.image.adjust_brightness(image, delta)
-    image = tf.clip_by_value(image, clip_value_min=0.0, clip_value_max=1.0)
+    image = tf.image.adjust_brightness(image / 255, delta) * 255
+    image = tf.clip_by_value(image, clip_value_min=0.0, clip_value_max=255.0)
     return image
 
 
@@ -888,11 +888,11 @@ def random_adjust_contrast(image,
                            preprocess_vars_cache=None):
   """Randomly adjusts contrast.
 
-  Makes sure the output image is still between 0 and 1.
+  Makes sure the output image is still between 0 and 255.
 
   Args:
     image: rank 3 float32 tensor contains 1 image -> [height, width, channels]
-           with pixel values varying between [0, 1].
+           with pixel values varying between [0, 255].
     min_delta: see max_delta.
     max_delta: how much to change the contrast. Contrast will change with a
                value between min_delta and max_delta. This value will be
@@ -913,8 +913,8 @@ def random_adjust_contrast(image,
         generator_func,
         preprocessor_cache.PreprocessorCache.ADJUST_CONTRAST,
         preprocess_vars_cache)
-    image = tf.image.adjust_contrast(image, contrast_factor)
-    image = tf.clip_by_value(image, clip_value_min=0.0, clip_value_max=1.0)
+    image = tf.image.adjust_contrast(image / 255, contrast_factor) * 255
+    image = tf.clip_by_value(image, clip_value_min=0.0, clip_value_max=255.0)
     return image
 
 
@@ -924,11 +924,11 @@ def random_adjust_hue(image,
                       preprocess_vars_cache=None):
   """Randomly adjusts hue.
 
-  Makes sure the output image is still between 0 and 1.
+  Makes sure the output image is still between 0 and 255.
 
   Args:
     image: rank 3 float32 tensor contains 1 image -> [height, width, channels]
-           with pixel values varying between [0, 1].
+           with pixel values varying between [0, 255].
     max_delta: change hue randomly with a value between 0 and max_delta.
     seed: random seed.
     preprocess_vars_cache: PreprocessorCache object that records previously
@@ -945,8 +945,8 @@ def random_adjust_hue(image,
     delta = _get_or_create_preprocess_rand_vars(
         generator_func, preprocessor_cache.PreprocessorCache.ADJUST_HUE,
         preprocess_vars_cache)
-    image = tf.image.adjust_hue(image, delta)
-    image = tf.clip_by_value(image, clip_value_min=0.0, clip_value_max=1.0)
+    image = tf.image.adjust_hue(image / 255, delta) * 255
+    image = tf.clip_by_value(image, clip_value_min=0.0, clip_value_max=255.0)
     return image
 
 
@@ -957,11 +957,11 @@ def random_adjust_saturation(image,
                              preprocess_vars_cache=None):
   """Randomly adjusts saturation.
 
-  Makes sure the output image is still between 0 and 1.
+  Makes sure the output image is still between 0 and 255.
 
   Args:
     image: rank 3 float32 tensor contains 1 image -> [height, width, channels]
-           with pixel values varying between [0, 1].
+           with pixel values varying between [0, 255].
     min_delta: see max_delta.
     max_delta: how much to change the saturation. Saturation will change with a
                value between min_delta and max_delta. This value will be
@@ -982,8 +982,8 @@ def random_adjust_saturation(image,
         generator_func,
         preprocessor_cache.PreprocessorCache.ADJUST_SATURATION,
         preprocess_vars_cache)
-    image = tf.image.adjust_saturation(image, saturation_factor)
-    image = tf.clip_by_value(image, clip_value_min=0.0, clip_value_max=1.0)
+    image = tf.image.adjust_saturation(image / 255, saturation_factor) * 255
+    image = tf.clip_by_value(image, clip_value_min=0.0, clip_value_max=255.0)
     return image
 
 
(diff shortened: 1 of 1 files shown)


drauh:
Hello @tombstone did you advance on this?

It seems that this fixes the problem experienced by some people here #2066

brianretford:
Ping? This would be useful to have in

plieningerweb:
Hey guys, this is a **really reallly important fix**!
Without this many many people try to find for hours their mistake in finetuning their networks, because thei get 0% mAP. I now know already more than 4 people how had the same issue and got stuck!

So this is **VERY IMPORTANT** and a very fast fix to keep users from getting frustrated!

So please review this!!!!!!!!!!

davidenitti:
@derekjchow could anyone verify this bugfix?
thank you very much

thaiat:
@davidenitti : why did you revert your change ?

davidenitti:
I had to do it temporarily now I put it back

daavoo:
I really don't understand why this is not already merged

mbenami:
Hi. 
any chance this commit will be merge sometimes?
just lost work day after updating the repo and forgetting this issue 

Thanks

liangxiao05:
@mbenami,thanks for reminding
This commit worked for me , this should be merged, anyone please come and see?

research/object_detection/core/preprocessor.py
```diff
@@ -739,6 +738,10 @@ def random_pixel_value_scale(image,
     image: image which is the same shape as input image.
   """
   with tf.name_scope('RandomPixelValueScale', values=[image]):
+    image = tf.convert_to_tensor(image, name='image')
+    # Remember original dtype to so we can convert back if needed
```
pkulzc: Please fix the typo.

research/object_detection/core/preprocessor.py
```diff
@@ -739,6 +738,10 @@ def random_pixel_value_scale(image,
...
+    orig_dtype = image.dtype
```
pkulzc: From line 728 [above](https://github.com/tensorflow/models/pull/2146/files#diff-8edc6c706ec1c943f095159ba7116e6cR728), image is already a float32 tensor so these convertions are not necessary.

You can either modify the function description or add a seperate function to deal with multiple input dtype.

drauh: I removed the conversion as it is not useful.

At the time of the writing I didn't see the conversion [here](https://github.com/tensorflow/models/blob/master/research/object_detection/trainer.py#L65). I think the main problem is that the image is in float32 but with range [0,255] so that i doesn't play well with clipping to [0,1] or tf.image.adjust_* ops.

I updated the commit with scaling by / 255 because tf.image.adjust_* need that (see the doc of [tf.image.rgb_to_hsv](https://www.tensorflow.org/api_docs/python/tf/image/rgb_to_hsv) used by tf.image.adjust_hue)

I can add a clipping to [0, 255] if you want.

research/object_detection/core/preprocessor.py
```diff
@@ -876,8 +870,7 @@ def random_adjust_brightness(image,
         preprocessor_cache.PreprocessorCache.ADJUST_BRIGHTNESS,
         preprocess_vars_cache)
 
-    image = tf.image.adjust_brightness(image, delta)
-    image = tf.clip_by_value(image, clip_value_min=0.0, clip_value_max=1.0)
+    image = tf.image.adjust_brightness(image / 255, delta) * 255
```
pkulzc: Please document that image pixel values should be in [0,255].

pkulzc: Same with all other functions.

liangxiao05: @drauh
hi，as data augamention functions needs input pixels to be [0,1] ,but the image pixel datas extracted from tf record files are between [0,255], do you know when does the ssd model precess this?I can't find the it , another thing I'm sure that the values outputs after  data augamention functions are still between [0,255]

drauh: @liangxiao05  just before images are feeded to the network [here](https://github.com/tensorflow/models/blob/master/research/object_detection/models/ssd_mobilenet_v1_feature_extractor.py#L85)

pkulzc: @drauh Right, are all the tests passing? If not, please also update preprocessor_test.py in this PR.

liangxiao05: @drauh no,it's not here .The augmentation functions are used before this 'preprocess' function. More ,this 'preprocess' function outputs are [-1,1] not [0, 1]  as  the augmentation functions required.

research/object_detection/core/preprocessor.py
```diff
@@ -982,20 +964,18 @@ def random_adjust_saturation(image,
         generator_func,
         preprocessor_cache.PreprocessorCache.ADJUST_SATURATION,
         preprocess_vars_cache)
-    image = tf.image.adjust_saturation(image, saturation_factor)
-    image = tf.clip_by_value(image, clip_value_min=0.0, clip_value_max=1.0)
+    image = tf.image.adjust_saturation(image / 255, saturation_factor) * 255
```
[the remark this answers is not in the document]
liangxiao05: any integer images to be in [0, MAX], representing fix point numbers in [0,1).
Got that ,thank you ,what do you mean representing fix point numbers in [0,1)? At calucate time , or encodes & decodes images with tf-record time?

research/object_detection/core/preprocessor.py
drauh: Sorry I misunderstood you, I should have said the conversion doesn't takes place and that was the problem.
But augmentation functions doesn't fundamentally require input to be in [0,1]. What is assumed by tf.image.adjust_* ops is that if the type is a float the range should be [0,1] and [0,255] if uint8.

One solution would be to replace the function [tf.to_float](https://github.com/tensorflow/models/blob/master/research/object_detection/trainer.py#L65) by convert_image_dtype but every preprocess functions should be changed, for instance / 255 should be removed [here](https://github.com/tensorflow/models/blob/master/research/object_detection/models/ssd_mobilenet_v1_feature_extractor.py#L85).
An other solution would be to remove this conversion, letting images to be in uint8 while doing data augmentions and then casting them to float just before feeding them to detection_model.preprocess function.

research/object_detection/core/preprocessor.py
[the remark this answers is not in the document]
FalcoGer: I don't know... [0,255]x3/4 seems a pretty standard, fast, space efficient and intuitive way to encode 24/32 bit images...
In any case, this would fix my issue with 0 precision, clearly the 'standard' way to use [0, 1) causes problems to plenty of people. can't you do some kind of format check? or force the image into a specific format when you load it so it's consistent all across? clearly it should've thrown an error right away when it read a 200 when it should be less than 1 to begin with. but instead it's clipped to 1 and i get a bunch of garbage that the network is supposed to make sense of.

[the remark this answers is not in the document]
martinwicke: Sorry that is confusing. For example, int8 images are stored as [0,255]. But conceptually, they represent values in [0,1) because they are fix-point numbers with an (implicit) factor of 1/256 (see also https://en.wikipedia.org/wiki/Fixed-point_arithmetic). 

That is how convert_image_dtype sees them, and how they're treated implicitly when handed to functions which accept multiple data types. Many preprocessing functions convert to float to perform their calculations, e.g. adjust_saturation will convert to float, perform a HSV conversion, adjust S, convert back to RGB, and then convert back to the original data type (or it used to, maybe we added direct int8 support for HSV conversions, but I don't think so).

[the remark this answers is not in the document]
martinwicke: So, the best way to deal with images is either: 

- Keep the image in int8. Be careful what you're doing and make sure that all preprocessing operations can actually work on int8 images (be careful, adjust_brightness looks like it can, but it's super-dangerous because of overflows). This is faster if you do it right, but only then.

- Immediately after reading, convert the images to float. This is generally the best options, few image augmentation pipelines are able to run well in int only.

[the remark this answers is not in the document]
pkulzc: Thanks @martinwicke for the explanation! Yes the second way is definitely prefered.

Ideally we should convert image to [0,1) at the beginning of the pipeline, but unfortunately current flow didn't follow this. We already have an ongoing effort to consolidation our interface and this issue will be gone soon.

So for now, I suggested to have this quick fix to make this model easier for git users to use.
Basically we should clearly document that these preprocessing function takes int8(or whatever) dtype as input and then use convert_image_dtype to do the conversion (instead of directly divided by 255).

Does this sound good to you @martinwicke ?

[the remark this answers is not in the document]
martinwicke: Totally. It's up to you. Just wanted to make sure the reasoning is clear. If you can move towards this it'll be good, but I definitely don't want to block this.

pkulzc:
@drauh, as the discussion above, could you please state that the input and output dtype of these functions are both int8? besides, please use convert_image_dtype functions to convert to float so that you can use tf.image.XXXX safely (and convert back to original type at the end of the function). Thank you for your patience!

drauh:
@pkulzc, I am not completely sure of what I need to do because:
- if I'm not mistaken currently image feeded to these functions are casted to float32 [here](https://github.com/tensorflow/models/blob/master/research/object_detection/trainer.py#L65) and [here](https://github.com/tensorflow/models/blob/master/research/object_detection/inputs.py#L146)
- convert_image_dtype logic (with safe saturate casting) is already implemented in tf.image.* functions, see [tf.image.adjust_brightness](https://github.com/tensorflow/tensorflow/blob/r1.7/tensorflow/python/ops/image_ops_impl.py#L1049) for instance
- because some preprocessing functions such as [random_image_scale](https://github.com/tensorflow/models/blob/master/research/object_detection/core/preprocessor.py#L767) without any casting will output float32 image (because [tf.image.resize_images](https://www.tensorflow.org/api_docs/python/tf/image/resize_images) will), i think it is better to keep everything in float32, in range [0,255] for now and in range [0,1] when you have refactored your code

In regards to that, tell me what should I do!

pkulzc:
@drauh You're right, I went through the current flow again and the data augmentations did take float [0,255] images as inputs, and it should also output float 0-255 since later feature extraction assumes images range to be [0,255]. 
Your current change looks good to me as a quick fix. We will officially fix this issue during our ongoing migration. Thanks for your patience!

pkulzc:
One more thing, please verify if preprocessor_test.py and model_lib_test.py are both passing.
I can approve this PR once confirmed. Thanks!

drauh:
I did not find model_lib_test.py but for preprocessor_test.py it is passing:
```
================================================ warnings summary =================================================
object_detection/core/preprocessor_test.py::PreprocessorTest::testRunRandomCropImageWithKeypointsOutsideCrop
  /home/malo/next/bin/anaconda3/envs/py2/lib/python2.7/site-packages/tensorflow/python/framework/test_util.py:1141: RuntimeWarning: invalid value encountered in greater
    np.abs(a - b) > atol + rtol * np.abs(b),

-- Docs: http://doc.pytest.org/en/latest/warnings.html
===================================== 96 passed, 1 warnings in 79.62 seconds ======================================
```

pkulzc:
Please sync your fork to latest.

drauh:
Thanks, all test are passing.

preprocessor_test.py
```
================================================= warnings summary =================================================
object_detection/core/preprocessor_test.py::PreprocessorTest::testRunRandomCropImageWithKeypointsOutsideCrop
  /home/malo/next/bin/anaconda3/envs/py2/lib/python2.7/site-packages/tensorflow/python/framework/test_util.py:1141: RuntimeWarning: invalid value encountered in greater
    np.abs(a - b) > atol + rtol * np.abs(b),

-- Docs: http://doc.pytest.org/en/latest/warnings.html
===================================== 100 passed, 1 warnings in 86.56 seconds ======================================
```

model_lib_test.py
```
=========================================== 13 passed in 157.81 seconds ============================================
```

pkulzc:
Thanks @drauh , will merge this once all our internal tests pass.

liangxiao05:
@pkulzc .Excuse, I have one more question in this script.It seems to be that the  'ssd_random_crop' function doesn't judge whether the center of the GT is still remaining in the cropped image as the ssd paper said 
> We keep the overlapped part of the ground truth box if the center of it is in the sampled patch
> 
The related codes  here https://github.com/tensorflow/models/blob/master/research/object_detection/core/preprocessor.py#L1106
```
    # remove boxes that are outside cropped image
    boxlist, inside_window_ids = box_list_ops.prune_completely_outside_window(
        boxlist, im_box_rank1)

    # remove boxes that are outside image
    overlapping_boxlist, keep_ids = box_list_ops.prune_non_overlapping_boxes(
        boxlist, im_boxlist, overlap_thresh)
```
Is that my  understanding  wrong and you have implement it somewhere else,or you find that it's unnecessary to implement it ?

PR #2152, opened 2017-08-08
Fix import paths and GPUOptions initialization

elibixby:
Import paths are necessary for this code to work when packaged as a tar with sdist (which is the distribution strategy for CMLE).

Also explicit relative imports are considered best practices (and required when `__future__.absolute_import` is used)

Finally there was a broken property assignment before message initialization in the config options.

Would it be possible to add an `__init__.py` to this directory as well? (this would prevent users from having to add their own for CMLE).

With all these working, I've verified that this works using the following `job_config.yaml`:

```
trainingInput:
  scaleTier: CUSTOM
  masterType: complex_model_m_gpu
  workerType: complex_model_m_gpu
  parameterServerType: complex_model_m
```

And the following command:

```
gcloud ml-engine jobs submit training cifarmultigpu6 \
    --runtime-version 1.2 \
    --staging-bucket gs://cifarmultigpu \
    --config cifar10_estimator/job_config.yaml \
    --package-path cifar10_estimator/ \
    --region us-central1 \
    --module-name cifar10_estimator.cifar10_main \
    -- \
    --data_dir=gs://cifarmultigpu/train_data \
    --model_dir=gs://cifarmultigpu/modeldirs/6 \
    --is_cpu_ps=True \
    --force_gpu_compatible=True \
    --num_gpus=4 \
    --train_steps=1000 \
    --run_experiment=True
```

Some other recommendations to make this more idiomatic:
* use argparse not FLAGS, so that flags will use `-`
* What's the purpose of run_experiment? It should work fine on single worker so no flag is necessary
* Can GPU number not be detected from the environment? Maybe can be passed as a Tensor using `tf.Scaffold`?

@mari-linhares

### DIFF
--- a/tutorials/image/cifar10_estimator/README.md
+++ b/tutorials/image/cifar10_estimator/README.md
@@ -17,62 +17,74 @@ Before trying to run the model we highly encourage you to read all the README.
 2. Download the CIFAR-10 dataset.
 
 ```shell
-$ curl -o cifar-10-python.tar.gz https://www.cs.toronto.edu/~kriz/cifar-10-python.tar.gz
-$ tar xzf cifar-10-python.tar.gz
+curl -o cifar-10-python.tar.gz https://www.cs.toronto.edu/~kriz/cifar-10-python.tar.gz
+tar xzf cifar-10-python.tar.gz
 ```
 
 After running the commands above, you should see the following files in the folder where the data was downloaded.
 
 ``` shell
-$ ls -R cifar-10-batches-py
+ls -R cifar-10-batches-py
+```
 
+The output should be:
+
+
+```
 batches.meta  data_batch_1  data_batch_2  data_batch_3
 data_batch_4  data_batch_5  readme.html  test_batch
 ```
 
 3. Generate TFRecord files.
+This will generate a tf record for the training and test data available at the input_dir.
+You can see more details in `generate_cifar10_tf_records.py`
+
 ```shell
-# This will generate a tf record for the training and test data available at the input_dir.
-# You can see more details in generate_cifar10_tf_records.py
-$ python generate_cifar10_tfrecords.py --input-dir=/prefix/to/downloaded/data/cifar-10-batches-py \
-                                       --output-dir=/prefix/to/downloaded/data/cifar-10-batches-py
+python generate_cifar10_tfrecords.py --input-dir=${PWD}/cifar-10-batches-py \
+                                     --output-dir=${PWD}/cifar-10-batches-py
 ```
 
 After running the command above, you should see the following new files in the output_dir.
 
 ``` shell
-$ ls -R cifar-10-batches-py
+ls -R cifar-10-batches-py
+```
 
+```
 train.tfrecords validation.tfrecords eval.tfrecords
 ```
 
 ## How to run on local mode
 
+Run the model on CPU only. After training, it runs the evaluation.
+
+```
+python cifar10_main.py --data-dir=${PWD}/cifar-10-batches-py \
+                       --job-dir=/tmp/cifar10 \
+                       --num-gpus=0 \
+                       --train-steps=1000
 ```
 
-# Run the model on CPU only. After training, it runs the evaluation.
-$ python cifar10_main.py --data-dir=/prefix/to/downloaded/data/cifar-10-batches-py \
-                         --job-dir=/tmp/cifar10 \
-                         --num-gpus=0 \
-                         --train-steps=1000
-
-# Run the model on 2 GPUs using CPU as parameter server. After training, it runs the evaluation.
-$ python cifar10_main.py --data-dir=/prefix/to/downloaded/data/cifar-10-batches-py \
-                         --job-dir=/tmp/cifar10 \
-                         --num-gpus=2 \
-                         --train-steps=1000
-
-# Run the model on 2 GPUs using GPU as parameter server.
-# It will run an experiment, which for local setting basically means it will run stop training
-# a couple of times to perform evaluation.
-$ python cifar10_main.py --data-dir=/prefix/to/downloaded/data/cifar-10-batches-bin \
-                         --job-dir=/tmp/cifar10 \
-                         --variable-strategy GPU \
-                         --num-gpus=2 \
-
-
-# There are more command line flags to play with; check cifar10_main.py for details.
+Run the model on 2 GPUs using CPU as parameter server. After training, it runs the evaluation.
 ```
+python cifar10_main.py --data-dir=${PWD}/cifar-10-batches-py \
+                       --job-dir=/tmp/cifar10 \
+                       --num-gpus=2 \
+                       --train-steps=1000
+```
+
+Run the model on 2 GPUs using GPU as parameter server.
+It will run an experiment, which for local setting basically means it will run stop training
+a couple of times to perform evaluation.
+
+```
+python cifar10_main.py --data-dir=${PWD}/cifar-10-batches-bin \
+                       --job-dir=/tmp/cifar10 \
+                       --variable-strategy GPU \
+                       --num-gpus=2 \
+```
+
+There are more command line flags to play with; run `python cifar10_main.py --help` for details.
 
 ## How to run on distributed mode
 
@@ -86,7 +98,7 @@ You'll also need a Google Cloud Storage bucket for the data. If you followed the
 
 ```
 MY_BUCKET=gs://<my-bucket-name>
-gsutil cp -r cifar-10-batches-py $MY_BUCKET/
+gsutil cp -r ${PWD}/cifar-10-batches-py $MY_BUCKET/
 ```
 
 Then run the following command from the `tutorials/image` directory of this repository (the parent directory of this README):
@@ -172,18 +184,19 @@ By the default environment is *local*, for a distributed setting we need to chan
 Once you have a `TF_CONFIG` configured properly on each host you're ready to run on distributed settings.
 
 #### Master
+Run this on master:
+Runs an Experiment in sync mode on 4 GPUs using CPU as parameter server for 40000 steps.
+It will run evaluation a couple of times during training.
+The num_workers arugument is used only to update the learning rate correctly.
+Make sure the model_dir is the same as defined on the TF_CONFIG.
+
 ```shell
-# Run this on master:
-# Runs an Experiment in sync mode on 4 GPUs using CPU as parameter server for 40000 steps.
-# It will run evaluation a couple of times during training.
-# The num_workers arugument is used only to update the learning rate correctly.
-# Make sure the model_dir is the same as defined on the TF_CONFIG.
-$ python cifar10_main.py --data-dir=gs://path/cifar-10-batches-py \
-                         --job-dir=gs://path/model_dir/ \
-                         --num-gpus=4 \
-                         --train-steps=40000 \
-                         --sync \
-                         --num-workers=2
+python cifar10_main.py --data-dir=gs://path/cifar-10-batches-py \
+                       --job-dir=gs://path/model_dir/ \
+                       --num-gpus=4 \
+                       --train-steps=40000 \
+                       --sync \
+                       --num-workers=2
 ```
 
 *Output:*
@@ -313,16 +326,17 @@ INFO:tensorflow:Saving dict for global step 1: accuracy = 0.0994, global_step = 
 
 #### Worker
 
+Run this on worker:
+Runs an Experiment in sync mode on 4 GPUs using CPU as parameter server for 40000 steps.
+It will run evaluation a couple of times during training.
+Make sure the model_dir is the same as defined on the TF_CONFIG.
+
 ```shell
-# Run this on worker:
-# Runs an Experiment in sync mode on 4 GPUs using CPU as parameter server for 40000 steps.
-# It will run evaluation a couple of times during training.
-# Make sure the model_dir is the same as defined on the TF_CONFIG.
-$ python cifar10_main.py --data-dir=gs://path/cifar-10-batches-py \
-                         --job-dir=gs://path/model_dir/ \
-                         --num-gpus=4 \
-                         --train-steps=40000 \
-                         --sync
+python cifar10_main.py --data-dir=gs://path/cifar-10-batches-py \
+                       --job-dir=gs://path/model_dir/ \
+                       --num-gpus=4 \
+                       --train-steps=40000 \
+                       --sync
 ```
 
 *Output:*
@@ -428,12 +442,11 @@ INFO:tensorflow:loss = 27.8453, step = 179 (18.893 sec)
 
 #### PS
 
-```shell
-# Run this on ps:
-# The ps will not do training so most of the arguments won't affect the execution
-$ python cifar10_main.py --job-dir=gs://path/model_dir/
+Run this on ps:
+The ps will not do training so most of the arguments won't affect the execution
 
-# There are more command line flags to play with; check cifar10_main.py for details.
+```shell
+python cifar10_main.py --job-dir=gs://path/model_dir/
 ```
 
 *Output:*
@@ -460,11 +473,12 @@ When using Estimators you can also visualize your data in TensorBoard, with no c
 
 You'll see something similar to this if you "point" TensorBoard to the `model_dir` you used to train or evaluate your model.
 
+Check TensorBoard during training or after it.
+Just point TensorBoard to the model_dir you chose on the previous step
+by default the model_dir is "sentiment_analysis_output"
+
 ```shell
-# Check TensorBoard during training or after it.
-# Just point TensorBoard to the model_dir you chose on the previous step
-# by default the model_dir is "sentiment_analysis_output"
-$ tensorboard --log-dir="sentiment_analysis_output"
+tensorboard --log-dir="sentiment_analysis_output"
 ```
 
 ## Warnings
--- a/tutorials/image/cifar10_estimator/README.md
+++ b/tutorials/image/cifar10_estimator/README.md
@@ -73,8 +73,7 @@ $ python cifar10_main.py --data-dir=/prefix/to/downloaded/data/cifar-10-batches-
                          --is-cpu-ps=False \
                          --force-gpu-compatible=True \
                          --num-gpus=2 \
-                         --train-steps=1000
-                         
+
 
 # There are more command line flags to play with; check cifar10_main.py for details.
 ```
@@ -99,19 +98,16 @@ Then run the following command from the `tutorials/image` directory of this repo
 ```
 gcloud ml-engine jobs submit training cifarmultigpu \
     --runtime-version 1.2 \
-    --staging-bucket $MY_BUCKET \
-    --config cifar10_estimator/job_config.yaml \
+    --job-dir=$MY_BUCKET/model_dirs/cifarmultigpu \
+    --config cifar10_estimator/cmle_config.yaml \
     --package-path cifar10_estimator/ \
-    --region us-central1 \
     --module-name cifar10_estimator.cifar10_main \
     -- \
     --data-dir=$MY_BUCKET/cifar-10-batches-py \
-    --job-dir=$MY_BUCKLET/model_dirs/cifarmultigpu \
     --is-cpu-ps=True \
     --force-gpu-compatible=True \
     --num-gpus=4 \
-    --train-steps=1000 \
-    
+    --train-steps=1000
 ```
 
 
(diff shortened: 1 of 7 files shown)


mari-linhares:
* *use argparse not FLAGS, so that flags will use -*

  I don't have an opinion about this...

* *What's the purpose of run_experiment? It should work fine on single worker so no flag is necessary*
  
  We can change it to be True by default (I think I forgot to this at some point), I think is good to have a flag to let people know this is not mandatory. Feel free to make this change on this patch.

* *Can GPU number not be detected from the environment? Maybe can be passed as a Tensor using tf.Scaffold?*

Other TensorFlow code implementations that do multi-gpu use a FLAG, I think we can keep it. But if everybody agrees to check from the environment we can do it.

elibixby:
Any thoughts on adding an empty `__init__.py` file to this directory? Otherwise CMLE users will have to add their own.

mari-linhares:
@elibixby once you try the model on CMLE add these instructions on README

elibixby:
Verified this works multi-GPU + Data Parallel. I'll add the job_config.yaml to the repo and the command line instructions to the README if that works for you?

mari-linhares:
@tfboyd any opinions on these changes?

elibixby:
on dashes vs underscores in flags, CMLE uses `--job-dir` rather than `--model_dir` which it validates access to in GCS, and namespaces automatically in the case of a hyperparameter tuning job, then passes through to the user code as `--job-dir`. So for hyperparameter tuning to work on CMLE (without a crufty workaround in the code) Need to use `--job-dir` rather than `--model_dir`. AFAIK `tf.flags` doesn't support dashes in flag names (besides being inferior to argparse in a bunch of other ways :P

elibixby:
@tfboyd @mari-linhares ping on this?

tfboyd:
Sorry, looking at it now.  I am not use to being in the critical review path.  :-)

tfboyd:
Yes we should use argparse that is what the garden used and I (we) want to
follow their lead.  That would be great thank you Eli.

elibixby:
Additionally something I'd love to see added to more examples is an export strategy (and a slimmed down graph for serving). We're slowly trying to move towards "If it doesn't demonstrate inference it's not a complete example" world, although this example is pretty full anyway. I'll try to add in a followup PR when I get around to it.

tfboyd:
Total agreement.  If we fill this example (and then future examples) with
all the necessary parts then it is easier for people to copy the style for
their examples.  As well as push for simpler high level APIs to make these
things more automatic and less code.  Thank you so much Eli.

elibixby:
Hey, converting this to argparse has lead me down a pretty deep rabbit hole. One question I have. Is `--is-cpu-ps=False` even compatible with a multi-worker scenario? My suspicion is no... need to confirm, AFAIK multi-worker TF jobs require a dedicated parameter server to serve as the source of truth for variables and variable updates.

Given this, and to simplify this tutorial greatly, I would strongly suggest completely removing both the `--run-experiment` and `--is-cpu-ps` flags entirely. 

Thoughts?

mari-linhares:
Hi @elibixby,

 I think `--is_cpu_ps=False` can work with multi-worker (You can try to check), what would happen is that for each worker the local parameters would be stored on the GPUs instead of the CPU, but an external PS would still be needed.

tfboyd:
is_cpu_ps should be changed to variable update style or a better name and
not true false.  You could be using All Reduce, replicated, single GPU, GPU
balance, CPU and then getting interesting you can aggregate locally and
then update the remote parameter server.  I am honestly not sure what
experiments hijacks but I suspect it overrides the device setter and all of
the parametsr are placed on the ps_server.

If Cloud forces people to have a remote parameter server that is a separate
process you may have some problems with performance if running on 8 GPUs or
less.  This would likely get worse with P100s unless other strategies for
variable update appear than what we are currently using.

mari-linhares:
Oh @elibixby about `--run_experiments` I think this can be removed.

elibixby:
> is_cpu_ps should be changed to variable update style or a better name and
not true false.

Yep, in my draft commit I changed it to `--avg-on-gpu` (flipping the boolean to allow the flag to be argumentless). Does this seem like a reasonable name? Help text is `"Average gradients on GPU"`

>I am honestly not sure what
>experiments hijacks but I suspect it overrides the device setter and all of
>the parametsr are placed on the ps_server

Yes it sets the device setter to `tf.replica_device_setter` here: 
https://github.com/tensorflow/tensorflow/blob/master/tensorflow/python/estimator/estimator.py#L627 

Not sure how that interacts with existing device contexts? Do they stack? It's possible if the replica device setter only enforces placement on a task level, that this could still be fine. But seems unlikely to me. 

>If Cloud forces people to have a remote parameter server that is a separate
>process you may have some problems with performance if running on 8 GPUs or
>less. 

Is there anyway to do multi-machine training without a separate PS? Not sure what the alternative is here. For the BASIC tier I don't think a remote PS is used, so that's an option, for small numbers of GPUs, (I think the smallest distributed GPU tier is 4 GPUs anyway).

In any case, it doesn't look like `--avg-on-gpu` works with `--run-experiments` anyway, I'll investigate to make sure, but if it doesn't I'd argue for removing both (makes it much easier anyway).

tfboyd:
I am 99% sure it will overwrite and the variables will be placed on the
remote ps_server.  If not the script would not work in distributed.

--"Average gradients on GPU" does not really descibe what the option does
for multi-GPU single node.

In regard to the variable update name you are hitting the same problem we
have with our benchmarks and our solution can confuse users and is as
follows
-  variable_update:  which can be parameter_server or replicated or
distributed_replicated
- local_parameter_device: which can be gpu or cpu
- and use_nccl if we want that option.

The remote ps-server is always CPU (although we have run them GPU before)
as we usually just run it local with the worker as K80s usually are paired
with more than enough CPU.

For this script we are only supporting a couple different variable_update
options but there is work to add the others and it would be nice to not
have to change the signature of the script.  I will think about how to
better handle the variable update, e.g. the device where the variables are
set.  When we created the script it was not setup to do distributed.  I
think this shows how complex the story is and the high level APIs do not
seem aligned unless I am not following them.

elibixby:
Yes. My reading is it will be placed on `'/job:ps/task:X/gpu:Y'` if it's multi-worker and `--is_cpu_ps=False`. Because the `replica_device_setter` will set is to `'/job:ps/task:X'` and then the inner context will set the device to `/gpu:0`.

But this seems like non-desired behavior.

elibixby:
@tfboyd Just committed https://github.com/tensorflow/models/pull/2152/commits/8b829873a58ec0e353a243d656fc9291fef432d0 PTAL. Made some additional stylistic changes I'll happily reverse if you don't like. Will try to keep the style changes in separate commits in the future so they are easy to revert in the future.

tutorials/image/cifar10_estimator/cifar10_main.py
```diff
@@ -220,147 +145,172 @@ def __call__(self, op):
     return device_name
 
 
-def _create_device_setter(is_cpu_ps, worker, num_gpus):
+def _create_device_setter(avg_on_gpu, worker, num_gpus):
```
tfboyd: While silly, although not always silly you can put the variables on the GPUs and do the ops on CPU.  When you do replicated variables that is totally an option and in some amusing cases faster.

tutorials/image/cifar10_estimator/README.md
```diff
@@ -154,15 +181,14 @@ Once you have a `TF_CONFIG` configured properly on each host you're ready to run
...
-                         --sync=True \
-                         --run_experiment=True \
-                         --num_workers=2
+$ python cifar10_main.py --data-dir=gs://path/cifar-10-batches-py \
+                         --job-dir=gs://path/model_dir/ \
+                         --force-gpu-compatible \
+                         --num-gpus=4 \
+                         --train-steps=40000 \
+                         --sync \
+                          \
```
mari-linhares: Do we need this line?

tutorials/image/cifar10_estimator/cifar10_main.py
```diff
@@ -460,86 +459,196 @@ def _experiment_fn(run_config, hparams):
...
+      help="""\
+      If present when running in a distributed environment will run on sync mode.\
+      """
+  )
+  parser.add_argument(
+      '--num-intra-threads',
+      type=int,
+      default=1,
+      help="""\
+      Number of threads to use for intra-op parallelism. If set to 0, the
```
elibixby: Looks fine to me:

```
usage: cifar10_main.py [-h] --data-dir DATA_DIR --job-dir JOB_DIR
                       [--avg-on-gpu] [--num-gpus NUM_GPUS]
                       [--num-layers NUM_LAYERS] [--train-steps TRAIN_STEPS]
                       [--train-batch-size TRAIN_BATCH_SIZE]
                       [--eval-batch-size EVAL_BATCH_SIZE]
                       [--momentum MOMENTUM] [--weight-decay WEIGHT_DECAY]
                       [--learning-rate LEARNING_RATE]
                       [--use-distortion-for-training USE_DISTORTION_FOR_TRAINING]
                       [--sync] [--num-intra-threads NUM_INTRA_THREADS]
                       [--num-inter-threads NUM_INTER_THREADS]
                       [--force-gpu-compatible] [--log-device-placement]
                       [--batch_norm_decay BATCH_NORM_DECAY]
                       [--batch_norm_epsilon BATCH_NORM_EPSILON]

optional arguments:
  -h, --help            show this help message and exit
  --data-dir DATA_DIR   The directory where the CIFAR-10 input data is stored.
  --job-dir JOB_DIR     The directory where the model will be stored.
  --avg-on-gpu          If present, use GPU to average gradients.
  --num-gpus NUM_GPUS   The number of gpus used. Uses only CPU if set to 0.
  --num-layers NUM_LAYERS
                        The number of layers of the model.
  --train-steps TRAIN_STEPS
                        The number of steps to use for training.
  --train-batch-size TRAIN_BATCH_SIZE
                        Batch size for training.
  --eval-batch-size EVAL_BATCH_SIZE
                        Batch size for validation.
  --momentum MOMENTUM   Momentum for MomentumOptimizer.
  --weight-decay WEIGHT_DECAY
                        Weight decay for convolutions.
  --learning-rate LEARNING_RATE
                        This is the inital learning rate value. The learning
                        rate will decrease during training. For more details
                        check the model_fn implementation in this file.
  --use-distortion-for-training USE_DISTORTION_FOR_TRAINING
                        If doing image distortion for training.
  --sync                If present when running in a distributed environment
                        will run on sync mode.
  --num-intra-threads NUM_INTRA_THREADS
                        Number of threads to use for intra-op parallelism. If
                        set to 0, the system will pick an appropriate number.
                        The default is 1 since in this example CPU only
                        handles the input pipeline and gradient aggregation
                        (when --is-cpu-ps). Ops that could potentially benefit
                        from intra-op parallelism are scheduled to run on
                        GPUs.
  --num-inter-threads NUM_INTER_THREADS
                        Number of threads to use for inter-op parallelism. If
                        set to 0, the system will pick an appropriate number.
  --force-gpu-compatible
                        Whether to enable force_gpu_compatible in GPU_Options.
                        Check tensorflow/core/protobuf/config.proto#L69 for
                        details.
  --log-device-placement
                        Whether to log device placement.
  --batch_norm_decay BATCH_NORM_DECAY
                        Decay for batch norm.
  --batch_norm_epsilon BATCH_NORM_EPSILON
                        Epsilon for batch norm.
```

tutorials/image/cifar10_estimator/README.md
elibixby: Woops, no good catch. Removed

tutorials/image/cifar10_estimator/cifar10_main.py
elibixby: Yes, in fact, I think this bit of code could be replaced by the following, which I believe is equivalent! 
```
placement_strat = tf.contrib.training.GreedyLoadBalancingStrategy(
  num_ps, load_fn=tf.contrib.training.byte_size_load_fn)
with tf.device(tf.train.replica_device_setter(worker_device=worker,
                                                                    ps_strategy=placement_strat)):
   # tower
```
Which I think should be almost identical (your device_fn uses number of elements rather than byte size, but if they are all the same type those should be equivalent).

tutorials/image/cifar10_estimator/cifar10_main.py
tfboyd: Cool I can tell the person WTF next time.  Looks good to me as well.

tutorials/image/cifar10_estimator/cifar10_main.py
tfboyd: I looked at that before and I could not see where it was passing in the possible ps_servers to use for the balancing.  I am fine with bytes, that might even work better depending on the situation.  If it works I am ok simplifying my examples and the code

elibixby: RE: `--avg-on-gpu` am I correct in noting that there are two separate things being controlled by this flag?

(1) whether variables are placed on GPU or not.
(2) Whether the averaging ops are placed on GPU or not.

Note for (2) I'm still a little confused, because there appear to be two different device setters, one at 236, and one that uses `var.device` at 243. If I understand device contexts properly using explicit devices rather than device functions makes the implicit device function `merge_devices`, which is conservative. So the averaging will always happen on the device the variables are located on (which makes sense, so that the final gradients are colocated with the device on which the update op takes place). So I'm not quite sure what the outer device context is doing? Maybe a few concats and an average of loss? But I think it would be clearer if they weren't nested. Given that the variable devices are going to be fully specified, the outer context will always be overridden by the inner context.

tutorials/image/cifar10_estimator/README.md
```diff
@@ -34,8 +34,8 @@ data_batch_4  data_batch_5  readme.html  test_batch
 ```shell
 # This will generate a tf record for the training and test data available at the input_dir.
 # You can see more details in generate_cifar10_tf_records.py
-$ python generate_cifar10_tfrecords.py --input_dir=/prefix/to/downloaded/data/cifar-10-batches-py \
-                                       --output_dir=/prefix/to/downloaded/data/cifar-10-batches-py
+$ python generate_cifar10_tfrecords.py --input-dir=/prefix/to/downloaded/data/cifar-10-batches-py \
+                                       --output-dir=/prefix/to/downloaded/data/cifar-10-batches-py
```
tfboyd: I know this was here but I would rather this happened automatically in one step rather than two. Or at a minimum if you copy and paste all the commands you get a result that works.

tutorials/image/cifar10_estimator/cifar10_main.py
```diff
@@ -460,86 +364,186 @@ def _experiment_fn(run_config, hparams):
...
+      help='The directory where the CIFAR-10 input data is stored.'
+  )
+  parser.add_argument(
+      '--job-dir',
+      type=str,
+      required=True,
+      help='The directory where the model will be stored.'
+  )
+  parser.add_argument(
+      '--variable_strategy',
```
tfboyd: should this be `variable-strategy` to follow the format you were using.

tutorials/image/cifar10_estimator/README.md
elibixby: Agree. Lets replace this prefix with $PWD by default.

tutorials/image/cifar10_estimator/cifar10_utils.py
```diff
@@ -0,0 +1,137 @@
+import six
+
```
tfboyd: need to add.  Not sure why it did not error on you.  
import collections  
import tensorflow as tf  <-  due to tf.contrib.learn.RunConfig

tutorials/image/cifar10_estimator/cifar10_main.py
```diff
@@ -438,108 +279,261 @@ def input_fn(subset, num_shards):
...
+  if args.num_gpus == 0 and args.variable_strategy == 'GPU':
     raise ValueError(
-        'No GPU available for use, must use CPU as parameter server.')
-  if (FLAGS.num_layers - 2) % 6 != 0:
+        'No GPU available for use, must use CPU to average gradients.')
+  if (args.num_layers - 2) % 6 != 0:
     raise ValueError('Invalid num_layers parameter.')
-  if FLAGS.num_gpus != 0 and FLAGS.train_batch_size % FLAGS.num_gpus != 0:
+  if args.num_gpus != 0 and args.train_batch_size % args.num_gpus != 0:
     raise ValueError('train_batch_size must be multiple of num_gpus.')
```
elibixby: Argparse automatically converts `-` to `_` in variable names.

tutorials/image/cifar10_estimator/cifar10_main.py
```diff
@@ -25,342 +25,187 @@
...
-
-def _tower_fn(is_training, weight_decay, feature, label, tower_losses,
-              tower_gradvars, tower_preds, is_cpu):
+
+    # Device that runs the ops to apply global gradient updates.
+    consolidation_device = '/gpu:0' if variable_strategy == 'GPU' else '/cpu:0'
+    with tf.device(consolidation_device):
+      # Suggested learning rate scheduling from
+      # https://github.com/ppwwyyxx/tensorpack/blob/master/examples/ResNet/cifar10-resnet.py#L155
+      # users could apply other scheduling.
```
elibixby: Done

tutorials/image/cifar10_estimator/cifar10_main.py
```diff
@@ -438,108 +279,261 @@ def input_fn(subset, num_shards):
...
+  )
+  parser.add_argument(
+      '--num-intra-threads',
+      type=int,
+      default=1,
+      help="""\
+      Number of threads to use for intra-op parallelism. If set to 0, the
+      system will pick an appropriate number. The default is 1 since in this
+      example CPU only handles the input pipeline and gradient aggregation
+      (when --is-cpu-ps). Ops that could potentially benefit from intra-op
```
elibixby: Done

tutorials/image/cifar10_estimator/cifar10_main.py
```diff
@@ -438,108 +279,261 @@ def input_fn(subset, num_shards):
...
+
+  if args.num_gpus < 0:
     raise ValueError(
         'Invalid GPU count: \"num_gpus\" must be 0 or a positive integer.')
-  if FLAGS.num_gpus == 0 and not FLAGS.is_cpu_ps:
+  if args.num_gpus == 0 and args.variable_strategy == 'GPU':
     raise ValueError(
-        'No GPU available for use, must use CPU as parameter server.')
-  if (FLAGS.num_layers - 2) % 6 != 0:
+        'No GPU available for use, must use CPU to average gradients.')
```
elibixby: Done.

tutorials/image/cifar10_estimator/cifar10_main.py
```diff
@@ -438,108 +279,261 @@ def input_fn(subset, num_shards):
...
+      '--batch-norm-epsilon',
+      type=float,
+      default=1e-5,
+      help='Epsilon for batch norm.'
+  )
+  args = parser.parse_args()
+
+  if args.num_gpus < 0:
     raise ValueError(
         'Invalid GPU count: \"num_gpus\" must be 0 or a positive integer.')
```
elibixby: Done

tutorials/image/cifar10_estimator/cifar10_main.py
```diff
@@ -438,108 +279,261 @@ def input_fn(subset, num_shards):
...
-  if (FLAGS.num_layers - 2) % 6 != 0:
+        'No GPU available for use, must use CPU to average gradients.')
+  if (args.num_layers - 2) % 6 != 0:
     raise ValueError('Invalid num_layers parameter.')
-  if FLAGS.num_gpus != 0 and FLAGS.train_batch_size % FLAGS.num_gpus != 0:
+  if args.num_gpus != 0 and args.train_batch_size % args.num_gpus != 0:
     raise ValueError('train_batch_size must be multiple of num_gpus.')
-  if FLAGS.num_gpus != 0 and FLAGS.eval_batch_size % FLAGS.num_gpus != 0:
+  if args.num_gpus != 0 and args.eval_batch_size % args.num_gpus != 0:
     raise ValueError('eval_batch_size must be multiple of num_gpus.')
```
elibixby: Done.

tfboyd:
I tested the changes and here are the results, which are inline with the pre-pr results.

Run on a GTX 1080
- 2 GPU variable_strategy=GPU batch-size 64= .927
- 1 GPU variable_strategy=GPU batch-size 64 = .9323
- 1 GPU batch size 128 = .914
- 2 GPU batch size 128 = .9258

tfboyd:
I will fix the format and lint issues in a PR ASAP.

Issue #2157, opened 2017-08-08
Can't load the pretrained encoder fo training the decoder of Perspective Transformer Nets


mees:
### System information
- **What is the top-level directory of the model you are using**: Ptn directory
- **Have I written custom code (as opposed to using a stock example script provided in TensorFlow)**: No
- **OS Platform and Distribution (e.g., Linux Ubuntu 16.04)**: Ubuntu 14.04
- **TensorFlow installed from (source or binary)**: Source 
- **TensorFlow version (use command below)**: ('v1.3.0-rc1-27-g2784b1c', '1.3.0-rc2')
- **Bazel version (if compiling from source)**: 0.5.2
- **CUDA/cuDNN version**: 8.0/5.1
- **GPU model and memory**: Titan X
- **Exact command to reproduce**: run -c opt :train_ptn -- --init_model='/home/meeso/models/ptn/my_models/deeprotator_pretrain/train/' --checkpoint_dir='/home/meeso/models/ptn/my_models/' --inp_dir='/home/meeso/Datasets/shapenet_tf/

### Describe the problem

the init_model parameter is supposed to be absolute path of the folder where the pre-trained encoder is located. However, when you try to train the volumetric decoder, it doesn't find the model and crashes with following error.

 
### Source code / logs
```
INFO: Running command line: bazel-bin/train_ptn '--init_model=/home/meeso/models/ptn/my_models/deeprotator_pretrain/train/' '--checkpoint_dir=/home/meeso/models/ptn/my_models/' '--inp_dir=/home/meeso/Datasets/shapenet_tf/'
Traceback (most recent call last):
  File "/home/meeso/.cache/bazel/_bazel_meeso/3a6bfcc5e3fa6521c178dee287de79bc/execroot/__main__/bazel-out/local-opt/bin/train_ptn.runfiles/__main__/train_ptn.py", line 230, in <module>
    app.run()
  File "/home/meeso/.virtualenvs/2d_to_3d/local/lib/python2.7/site-packages/tensorflow/python/platform/app.py", line 48, in run
    _sys.exit(main(_sys.argv[:1] + flags_passthrough))
  File "/home/meeso/.cache/bazel/_bazel_meeso/3a6bfcc5e3fa6521c178dee287de79bc/execroot/__main__/bazel-out/local-opt/bin/train_ptn.runfiles/__main__/train_ptn.py", line 210, in main
    init_fn = model.get_init_fn(init_scopes)
[...]

PR #2159, opened 2017-08-08
fixes bug #2157, loads the pretrained encoder to finetune the decoder

mees:
bug #2157 didn't load the pretrained decoder when training  the volumetric decoder. This patch loads the latest checkpoint model from the given folder.


mees:
@xcyan, here the pull request

PR #2167, opened 2017-08-09
Update Attention OCR README.md to point to a fixed checkpoint file.

alexgorban:
The existing checkpoint for the model
http://download.tensorflow.org/models/attention_ocr_2017_05_17.tar.gz
is now broken because of tensor name changes in the attention decoder.
I applied the fix and sent request to @martinwicke to upload it as http://download.tensorflow.org/models/attention_ocr_2017_05_17.fixed.tar.gz


martinwicke:
The new filename has a new date (8/9). It's been uploaded.

PR #2177, opened 2017-08-10
Adv_Txt: from six.moves import range for Python 3

cclauss:


cclauss:
@rsepassi I have made the changes that you requested but please allow me to explain the logic behind approach that I took with my original changes.  Six makes range() and xrange() interchangeable while it remains a project-wide dependency.  However, the lack of conversation around #2105 and #2024 lead me to conclude that this community will want to remove the dependency on Six as soon as Python 2 is no longer needed.

xrange() is the legacy Python formulation which has been deprecated and removed in favor of  range() in the [present and future](https://wiki.python.org/moin/Python2orPython3) form of the language.  There are currently more than 80 Python files in this repo that contain xrange().  Given that we need to modify those files to be compatible with both versions of Python, my sense was that it was better to move to the future-proof formulation so that dropping the six dependency in the future would merely involve removing the import line and using Python 3+.  This approach also builds up the muscle memory in developers to write the future-ready formulation rather than of the legacy formulation.

rsepassi:
Thanks for the explanation @cclauss. Let's keep things consistent with the rest of the repo here and internally for now. This maintains compatibility both ways, whereas switching to py3's range actually changes py2 functionality (i.e. uses of range now become xrange, which may not have been intended). You're right that this change will eventually need to be made; let's leave it for a wholesale change later. Thanks again.

PR #2109, opened 2017-08-03
adversarial_crypto: ac.eve_optimizer, ac.reset_eve_vars

cclauss:
* eve_optimizer --> ac.eve_optimizer
* reset_eve_vars --> ac.reset_eve_vars
* eve_optimizer --> ac.eve_optimizer

Why make these changes?  flake8 testing of https://github.com/tensorflow/models on Python 2.7.13

$ __flake8 . --count --select=E901,E999,F821,F822,F823 --show-source --statistics__

```
./adversarial_crypto/train_eval.py:252:15: F821 undefined name 'eve_optimizer'
        s.run(eve_optimizer)
              ^

./adversarial_crypto/train_eval.py:257:15: F821 undefined name 'reset_eve_vars'
        s.run(reset_eve_vars)
              ^

./adversarial_crypto/train_eval.py:262:19: F821 undefined name 'eve_optimizer'
            s.run(eve_optimizer)
                  ^
```

PR #1981, opened 2017-07-18
Cog_Map: Resolve 11 undefined names that could raise NameError at runtime

cclauss:
flake8 testing of https://github.com/tensorflow/models on Python 2.7.13

$ __flake8 . --count --select=E901,E999,F821,F822,F823 --show-source --statistics__

```
./cognitive_mapping_and_planning/cfgs/config_distill.py:94:33: F821 undefined name 'resnet_v2_50_path'
  args.solver.pretrained_path = resnet_v2_50_path
                                ^

./cognitive_mapping_and_planning/src/depth_utils.py:26:19: F821 undefined name 'utils'
  camera_matrix = utils.Foo(xc=xc, zc=zc, f=f)
                  ^

./cognitive_mapping_and_planning/src/file_utils.py:36:14: F821 undefined name 'file_name'
  with fopen(file_name, 'r') as f:
             ^

./cognitive_mapping_and_planning/src/file_utils.py:38:10: F821 undefined name 'np'
    II = np.array(I)
         ^

./cognitive_mapping_and_planning/src/graph_utils.py:453:7: F821 undefined name 'logging'
      logging.error('Did not find any good nodes.')
      ^

./cognitive_mapping_and_planning/src/graph_utils.py:468:14: F821 undefined name 'get_path_ids'
      path = get_path_ids(start_node_ids[i], end_node_ids[i], pred_map)
             ^

./cognitive_mapping_and_planning/src/graph_utils.py:511:7: F821 undefined name 'logging'
      logging.error('Did not find any good nodes.')
      ^

./cognitive_mapping_and_planning/src/graph_utils.py:517:14: F821 undefined name 'get_path_ids'
      path = get_path_ids(start_node_ids[i], end_node_ids[i], pred_map)
             ^

./cognitive_mapping_and_planning/src/graph_utils.py:547:14: F821 undefined name 'get_path_ids'
      path = get_path_ids(start_node_ids[i], end_node_ids[i], pred_map)
             ^

./cognitive_mapping_and_planning/src/map_utils.py:143:7: F821 undefined name 'logging'
      logging.error('Unknown resizing method')
      ^

./cognitive_mapping_and_planning/tfcode/vision_baseline_lstm.py:152:21: F821 undefined name 'is_training'
        is_training=is_training)
                    ^
```

PR #2180, opened 2017-08-10
Tutorials should be Python 3 compatible

cclauss:


cclauss:
Given the lack of response to #2105 , I have been reluctant to use `six`.

In this instance, we are __merely counting to 4__ so there is no significant difference between xrange() and range() so I elected to move to the new formulation.

tfboyd:
Keep in mind in models there is no single owner so that make it very hard to fix anything in mass.  Soon there will be some changes to models where models that are very supported will be on one section and "other" stuff in another area.  I do not have a python3 environment to test in, a lame excuse but time is time.  I will try to do a quick test of your PR and merge it in as I do want our example to work on Python 3.  Thank you for your patience.

tfboyd:
Yeah I am not allowed to run docker for security reasons.  but I can get
around that. Good idea.  I should have thought of it.  Thanks.  I should
get some rest.

tfboyd:
The example is using six for other things so that will be fine for the example.  When they create the maintained area they can determine the guidelines.  I am going to keep this open but I will likely not merge it as we did a big refactor (more than we should have in one PR) to make the example also work with CloudML and some other nice features.  Part of this example is also to work with the high level API team to reduce boilerplate and move and improve code while moving it from contrib to core.  

Also thank you again for pointing out this issue.

cclauss:
If your team is pressed for time then you should implement automated testing (#1980 or something similar) on PRs so that submissions that break things are automatically flagged.  This repo is now just a few PRs (#1981, #2109, #2116) away from having a passing flake8 score on Python 2.  We will get there.

cclauss:
@tfboyd How is the review process going on this change?

PR #2187, opened 2017-08-11
Fix MobileNet to only use dropout at train time

MyHumbleSelf:
and add clarifying comment about the last layer bias terms


derekjchow:
@MyHumbleSelf I don't think this change is necessary. is_training gets applied through the slim arg_scope. Have you noticed actual issues with this code?

MyHumbleSelf:
You're totally right! I missed that and just confirmed that the arg does indeed get applied. Sorry for the confusion!

PR #2190, opened 2017-08-11
Assert height or width of preprocessed_inputs is not less than 33 

rolai:
In `FasFasterRCNN`, the input image spatial size (height or width) should be not less than 33.  It is checked in `FasterRCNNResnetV1FeatureExtractor._extract_proposal_features`, but missed in `FasterRCNNInceptionResnetV2FeatureExtractor`. 
Furthermore, the comments of `FasterRCNNInceptionResnetV2FeatureExtractor._extract_proposal_features` contains this assertion, but the code missed.


rolai:
I signed it!

rolai:
I fixed it!

PR #2196, opened 2017-08-11
Do not skip download if image file already exists

dzautner:
As the script uses the '-c' flag with wget, it allows wget to continue downloading from where the previous download stopped. By skipping the download all-together when the file already exists it renders this flag meaningless, and would render the script broken if a previous download has failed before finishing.


dzautner:
I signed it!

Issue #2197, opened 2017-08-11
Visualization in Perspective Transformer Nets is broken


mees:
### System information
- **What is the top-level directory of the model you are using**: Ptn directory
- **Have I written custom code (as opposed to using a stock example script provided in TensorFlow)**: No
- **OS Platform and Distribution (e.g., Linux Ubuntu 16.04)**: Ubuntu 14.04
- **TensorFlow installed from (source or binary)**: Source 
- **TensorFlow version (use command below)**: ('v1.3.0-rc1-27-g2784b1c', '1.3.0-rc2')
- **Bazel version (if compiling from source)**: 0.5.2
- **CUDA/cuDNN version**: 8.0/5.1
- **GPU model and memory**: Titan X
- **Exact command to reproduce**: run -c opt :train_ptn -- --init_model='/home/meeso/models/ptn/my_models/deeprotator_pretrain/train/' --checkpoint_dir='/home/meeso/models/ptn/my_models/' --inp_dir='/home/meeso/Datasets/shapenet_tf/
Then open tensorboard

### Describe the problem
The visualization is broken, because all 3D models are shown in blue, without depth.
![step234](https://user-images.githubusercontent.com/818535/29213853-56f58cec-7ea5-11e7-8359-82c726b7fec3.png)

xcyan:
@skye Can you merge the PR submitted by @mees? We have verified it already. Thank you!

ymodak:
Closing this issue since PR has been merged. Please feel free to reopen the issue if it still persists. Thanks!

PR #2198, opened 2017-08-11
Fixes broken visualization in Perspective Transformer Nets, #2197

mees:
Fixes broken visualization in Perspective Transformer Nets, bug #2197.

### DIFF
--- a/ptn/utils.py
+++ b/ptn/utils.py
@@ -74,7 +74,7 @@ def visualize_voxel_spectral(points, vis_size=128):
   points = np.rint(points)
   points = np.swapaxes(points, 0, 2)
   fig = p.figure(figsize=(1, 1), dpi=vis_size)
-  verts, faces = measure.marching_cubes(points, 0, spacing=(0.1, 0.1, 0.1))
+  verts, faces = measure.marching_cubes_classic(points, 0, spacing=(0.1, 0.1, 0.1))
   ax = fig.add_subplot(111, projection='3d')
   ax.plot_trisurf(
       verts[:, 0], verts[:, 1], faces, verts[:, 2], cmap='Spectral_r', lw=0.1)


mees:
@xcyan @arkanath,  this is the fix I found for the colorization of the 3D shapes.

xcyan:
Did you test on both linux and mac?

mees:
just on Ubuntu, I don't have a mac. But it anyway complained that the api
changed for marching cubes. It recommended either switching
to marching_cubes_lewiner or  marching_cubes_classic.

arkanath:
Can you link the documentation which states that the API call already being used is deprecated?

mees:
@arkanath With marching_cubes I get the following error when running the perspective transformer:
/home/meeso/.virtualenvs/2d_to_3d/local/lib/python2.7/site-packages/skimage/measure/_marching_cubes_lewiner.py:167: skimage_deprecation: `marching_cubes` now uses a better and faster algorithm, and returns four instead of two outputs (see docstring for details). Backwards compatibility with 0.12 and prior is available with `marching_cubes_classic`. This function will be removed in 0.14, consider switching to `marching_cubes_lewiner`.
You have that warning here: https://github.com/scikit-image/scikit-image/blob/v0.13.x/skimage/measure/_marching_cubes_lewiner.py#L167

And here is the API change: https://github.com/scikit-image/scikit-image/blob/37ed59d2880fa96ff908608d7f723b98cd319c9b/doc/source/api_changes.rst (skimage.measure.marching_cubes has been removed in favor of skimage.measure.marching_cubes_lewiner.)

xcyan:
Looks good to me. Maybe it is better to mention the version of scikit-image library you are using in README.

xcyan:
@hellojas Can you merge this PR? Thanks!

PR #2208, opened 2017-08-14
python 3.x compatibility

Chaoscendence:
On default, GFile.read() returns a python string which would depend on the python version. Set GFile to bytes mode and the function will work on both python 2.x and 3.x.


Chaoscendence:
I signed it!

Issue #2140, opened 2017-08-06
Perspective Transformer Nets broken


mees:
### System information
- **What is the top-level directory of the model you are using**: Models directory
- **Have I written custom code (as opposed to using a stock example script provided in TensorFlow)**: No
- **OS Platform and Distribution (e.g., Linux Ubuntu 16.04)**: Ubuntu 14.04
- **TensorFlow installed from (source or binary)**: Source
- **TensorFlow version (use command below)**: ('v1.3.0-rc1-27-g2784b1c', '1.3.0-rc2')
- **Bazel version (if compiling from source)**:  0.5.2
- **CUDA/cuDNN version**: 8.0/5.1
- **GPU model and memory**: Tesla P100, 16GB
- **Exact command to reproduce**: bazel run -c opt ptn:pretrain_rotator -- --step_size={} --init_model={} --inp_dir='/home/meeso/Datasets/shapenet_tf/

### Describe the problem
When I try to run the Perspective Transformer Net, bazel doesn't find the BUILD file in the nets subdirectory.

(2d_to_3d) meeso@hpcgpu8:~/models$ bazel run -c opt ptn:pretrain_rotator -- --step_size={} --init_model={} --inp_dir='/home/meeso/Datasets/shapenet_tf/'
WARNING: Output base '/home/meeso/.cache/bazel/_bazel_meeso/85eec191ea1a70d349654e56e775424f' is on NFS. This may lead to surprising failures and undetermined behavior.
ERROR: /home/meeso/models/ptn/BUILD:30:1: no such package 'nets': BUILD file not found on package path and referenced by '//ptn:model_rotator'.
ERROR: Analysis of target '//ptn:pretrain_rotator' failed; build aborted.
INFO: Elapsed time: 0.281s
ERROR: Build failed. Not running target.

xcyan:
@mees Can you please close the issue and paste the answers here. Thank you!

mees:
Well, the solution is to create an empty file in the ptn directory with the name of WORKSPACE. I think a pull  with the fix is needed, so it's fixed for everybody.

mees:
@xcyan  I created a pull request, so you can merge it.

reedwm:
@mees The pull request is empty. Did you already merge it?

mees:
@reedwm, @xcyan  yeah sorry, somehow the old pull request got removed. I create a new pull request with the same patch.

xcyan:
Please close the issue if the problem has been resolved. Thanks!

PR #2218, opened 2017-08-15
Fixes bug #2140, that prevents running the perspective transformer net

mees:
Fixes bug #2140. Adds bazel workspace file to enable running perspective transformer networks.


arkanath:
Looks good.

PR #2242, opened 2017-08-17
Fix batch norm documentation for is_training

ankitvgupta:
Fixes the documentation to reflect the deprecation of is_training as an argument to resnet_arg_scope and moves is_training to be passed into the network directly.'


ankitvgupta:
I signed it!

PR #2249, opened 2017-08-18
Swivel: print(), basestring(), exception changes for Python 3

cclauss:
Used https://pythonhosted.org/six as discussed in #2105

### DIFF
--- a/swivel/nearest.py
+++ b/swivel/nearest.py
@@ -16,6 +16,7 @@
 
 """Simple tool for inspecting nearest neighbors and analogies."""
 
+from __future__ import print_function
 import re
 import sys
 from getopt import GetoptError, getopt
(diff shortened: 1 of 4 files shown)


swivel/nearest.py
```diff
@@ -55,21 +56,21 @@
...
 
   if not res:
     continue
 
   for word, sim in res[:20]:
-    print '%0.4f: %s' % (sim, word)
+    print('%0.4f: %s' % (sim, word))
 
-  print
+  print('')
```
waterson: How about just `print()`?

cclauss: `print()` will print `()` in Python 2 unless you `from __future__ import print_function`

waterson: ...but you do import print_function as part of this change, right?

PR #2250, opened 2017-08-18
file() was removed in Python 3, use open() instead

cclauss:
@yuyuz http://python-future.org/compatible_idioms.html?highlight=file#file

Also import Queue class in both Python 2 and Python 3  http://python-future.org/compatible_idioms.html?highlight=queue#queue

### DIFF
--- a/qa_kg/util/data_reader.py
+++ b/qa_kg/util/data_reader.py
@@ -14,7 +14,10 @@
 # ==============================================================================
 
 from collections import namedtuple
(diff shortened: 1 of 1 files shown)


cclauss:
@yuyuz This PR is still open and not merged.

yuyuz:
@cclauss Sorry, it seems that I don't have write access.

@panyx0718 Are you okay to merge this PR? Thanks!

Issue #2240, opened 2017-08-17
Perspective transformer rotator evaluation script broken due to missing function


mees:
### System information
- **What is the top-level directory of the model you are using**: Ptn directory
- **Have I written custom code (as opposed to using a stock example script provided in TensorFlow)**: No
- **OS Platform and Distribution (e.g., Linux Ubuntu 16.04)**: Ubuntu 14.04
- **TensorFlow installed from (source or binary)**: Source 
- **TensorFlow version (use command below)**: ('v1.3.0-rc1-27-g2784b1c', '1.3.0-rc2')
- **Bazel version (if compiling from source)**: 0.5.2
- **CUDA/cuDNN version**: 8.0/5.1
- **GPU model and memory**: Titan X
- **Exact command to reproduce**: CUDA_VISIBLE_DEVICES=0 bazel run -c opt :eval_rotator  -- --checkpoint_dir='/home/meeso/models/ptn/my_models/' --inp_dir='/tmp/shapenet_tf/'

### Describe the problem
The rotator evaluation scripts fails calls a function named rotator_metrics that is not implemented.

### Source code / logs
```
Traceback (most recent call last):
  File "/home/meeso/.cache/bazel/_bazel_meeso/3a6bfcc5e3fa6521c178dee287de79bc/execroot/__main__/bazel-out/local-opt/bin/eval_rotator.runfiles/__main__/eval_rotator.py", line 127, in <module>
/home/meeso/models/ptn/my_models/deeprotator_pretrain/train
    app.run()
  File "/home/meeso/.virtualenvs/2d_to_3d/local/lib/python2.7/site-packages/tensorflow/python/platform/app.py", line 48, in run
    _sys.exit(main(_sys.argv[:1] + flags_passthrough))
  File "/home/meeso/.cache/bazel/_bazel_meeso/3a6bfcc5e3fa6521c178dee287de79bc/execroot/__main__/bazel-out/local-opt/bin/eval_rotator.runfiles/__main__/eval_rotator.py", line 111, in main
    inputs, outputs, FLAGS)
  File "/home/meeso/models/ptn/model_rotator.py", line 193, in get_metrics
    names_to_values, names_to_updates = metrics.rotator_metrics(
AttributeError: 'module' object has no attribute 'rotator_metrics'
ERROR: Non-zero return code '1' from command: Process exited with status 1.
```

mees:
@xcyan, @arkanath

xcyan:
[...]

PR #2251, opened 2017-08-18
Bugfix for missing function rotator_metrics in models/ptn

xcyan:
This PR fixes the known bug when evaluating the rotator model in models/ptn. #2240 

Please check this out @arkanath @mees

### DIFF
--- a/ptn/model_rotator.py
+++ b/ptn/model_rotator.py
@@ -190,8 +190,34 @@ def get_train_op_for_scope(loss, optimizer, scopes, params):
 
 
 def get_metrics(inputs, outputs, params):
-  names_to_values, names_to_updates = metrics.rotator_metrics(
-      inputs, outputs, params)
+  """Aggregate the metrics for rotator model.
+  
+  Args:
+    inputs: Input dictionary of the rotator model.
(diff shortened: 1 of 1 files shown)


xcyan:
I signed it!

mees:
thanks @xcyan, the patch works in general. However, you also need to fix a small detail in eval_rotator.py to enable the script to run. You need to change gs.DEFINE_string('master', 'local', '') to gs.DEFINE_string('master', '', '') to avoid a session error. Last question: the rnn1-rnn24 represent the error when predicting directly from input to 15 degrees*rotation_step or do you chain the (rotation) predictions?

xcyan:
I think we accumulated the loss across rotations steps (actually average over the first K steps).
https://github.com/tensorflow/models/blob/master/ptn/losses.py#L49

xcyan:
@mees Please close the issue if the problem has been resolved. Thank you!

mees:
@xcyan @reedwm please add the fix flags.DEFINE_string('master', '', '') in eval_rotator.py . I didn't want to open a bug issue and create a PR, because I was hoping @xcyan could add this minor change before merging the whole patch.

PR #2259, opened 2017-08-19
bugfix for rnn tutorial on Python3

skwbc:
`str` object doesn't have `decode` method and so `tutorials/rnn/ptb/reader.py` doesn't work on Python3.

Actually we don't have to call `decode` method even if we use Python2 because the data used in the tutorial only include ascii characters.
I wrote the code which have `if` condition for `Py3` to apply this code to some texts include non-ascii characters.


skwbc:
I signed it!

PR #2268, opened 2017-08-22
Make the record scripts tests python 3 compatible and add further test.

datitran:
I added one more test where you have two classes on one image so that it is more explicit that the bounding boxes are a list of xmins, ymins, xmaxs, ymaxs etc and also made it python 3 compatible.

### DIFF
--- a/object_detection/create_pascal_tf_record_test.py
+++ b/object_detection/create_pascal_tf_record_test.py
@@ -36,7 +36,7 @@ class DictToTFExampleTest(tf.test.TestCase):
     proto_list = [p for p in proto_field]
     self.assertListEqual(proto_list, expectation)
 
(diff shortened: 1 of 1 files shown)


object_detection/create_pascal_tf_record_test.py
```diff
@@ -80,12 +80,12 @@ def test_dict_to_tf_example(self):
         example.features.feature['image/width'].int64_list.value, [256])
     self._assertProtoEqual(
         example.features.feature['image/filename'].bytes_list.value,
-        [image_file_name])
+        [image_file_name.encode('utf-8')])
```
derekjchow: Why not do something like:

image_file_name = 'tmp_image.jpg'.encode('utf-8')

PR #2276, opened 2017-08-23
Add different rnn implementation modes to ptb tutorial

bignamehyp:
Add different rnn implementation modes to ptb tutorial using CudnnLSTM and LSTMBlockCell. 

rnn_mode == CUDNN
===================================================
| config | epochs | train | valid  | test   | wps
====================================================
| small  | 13     | 40.50 | 116.60 | 112.13 | 49k
| medium | 39     | 22.58 | 123.83 | 118.69 | 24.2k
| large  | 55     |  8.03 | 129.78 | 126.03 | 10.5k
====================================================
In fact, the params tensor layout for cudnn_lstm is so different from canonical LSTM,
we need a complete different set of init_scale and learning rate parameters.

rnn_mode == BLOCK
===================================================
| config | epochs | train | valid  | test   | wps
====================================================
| small  | 13     | 40.55 | 120.70 | 115.52 | 17.2k
| medium | 39     | 45.68 |  86.97 |  83.47 | 13.6k
| large  | 55     | 37.94 |  82.75 |  78.49 | 5.0k
====================================================

Wps before this cl for small, medium, large models are
15.6k, 12.6k, and 5.0k, respectively.

Benchmarking platform: E5-2690 v4, Titan X.

### DIFF
--- a/tutorials/rnn/ptb/ptb_word_lm.py
+++ b/tutorials/rnn/ptb/ptb_word_lm.py
@@ -40,6 +40,9 @@ The hyperparameters used in the model:
 - keep_prob - the probability of keeping weights in the dropout layer
 - lr_decay - the decay of the learning rate for each epoch after "max_epoch"
 - batch_size - the batch size
+- rnn_mode - the low level implementation of lstm cell: one of CUDNN,
+             BASIC, or BLOCK, representing cudnn_lstm, basic_lstm, and
+             lstm_block_cell classes.
 
 The data required for this example is in the data/ dir of the
 PTB dataset from Tomas Mikolov's webpage:
@@ -56,13 +59,15 @@ from __future__ import absolute_import
 from __future__ import division
 from __future__ import print_function
 
-import inspect
 import time
 
+import google3
 import numpy as np
 import tensorflow as tf
 
-import reader
+from google3.third_party.tensorflow_models.tutorials.rnn.ptb import reader
+from google3.third_party.tensorflow_models.tutorials.rnn.ptb import util
+from google3.third_party.tensorflow.python.client import device_lib
 
 flags = tf.flags
 logging = tf.logging
@@ -76,8 +81,18 @@ flags.DEFINE_string("save_path", None,
                     "Model output directory.")
 flags.DEFINE_bool("use_fp16", False,
                   "Train using 16-bit floats instead of 32bit floats")
-
+flags.DEFINE_integer("num_gpus", 1,
+                     "If larger than 1, Grappler AutoParallel optimizer "
+                     "will create multiple training replicas with each GPU "
+                     "running one replica.")
+flags.DEFINE_string("rnn_mode", None,
+                    "The low level implementation of lstm cell: one of CUDNN, "
+                    "BASIC, and BLOCK, representing cudnn_lstm, basic_lstm, "
+                    "and lstm_block_cell classes.")
 FLAGS = flags.FLAGS
+BASIC = "basic"
+CUDNN = "cudnn"
+BLOCK = "block"
 
 
 def data_type():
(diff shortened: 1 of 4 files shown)


bignamehyp:
I just signed cla.

bignamehyp:
Run googlebot again.

tutorials/rnn/ptb/ptb_word_lm.py
```diff
@@ -56,13 +59,15 @@
...
 import time
 
+import google3
 import numpy as np
 import tensorflow as tf
 
-import reader
+from google3.third_party.tensorflow_models.tutorials.rnn.ptb import reader
+from google3.third_party.tensorflow_models.tutorials.rnn.ptb import util
+from google3.third_party.tensorflow.python.client import device_lib
```
nealwu: Can you remove the google3 imports? Also, we don't have a `util.py` file in this folder, and I'm not sure whether we can import `device_lib`.

bignamehyp: Done. PTAL.

bignamehyp:
PTAL.

nealwu:
Looks good at a glance. I'll try to run it on my machine tomorrow and see how it does.

nealwu:
:+1: Got the following results with the medium model:

Epoch: 39 Train Perplexity: 45.733
Epoch: 39 Valid Perplexity: 87.722
Test Perplexity: 83.632

janchorowski:
This cl is buggy and should be reverted:

First, the CUDNN backend doesn't make sense:
1. it doesn't do dropout between LSTM layers, which is required to reach the perplexities reported in the header of this file.
2. the code doesn't handle initial states properly, it is mandatory that the initial state of the RNN is taken from the last state of the previous iteration.

Second, the CL pulled old code for the basic LSTM, making it incompatible with the new Tensorflow, see e.g. #8191.

protoget:
Re janchorowski@
First
1. Cudnn takes dropout rate as one argument. It does dropout for each layer input. I think we tested the convergence before submitting the CL. 
2. This isn't true. For each example, the initial state is always 0 to begin with. The state is carried over to next step (not next example); after finishing all the steps you get the final state (for this example).

Second:
Not sure which part is problematic, could you point out more explicitly?

janchorowski:
Hi,

I aplogize for speaking too soon. I had two problems with running the new code and blamed it to this change:
- the one I described as second, in which this line: https://github.com/tensorflow/models/blob/7e9e15ad6b685a1d8f1048c2e39194292cf54d6a/tutorials/rnn/ptb/ptb_word_lm.py#L131 got changed into: https://github.com/tensorflow/models/blob/master/tutorials/rnn/ptb/ptb_word_lm.py#L222 Note that before the change a new cell object was created for each layer, while after the change the same cell object is reused. This triggered the error "ValueError: Attempt to reuse RNNCell". I then upgraded tensorflow to v 1.3 and can't reproduce this error, I can install some older versions and see at which versions the error is occurring, unfortunately I didn't record it.
- the bad perplexities achieved with the cudnn backend. The model still overfits (see the table in the OP message). I blamed it on lack of dropout in the cudnn wrapper and on incorrect recurrent state preservation. I reread the code and you are right about the hidden states being properly preserved from minibatch to minibatch. If dropout also works for all layers, then why changing the LSTM implementation has such a drastic impact on the perplexity? The cudnn backend shouldn't require a completely new set of hyperparameters. It is unfortunate to have a fast implementation which gives bad perplexities and a slow one that gives good perplexities.

touxian:
Re janchorowski@

Thanks for the details.

1. You're right the cell is reused, thanks for catching that. Would you create a PR since you've invested a lot of time in it? We can review it.
2. We have no control of the essential implementation of Cudnn RNN, since it's done by Nvidia as a close-source library. I agree it is unfortunate. That said, I think for the short time being, it would become common for users to use different hyper params depending on the impl. e.g. Nvidia has a [tutorial](http://on-demand.gputechconf.com/gtc/2017/presentation/s7218-training-with-mixed-precision-boris-ginsburg.pdf) for training with mixed precision for speedup. . You might need to adjust your learning rate in order to achieve convergence.

We have a way to initialize cudnn variables w/ the same distributions as that for platform-independent rnn cells. The change is pending review and would probably catch the 1.4 release. We will update the example with the new cudnn APIs and check the pplx again.

Thank you.

protoget:
Re janchorowski@

Thanks for the details.

You're right the cell is reused, thanks for catching that. Would you create a PR since you've invested a lot of time in it? We can review it.
We have no control of the essential implementation of Cudnn RNN, since it's done by Nvidia as a close-source library. I agree it is unfortunate. That said, I think for the short time being, it would become common for users to use different hyper params depending on the impl. e.g. Nvidia has a tutorial for training with mixed precision for speedup. . You might need to adjust your learning rate in order to achieve convergence.
We have a way to initialize cudnn variables w/ the same distributions as that for platform-independent rnn cells. The change is pending review and would probably catch the 1.4 release. We will update the example with the new cudnn APIs and check the pplx again.

Thank you.

Utumno:
Does the commit linked above fix the problem with reusing the same cell ? If yes I could submit a PR

nealwu:
@bignamehyp @protoget Does that change by @Utumno seem good to you? See https://github.com/tensorflow/models/pull/934 for some context.

protoget:
LGTM if it passes quality tests.
FYI new Cudnn RNN layer API has been submitted so later this example will be further refactored.

nealwu:
@Utumno could you submit your PR and @-mention us?

Utumno:
Thanks, will do ASAP but please merge https://github.com/tensorflow/models/pull/2403 first - it's needed so this even runs on python 3 and it will be in my pull otherwise

nealwu:
@Utumno Sure thing, merged.

cocosci:
Does anyone can tell me the difference between "canonical LSTM cells" and "block or CUDNN cells"?
My tf version is 1.2.1 and that leaves me with canonical LSTM cells only. The keyword "BASIC" sounds not so fancy...:)

Correct me if I am wrong. 
"All these three backend cells are LSTM units. The major difference lies in the hardware optimization. It's Okay to choose any of the three"

Issue #2239, opened 2017-08-17
Perspective transformer 3D evaluation script broken


mees:
### System information
- **What is the top-level directory of the model you are using**: Ptn directory
- **Have I written custom code (as opposed to using a stock example script provided in TensorFlow)**: Yes, I have fixed the first flag issue https://github.com/mees/models/commit/48ac9b0dfb2dc2ec4d2f70eb73bcb800a9377b9d

- **OS Platform and Distribution (e.g., Linux Ubuntu 16.04)**: Ubuntu 14.04
- **TensorFlow installed from (source or binary)**: Source 
- **TensorFlow version (use command below)**: ('v1.3.0-rc1-27-g2784b1c', '1.3.0-rc2')
- **Bazel version (if compiling from source)**: 0.5.2
- **CUDA/cuDNN version**: 8.0/5.1
- **GPU model and memory**: Titan X
- **Exact command to reproduce**: CUDA_VISIBLE_DEVICES=0 bazel-bin/eval_ptn   '--checkpoint_dir=/home/meeso/models/ptn/my_models/no_finetune' '--inp_dir=/tmp/shapenet_tf/' '--model_name=ptn_finetune'

### Describe the problem
Once I train a ptn network,  I run the evaluation script (once fixed the first problem with the wrong flags #2236) to compute the IOU. I get an error that the predictions are out of bound. The offending line seems to be here https://github.com/tensorflow/models/blob/master/ptn/metrics.py#L103 but I am not sure where this mapping comes from (3*times predictions -2) and if its correct.

### Source code / logs
```
2017-08-17 18:11:38.347664: I tensorflow/core/common_runtime/gpu/gpu_device.cc:1045] Creating TensorFlow device (/gpu:0) -> (device: 0, name: TITAN X (Pascal), pci bus id: 0000:02:00.0)
Traceback (most recent call last):
  File "/home/meeso/models/ptn/bazel-bin/eval_ptn.runfiles/__main__/eval_ptn.py", line 134, in <module>
    app.run()
  File "/home/meeso/.virtualenvs/2d_to_3d/local/lib/python2.7/site-packages/tensorflow/python/platform/app.py", line 48, in run
    _sys.exit(main(_sys.argv[:1] + flags_passthrough))
  File "/home/meeso/models/ptn/bazel-bin/eval_ptn.runfiles/__main__/eval_ptn.py", line 130, in main
    eval_interval_secs=FLAGS.eval_interval_secs)
[...]

PR #2282, opened 2017-08-24
 fixes Perspective Transformer IOU evaluation

mees:
fixes IOU computation for issue #2239

### DIFF
--- a/ptn/metrics.py
+++ b/ptn/metrics.py
@@ -99,8 +99,8 @@ def add_volume_iou_metrics(inputs, outputs):
   names_to_updates = dict()
   labels = tf.greater_equal(inputs['voxels'], 0.5)
   predictions = tf.greater_equal(outputs['voxels_1'], 0.5)
-  labels = 2 - tf.to_int32(labels)
-  predictions = 3 - tf.to_int32(predictions) * 2
+  labels = (2 - tf.to_int32(labels)) - 1
+  predictions = (3 - tf.to_int32(predictions) * 2) - 1
   tmp_values, tmp_updates = tf.metrics.mean_iou(
       labels=labels,
       predictions=predictions,


mees:
Here is the fix @arkanath @xcyan

xcyan:
Looks good to me.

xxxzhi:
what's about label 255 at ground truth.

xcyan:
@xxxzhi Can you be more specific?

xxxzhi:
@xcyan In segmentation groud truth label, there are some label who's value is 255 which means fuzzy tag.  This method tf.contrib.metrics.streaming_mean_iou will fail to call.

https://github.com/tensorflow/tensorflow/commit/6ac3efd42902d48d45d59128926110e6d5121a08

xcyan:
@xxxzhi Are you talking about a bug specific to this PR? or a bug specific to tensorflow/models/ptn? I don't think we have segmentation label who's value is 255.

xxxzhi:
@xcyan I'm taling about https://github.com/tensorflow/tensorflow/commit/6ac3efd42902d48d45d59128926110e6d5121a08 . I use a label who's value is 255, then I fail to call that function. This PR is nice.

xxxzhi:
@xcyan I often use label 255 for ignore label in segmentation.

mees:
@xxxzhi I don't really understand the discussion. @xcyan maps the prediction to 3(free) and 1(occupied) and the gt to 2(free) and 1(occupied) and thus gets the intersection for 1. My patch just makes so that it starts with 0 instead of 1, because that was causing problems. I don't think that with this code you will get a value of 255 as a label, but maybe I didn't understand your concern correctly.

xcyan:
Anyways, this PR looks good. Please file another bug/feature request for discussions not related to this PR.

xcyan:
@hellojas Please review this PR at your convenience. Thanks!

xcyan:
@hellojas Please review & merge this PR if possible. Thank you!

Best,
Xinchen

PR #2053, opened 2017-07-27
object_detection exporter updates

derekjchow:

PR #2295, opened 2017-08-28
Change exporting_models.md to reflect the updated command in export_inference_graph.py

yatendragoel:
Since the export_inference_graph.py was changed in [this pull request](https://github.com/tensorflow/models/pull/2053/files), the corresponding [markdown file](https://github.com/yatendragoel/models/blob/master/object_detection/g3doc/exporting_models.md) also had to be changed


yatendragoel:
I signed it!

PR #2300, opened 2017-08-28
Fix the call to build_imagenet_data.py

deepjedi:
The Python extension is missing in the BUILD_SCRIPT string.


deepjedi:
I signed it!

PR #2303, opened 2017-08-29
DOC: Change link for slim's imagenet instructions

scottclowe:
We now link to the instructions to download and convert imagenet to
TFrecords contained in the slim README instead of the inception
README.

The build script in the inception folder does not create a
labels.txt file and so it is no longer compatible with the slim
train script.

### DIFF
--- a/slim/README.md
+++ b/slim/README.md
@@ -126,7 +126,7 @@ from integer labels to class names.
 
 You can use the same script to create the mnist and cifar10 datasets.
 However, for ImageNet, you have to follow the instructions
-[here](https://github.com/tensorflow/models/blob/master/inception/README.md#getting-started).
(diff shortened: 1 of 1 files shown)


scottclowe:
The documentation is still incorrect.

You have a [README](https://github.com/tensorflow/models/blob/master/research/slim/README.md) with a section on Imagenet:
https://github.com/tensorflow/models/tree/master/research/slim#an-automated-script-for-processing-imagenet-data

And yet earlier in the very same README, there is an instruction telling the user to go and read some other, out-dated, incorrect documentation instead.
https://github.com/tensorflow/models/blame/master/research/slim/README.md#L135-L136
If a user trying to use slim to train an imagenet model reads this README in order, they won't get to the bit they actually need to read because they get sent to use some other code that doesn't work!

PR #2311, opened 2017-08-30
Update rules_closure

vladmos:
This change is required for compatibility with Bazel 0.6+

See also https://github.com/tensorflow/tensorflow/pull/12685


vladmos:
This commit broke tests, an attempt to fix: #2313

PR #2317, opened 2017-08-31
Update exporting_models.md

skulhare:
Just a simple change.


derekjchow:
Thanks for sending this patch, but this looks fixed in the most recent version.

PR #2318, opened 2017-08-31
Add option to configure max GPU memory fraction for training and evaluation in object_detection

ArjanSchouten:
**Subfolder**: object_detection
**Use Case**:
I want to follow the accuracy over time during training. I run train.py and eval.py at the same time. train.py is directly allocating all GPU VRAM so eval.py can't restore the graph anymore and results in OOM (since I use one GPU).

**Changed method definition:**
An additional parameter is added in eval_util.py (method repeated_checkpoint_run and run_checkpoint_once). If this is problematic then I would suggest leaving the config option for eval and only add the option to train.proto.


ArjanSchouten:
I signed it!

ArjanSchouten:
I signed it!

Briannairish:
I signed it!

ArjanSchouten:
Sorry everyone for this messed up PR. Tried to fix wrong email address for cla check.

Issue #2329, opened 2017-09-03
doc issue for exporting object_detection model


salisbury-espinosa:
### System information
- **What is the top-level directory of the model you are using**: object detection
- **Have I written custom code (as opposed to using a stock example script provided in TensorFlow)**: NO
- **OS Platform and Distribution**: Linux Ubuntu 16.04
- **TensorFlow installed from (source or binary)**:binary
- **TensorFlow version (use command below)**:1.3.0
- **CUDA/cuDNN version**:8.0, 5.1
- **GPU model and memory**:GTX 850M 4GB

Documentation 
https://github.com/tensorflow/models/blob/master/object_detection/g3doc/exporting_models.md
does not match example usage in export_inference_graph:
https://github.com/tensorflow/models/blob/master/object_detection/export_inference_graph.py#L51

PR: https://github.com/tensorflow/models/pull/2330

jart:
#2330 is now approved. Should be merged in the next few days. Thank you for the contribution.

PR #2330, opened 2017-09-03
fix exporting model doc for object detection

salisbury-espinosa:
see example usage in https://github.com/tensorflow/models/blob/master/object_detection/export_inference_graph.py#L51
issue https://github.com/tensorflow/models/issues/2329


salisbury-espinosa:
I signed it!

derekjchow:
Thank you for updating our docs!

PR #2333, opened 2017-09-04
Definite Tensor outside detection loop, change Tensor names

jzhugithub:
I move all detection_graph.get_tensor_by_name outside detection loop, since detection_graph.get_tensor_by_name only need to run once and do not need to run in every loop.
I change all Tensor's names, since names of Tenor and numpy array should be difference, otherwise, it will confuse someone with Tenor and numpy array.


derekjchow:
LGTM thanks!

PR #2341, opened 2017-09-05
Add option to configure max GPU memory fraction for training and evaluation in object_detection

ArjanSchouten:
**Subfolder**: object_detection
**Use Case**:
I want to follow the accuracy over time during training. I run train.py and eval.py at the same time. train.py is directly allocating all GPU VRAM so eval.py can't restore the graph anymore and results in OOM (since I use one GPU).

**Changed method definition**:
An additional parameter is added in eval_util.py (method repeated_checkpoint_run and run_checkpoint_once). If this is problematic then I would suggest leaving the config option for eval and only add the option to train.proto.

(Original [PR](https://github.com/tensorflow/models/pull/2318))

### DIFF
--- a/object_detection/eval_util.py
+++ b/object_detection/eval_util.py
@@ -295,7 +295,8 @@ def run_checkpoint_once(tensor_dict,
                         save_graph=False,
                         save_graph_dir='',
                         metric_names_to_values=None,
-                        keys_to_exclude_from_results=()):
+                        keys_to_exclude_from_results=(),
+                        eval_config=None):
   """Evaluates both python metrics and tensorflow slim metrics.
 
   Python metrics are processed in batch by the aggregated_result_processor,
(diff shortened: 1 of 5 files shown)


ArjanSchouten:
I signed it!

object_detection/eval_util.py
```diff
@@ -295,7 +295,8 @@ def run_checkpoint_once(tensor_dict,
                         save_graph=False,
                         save_graph_dir='',
                         metric_names_to_values=None,
-                        keys_to_exclude_from_results=()):
+                        keys_to_exclude_from_results=(),
+                        eval_config=None):
```
derekjchow: Change this to per_process_gpu_memory_fraction

object_detection/eval_util.py
```diff
@@ -356,7 +358,11 @@ def run_checkpoint_once(tensor_dict,
   """
   if save_graph and not save_graph_dir:
     raise ValueError('`save_graph_dir` must be defined.')
-  sess = tf.Session(master, graph=tf.get_default_graph())
+  session_config = tf.ConfigProto()
```
derekjchow: Default session_config to None, and create if per_process_gpu_memory_fraction is passed in.

PR #2343, opened 2017-09-06
Fix incorrect SyncReplicasOptimizer usage

elibixby:
If I understand correctly, `SyncReplicasOptimizer.make_session_run_hook`s output is supposed to be run on every worker, with the provided `is_chief` flag determining whether aggregation is performed. 

This PR fixes that (it also moves other hooks inside the model_fn so that you can refer to logging tensors by reference rather than by name which is fragile). 

@tfboyd @isaprykin

### DIFF
--- a/tutorials/image/cifar10_estimator/cifar10_main.py
+++ b/tutorials/image/cifar10_estimator/cifar10_main.py
@@ -153,18 +153,26 @@ def get_model_fn(num_gpus, variable_strategy, num_workers):
 
       learning_rate = tf.train.piecewise_constant(tf.train.get_global_step(),
                                                   boundaries, staged_lr)
-      # Create a nicely-named tensor for logging
-      learning_rate = tf.identity(learning_rate, name='learning_rate')
 
+      loss = tf.reduce_mean(tower_losses, name='loss')
       optimizer = tf.train.MomentumOptimizer(
           learning_rate=learning_rate, momentum=momentum)
 
-      chief_hooks = []
+
+      examples_sec_hook = cifar10_utils.ExamplesPerSecondHook(
+          params.train_batch_size, every_n_steps=10)
+
+      tensors_to_log = {'learning_rate': learning_rate, 'loss': loss}
+
+      logging_hook = tf.train.LoggingTensorHook(
+          tensors=tensors_to_log, every_n_iter=100)
+
+      train_hooks = [logging_hook, examples_sec_hook]
       if params.sync:
         optimizer = tf.train.SyncReplicasOptimizer(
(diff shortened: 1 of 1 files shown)


tutorials/image/cifar10_estimator/cifar10_main.py
```diff
@@ -185,14 +193,13 @@ def _resnet_model_fn(features, labels, mode, params):
...
               tf.metrics.accuracy(stacked_labels, predictions['classes'])
       }
-      loss = tf.reduce_mean(tower_losses, name='loss')
 
     return tf.estimator.EstimatorSpec(
         mode=mode,
         predictions=predictions,
         loss=loss,
         train_op=train_op,
-        training_chief_hooks=chief_hooks,
```
isaprykin: Just curious, does this mean that previously non-chief-workers did not have sync replicas optimizer initialized with a hook?  Just based on the name.

tutorials/image/cifar10_estimator/cifar10_main.py
```diff
@@ -336,32 +343,24 @@ def _experiment_fn(run_config, hparams):
...
-        hparams.train_batch_size, every_n_steps=10)
-
-    tensors_to_log = {'learning_rate': 'learning_rate', 'loss': 'loss'}
 
-    logging_hook = tf.train.LoggingTensorHook(
-        tensors=tensors_to_log, every_n_iter=100)
-
-    hooks = [logging_hook, examples_sec_hook]
+    if run_config.num_worker_replicas:
+      num_workers = run_config.num_worker_replicas + 1
```
isaprykin: Well.   But this +1 is due to a bug.  I could just fix the bug today or tomorrow?

Alternatively, I'll have to remove this +1 when the bug is fixed.

--
Bug: third_party/tensorflow/contrib/learn/python/learn/estimators/run_config.py line=393 needs count in 'master' in addition to 'worker'.

tutorials/image/cifar10_estimator/cifar10_main.py
elibixby: Yep that's correct

tutorials/image/cifar10_estimator/cifar10_main.py
elibixby: Yeah. Either way. Up to you.

isaprykin: Eli, this +1 shouldn't be needed as of 168702095.

tfboyd:
If you are good Igor it is fine with me.  I am trying to leave this to you
and Eli.

isaprykin:
I removed the +1 using the github editor.

PR #2346, opened 2017-09-06
Fix wrong base path from exporting_model.md

lime62:
The base path for this should be `tensorflow/models` not `tensorflow/models/object_detection`


derekjchow:
Thanks for sending the fix, but I've updated it [here](https://github.com/tensorflow/models/pull/2940).

PR #2349, opened 2017-09-06
Added code for Learned Optimizers that Scale and Generalize

olganw:


hot9cups:
I can't find these files in the codebase anymore. I came here looking for code to [this](https://research.google/pubs/pub46116/).

Has this been migrated somewhere else? Every other link I find online is broken, for example [this](https://github.com/tensorflow/models/tree/master/research/learned_optimizer)

Any ideas @martinwicke ?

olganw:
This code was deleted as it is no longer guaranteed to work with current versions of TF and other deps.

hot9cups:
Thank you for replying. 
Would you then know where I might get code for this [paper](https://research.google/pubs/pub46116/) then?

olganw:
Unfortunately both of us authors moved on to other projects and are no longer actively maintaining that code. As you can imagine, keeping every old research project up to date with the latest deep learning frameworks is a losing battle :)

Thanks for your interest in this work though! It's still an exciting area and if you're interested in where some of the learned optimizer work has gone, I recommend looking at e.g. https://arxiv.org/abs/2009.11243 from some of the same people.

PR #2365, opened 2017-09-11
fixed "learning rate is illegal" warning

leacoleaco:
when I training the data with tensorflow 1.3， the console print the warning text ”learning rate is illegal“ .  so i find and changed this word


leacoleaco:
@googlebot I signed it!

PR #2371, opened 2017-09-12
Reference evaluation loop example.

williford:
In order to help address issue #5987, I'm adding references to the evaluation loop example. I'm also making it more obvious that this is not the main TF-Slim page and adding another reference to that page in the introductory paragraph.


williford:
I signed it!

PR #2375, opened 2017-09-13
fix WORK_DIR

zhimengfan1990:
the WORK_DIR seems incorrect, it was inconsistent with the directory structure, maybe it should be 
`".."`


nealwu:
Hi @zhimengfan1990, please run this file using bazel as described in the README: https://github.com/tensorflow/models/tree/master/inception

PR #2380, opened 2017-09-13
Fix wrong explanation

solaris33:


nealwu:
@cshallue does this look right to you?

cshallue:
@nealwu  LGTM

PR #2396, opened 2017-09-16
Move up the message in the inception README clarifying the folder is mostly outdated

nealwu:


nealwu:
@tfboyd @shlens should we change the description of the distributed training files?

PR #2398, opened 2017-09-16
Fix python3 numpy load in skip_thoughts

### LINKED ISSUE #1262: Loading Skip Thoughts Model in Python 3
Model: skip_thoughts
Issue type: Update for Python 3.5+ support
Excellent job with the implementation of Skip-Thoughts in TF.
Just want to bring to your attention about loading the embedding matrix .npy file in Python 3.5.3.
These lines in encoder_manager.py:
```
with open(embedding_matrix_file, "r") as f:
      embedding_matrix = np.load(f)
```
need to be changed to the following:
`embedding_matrix = np.load(embedding_matrix_file)`
as Python 3 seems to have issues loading the file object but seems fine loading directly.
The original code raises an error as: _startswith first arg must be str or a tuple of str, not bytes_
when npyio.py file reads the file object at line 393: `magic.startswith(_ZIP_PREFIX)`

Thanks

YuMS:
fix #1262


karmel:
@cshallue Can you review?

YuMS:
@karmel finished

PR #2403, opened 2017-09-17
tutorials/rnn/ptb: Small fix to make Python 3-compatible.

gnprice:
Replace dict.iteritems with dict.items.  In both cases the dict is
very small, so the optimization that `iteritems` provides on Python 2
doesn't matter.

With this fix, the model works smoothly for me on CPython 3.5.3.


gnprice:
CLA signed!

Utumno:
Please merge this

gnprice:
Thanks Neal!

PR #2437, opened 2017-09-22
Fixed the installation instructions URL

rtjarvis:
There is a broken link (maybe more than one) because of the move of the models into the research subdirectory.


rtjarvis:
I've signed it

nealwu:
Thanks @rtjarvis; this was fixed in https://github.com/tensorflow/models/pull/2445/files#diff-30a678190e866b84831e42ecb7f27d15.

PR #2438, opened 2017-09-22
object_detection docs: Fixed reference to directory in protobuff compilation instructions.

MatthewShotton:
The installation instructions are missing the correct path to the Protobuff libraries following the move of models into the research folder.


MatthewShotton:
Apologies, looks like #2434 already fixes this and more.

PR #2441, opened 2017-09-22
Refactor object detection tutorial to use recommended PIL to numpy conversion

MatthewShotton:
This PR simplifies the object detection tutorial a little by removing the helper function for converting PIL/Pillow images to numpy arrays by using the numpy array interface on the PIL.Image object (introduced PIL 1.1.6). Which seems to be the preferred method for converting between the two [1] [2].

[1] [https://stackoverflow.com/a/384926](https://stackoverflow.com/a/384926)
[2] [http://effbot.org/zone/pil-changes-116.htm](http://effbot.org/zone/pil-changes-116.htm)

### DIFF
--- a/research/object_detection/object_detection_tutorial.ipynb
+++ b/research/object_detection/object_detection_tutorial.ipynb
@@ -194,27 +194,6 @@
    "cell_type": "markdown",
    "metadata": {},
    "source": [
-    "## Helper code"
-   ]
-  },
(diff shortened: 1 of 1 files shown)


MalteOeljeklaus:
This should be merged because the original code is super slow, it shows especially on embedded platforms. I searched way too long before I found out what was causing this.

PR #2443, opened 2017-09-22
Flags

itsmeolivia:
removing tf.app.flags.FLAGS in lieu of argparse

### DIFF
--- a/tutorials/image/cifar10/cifar10.py
+++ b/tutorials/image/cifar10/cifar10.py
@@ -40,21 +40,20 @@ import re
 import sys
 import tarfile
 
+import argparse
 from six.moves import urllib
 import tensorflow as tf
 
 import cifar10_input
 
-FLAGS = tf.app.flags.FLAGS
+parser = argparse.ArgumentParser()
 
 # Basic model parameters.
-tf.app.flags.DEFINE_integer('batch_size', 128,
-                            """Number of images to process in a batch.""")
-tf.app.flags.DEFINE_string('data_dir', '/tmp/cifar10_data',
-                           """Path to the CIFAR-10 data directory.""")
-tf.app.flags.DEFINE_boolean('use_fp16', False,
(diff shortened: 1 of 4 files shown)


tutorials/image/cifar10/cifar10.py
```diff
@@ -40,21 +40,25 @@
 import sys
 import tarfile
 
+import argparse
```
nealwu: Let's move this to the section above since this is a native Python import (right before `import os`)

tutorials/image/cifar10/cifar10.py
```diff
@@ -40,21 +40,25 @@
...
-tf.app.flags.DEFINE_string('data_dir', '/tmp/cifar10_data',
-                           """Path to the CIFAR-10 data directory.""")
-tf.app.flags.DEFINE_boolean('use_fp16', False,
-                            """Train the model using fp16.""")
+parser.add_argument('--batch_size', type=int, default=128,
+                    help='Number of images to process in a batch.')
 
+parser.add_argument('--data_dir', type=str, default='/tmp/cifar10_data',
+                    help='Path to the CIFAR-10 data directory.')
+
```
nealwu: We should either remove this blank line or add one above for consistency

tutorials/image/cifar10/cifar10.py
```diff
@@ -40,21 +40,25 @@
...
+parser.add_argument('--use_fp16', type=bool, default=False,
+                    help='Train the model using fp16.')
+
+FLAGS = parser.parse_args()
```
nealwu: Add a blank line after this one.

Also, should we have this under `if __name__ == '__main__'` instead? Otherwise this will be run whenever this file is imported.

itsmeolivia: yeah, I want the flags to be imported into the other files so that the flags can be used across files

PR #2444, opened 2017-09-22
pylint errors

itsmeolivia:
removing unnecessary imports

### DIFF
--- a/official/resnet/imagenet.py
+++ b/official/resnet/imagenet.py
@@ -83,11 +83,11 @@ def create_readable_names_for_imagenet_labels():
   (since 0 is reserved for the background class).
 
   Code is based on
-  https://github.com/tensorflow/models/blob/master/inception/inception/data/build_imagenet_data.py#L463
+  https://github.com/tensorflow/models/blob/master/research/inception/inception/data/build_imagenet_data.py
   """
 
   # pylint: disable=g-line-too-long
-  base_url = 'https://raw.githubusercontent.com/tensorflow/models/master/inception/inception/data/'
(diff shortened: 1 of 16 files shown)


nealwu:
@itsmeolivia Can you pull and rebase? Your change pulled in two other changes here for some reason.

official/resnet/imagenet.py
```diff
@@ -83,11 +83,11 @@ def create_readable_names_for_imagenet_labels():
   (since 0 is reserved for the background class).
 
   Code is based on
-  https://github.com/tensorflow/models/blob/master/inception/inception/data/build_imagenet_data.py#L463
+  https://github.com/tensorflow/models/blob/master/research/inception/inception/data/build_imagenet_data.py
```
shlens: This is deprecated and should instead point to https://github.com/tensorflow/models/blob/master/research/slim/datasets/build_imagenet_data.py

official/resnet/imagenet.py
```diff
@@ -83,11 +83,11 @@ def create_readable_names_for_imagenet_labels():
...
   """
 
   # pylint: disable=g-line-too-long
-  base_url = 'https://raw.githubusercontent.com/tensorflow/models/master/inception/inception/data/'
+  base_url = 'https://raw.githubusercontent.com/tensorflow/models/master/research/inception/inception/data/'
```
shlens: Ditto from above

official/resnet/imagenet.py
nealwu: Thanks Jon! Just set up a separate pull request for this: https://github.com/tensorflow/models/pull/2455

PR #2450, opened 2017-09-24
Fixing the filename check for PNG files

niyazpk:
I have a large number of files where the file-names are autogenerated using a script and they contain the name of the PNG file they were generated from.

So a filename could look like: "xxxxxxxxx.png_xxxxxxxxx.jpg"

This breaks the existing check. So tweaking it a bit to fix this.


niyazpk:
Signed Individual Contributor License Agreement.

nealwu:
Thanks @niyazpk.

PR #2452, opened 2017-09-25
Update resnet_model.py

ngujdi:
tf.nn.softmax_cross_entropy_with_logits(logits=logits, labels=self.labels) 
===> 
tf.nn.softmax_cross_entropy_with_logits(logits=self.predictions, labels=self.labels)


ngujdi:
close

PR #2460, opened 2017-09-26
Fixed broken paths in object_detection/protos

mahmoudakl:
This patch fixes the paths of imported modules in research/object_detection/protos. Without this patch, the following command fails:
$ protoc research/object_detection/protos/*.proto --python_out=.
This is due to a restructuring that occurred in this repository that moved libraries from the root directory.


derekjchow:
I think the correct thing to do would be to run the protoc step from the research directory. jch1@ WYDT?

jkurz:
I agree. Just updated Readme to export PythonPath from the research folder.

PR #2522, opened 2017-10-11
BNS=>DNS in arg description string

crizCraig:


derekjchow:
Thank you for the patch! LGTM

Issue #2505, opened 2017-10-06
Clarifications/suggestions for models/tutorials/rnn/ptb


Utumno:
I have been trying to adapt the models/tutorials/rnn/ptb to my needs. Along the way I run across some questions and suggestions for improvement - so this classifies as feature request :)

- why are we [exporting then importing the metagraph](https://github.com/tensorflow/models/blob/1464a721fef7da67e87b0f1f1a710b1e904e8764/tutorials/rnn/ptb/ptb_word_lm.py#L486-L496) ? This needs comments as it adds complexity to already complex code (hey it's a tutorial)
- since Supervisor is on its way to deprecation (wise, we have countless APIs for managed sessions, it only adds to learning curve) shouldn't we change [the session](https://github.com/tensorflow/models/blob/1464a721fef7da67e87b0f1f1a710b1e904e8764/tutorials/rnn/ptb/ptb_word_lm.py#L502-L504) to a MonitoredTrainingSession ?
- the [comment on using a static rnn](https://github.com/tensorflow/models/blob/1464a721fef7da67e87b0f1f1a710b1e904e8764/tutorials/rnn/ptb/ptb_word_lm.py#L226-L234) is badly outdated as it refers to the `tensorflow_models/tutorials/rnn/rnn.py's rnn()`. How should this be rephrased as of now ?
- some variables need renaming (max_max_epoch input_ etc) to more descriptive names

I can take this up as pull request but I need some feedback on the points above I can then incorporate into the pull.

Thanks

Utumno:
And a bug  - if run on a machine with gpus == 0 it [raises](https://github.com/tensorflow/models/blob/1464a721fef7da67e87b0f1f1a710b1e904e8764/tutorials/rnn/ptb/ptb_word_lm.py#L449) for num_gpus == 1 (the default value). However if setting num_gpus == 0 it will silently zero the tuples in ptb.util.import_state_tuples. This can lead to nasty goose chases. Here is a fix:

https://github.com/Utumno/models/commit/7d9ddc794006c7823cf59bce3f66da36bb269616

I switched to argparse as is [done in cifar](https://github.com/tensorflow/models/commits/master/tutorials/image/cifar10) and corrected some minor control flow issues.

[...]

PR #2524, opened 2017-10-11
Ptb fixups and using argparse

Utumno:
@nealwu, @protoget

- https://github.com/Utumno/models/commit/34e9041f196b412a1b0930ee74580786a6de8dae addresses https://github.com/tensorflow/models/pull/2276#issuecomment-329885088
 - https://github.com/Utumno/models/commit/cdd82c7bf7de8a9069a8cc1f9607936ff70132bd addresses this: https://github.com/tensorflow/models/issues/2505#issuecomment-334936514 and also adds argparse

### DIFF
--- a/tutorials/rnn/ptb/ptb_word_lm.py
+++ b/tutorials/rnn/ptb/ptb_word_lm.py
@@ -59,7 +59,9 @@ from __future__ import absolute_import
 from __future__ import division
 from __future__ import print_function
 
+import argparse
 import time
+import sys
 from functools import partial
 
 import numpy as np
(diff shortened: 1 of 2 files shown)


tutorials/rnn/ptb/ptb_word_lm.py
```diff
@@ -60,6 +60,7 @@
 from __future__ import print_function
 
 import time
+from functools import partial
```
protoget: I'd prefer using lambda config, is_training: tf.contrib..XXX()

tutorials/rnn/ptb/util.py
```diff
@@ -33,7 +33,7 @@ def export_state_tuples(state_tuples, name):
 
 def import_state_tuples(state_tuples, name, num_replicas):
   restored = []
-  for i in range(len(state_tuples) * num_replicas):
+  for i in range(len(state_tuples) * (num_replicas or 1)):
```
protoget: how about putting assert num_replicas > 0?
I think the caller of import_state_tuples should pass a valid num_replicas based on num of cpu/gpus.

Utumno: As  I explained https://github.com/tensorflow/models/issues/2505#issuecomment-334936514 "on a machine with gpus == 0 it [raises](https://github.com/tensorflow/models/blob/1464a721fef7da67e87b0f1f1a710b1e904e8764/tutorials/rnn/ptb/ptb_word_lm.py#L449) for num_gpus == 1 (the default value). "

Utumno: We should probably redo all the error checking but I believe this will do for now

tutorials/rnn/ptb/ptb_word_lm.py
Utumno: I believe that config is not a good idea to begin with. As I was trying to grok the code, having config passed as argument from method to method building the graph had one simply wonder what exactly the methods need from config. IOW this:

    def _build_rnn_graph_lstm(self, inputs, rnn_mode, hidden_size, num_layers,
                                           keep_prob, batch_size, is_training):

is much more easy to follow than:
 
      def _build_rnn_graph_lstm(self, inputs, config, is_training):

plus it shows that our model depends on the batch_size (not good ?). `is_training` calls for a subclass.

I would suggest going all the way to argparse dropping the configs - maybe keeping them as "presets"

PR #2536, opened 2017-10-13
Update data URL for CIFAR dataset.

sbrodehl:
CIFAR website is now using https.
Since urllib does not follow the http to https redirection and thus never retrieves any data, the data URL needs an update.


sbrodehl:
I signed it!

PR #2552, opened 2017-10-18
Updates to resnet, including using epochs in flags rather than steps

nealwu:


nealwu:
Good suggestion. I ended up cleaning up all the constants for consistency.

PR #2555, opened 2017-10-19
Correct 3x3 convolutions to 5x5 convolutions

### LINKED ISSUE #2545: Inception v1 missing 5x5 Convolutions
According to [Going Deeper with Convolutions](https://arxiv.org/pdf/1409.4842v1.pdf) the original inception modules had the following structure: 
image

However, while reviewing `models/research/slim/nets/inception_v1.py` I noticed that the inception modules were missing 5x5 convolutions and instead had an extra 3x3 convolution in `Branch_2`:
https://github.com/tensorflow/models/blob/6263692bf175b0e9160faa5614d5bdb1b237c878/research/slim/nets/inception_v1.py#L84-L97

Would you like me to submit a PR to correct this? If so, how would you like to handle the pre-trained networks that you have available for download?

/cc @nealwu @sguada

JoshVarty:
Fixes #2545 

This PR corrects the 3x3 convolutions to 5x5 convolutions in Inception v1.

I have run the tests in `inception_v1_test.py` and ensured that they pass following these changes.

### DIFF
--- a/research/slim/nets/inception_v1.py
+++ b/research/slim/nets/inception_v1.py
@@ -89,7 +89,7 @@ def inception_v1_base(inputs,
             branch_1 = slim.conv2d(branch_1, 128, [3, 3], scope='Conv2d_0b_3x3')
           with tf.variable_scope('Branch_2'):
             branch_2 = slim.conv2d(net, 16, [1, 1], scope='Conv2d_0a_1x1')
-            branch_2 = slim.conv2d(branch_2, 32, [3, 3], scope='Conv2d_0b_3x3')
+            branch_2 = slim.conv2d(branch_2, 32, [5, 5], scope='Conv2d_0b_5x5')
           with tf.variable_scope('Branch_3'):
             branch_3 = slim.max_pool2d(net, [3, 3], scope='MaxPool_0a_3x3')
             branch_3 = slim.conv2d(branch_3, 32, [1, 1], scope='Conv2d_0b_1x1')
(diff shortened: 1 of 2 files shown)


research/slim/nets/inception_v1.py
```diff
@@ -225,7 +225,7 @@ def inception_v1_base(inputs,
             branch_1 = slim.conv2d(branch_1, 320, [3, 3], scope='Conv2d_0b_3x3')
           with tf.variable_scope('Branch_2'):
             branch_2 = slim.conv2d(net, 32, [1, 1], scope='Conv2d_0a_1x1')
-            branch_2 = slim.conv2d(branch_2, 128, [3, 3], scope='Conv2d_0a_3x3')
+            branch_2 = slim.conv2d(branch_2, 128, [5, 5], scope='Conv2d_0b_5x5')
```
JoshVarty: There was also a scope name change here. (`0a` -> `0b`)

research/slim/nets/inception_v1_test.py
```diff
@@ -114,7 +114,7 @@ def testModelHasExpectedNumberOfParameters(self):
       inception.inception_v1_base(inputs)
     total_params, _ = slim.model_analyzer.analyze_vars(
         slim.get_model_variables())
-    self.assertAlmostEqual(5607184, total_params)
+    self.assertAlmostEqual(5988112, total_params)
```
JoshVarty: I've replaced the old value with the value returned by `slim.model_analyzer.analyze_vars()`.

JoshVarty:
This PR doesn't address the missing LocalResponseNorm operations or the redundant 1x1 Conv layer at the head of the network. If you think these are worth addressing, let me know and I can add them too.

nealwu:
Nice, let me loop in a few people for review.

sguada:
The problem with this change is that it would break all the previous checkpoints.

JoshVarty:
Yes it would break them, but I believe they're incorrect right now. Is it a problem to regenerate them and release the new checkpoints?

PR #2567, opened 2017-10-20
Fix flags issue in cifar10 example

svebk:
The way flags were handled actually did not allow to pass the `--train_dir` parameter to the script `cifar10_train.py` because the flags were parsed in `cifar10.py` before this argument was added.

I use explicitly the `tf.flags._global_parser` as parser and I am able to pass both the `--train_dir` and the `--data_dir` parameters to the script `cifar10_train.py`.


levis2871:
mark

danaodai:
Thx, Mark

PR #2569, opened 2017-10-21
Upgrade input pipeline to 1.4, fix shuffle.

MarkDaoust:
Switch Dataset from contrib.data to tf.data.

Fix shuffle: The shuffle is only executed in graph, not when the method is called. If you create two iterators you get two different shuffles.

It's also important to have the shuffle before the batch so that you:

Shuffle 1e5 examples, not 1e5 batches.
Different items are batched together each epoch.


nealwu:
Thanks Mark. This should be covered by https://github.com/tensorflow/models/pull/2690 now, so I'll close.

PR #2583, opened 2017-10-25
Add a data_format flag to the models + other fixes

nealwu:


official/mnist/mnist.py
```diff
@@ -35,8 +35,19 @@
 parser.add_argument('--model_dir', type=str, default='/tmp/mnist_model',
                     help='The directory where the model will be stored.')
 
-parser.add_argument('--steps', type=int, default=20000,
-                    help='Number of steps to train.')
+parser.add_argument('--train_epochs', type=int, default=40,
+                    help='Number of epochs to train.')
+
+parser.add_argument(
+    '--data_format', type=str, default=None,
```
k-w-w: Should we set a default here? If not, then the README needs to be updated

official/resnet/imagenet_main.py
```diff
@@ -53,6 +53,12 @@
     '--batch_size', type=int, default=32,
     help='Batch size for training and evaluation.')
 
+parser.add_argument(
+    '--data_format', type=str, default=None,
```
k-w-w: Same here

official/mnist/mnist.py
nealwu: The default is `None`, in which case it sets `data_format` based on a check to `tf.test.is_built_with_cuda()`. This is just an option to override that (e.g., run with NHWC even though you have a GPU).

nealwu: I can add to the help message to explain the behavior when the flag is unspecified.

nealwu: Added.

k-w-w: Ohhh I see! Thanks for the changes

PR #2607, opened 2017-10-27
Allow data_dir to be a remote location like GCS

jlewi:
* Download the data to a local temporary directory so that data_dir
  can be a remote path like GCS.


jlewi:
@elibixby Could you review this?

jlewi:
Ping is there anyone that can review this?

PR #2615, opened 2017-10-27
add label maps for kitti and open image datasets.

tombstone:


jasmine-ren:
There are only 2 classes in models/research/object_detection/data/kitti_label_map.pbtxt. Is this as intended?

PR #2616, opened 2017-10-27
Allow ssd anchor generator to specify scales.

tombstone:
add interpolated scale at a give aspect ratio. This change breaks
the existing checkpoints because the anchor orders are different.

This change should be merged along newly generated frozen ssd models.


tombstone:
I signed it!

PR #2621, opened 2017-10-28
Use config parsing routines from utils/config_utils.py 

tombstone:
Use config parsing routines from utils/config_utils.py in train.py


tombstone:
Thanks Derek!

PR #2640, opened 2017-10-30
Fixing visualization_utils, updating draw_boxes

Davidnet:
Removing unnecessary parenthesis in function definition


Davidnet:
I signed it!

derekjchow:
Thanks for the PR, but we've already merged a patch with a similar fix.

PR #2643, opened 2017-10-30
Evaluate multiple checkpoints.

Adrrei:
It may occasionally not be possible to evaluate the various models from start (first checkpoint) to finish (last checkpoint) while simultaneously training the model (limited resources). I've added a new flag, 'eval_from_ckpt', which when set will evaluate all checkpoints from and including the specified one.

Example usage:
`python C:/.../object_detection/eval.py \ --logtostderr \ --pipeline_config_path=C:/.../models/name.config \ --eval_dir=C:/.../testing \ --checkpoint_dir=C:/.../training \ --eval_from_ckpt=840`

In this case, all checkpoints from checkpoint 840 to 'latest_checkpoint' will be evaluated.


MarkDaoust:
Hi, thanks for taking the time to make a PR. 

But we can't submit this like this. It looks like you replayed master on top of your PR, instead of the other way around (?).

Probably the easiest thing to do now to get a submittable PR is to copy your branch (`git checkout -b rebase-test`), and  `git rebase master` to replay your changes on master, and make a new PR from that branch.

Adrrei:
Sorry, this was unintended.
Please disregard the latest pull request.

Issue #2571, opened 2017-10-22
Reason tf_models/research/slim/BUILD includes explicit dependency on //tensorflow recently?


gyang274:
I submitted an issue regarding the models/research/slim/BUILD file, specifically, why several libraries, tests and binaries have `deps = ["//tensorflow", ], `?

From my limited knowledge, is this deps looking for a directory under slim, e.g., models/research/slim/tensorflow?

To avoid repeat, would you kindly look at the issue reported at https://github.com/tensorflow/serving/issues/628, which more details?

Please let me know which is the better place, I will move.

Thank you.

drpngx:
@sguada any comment on that?

kirilg:
I think it's a bug in syncing the repos, this dep should not be there. This package fails to build on its own as well. e.g.

```
git clone --recursive https://github.com/tensorflow/models
cd models/research/slim
bazel build ...
```

ERROR: /tmp/slim/models/research/slim/BUILD:51:1: no such package 'tensorflow': BUILD file not found on package path and referenced by '//:download_and_convert_flowers'.
ERROR: Analysis of target '//:download_and_convert_flowers' failed; build aborted.

PR #2680, opened 2017-11-01
comment out deps on "//tensorflow"

gyang274:
There are several py_library(), py_binary() and py_test() claimed deps on "//tensorflow" explicitly. This causes the BUILD failed. This is discussed in the issue submitted earlier at [tensorflow/models 2571](https://github.com/tensorflow/models/issues/2571).

As my understanding, the `deps = [ "//tensorflow" ]` is looking for a directory tensorflow under the slim, which doesn't exist. Remove the deps fixed the issue, as long as the tensorflow is built and available.

Another approach might define the tensorflow library in workspace, by linking it to outside github repo? However, in this way, it could lead to replicated tensorflow modules?

Thank you.

### DIFF
--- a/research/slim/BUILD
+++ b/research/slim/BUILD
@@ -13,7 +13,7 @@ py_library(
     name = "dataset_utils",
     srcs = ["datasets/dataset_utils.py"],
     deps = [
-        "//tensorflow",
+        # "//tensorflow",
     ],
 )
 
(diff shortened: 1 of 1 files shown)


gyang274:
I signed it!

sguada:
Fixes #2571

PR #2685, opened 2017-11-02
Fix nasnet image classification and object detection

shlens:
Fix nasnet image classification and object detection by moving the
option to turn ON or OFF batch norm training into it's own arg_scope
used only by detection


shlens:
Jenkins, test this please.

PR #2686, opened 2017-11-02
fix input_shape parse error

scotthuang1989:
the code to parse a string: [-1, 100, 200, 3] to list is this:

```
   input_shape = [
        int(dim) if dim != '-1' else None
        for dim in FLAGS.input_shape.split(',')
```
but by splitting this with split, this will end up with [-1, 100, 200, 3. I use RE to parse this input.


scotthuang1989:
I signed it

scotthuang1989:
@tombstone  . I try to export my own model today, but 

> export_inference_graph.py

 still raise an exception when I try to specify input_shape parameter.  So I check the code, it still not be fixed.  please take a look at this PR, it can fix it.

PR #2690, opened 2017-11-02
Changing tf.contrib.data to tf.data for release of tf 1.4

k-w-w:

PR #2691, opened 2017-11-02
Update version string to 1.4 in docker_test.sh.

case540:


case540:
I think this is the version update you mentioned from https://github.com/tensorflow/models/pull/2690

PR #2717, opened 2017-11-06
Add simplifications requested by get_started writers.

MarkDaoust:
cc: @jugglerix


MarkDaoust:
This is included in the second PR.

Issue #2715, opened 2017-11-06
SyntaxNet dockerfile not building


MathiasDesch:
I am unable to build the current version of Syntaxnet using the Dockerfile.
A first issue is that `libpng12-dev` does not exist anymore. 
A second issue (is I replace `libpng12-dev` to `libpng-dev`) is the during the build using syntaxnet.

I think there might have been some update of tensorflow without checking if everything is working properly. 

Does anyone succeed to build syntaxnet ? @bogatyy is the commit 4594a1ea58f7a8df67d4c1112f34dd4c12c35225 might be the cause of the problem.

I have tried to checkout previous commit without success.

angersson:
Please provide details about what platform you are using  (operating system, architecture). Also include your TensorFlow version. Also, did you compile from source or install a binary?  Make sure you also include the exact command if possible to produce  the output included in your test case. If you are unclear what to include  see the issue template displayed in  [the Github new issue template](https://github.com/tensorflow/tensorflow/issues/new).

 We ask for this in the issue submission template, because    it is really difficult to help without that information. Thanks!

bogatyy:
@MathiasDesch thanks for the fix!

PR #2723, opened 2017-11-07
Fix Dockerfile build

MathiasDesch:
Dockerfile build was failing. Detail of the fix is  :

* `libpng12-dev` is not present in repositories anymore. My guess is a change from based image `open-jdk8`
* add `patch` package to make bazel work with @opt
* CI configured is trying to get git version, which leads to failing since the parents' folders are not copied. I changed to use the simpler configure script
* bazel version was too old for the new tensorflow version
* bazel not finding PYTHON_BIN_PATH, add ENV command as a workaround

Docker builds and can run the image on this PR. However, the common error with `Duplicate tensor` is still present, preventing from a correct usage, the goal of this PR is not to solve this problem since it has already been worked on in different issues.

This PR obviously fix my issue #2715

### DIFF
--- a/research/syntaxnet/Dockerfile
+++ b/research/syntaxnet/Dockerfile
@@ -71,7 +71,7 @@ ENV PYTHON_BIN_PATH=/usr/bin/python
 # source. This makes it more convenient to re-compile DRAGNN / SyntaxNet for
 # development (though not as convenient as the docker-devel scripts).
 RUN cd $SYNTAXNETDIR/syntaxnet/tensorflow \
-    && ./configured CPU \
+    && ./configure CPU \
     && cd $SYNTAXNETDIR/syntaxnet \
     && bazel build -c opt @org_tensorflow//tensorflow:tensorflow_py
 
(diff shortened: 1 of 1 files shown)


MathiasDesch:
I signed it!

bogatyy:
This is great, thank you for contributing the fix!

research/syntaxnet/Dockerfile
```diff
@@ -55,20 +56,22 @@ RUN python -m pip install \
...
 COPY tensorflow $SYNTAXNETDIR/syntaxnet/tensorflow
 
+# Workaround solving the PYTHON_BIN_PATH not found problem
+ENV PYTHON_BIN_PATH=/usr/bin/python
 # Compile common TensorFlow targets, which don't depend on DRAGNN / SyntaxNet
 # source. This makes it more convenient to re-compile DRAGNN / SyntaxNet for
 # development (though not as convenient as the docker-devel scripts).
 RUN cd $SYNTAXNETDIR/syntaxnet/tensorflow \
-    && tensorflow/tools/ci_build/builds/configured CPU \
+    && ./configure CPU \
```
bogatyy: This seems to be a typo?
`./configure` does not really take "CPU" as a command-line argument.

MathiasDesch: That's true. I can do another PR if you wish to fix that. It's working properly with it but it is kind of misleading the user.

PR #2732, opened 2017-11-08
Shuffle an exact epoch of 32561 examples in wide_deep.py rather than 100k

nealwu:


nealwu:
@av8ramit it looks like the tests have passed on Jenkins, but it's not reporting the results back here. Any clue what's going on?

av8ramit:
Jenkins, test this please.

PR #2734, opened 2017-11-08
Use range rather than xrange for compatibility with Python3

cancan101:


research/slim/nets/cyclegan.py
```diff
@@ -223,7 +223,7 @@ def cyclegan_generator_resnet(images,
           stride=1,
           activation_fn=tf.nn.relu,
           padding='VALID'):
-        for block_id in xrange(num_resnet_blocks):
+        for block_id in range(num_resnet_blocks):
```
cclauss: The preferred approach is to put __from six.moves import xrange__ into the imports section at the top of the file and then leave this line as xrange().  

The reason for this is that range() in Python 2 will actually build a list at runtime of all numbers from 0 thru num_resnet_blocks-1.  This is fine when num_resnet_blocks is a small number but will take up a lot of memory if num_resnet_blocks gets massive.  The import above will mean that _an iterator will be used in both Py2 and Py3_.

research/slim/nets/cyclegan.py
```diff
@@ -19,6 +19,7 @@
 
 import google3
 import numpy as np
+from six.moves import xrange
```
cclauss: LGTM

PR #2747, opened 2017-11-09
fix: create_pascal_tf_record_test don't compatible with py3

scotthuang1989:
create_pascal_tf_record_test.py will fail on python 3 due to comparing bytes with unicode issue.

### DIFF
--- a/research/object_detection/create_pascal_tf_record_test.py
+++ b/research/object_detection/create_pascal_tf_record_test.py
@@ -16,6 +16,7 @@
 """Test for create_pascal_tf_record.py."""
 
 import os
+import sys
 
 import numpy as np
 import PIL.Image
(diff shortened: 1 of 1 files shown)


research/object_detection/create_pascal_tf_record_test.py
```diff
@@ -80,12 +94,12 @@ def test_dict_to_tf_example(self):
...
         example.features.feature['image/filename'].bytes_list.value,
-        [image_file_name])
+        [self._cvt_bytes_py3(image_file_name)])
     self._assertProtoEqual(
         example.features.feature['image/source_id'].bytes_list.value,
-        [image_file_name])
+        [self._cvt_bytes_py3(image_file_name)])
     self._assertProtoEqual(
-        example.features.feature['image/format'].bytes_list.value, ['jpeg'])
+        example.features.feature['image/format'].bytes_list.value, [self._cvt_bytes_py3('jpeg')])
```
cclauss: Why not __[b'jpeg']__?

PR #2756, opened 2017-11-09
Add savedmodel example.

MarkDaoust:
Factor out `iris_data` and add `savedmodel_estimator`. 

cc: @jugglerix


MarkDaoust:
Change of plans.

PR #2772, opened 2017-11-11
Fix use of `filter` in `filter_variables`

rstebbing:
The current use of `filter` is *not* Python 3 compatible.

### DIFF
--- a/research/object_detection/utils/variables_helper.py
+++ b/research/object_detection/utils/variables_helper.py
@@ -42,7 +42,7 @@ def filter_variables(variables, filter_regex_list, invert=False):
     a list of filtered variables.
   """
   kept_vars = []
-  variables_to_ignore_patterns = filter(None, filter_regex_list)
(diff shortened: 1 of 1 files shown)


research/object_detection/utils/variables_helper.py
```diff
@@ -42,7 +42,7 @@ def filter_variables(variables, filter_regex_list, invert=False):
     a list of filtered variables.
   """
   kept_vars = []
-  variables_to_ignore_patterns = filter(None, filter_regex_list)
+  variables_to_ignore_patterns = list(filter(None, filter_regex_list))
```
murarugeorgec: I know this is an old issue, but I think this PR can be closed.
(it is ok with only the filter -  it would create a filter object and you can iterate through it - it gives you back the next element)

rstebbing: Ok, sounds good :+1:.

PR #2789, opened 2017-11-14
Fix unrecognized arguments

shaofan:
Fix unrecognized arguments when run with arguments defined in cifar10_train or cifar10_eval


shaofan:
I signed it!

PR #2792, opened 2017-11-14
Add autograd v1.1.13 dependency to manual install

strubell:
New bulk component tests fail with the latest version of autograd (1.2), which was rewritten and no longer contains the `containers` module. They pass with v1.1.13, which is the latest pre-1.2 version available in pip.


strubell:
CLA signed

bogatyy:
Thanks Emma!

Issue #2778, opened 2017-11-13
Odd batch_size specific behaviour with nasnet_large on ImageNet validation


rwightman:
I've been trying to reproduce the ImageNet validation accuracies posted with the pretrained weights.

I cannot reproduce the results with an image size of 331 and the default batch size of 100 used in the eval_image_classifer script. With defaults I get roughly 64% top-1 and 74% top-5 using the command:

`python eval_image_classifier.py --dataset_dir ~/imagenet/ --dataset_split_name validation --model_name nasnet_large --checkpoint_path ~/nasnet_large/model.ckpt  --dataset_name imagenet --moving_average_decay=0.9999`

However, if I change the batch size down to 50, I hit the posted results dead on, 82.7% top-1 and 92.6% top-5. This seems rather odd to me. I've never experienced that significant of a performance drop by changing the batch size by a factor of two at eval. 

I noticed that with the batch size of 100, I could improve performance by dropping the eval image size down. There is an optimal value somewhere in the 280-290 range before it starts to fall off. My best result with batch size of 100 was at 288x288 with a 80.8% top-1 and 95.2% top-5, it falls off quickly as the image size increases from there. 

Is this possibly some sort of overflow in batch norm or another op in the network?

Note, I'm using latest released pip install of Tensorflow 1.4 for GPU. CUDA 8.0, cuDNN 6.

chenghuige:
@rajatmonga , I face this also... Ando when train I use eval_batch_size 50 and eval result during train is as expected. I frozen ops for inference, and when inference from pb model saved and set inference buffer size 100, I got worse inference result. But when set inference batch size to 50 the result is ok.
Is it due to moving average decay 0.9999  as other models like incpetion V4 do not met this problem.
Also even if use batch size 50 for different run I will get slightly different inference result that is confusing.

chenghuige:
1. I verified on two machines, there might be some env idff, running same code inference using same model. I have tested both tf 1.3 and tf 1.4 on one machine batch size 100 and 50 all get cider 1.73, 
on another machine batch size 100 can only get cider 1.49.
2. On both machine, different inference run will get slightly different result.

chenghuige:
Is it the problem related to below code in nasnet.py ? seems is_batchnorm_training is always True, even
when inference, if using nets_factory.py as we only call
 func(images, num_classes, is_training=is_training, **kwargs)

def build_nasnet_large(images, num_classes,
                       is_training=True, is_batchnorm_training=True,
[...]

PR #2807, opened 2017-11-15
Update eval_image_classifier.py default batch_size from 100 to 50 to avoid bug.

BarretZoph:
Right now the large NASNet-A model, when evaluated with the default batch_size of 100, runs into a bug. This change will allow the NASNet-A model to work with the default values. See #2778 for details.


shlens:
@sguada The reason for this change is that there seems to be weird [bug](https://github.com/tensorflow/models/issues/2778) in a low-level library that only occurs for batch_size = 100. This small change might save quite a bit of user head-ache.

PR #2810, opened 2017-11-16
Update exporting_models.md

aristeiaa:
Updated so that commented path matches path in markdown.


aristeiaa:
I signed it!

research/object_detection/g3doc/exporting_models.md
```diff
@@ -11,8 +11,8 @@ After you've identified a candidate checkpoint to export, run the following
 command from tensorflow/models/research/object_detection:
```
derekjchow: Remove object_detection instead

derekjchow:
Fixed [here](https://github.com/tensorflow/models/pull/2940)

PR #2858, opened 2017-11-21
Fix doc-string for custom_estimator.py

MarkDaoust:
Fix doc-string for samples/core/get_started/custom_estimator.py


MarkDaoust:
needs sync

PR #2859, opened 2017-11-21
Project import generated by Copybara.

joel-shor:
PiperOrigin-RevId: 175826086

I thought this was already submitted, but I can't find any history of it. Launch entry #207179

@martinwicke @nealwu


joel-shor:
@nealwu I think I've appropriately changed the history so the commits are from me. Please let me know if this still is incorrect!

PR #2868, opened 2017-11-22
Don't reuse the same cell when creating MultiRNNCell in ptb_word_lm.py

nealwu:


nealwu:
Jenkins, test this please.

PR #2878, opened 2017-11-22
Fixes bug where unwrapped cells are shared between layers

duncanriach-nvidia:
The bug is causing the weights to be shared shared between layers.

Without this patch, if you train the model (--size=small) and take a look at tf.trainable_variables() after each epoch, you'll see that there is only one set of weights, even though the model has two layers. Note that the default configuration of the model will not currently run (at least it doesn't for me), you have to add --rnn_mode=basic or --rnn_mode=block to the command line (that is an independent issue).

When the medium and large (--size=medium and --size=large) size models are trained, there *does* seem to be an independent set of weights per layer, at least in the training and validation models, and so the bug seems to be masked by the LSTM cells being wrapped with DropoutWrapper in those cases. However, I believe that there is only one set of weights in the test model (which doesn't apply DropoutWrapper), but I have not confirmed this. This bug-masking effect seems to be related to a difference between how DropoutWrapper and BasicLSTMCell (and the other LSTM cells) inherit from RNNCell -- DropoutWrapper overrides `__call__()` whereas the LSTM cells do not. I don't understand how wrapping with DropoutWrapper masks the bug. The way it currently is, the model should consistently, and incorrectly, share weights between layers. 

I believe that this patch makes the model use the LSTM cells and DropoutWrapper correctly, as they are used in other models, including tensorflow/models/translate/seq2seq_model.py and tensorflow/nmt, by instantiating an independent instance of the cell for each layer. *Therefore I think that this patch should be applied even though the bug is masked in model configurations that use dropout*.

### DIFF
--- a/tutorials/rnn/ptb/ptb_word_lm.py
+++ b/tutorials/rnn/ptb/ptb_word_lm.py
@@ -213,13 +213,21 @@ class PTBModel(object):
     # Slightly better results can be obtained with forget gate biases
     # initialized to 1 but the hyperparameters of the model would need to be
     # different than reported in the paper.
-    cell = self._get_lstm_cell(config, is_training)
+    def get_unwrapped_single_cell():
+      return self._get_lstm_cell(config, is_training)
+    def get_wrapped_single_cell():
+      return tf.contrib.rnn.DropoutWrapper(
+          get_unwrapped_single_cell(), output_keep_prob=config.keep_prob)
     if is_training and config.keep_prob < 1:
-      cell = tf.contrib.rnn.DropoutWrapper(
-          cell, output_keep_prob=config.keep_prob)
+      def get_single_cell():
+        return get_wrapped_single_cell()
+    else:
+      def get_single_cell():
+        return get_unwrapped_single_cell()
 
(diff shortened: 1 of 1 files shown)


duncanriach-nvidia:
Update: I trained the medium model both with and without the patch and I got roughly the same perplexity for the training, validation, *and* the test datasets. These numbers also match those reported in the original paper that the model is based on. This suggests to me that in the test model there *are* independent weights for the two layers. I don't understand how that's possible. This doesn't invalidate the patch however, as it definitely is needed to make the small configuration, or any configuration that uses non-wrapped cells, work.

Medium model without patch
Train: 45.678
Valid: 87.435
Test: 83.928

Medium model with patch
Train: 45.763
Valid: 87.191
Test: 83.372

k-w-w:
Thanks for noticing this bug, and for submitting your fixes! We also noticed this issue recently, and [fixed it in this commit](https://github.com/tensorflow/models/commit/25a16a2940b952d3f899abc55d107a5106d7790c#diff-c0c93e71f6fd8d214ab2489fbb6fc7e2).

PR #2893, opened 2017-11-25
Add github-account owner to `tensorflow/models/gan` CODEOWNERS.

joel-shor:
@nealwu @martinwicke  Add my github account to CODEOWNERS.

Note: I think some of the entries in CODEOWNERS are `@google` email addresses and some are github accounts. It might make sense to standardize at some point.


nealwu:
The entries are all intended to be GitHub accounts. Which ones do you think aren't? We should fix them.

PR #2896, opened 2017-11-25
Fix TFGAN tutorial table of contents

joel-shor:
@nealwu This corresponds to cl/176659312.

Note that this is the same convention as the slim tutorial's TOC. It renders properly, but doesn't actually like to sections in the jupyter notebook.


joel-shor:
It *should* work according to: http://sebastianraschka.com/Articles/2014_ipython_internal_links.html, and it works when I run it locally. I tried the other options from that website, and they all work locally but don't link properly when github renders it. I don't know how to debug from here, so I think I'd prefer to submit as is and fix it later if a solution presents itself.

Sound ok?

nealwu:
I see, sounds good then.

PR #2898, opened 2017-11-26
change to lfads to allow training of encoder weights only

cpandar:
Changes allow one to train an lfads model while only allowing modifications to the weights of the encoder. This could be useful to training a model on a certain set of conditions, and then only training the encoder on a set of held-out conditions, to see if the original generator's dynamics are capable of describing the held-out conditions.

ATTN: @sussillo


sussillo:
LGTM - permits an experiment to train encoders only for testing held out conditions.

Issue #1497, opened 2017-05-22
Resnet_v2 wrong default preprocessing and default_image_size


asanakoy:
In the readme file it is specified that 

> ^ ResNet V2 models use Inception pre-processing and input image size of 299 (use --preprocessing_name inception --eval_image_size 299 when using eval_image_classifier.py).

But in the source code it is different:
1. ResNet V2 preprocessing is set to `vgg_preprocessing`  [preprocessing_factory.py#L61](https://github.com/tensorflow/models/blob/master/slim/preprocessing/preprocessing_factory.py#L61)
2. Default image size is set to 224 [https://github.com/tensorflow/models/blob/master/slim/nets/resnet_v2.py#L219](resnet_v2.py#L219)

I don't know which source to believe. One of this is wrong.

micklexqg:
same puzzled

shastakr:
same puzzled 2

tensorflowbutler:
Hi There,
 We are checking to see if you still need help on this, as this seems to be considerably old issue. Please update this issue with the latest information, code snippet to reproduce your issue and error you are seeing.
 If we don't hear from you in the next 7 days, this issue will be closed automatically. If you don't need help on this issue any more, please consider closing this.

PR #2901, opened 2017-11-27
Change resnet_v2's wrong default preprocessing & default_image_size

shastakr:
Related with issue #1497 

I tested with resnet_v2_152 on my local environment.
I found out that document is correct and code is wrong.


ZERODARK00:
You are right, thx. Help a lot

PR #2902, opened 2017-11-27
Fix two issues with tensorflow_models/slim downloading imagenet.

joel-shor:
@nealwu @sguada 
1) Fix word dir
2) Fix download url

Corresponds to cl/176965289.

### DIFF
--- a/research/slim/datasets/download_and_convert_imagenet.sh
+++ b/research/slim/datasets/download_and_convert_imagenet.sh
@@ -58,7 +58,7 @@ DATA_DIR="${1%/}"
 SCRATCH_DIR="${DATA_DIR}/raw-data/"
 mkdir -p "${DATA_DIR}"
 mkdir -p "${SCRATCH_DIR}"
-WORK_DIR="$0.runfiles/third_party/tensorflow_models/research/slim"
+WORK_DIR="$0.runfiles/__main__"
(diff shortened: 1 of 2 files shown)


joel-shor:
@shlens as well

scotthuang1989:
@joel-shor  I apply your patch. still get this error: 

> ./download_and_convert_imagenet.sh: line 67: ./download_and_convert_imagenet.sh.runfiles/__main__/datasets/download_imagenet.sh: No such file or directory

I guess it is related to bazel. but I don't find any instruction in the readme.md to work around this. how do you do it?

joel-shor:
With the code in the PR, I *don't* get that error. '__main__' renders fine.

Is there someone who owns this code and can get it to work? If not, it should probably be deleted.

joel-shor:
I don't get that error. Also, to be sure, the word `main` in your code is actually surrounded by two underscores, correct?

What was the command you used?

scotthuang1989:
@joel-shor   I just change my working directory to  

> research/slim/datasets

 and execute  

> download_and_convert_imagenet.sh
I don't do any other actions. I miss something?

joel-shor:
The instructions are wrong. It needs to be built with bazel. I'll make the
change shortly.

PR #2910, opened 2017-11-27
Fix create pet

sheromon:
I had an issue running the pet tutorial having to do with the new faces_only option. I think my changes resolve the issue and give the desired behavior? Updating the tutorial to reflect the updated path the create_pet_tf_records.py file while I'm at it.


sheromon:
I signed it!

PR #2934, opened 2017-12-01
Modify the inception code to work with TfJob

jlewi:
To make the inception model work with TfJob we need to configure it based on TF_CONFIG as opposed to command line arguments.


jlewi:
Sorry didn't mean to create this pull request.

PR #2935, opened 2017-12-01
Inception kubeflow

jlewi:
To make the inception model work with TfJob we need to configure it based on TF_CONFIG as opposed to command line arguments.


jlewi:
Didn't mean to create this pull request.

PR #2956, opened 2017-12-04
add feature_columns directory

MarkDaoust:
to samples/core/get_started


MarkDaoust:
This needs some rework

PR #2974, opened 2017-12-07
Use /usr/bin/env python instead of /usr/bin/python

jinmel:
For python users using virtualenv or pyenv to use their own python this causes unexpected error.

Using python set by environment variable would fix


jinmel:
I signed it!

PR #2990, opened 2017-12-11
fix #327 nccl not found error

markroxor:
I spent a lot of time looking for a fix for -
`fatal error: external/nccl_archive/src/nccl.h: No such file or directory `

Documenting the installation of `nvcc` saves a lot of set-up time for beginners.


markroxor:
I signed it!

PR #2993, opened 2017-12-12
Update tutorials/rnn/README.md with the correct Quick Draw link

nealwu:
cc @vcarbune


nealwu:
The existing link gives a 404 since the tutorial is not available in r1.4, but the master version is up.

PR #3007, opened 2017-12-14
fix eval_training_data bug

jzhugithub:
`eval_training_data` flag seems no effect.
Since no code handles `eval_training_data` flag, `input_config = configs['eval_input_config']` will be always executed, even though ‘eval_training_data’ is set to `Ture`.
When ‘eval_training_data’ is set to `Ture`, `input_config` should be `configs['train_input_config']`.
I fix the bug in pull request.


derekjchow:
Good catch! LGTM

PR #3024, opened 2017-12-19
[mnist]: Updates

asimshankar:
- Use the object-oriented tf.layers API instead of the functional one.
  The object-oriented API is particularly useful when using the model
  with eager execution.
- Update unittest to train, evaluate, and predict using the model.
- Add a micro-benchmark for measuring step-time.
  The parameters (batch_size, num_steps etc.) have NOT been tuned,
  the purpose with this code is mostly to illustrate how model
  benchmarks may be written.

These changes are made as a step towards consolidating model definitions
for different TensorFlow features (like eager execution and support for
TPUs in
https://github.com/tensorflow/tensorflow/tree/master/tensorflow/contrib/eager/python/examples/mnist
and
https://github.com/tensorflow/tpu-demos/tree/master/cloud_tpu/models/mnist)

CC @martinwicke @jhseu @ispirmustafa

### DIFF
--- a/official/mnist/README.md
+++ b/official/mnist/README.md
@@ -21,6 +21,13 @@ python mnist.py
 The model will begin training and will automatically evaluate itself on the
 validation data.
 
+Illustrative unittests and benchmarks can be run with:
+
+```
+python mnist_test.py
+python mnist_test.py --benchmarks=.
+```
+
 ## Exporting the model
 
 You can export the model into Tensorflow [SavedModel](https://www.tensorflow.org/programmers_guide/saved_model) format by using the argument `--export_dir`:
--- a/official/mnist/README.md
+++ b/official/mnist/README.md
@@ -21,7 +21,7 @@ python mnist.py
 The model will begin training and will automatically evaluate itself on the
 validation data.
 
-Illustrative unittests and benchmarks can be run with:
+Illustrative unit tests and benchmarks can be run with:
 
 ```
 python mnist_test.py
--- a/official/mnist/README.md
+++ b/official/mnist/README.md
@@ -33,7 +33,7 @@ python mnist_test.py --benchmarks=.
 You can export the model into Tensorflow [SavedModel](https://www.tensorflow.org/programmers_guide/saved_model) format by using the argument `--export_dir`:
 
 ```
-python mnist.py --export_dir /tmp/mnist_saved_model 
+python mnist.py --export_dir /tmp/mnist_saved_model
 ```
 
 The SavedModel will be saved in a timestamped directory under `/tmp/mnist_saved_model/` (e.g. `/tmp/mnist_saved_model/1513630966/`).
@@ -42,7 +42,7 @@ The SavedModel will be saved in a timestamped directory under `/tmp/mnist_saved_
 Use [`saved_model_cli`](https://www.tensorflow.org/programmers_guide/saved_model#cli_to_inspect_and_execute_savedmodel) to inspect and execute the SavedModel.
 
 ```
-saved_model_cli run --dir /tmp/mnist_saved_model/TIMESTAMP --tag_set serve --signature_def classify --inputs image_raw=examples.npy
+saved_model_cli run --dir /tmp/mnist_saved_model/TIMESTAMP --tag_set serve --signature_def classify --inputs image=examples.npy
 ```
 
 `examples.npy` contains the data from `example5.png` and `example3.png` in a numpy array, in that order. The array values are normalized to values between 0 and 1.
--- a/official/mnist/mnist.py
+++ b/official/mnist/mnist.py
@@ -63,6 +63,7 @@ parser.add_argument(
     type=str,
     help='The directory where the exported SavedModel will be stored.')
 
+
 def train_dataset(data_dir):
   """Returns a tf.data.Dataset yielding (image, label) pairs for training."""
   data = input_data.read_data_sets(data_dir, one_hot=True).train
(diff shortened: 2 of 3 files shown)


official/mnist/mnist.py
```diff
@@ -75,141 +76,128 @@ def eval_dataset(data_dir):
...
+    y = tf.reshape(inputs, self._input_shape)
+    y = self.conv1(y)
+    y = self.max_pool2d(y)
+    y = self.conv2(y)
+    y = self.max_pool2d(y)
+    y = tf.layers.flatten(y)
+    y = self.fc1(y)
+    y = self.dropout(y, training=training)
+    return self.fc2(y)
+
```
martinwicke: I suppose the model part would stop here, the rest would be part of the (hopefully generic) Estimator executor?

martinwicke: (pending some pulling out of optimizer, etc., and more generic naming of the inputs)

official/mnist/mnist.py
```diff
@@ -75,141 +76,128 @@ def eval_dataset(data_dir):
...
+def model_fn(features, labels, mode, params):
+  """The model_fn argument for creating an Estimator."""
+  model = Model(params['data_format'])
+  if mode == tf.estimator.ModeKeys.PREDICT:
+    return predict_spec(model, features)
+  if mode == tf.estimator.ModeKeys.TRAIN:
+    return train_spec(model, features, labels)
+  if mode == tf.estimator.ModeKeys.EVAL:
+    return eval_spec(model, features, labels)
```
martinwicke: It makes me a little bit sad that this is split up into three separate functions. Do you need the separate functions? 

Would it make sense to simply always fill out all fields? Or did you try to show what is needed in which case more explicitly?

Even if you want to show it explicitly, I would probably prefer to inline these, but I haven't seen the resulting function, maybe it is too big.

official/mnist/mnist.py
```diff
@@ -233,10 +225,11 @@ def eval_input_fn():
 
   # Export the model
   if FLAGS.export_dir is not None:
-    image = tf.placeholder(tf.float32,[None, 28, 28])
-    serving_input_fn = tf.estimator.export.build_raw_serving_input_receiver_fn(
-        {"image_raw":image})
-    mnist_classifier.export_savedmodel(FLAGS.export_dir, serving_input_fn)
+    image = tf.placeholder(tf.float32, [None, 28, 28])
```
martinwicke: Why is image here and another placeholder is created in the dict? I think this one is left over from an earlier life.

official/mnist/README.md
```diff
@@ -21,6 +21,13 @@ python mnist.py
...
 validation data.
 
+Illustrative unittests and benchmarks can be run with:
+
+```
+python mnist_test.py
+python mnist_test.py --benchmarks=.
+```
+
 ## Exporting the model
```
k-w-w: Can you change `image_raw` to `image` in the snippet below?

official/mnist/mnist.py
asimshankar: The model produces logits, the (hopefully generic) executor will need a loss function and optimizer that are independent of the "forward pass" of the model. Hence, stopping here.

official/mnist/mnist.py
asimshankar: As a newbie to using estimators, I found it helpful to understand what is needed when. So I would like to avoid filling out all the fields. That way one doesn't have to think about loss functions to get a predictor up and running fast and other such things.

Separate functions or inlining - my personal preference was to have them separate, but I don't have a strong argument either way. Will inline.

official/mnist/mnist.py

official/mnist/README.md
asimshankar: Done

official/mnist/mnist.py
(Though, after doing it I feel slightly more strongly that the separate functions made it more readable, but I realize that I'm not the authority on how other people should read code :)

martinwicke: Another alternative is to always fill in all fields. This is what we're recommending, and since this is graph mode, it is harmless (that may change if Estimators ever support Eager).

If you wanted to make clear what is needed in which mode, that's not a great option of course, but it makes it more concise, and the cost in the graph is trivial.

asimshankar: I agree the cost in the graph is trivial.
However, I feel it's conceptually cleaner to not have to fill-in all fields. If I'm interested only in predictions, would be nice to not have to read through code about optimizers and losses. If I'm interested only in training, would be nice to not have to read through code around metrics.

Filling all fields won't make it much more concise either - the 2-3 duplicated lines are worth it, IMHO.

Let me know if you strongly disagree :)

official/mnist/mnist.py
```diff
@@ -75,141 +76,115 @@ def eval_dataset(data_dir):
...
+class Model(object):
+  """Class that defines a graph to recognize digits in the MNIST dataset."""
+
+  def __init__(self, data_format):
+    """Creates a model for classifying a hand-written digit.
+
+    Args:
+      data_format: Either 'channels_first' or 'channels_last'.
+        'channels_first' is typically faster on GPUs while 'channels_last' is
+        typically faster on CPUs. See
```
nealwu: I think the main reason for running channels_last on CPU is that channels_first doesn't actually work.

official/mnist/mnist.py
```diff
@@ -75,141 +76,115 @@ def eval_dataset(data_dir):
...
+      assert data_format == 'channels_last'
+      self._input_shape = [-1, 28, 28, 1]
+
+    self.conv1 = tf.layers.Conv2D(
+        32, 5, padding='same', data_format=data_format, activation=tf.nn.relu)
+    self.conv2 = tf.layers.Conv2D(
+        64, 5, padding='same', data_format=data_format, activation=tf.nn.relu)
+    self.fc1 = tf.layers.Dense(1024, activation=tf.nn.relu)
+    self.fc2 = tf.layers.Dense(10)
+    self.dropout = tf.layers.Dropout(0.5)
```
nealwu: I'm noticing this was 0.4 before, though as long as the accuracy is similar to before it doesn't make a big difference.

official/mnist/mnist.py
```diff
@@ -75,141 +76,115 @@ def eval_dataset(data_dir):
...
+
+    Returns:
+      A logits Tensor with shape [<batch_size>, 10].
+    """
+    y = tf.reshape(inputs, self._input_shape)
+    y = self.conv1(y)
+    y = self.max_pool2d(y)
+    y = self.conv2(y)
+    y = self.max_pool2d(y)
+    y = tf.layers.flatten(y)
```
nealwu: Why not use [`tf.layers.Flatten`](https://www.tensorflow.org/api_docs/python/tf/layers/Flatten)?

official/mnist/mnist_test.py
```diff
@@ -18,52 +18,76 @@
...
+  image = tf.random_uniform([BATCH_SIZE, 784])
+  labels = tf.random_uniform([BATCH_SIZE], maxval=9, dtype=tf.int32)
+  return image, tf.one_hot(labels, 10)
 
-  def input_fn(self):
-    features = tf.random_uniform([55000, 784])
-    labels = tf.random_uniform([55000], maxval=9, dtype=tf.int32)
-    return features, tf.one_hot(labels, 10)
 
-  def mnist_model_fn_helper(self, mode):
```
nealwu: What was the reason for deleting this test?

asimshankar: My thinking was that this was testing an implementation detail, not testing the interface expectations.

For the end-user of the `mnist.py` model, what we want from `model_fn` is to create an `Estimator` that can successfully train, predict, and evaluate. And that the predictions produce a scalar class and a vector of probabilities. And that the evaluation produces ...

The added tests to validate those end expectations. So I wasn't clear on what testing the `model_fn` directly is buying.

Thoughts?

official/mnist/mnist.py
martinwicke: In order to reuse this for Keras, we want to use layers.Flatten, even though it is true, it's not a Layer if the difference is state.

official/mnist/mnist_test.py
martinwicke: model_fn is easier to test. 

However, in a second step I think it would be nice to make this part of the code generic (independent of the model_fn), and actually test it with a mocked out model_fn.

official/mnist/mnist.py
asimshankar: As discussed offline, leaving as is for now. We'll have to figure out what it means to reuse this for Keras (and how we may want to handle direct usage of TensorFlow ops in that case)

official/mnist/mnist_test.py
asimshankar: `model_fn` is insufficient to test no? For example, testing the output of `model_fn` won't result in executing the graph, so we won't catch any errors thrown at runtime because say the graph was invalid.

As discussed offline, it seems we want to keep the testing proposed in this PR?

martinwicke: sgtm

official/mnist/mnist.py
martinwicke: yeah, we can deal with this once we get to Keras execution, likely after we have Network.

official/mnist/README.md
```diff
@@ -21,12 +21,19 @@ python mnist.py
 The model will begin training and will automatically evaluate itself on the
 validation data.
 
+Illustrative unittests and benchmarks can be run with:
```
nealwu: nit: unittests -> unit tests

official/mnist/mnist.py
nealwu: I don't feel too strongly about this; just don't want a user who tries to run NCHW on CPU to get confused.

I gave it a try and it gives a reasonable error, so it should be fine:

```
InvalidArgumentError (see above for traceback): Default MaxPoolingOp only supports NHWC.
	 [[Node: max_pooling2d/MaxPool = MaxPool[T=DT_FLOAT, data_format="NCHW", ksize=[1, 1, 2, 2], padding="VALID", strides=[1, 1, 2, 2], _device="/job:localhost/replica:0/task:0/device:CPU:0"](conv2d/Relu)]]
```

(doesn't mention CPU vs. GPU distinction though which is mildly disappointing)

official/mnist/mnist_test.py
nealwu: I agree that adding the new test is good, but I'm not sure there's anything wrong with this one. I believe everything it checks for are things that the model_fn should abide by. Is there an assert in here that you think is bad?

official/mnist/README.md
asimshankar: Done

official/mnist/mnist_test.py
asimshankar: Restored the tests, PTAL

PR #3029, opened 2017-12-20
compatible with tensorflow 1.4.1

tidyjiang8:
Make sure it's compatible with tensorflow latest version 1.4.1.


derekjchow:
LGTM, thank you for keeping this compatible for newer versions of TF!

PR #3034, opened 2017-12-21
[mnist]: Make flags function scoped (not module scoped).

asimshankar:
This will make it easier to share the model definition with
eager execution and TPU demos without any side effects of running
unnecessary code on module import.

### DIFF
--- a/official/mnist/mnist.py
+++ b/official/mnist/mnist.py
@@ -24,45 +24,6 @@ import sys
 import tensorflow as tf
 from tensorflow.examples.tutorials.mnist import input_data
 
-parser = argparse.ArgumentParser()
-
-# Basic model parameters.
-parser.add_argument(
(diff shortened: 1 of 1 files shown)


nealwu:
It looks like the tests weren't triggered here. I've noticed this happening on my PRs occasionally as well.

@av8ramit do we know what could cause this?

karmel:
I'm late to this party (sorry), but many of these options are things that will be shared across all of the models, correct? Is it feasible to create a base parser class with all the expected/default options, and thereby standardize at least some parts of the runnable API for each of these models? That will make automated testing much easier.

asimshankar:
Let's discuss offline - but, yeah, factoring common options out may be a reasonable follow up. Let's perhaps do that only after we're in a position to have more than say 3 uses of the common base parser class?

PR #3061, opened 2017-12-26
add log info about preprocess data augmentation

plieningerweb:
it is easy to make mistakes in protobuf config and therefore mistakenly think, that augmentation was used but acutally was never applied to training. Logging will prevent this mistake more easily. 

Root cause:
In protobuf config, data_augmentation_options can only contain one option per item, but if more, only last one silently applied, all others are ignored.

### DIFF
--- a/research/object_detection/trainer.py
+++ b/research/object_detection/trainer.py
@@ -70,11 +70,14 @@ def create_input_queue(batch_size_per_clone, create_tensor_dict_fn,
   include_keypoints = (fields.InputDataFields.groundtruth_keypoints
                        in tensor_dict)
   if data_augmentation_options:
+    tf.logging.info("used preprocess augmentation options are: %s", data_augmentation_options)
     tensor_dict = preprocessor.preprocess(
         tensor_dict, data_augmentation_options,
         func_arg_map=preprocessor.get_default_func_arg_map(
             include_instance_masks=include_instance_masks,
             include_keypoints=include_keypoints))
+  else:
+    tf.logging.info("no preprocess augmentation options used")
 
   input_queue = batcher.BatchQueue(
       tensor_dict,


plieningerweb:
I signed it!

plieningerweb:
train.config
```
...
  data_augmentation_options {
    random_horizontal_flip {
    }
  }
  data_augmentation_options {
    ssd_random_crop {
    }
  }
  data_augmentation_options {
    random_adjust_brightness {
      max_delta: 0.5
    }
    random_adjust_saturation {
      min_delta: 0.4
      max_delta: 1.6
    }
    random_adjust_hue{
      max_delta: 0.4
    }
    random_adjust_contrast{
      min_delta: 0.4
      max_delta: 1.6
    }
  }
...
```

Log will look like:
```
...
preprocess augmentation options are: [(<function random_horizontal_flip at 0x7fa976e1e6a8>, {'keypoint_flip_permutation': ()}), (<function ssd_random_crop
 at 0x7fa976e20730>, {}), (<function random_adjust_contrast at 0x7fa976e1ea60>, {'max_delta': 1.600000023841858, 'min_delta': 0.4000000059604645})]
...
```

plieningerweb:
Maybe we should fix it and raise an error instead or preprocess better, e.g.

options are built here:
https://github.com/vertexai/models/blob/fe257a40d94b5c0d0234280a90abfd68f754eaec/research/object_detection/trainer.py#L194

Using this step type:
https://github.com/vertexai/models/blob/fe257a40d94b5c0d0234280a90abfd68f754eaec/research/object_detection/builders/preprocessor_builder.py#L116

so instead it should be something like
```
step_type = preprocessor_step_config.WhichOneof('preprocessing_step')

# from https://developers.google.com/protocol-buffers/docs/reference/python/google.protobuf.message.Message-class
nonempty_fields = preprocessor_step_config.ListFields()

if len(nonempty_fields) != 1:
raise Exception('only one preprocessing step per "data_augmentation_options" can be specified. To use multiple, repeat "data_augmentation_options" with different preprocessing steps')
```

PR #3062, opened 2017-12-26
updated path to im2txt directory in readme

DLinNYC:
The current readme refers to a directory tensorflow-models which is not in the current directory structure


DLinNYC:
fixing typo in later pull request

PR #3070, opened 2017-12-27
updated eval for python 3

DecentGradient:
Because its nearly 2018 :)


DLinNYC:
I figure that supporting python 3 is worth the performance hit in an edge case, also happy 2018

PR #3082, opened 2017-12-29
a tetris ai

yaoGreat:
Tetris AI trained using the deepQ algorithm.
Contains model code and training code.


yaoGreat:
I am a beginner in machine learning, this is one of my practice procedures, hoping to get guidance. At the same time it may be used as a reference for other beginners :)

PR #3088, opened 2018-01-01
fix syntax

BL4CKvGHOST:


cclauss:
r-strings are valid in both Python 2 and Python 3.
* https://docs.python.org/2/reference/lexical_analysis.html#string-literals

PR #3093, opened 2018-01-02
[mnist]: Use FixedLengthRecordDataset

### LINKED ISSUE #3017: Official MNIST example should avoid huge constants
Right now MNIST model loads dataset as a tf.constant
https://github.com/tensorflow/models/blob/edb6ed22a801665946c63d650ab9a0b23d98e1b1/official/mnist/mnist.py#L65

This makes graphdef > 1 GB in size. That causes slowness when trying to visualize graph/dump graphdef. Better to load dataset into tf.Variable

### LINKED ISSUE #3046: the mnist under official model has OOM issue
the training process is OK, output message as following:
INFO:tensorflow:train_accuracy = 0.990091 (1.756 sec)
INFO:tensorflow:step = 21901, loss = 0.000445204 (1.756 sec)
INFO:tensorflow:Saving checkpoints for 22000 into /tmp/mnist_model/model.ckpt.
but evaluating process goes wrong, output message as following:
ResourceExhaustedError: OOM when allocating tensor with shape[10000,32,28,28]

btw, my GPU has only 2G memory

asimshankar:
- Prior to this change, the use of tf.data.Dataset essentially embedded
  the entire training/evaluation dataset into the graph as a constant,
  leading to unnecessarily humungous graphs (Fixes #3017)
- Also, use batching on the evaluation dataset to allow
  evaluation on GPUs that cannot fit the entire evaluation dataset in
  memory (Fixes #3046)

Using FixedLengthRecordDataset also provides an opportunity to use the same input pipeline code for the TPU demos (https://github.com/tensorflow/tpu-demos/tree/42a987e/cloud_tpu/models/mnist) without having to convert the raw data to TFRecords.

### DIFF
--- a/official/mnist/dataset.py
+++ b/official/mnist/dataset.py
@@ -57,7 +57,7 @@ def check_labels_file_header(filename):
 
 
 def maybe_download(directory, filename):
-  """Download a file from the MNIST dataset, if it doesn't already exist."""
+  """Download (and unzip) a file from the MNIST dataset, if it doesn't already exist."""
   if not tf.gfile.Exists(directory):
     tf.gfile.MakeDirs(directory)
   filepath = os.path.join(directory, filename)
--- a/official/mnist/dataset.py
+++ b/official/mnist/dataset.py
@@ -21,6 +21,7 @@ import os
 import shutil
 import gzip
 import numpy as np
+from six.moves import urllib
 import tensorflow as tf
 
 
@@ -56,7 +57,7 @@ def check_labels_file_header(filename):
                                                                      f.name))
 
 
-def maybe_download(directory, filename):
+def download(directory, filename):
   """Download (and unzip) a file from the MNIST dataset, if it doesn't already exist."""
   if not tf.gfile.Exists(directory):
     tf.gfile.MakeDirs(directory)
@@ -65,19 +66,18 @@ def maybe_download(directory, filename):
     return filepath
   # CVDF mirror of http://yann.lecun.com/exdb/mnist/
   url = 'https://storage.googleapis.com/cvdf-datasets/mnist/' + filename + '.gz'
-  zipped_filename = filename + '.gz'
-  zipped_filepath = os.path.join(directory, zipped_filename)
-  tf.contrib.learn.datasets.base.maybe_download(zipped_filename, directory, url)
-  with gzip.open(os.path.join(zipped_filepath), 'rb') as f_in, open(
-      filepath, 'wb') as f_out:
+  zipped_filepath = filepath + '.gz'
+  print('Downloading %s to %s' % (url, zipped_filepath))
+  urllib.request.urlretrieve(url, zipped_filepath)
+  with gzip.open(zipped_filepath, 'rb') as f_in, open(filepath, 'wb') as f_out:
     shutil.copyfileobj(f_in, f_out)
   os.remove(zipped_filepath)
   return filepath
 
 
 def dataset(directory, images_file, labels_file):
-  images_file = maybe_download(directory, images_file)
-  labels_file = maybe_download(directory, labels_file)
+  images_file = download(directory, images_file)
+  labels_file = download(directory, labels_file)
 
   check_image_file_header(images_file)
   check_labels_file_header(labels_file)
--- a/official/mnist/dataset.py
+++ b/official/mnist/dataset.py
@@ -1,0 +1,112 @@
+#  Copyright 2018 The TensorFlow Authors. All Rights Reserved.
+#
+#  Licensed under the Apache License, Version 2.0 (the "License");
+#  you may not use this file except in compliance with the License.
+#  You may obtain a copy of the License at
+#
+#   http://www.apache.org/licenses/LICENSE-2.0
+#
+#  Unless required by applicable law or agreed to in writing, software
+#  distributed under the License is distributed on an "AS IS" BASIS,
+#  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+#  See the License for the specific language governing permissions and
+#  limitations under the License.
+"""tf.data.Dataset interface to the MNIST dataset."""
+
+from __future__ import absolute_import
+from __future__ import division
+from __future__ import print_function
+
+import os
+import shutil
+import gzip
+import numpy as np
+import tensorflow as tf
+
+
+def read32(bytestream):
+  """Read 4 bytes from bytestream as an unsigned 32-bit integer."""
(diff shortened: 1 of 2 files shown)


official/mnist/dataset.py
```diff
@@ -0,0 +1,112 @@
...
+  with open(filename) as f:
+    magic = read32(f)
+    num_items = read32(f)
+    if magic != 2049:
+      raise ValueError('Invalid magic number %d in MNIST file %s' % (magic,
+                                                                     f.name))
+
+
+def maybe_download(directory, filename):
+  """Download a file from the MNIST dataset, if it doesn't already exist."""
```
mrry: Perhaps mention that this gunzips the file as well?

official/mnist/mnist.py
```diff
@@ -151,10 +139,10 @@ def train_input_fn():
     # When choosing shuffle buffer sizes, larger sizes result in better
     # randomness, while smaller sizes use less memory. MNIST is a small
     # enough dataset that we can easily shuffle the full epoch.
-    dataset = train_dataset(FLAGS.data_dir)
-    dataset = dataset.shuffle(buffer_size=50000).batch(FLAGS.batch_size).repeat(
+    ds = dataset.train(FLAGS.data_dir)
+    ds = ds.cache().shuffle(buffer_size=50000).batch(FLAGS.batch_size).repeat(
         FLAGS.train_epochs)
-    (images, labels) = dataset.make_one_shot_iterator().get_next()
+    (images, labels) = ds.make_one_shot_iterator().get_next()
```
mrry: While we're in here, would it make sense to switch to the new style of returning a `Dataset` directly? (Or perhaps, since 1.5 hasn't landed yet, we should have a TODO to make that switch?)

(Same applies to `eval_input_fn()` below.)

official/mnist/dataset.py
asimshankar: Done

official/mnist/mnist.py
asimshankar: Yup, was waiting for 1.5 to land.
Tempted to avoid TODOs in these "best practices" samples, unless you feel strongly.

k-w-w:
@mrry Does `Dataset.cache()` cache all the examples into memory (which would defeat the purpose of trying to use less memory)?

asimshankar:
@k-w-w - It caches them in CPU memory, not GPU memory. (That said, we could remove the use of `cache()` here).

k-w-w:
@asimshankar ohh, I see. Thanks for the response!

mrry:
Even with caching it will tend to use less memory, because the payloads of `tf.constant()` ops end up being stored multiple times in RAM (something that would be good to fix independently...). While you *could* disable caching, that moves the reading and parsing onto the critical path, and there's a good chance it would make the training process I/O-bound.

official/mnist/mnist.py
mrry: Fair enough!

k-w-w:
@mrry good to know, thanks! (looking forward to the dataset performance guide to learn more)

mrry:
@k-w-w On that topic, you can get a sneak peek at the guide here:

https://github.com/tensorflow/tensorflow/blob/master/tensorflow/docs_src/performance/datasets_performance.md

It should be on tensorflow.org once the 1.5 release is published.

official/mnist/dataset.py
```diff
@@ -0,0 +1,112 @@
+#  Copyright 2018 The TensorFlow Authors. All Rights Reserved.
```
nealwu: 2018! Nice :)

official/mnist/dataset.py
```diff
@@ -0,0 +1,112 @@
...
+  """Validate that filename corresponds to labels for the MNIST dataset."""
+  with open(filename) as f:
+    magic = read32(f)
+    num_items = read32(f)
+    if magic != 2049:
+      raise ValueError('Invalid magic number %d in MNIST file %s' % (magic,
+                                                                     f.name))
+
+
+def maybe_download(directory, filename):
```
nealwu: nit: I don't like the name `maybe_download` since the 'maybe' part seems very ambiguous. Let's call this `attempt_download` or just `download` instead?

official/mnist/dataset.py
```diff
@@ -0,0 +1,112 @@
...
+  if not tf.gfile.Exists(directory):
+    tf.gfile.MakeDirs(directory)
+  filepath = os.path.join(directory, filename)
+  if tf.gfile.Exists(filepath):
+    return filepath
+  # CVDF mirror of http://yann.lecun.com/exdb/mnist/
+  url = 'https://storage.googleapis.com/cvdf-datasets/mnist/' + filename + '.gz'
+  zipped_filename = filename + '.gz'
+  zipped_filepath = os.path.join(directory, zipped_filename)
+  tf.contrib.learn.datasets.base.maybe_download(zipped_filename, directory, url)
```
nealwu: Is there a way we can do this without contrib?

official/mnist/mnist.py
```diff
@@ -165,7 +153,8 @@ def train_input_fn():
 
   # Evaluate the model and print results
   def eval_input_fn():
-    return eval_dataset(FLAGS.data_dir).make_one_shot_iterator().get_next()
+    return dataset.test(FLAGS.data_dir).batch(
+        FLAGS.batch_size).make_one_shot_iterator().get_next()
```
nealwu: nit: I don't think this is the right code style. Can we split this onto two lines instead?

official/mnist/dataset.py
asimshankar: Renamed to `download`.

official/mnist/dataset.py
asimshankar: Changed to use `urllib.request.urlretrieve`. This means that the retry logic implemented in
https://github.com/tensorflow/tensorflow/blob/master/tensorflow/contrib/learn/python/learn/datasets/base.py#L189 no longer applies, but I suspect that this retry business is less relevant now that we're using a CVDF mirror.

official/mnist/mnist.py
asimshankar: This was the result of running the Python formatter (https://github.com/google/yapf), so it should be right? :)

nealwu: That's surprising; the part that seemed weird to me was indenting the arguments to the next line and then following up with more function calls. If the formatter says it's good though, should be fine.

official/mnist/dataset.py
nealwu: Sounds good!

PR #3113, opened 2018-01-06
[mnist]: Training and evaluation with TPUs.

asimshankar:
This is a step towards merging the example in
https://github.com/tensorflow/tpu-demos/tree/master/cloud_tpu/models/mnist
with this repository, so we have a single model definition for
training across CPU/GPU/eager execution/TPU.

The change to dataset.py is so that the raw data can be read
from cloud storage systems (like GCS and S3).

### DIFF
--- a/official/mnist/mnist_tpu.py
+++ b/official/mnist/mnist_tpu.py
@@ -1,0 +1,140 @@
+# Copyright 2017 The TensorFlow Authors. All Rights Reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+# ==============================================================================
+"""MNIST model training using TPUs.
+
+This program demonstrates training of the convolutional neural network model
+defined in mnist.py on Google Cloud TPUs (https://cloud.google.com/tpu/).
+
+If you are not interested in TPUs, you should ignore this file.
+"""
+from __future__ import absolute_import
+from __future__ import division
+from __future__ import print_function
+
+import tensorflow as tf
+import dataset
+import mnist
+
+tf.flags.DEFINE_string("data_dir", "",
+                       "Path to directory containing the MNIST dataset")
+tf.flags.DEFINE_string("model_dir", None, "Estimator model_dir")
+tf.flags.DEFINE_integer("batch_size", 128,
+                        "Mini-batch size for the training. Note that this "
+                        "is the global batch size and not the per-shard batch.")
+tf.flags.DEFINE_integer("train_steps", 1000, "Total number of training steps.")
+tf.flags.DEFINE_integer("eval_steps", 0,
+                        "Total number of evaluation steps. If `0`, evaluation "
+                        "after training is skipped.")
+tf.flags.DEFINE_float("learning_rate", 0.05, "Learning rate.")
+
+tf.flags.DEFINE_bool("use_tpu", True, "Use TPUs rather than plain CPUs")
+tf.flags.DEFINE_string("master", "local", "GRPC URL of the Cloud TPU instance.")
+tf.flags.DEFINE_integer("iterations", 50,
+                        "Number of iterations per TPU training loop.")
+tf.flags.DEFINE_integer("num_shards", 8, "Number of shards (TPU chips).")
+
+FLAGS = tf.flags.FLAGS
+
+
+def metric_fn(labels, logits):
+  accuracy = tf.metrics.accuracy(
+      labels=tf.argmax(labels, axis=1), predictions=tf.argmax(logits, axis=1))
+  return {"accuracy": accuracy}
+
+
+def model_fn(features, labels, mode, params):
+  del params
+  if mode == tf.estimator.ModeKeys.PREDICT:
+    raise RuntimeError("mode {} is not supported yet".format(mode))
+  image = features
+  if isinstance(image, dict):
+    image = features["image"]
+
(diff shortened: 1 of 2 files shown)


official/mnist/mnist_tpu.py
```diff
@@ -0,0 +1,140 @@
...
+      eval_batch_size=FLAGS.batch_size,
+      params={"data_dir": FLAGS.data_dir},
+      config=run_config)
+  # TPUEstimator.train *requires* a max_steps argument.
+  estimator.train(input_fn=train_input_fn, max_steps=FLAGS.train_steps)
+  # TPUEstimator.evaluate *requires* a steps argument.
+  # Note that the number of examples used during evaluation is
+  # --eval_steps * --batch_size.
+  # So if you change --batch_size then change --eval_steps too.
+  estimator.evaluate(input_fn=eval_input_fn, steps=FLAGS.eval_steps)
```
k-w-w: Should the evaluation results be printed?

official/mnist/mnist_tpu.py
```diff
@@ -0,0 +1,140 @@
...
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+# ==============================================================================
+"""MNIST model training using TPUs.
+
+This program demonstrates training of the convolutional neural network model
+defined in mnist.py on Google Cloud TPUs (https://cloud.google.com/tpu/).
+
+If you are not interested in TPUs, you should ignore this file.
```
k-w-w: Perhaps add instructions of how to run this file? Judging from the [instructions of the wide deep model](https://cloud.google.com/ml-engine/docs/samples), it might be too lengthy though.

k-w-w: (oops, correct link here: https://cloud.google.com/ml-engine/docs/getting-started-training-prediction)

official/mnist/mnist_tpu.py
```diff
@@ -0,0 +1,140 @@
...
+                        "Total number of evaluation steps. If `0`, evaluation "
+                        "after training is skipped.")
+tf.flags.DEFINE_float("learning_rate", 0.05, "Learning rate.")
+
+tf.flags.DEFINE_bool("use_tpu", True, "Use TPUs rather than plain CPUs")
+tf.flags.DEFINE_string("master", "local", "GRPC URL of the Cloud TPU instance.")
+tf.flags.DEFINE_integer("iterations", 50,
+                        "Number of iterations per TPU training loop.")
+tf.flags.DEFINE_integer("num_shards", 8, "Number of shards (TPU chips).")
+
```
karmel: Do you anticipate having these separately defined not just for each model, but for each API as well? Or is this a temporary measure until the rest is in place?

official/mnist/mnist_tpu.py
```diff
@@ -0,0 +1,140 @@
...
+  loss = tf.losses.softmax_cross_entropy(onehot_labels=labels, logits=logits)
+
+  if mode == tf.estimator.ModeKeys.TRAIN:
+    learning_rate = tf.train.exponential_decay(
+        FLAGS.learning_rate,
+        tf.train.get_global_step(),
+        decay_steps=100000,
+        decay_rate=0.96)
+    optimizer = tf.train.GradientDescentOptimizer(learning_rate=learning_rate)
+    if FLAGS.use_tpu:
```
karmel: What does it mean if this is False? Is that a valid possibility we should consider here, or does that mean the user has done something wrong?

official/mnist/mnist_tpu.py
```diff
@@ -0,0 +1,140 @@
...
+        mode=mode,
+        loss=loss,
+        train_op=optimizer.minimize(loss, tf.train.get_global_step()))
+
+  if mode == tf.estimator.ModeKeys.EVAL:
+    return tf.contrib.tpu.TPUEstimatorSpec(
+        mode=mode, loss=loss, eval_metrics=(metric_fn, [labels, logits]))
+
+
+def train_input_fn(params):
```
karmel: It occurs to me that if we want people to use these models, and in particular use them as starting points, really good docstrings explaining/instructing will be critical. That doesn't have to happen here and now, but perhaps we can have someone do a second pass to comment liberally after things are stable...

official/mnist/mnist_tpu.py
karmel: And returned in some way? We will want to be able to collect output and stats, but ideally it's not the responsibility of the train/eval functions to figure out what to do with them or where to put them.

official/mnist/mnist_tpu.py
```diff
@@ -0,0 +1,140 @@
...
+from __future__ import print_function
+
+import tensorflow as tf
+import dataset
+import mnist
+
+tf.flags.DEFINE_string("data_dir", "",
+                       "Path to directory containing the MNIST dataset")
+tf.flags.DEFINE_string("model_dir", None, "Estimator model_dir")
+tf.flags.DEFINE_integer("batch_size", 128,
```
jhseu: nit: default to 1024 so it's 128 per TPU core?

official/mnist/mnist_tpu.py
jhseu: We allow disabling TPUs in TPUEstimator for testing between TPU and GPU, so it's valid.

official/mnist/mnist_tpu.py
asimshankar: Sounds reasonable.`input_fn` is effectively implementing an interface required by `Estimator.train` or `Estimator.evaluate`.

The TPU code is admittedly not as polished yet, let's do this in a follow up pass.

official/mnist/mnist_tpu.py
asimshankar: Temporary measure till we have a few models and establish a pattern.

official/mnist/mnist_tpu.py
asimshankar: @jhseu can comment - but for now I believe this just ends up being output by the call to `Estimator.evaluate` in stdout.

official/mnist/mnist_tpu.py
asimshankar: Will update when CloudTPUs are out of Alpha and are more accessible.

PR #3119, opened 2018-01-07
adds second_stage_batch_size to faster_rcnn_nas_coco.config

### LINKED ISSUE #2668: ValueError: second_stage_batch_size should be no greater than first_stage_max_proposals.
- **What is the top-level directory of the model you are using**:
object_detection
- **Have I written custom code (as opposed to using a stock example script provided in TensorFlow)**:
No
- **OS Platform and Distribution (e.g., Linux Ubuntu 16.04)**:
Ubuntu 16.04
- **TensorFlow installed from (source or binary)**:
Binary
- **TensorFlow version (use command below)**:
('v1.3.0-rc2-20-g0787eee', '1.3.0')
- **Bazel version (if compiling from source)**:
- **CUDA/cuDNN version**:
8/6
- **GPU model and memory**:
1070 8105MiB Driver 384.90
- **Exact command to reproduce**:
train.py

Hi,

Using the default config provided in the samples dir I get this error:

ValueError: second_stage_batch_size should be no greater than first_stage_max_proposals.

Edit: Using the new faster_rcnn_nas pre trained model.

Thanks

Matt

sangeet259:
The parameter `second_stage_batch_size` is required in configuration files but is missing.
A Value error is being raised as a result of which.
Fixes #2668
Thanks to @mattryles for the solution.

### DIFF
--- a/research/object_detection/samples/configs/faster_rcnn_nas_coco.config
+++ b/research/object_detection/samples/configs/faster_rcnn_nas_coco.config
@@ -46,6 +46,7 @@ model {
     first_stage_max_proposals: 50
     first_stage_localization_loss_weight: 2.0
     first_stage_objectness_loss_weight: 1.0
+    second_stage_batch_size: 49
     initial_crop_size: 17
     maxpool_kernel_size: 1
     maxpool_stride: 1


sangeet259:
@googlebot I signed it!

PR #3120, opened 2018-01-07
Drop a dummy variable

grzegorznowak:
Let's not create a `raw_data` variable just for sake of creating an extra variable.


grzegorznowak:
`I signed it!`

PR #3140, opened 2018-01-10
Fix bug in gym_wrapper.

ashaw596:
Added an 'env' parameter to env_step function since one an unreferenced one is used in the function.


ashaw596:
I signed it!

PR #3154, opened 2018-01-12
Add RNN PTB to official models

k-w-w:
Hi Asim, can you review the PTB model? I followed the object-oriented style, tests and benchmarks from the MNIST model. 

Thanks!

### DIFF
--- a/official/rnn_ptb/README.md
+++ b/official/rnn_ptb/README.md
@@ -1,0 +1,126 @@
+# PTB Language Model
+This example code trains a [language model](https://en.wikipedia.org/wiki/Language_model) on the [Penn TreeBank](https://catalog.ldc.upenn.edu/ldc99t42) dataset ([available for download here](http://www.fit.vutbr.cz/~imikolov/rnnlm/simple-examples.tgz)).
+
+The model is implemented based on [*Recurrent Neural Network Regularization* (Zaremba et. al, 2014)](https://arxiv.org/abs/1409.2329). It incorporates 2 layers of LSTM cells with dropout applied as described in the paper.
+
+## Running the model
+
+1. **Download and extract the PTB dataset: http://www.fit.vutbr.cz/~imikolov/rnnlm/simple-examples.tgz**
+
+   The training, validation, and testing files are located in the `data` folder.
+
+2. **Train the model with [`run_training.py`](run_training.py):**
+   ```
+   python run_training.py --data_dir={DATA PATH}
+   ```
+   Required arguments:
+   * `--data_dir`: path to the training/validation/testing files. (e.g. `.../Downloads/simple-examples/data`)
+
+   Other arguments:
+   * `--model`: The model hyperparameters/configuration to use.
+         Options are: small, `medium` (default), and `large`.
+   * `--model_dir`: directory where the model will be saved. The model will be saved to `/tmp/ptb_model` by default.
+   * `--reset_training`: Whether to clear the model directory before training (default `false`).
+
+   **Visualizing training progress with TensorBoard**
+
+   At any point during/after the training, you can see the current training and validation perplexity using TensorBoard:
+   ```
+   tensorboard --logdir={MODEL DIR}
+   ```
+3. **Generate text with [`run_predict.py`](run_predict.py)**:
+   ```
+   python run_predict.py --data_dir={DATA PATH}
+   ```
+
+   The arguments `--data_dir` (required), `--model`, and `--model_dir` should be set to the same values as in the previous step.
+
+   Prediction arguments:
+   * `--input`: Starting string that will be used to warm up the state of the RNN.
+   * `--num_predictions`: Number of words to generate.
+
+### Tests and benchmarks
+[`ptb_test.py`](ptb_test.py) defines tests and benchmarks for the model.
+* The test class `TestModel` ensures that the outputs from evaluation and prediction have the correct shapes. In addition, it trains the model for 20 steps on dummy data to test that the loss decreases as expected.
+* The `Benchmarks` class times the average number of examples that are trained per second, averaged over 100 or 1000 iterations (100 iterations for cpu, 1000 for gpu).
+
+Use the commands below to run just the tests, or both tests and benchmarks:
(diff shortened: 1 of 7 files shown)


official/rnn_ptb/run_training.py
```diff
@@ -0,0 +1,244 @@
...
+# number steps to train the model).
+TRAIN_FILE_WORDS = 929589
+
+
+def input_fn(input_file, vocab_dict, num_epochs=1):
+  """Generates batches of data from the input file."""
+  def _input_fn(params):
+    batch_size, unrolled_count = params.batch_size, params.unrolled_count
+
+    def batch_data(data, steps_per_epoch):
```
karmel: Nesting functions three levels deep seems to introduce unnecessary complexity; could just be my misunderstanding, but any reason not to extract these and have them take the args they need? Is there an advantage to having them defined in the _input_fn?

official/rnn_ptb/run_training.py
```diff
@@ -0,0 +1,244 @@
...
+        steps_per_epoch: The number of steps it take to complete an epoch.
+      """
+      data = data.reshape([batch_size, steps_per_epoch, unrolled_count])
+      return np.swapaxes(data, 0, 1)
+
+    def input_generator():
+      """Yield batches of the input data."""
+      # Read all words from the input file, and append <eos> at every new line.
+      with tf.gfile.GFile(input_file, 'r') as f:
+        data = f.read().replace('\n', '<eos>').split()
```
karmel: Do we anticipate users trying files/input with non-unix line endings? Might be worth at least extracting these to module-level vars (something like `INPUT_EOL`, `EOL_TOKEN` or what-have-you), just to make it easier for eventual users to edit. (I speak from the experience of having tried to adapt a similar model in the past...)

Edit: I see below that EOS is defined in `util`; should that be used here as well?

official/rnn_ptb/util.py
```diff
@@ -0,0 +1,86 @@
...
+def build_vocab_id_dict(data_file):
+  """Create a dictionary mapping all words in the vocabulary to integer IDs."""
+  assert tf.gfile.Exists(data_file), 'File %s does not exist.' % data_file
+  with tf.gfile.Open(data_file, 'r') as f:
+    unique_words = set(f.read().split())
+
+    # Remove special symbols from the set.
+    for symbol in SPECIAL_SYMBOLS:
+      if symbol in unique_words:
+        unique_words.remove(symbol)
```
karmel: nit: I think this loop is equivalent to `unique_words = unique_words - set(SPECIAL_SYMBOLS)`

official/rnn_ptb/util.py
```diff
@@ -0,0 +1,86 @@
...
+    # is mapped to 0 (corresponding to '<unk>').
+    vocab_dict = collections.defaultdict(
+        int, zip(unique_words, range(len(unique_words))))
+    return vocab_dict
+
+
+def build_reverse_vocab_dict(vocab_dict):
+  """Build a dictionary of id->word."""
+  reverse_dict = {i: word for word, i in vocab_dict.iteritems()}
+  reverse_dict[3] = '.\n'  # Replace '<eos>' with a period with a new line.
```
karmel: Why append a period here? Seems like this could lead to unexpected behavior if we necessarily add punctuation. Also, hard-coding `3` here might be fragile if someone adds/changes the SPECIAL_SYMBOLS, which seems like a reasonable thing to do. Maybe at least do a quick lookup of the index of EOS in SPECIAL_SYMBOLS?

official/rnn_ptb/README.md
```diff
@@ -0,0 +1,126 @@
...
+2. **Train the model with [`run_training.py`](run_training.py):**
+   ```
+   python run_training.py --data_dir={DATA PATH}
+   ```
+   Required arguments:
+   * `--data_dir`: path to the training/validation/testing files. (e.g. `.../Downloads/simple-examples/data`)
+
+   Other arguments:
+   * `--model`: The model hyperparameters/configuration to use.
+         Options are: small, `medium` (default), and `large`.
```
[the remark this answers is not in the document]
k-w-w: Thanks for catching this! 

What do you mean by example/format?

official/rnn_ptb/README.md
```diff
@@ -0,0 +1,126 @@
...
+   * `--model_dir`: directory where the model will be saved. The model will be saved to `/tmp/ptb_model` by default.
```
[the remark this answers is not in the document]
k-w-w: Done, I left `_model` part in (`/tmp/rnn_ptb_model`), because the other models have `_model` as part of the directory name.

official/rnn_ptb/README.md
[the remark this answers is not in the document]
karmel: Ah, wait, I think I misread initially and thought you could pass in arbitrary configs. Disregard!

official/rnn_ptb/run_training.py
k-w-w: I originally had them under `_input_fn` for convenience. I agree that 3 nested functions is a bit too much, so I've moved them out.

official/rnn_ptb/run_training.py
k-w-w: Good point, I added a constant for `\n`, and reuse util.EOS for `<eos>`.

official/rnn_ptb/util.py
k-w-w: Nice! I've made this change.

official/rnn_ptb/util.py
k-w-w: I've changed this to replace `<eos>` with the NEWLINE symbol, and added a lookup for the index of EOS. I agree it's better with less hardcoding.

official/rnn_ptb/README.md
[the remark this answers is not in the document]
k-w-w: Ok!

tfboyd:
This is about a year old at this point. Closing.  Ping me to reopen if desired.  Cleaning up all PRs targeting /official.

PR #3163, opened 2018-01-14
Yolov1

VastoLorde95:
This is a partially complete work of the YOLO object detection model. Since the YOLO model does not fit neatly into the meta architecture / feature extractor format, we have done our best to follow the same - the actual model, has been implemented as yolov1_feature_extractor.py and the preprocessing, postprocessing and loss functions have been implemented as yolov1_meta_arch.py. All of our implementation is based off the origin paper by Joseph Redmon.

We have not implemented the pipeline that allows to train this model hence, we are opening this pull request so that others who have better knowledge about the same can add on to our work and complete the same.


podhrmic:
@VastoLorde95 thanks for the PR, I was just wondering why YOLO is not available as one of the models. Do you expect the weights from the pretrained original model (here: https://pjreddie.com/darknet/yolo/ ) to be transferable to TF YOLO?

YijinLiu:
@VastoLorde95 Is this still your intention to submit the PR?

PR #3175, opened 2018-01-17
Add MNIST Multi GPU loop

karmel:
Allow MNIST to run on multiple GPUs.

* Adds a script that runs MNIST in multi-GPU mode using `replicate_model_fn`, currently in contrib. 
* To facilitate running with multiple GPUs, minimally refactors the existing MNIST script.

CC @petermattson , @isaprykin


official/mnist/mnist_multi_gpu.py
```diff
@@ -0,0 +1,80 @@
...
+from __future__ import print_function
+
+import sys
+
+import mnist
+import tensorflow as tf
+from tensorflow.python.client import device_lib
+
+
+def model_fn_with_tower_optimizer(features, labels, mode, params):
```
asimshankar: Instead of a separate driving file (and the resulting refactoring of `mnist.py`), would it make sense to merge this into `mnist.py`? Some ideas off the top of my head:

- Add a `experimental_use_multiple_gpus` flag to `mnist.py` (which defaults to `False`). With the default value, no `contrib` APIs will be invoked.
- When that flag value is `True`, then we wrap `model_fn` with `tf.contrib.estimator.replicated_model_fn`
- And we also translate that flag to a value in the `params` dictionary to `model_fn`, so that the `TowerOptimizer` wrapping only happens when that param is set.

My gut feeling is that this will reduce the amount of code overall and also make what changes between single and multi GPU to the reader - while avoiding the use of `contrib` APIs by default.

Thoughts?

karmel: Maybe better discussed offline, but I tend to prefer the separation. Pros:

* Separates out contrib imports without introducing logic/conditionals.
* This puts the responsibility on the main loop to be reusable by other loops, whereas one file would push us towards the world where each run flavor introduces a bunch of conditionals into the main loop.
* This puts the responsibility on the multi-gpu loop to not affect the default run.
* This makes it more obvious to users what is doing what. Someone can use mnist.py as an example for running on single GPU without copying over a bunch of code that is useless to them, and someone looking for an example of how to run multi-GPU will find it much easier to understand exactly the set of changes they will need to make in their own models.
* In the world where there are many cooks in the Model Garden kitchen, more modules is easier than one big module.

(If I have not convinced you, let's discuss offline to avoid the inevitable back and forth.)

karmel:
Discussed offline; closing for now and will re-open with a joined API for multi- and single-modes.

isaprykin:
Is the "a joined API for multi- and single-modes." the future core API for replicate_model_fn or some re-work of the current set of APIs?  Just curious, thank you.

asimshankar:
@isaprykin : The intent will be to add a flag `--experimental_use_multiple_gpus` flag. That flag will be used to gate the wrapping of the optimizer with `TowerOptimizer` and calling `replicated_model_fn`. 

So, preparing for the future core API of `replicate_model_fn` by gating ~10 lines of code with the flag :)

PR #3199, opened 2018-01-19
Add multi-GPU to MNIST, take two

karmel:
Adding multi-GPU functionality to MNIST, all in the main loop now. I extracted/refactored some functions in order to keep consistent with some refactoring that will be necessary with what I do with Resnet.


official/mnist/mnist.py
```diff
@@ -95,7 +107,7 @@ def model_fn(features, labels, mode, params):
             'classify': tf.estimator.export.PredictOutput(predictions)
         })
   if mode == tf.estimator.ModeKeys.TRAIN:
-    optimizer = tf.train.AdamOptimizer(learning_rate=1e-4)
+    optimizer = optimizer_fn(learning_rate=LEARNING_RATE)
```
asimshankar: Why the change to a named constant?
The value is used in precisely one place (here), so inlining it avoids indirection when reading it.

If we want to parameterize the learning rate, then perhaps we should do so with a flag and a value in the `params` dictionary. That way everything that is "configurable" is configurable via flags rather than some things by flags and some by named constants in the source file.

official/mnist/mnist.py
```diff
@@ -75,7 +78,16 @@ def __call__(self, inputs, training):
...
 
 
-def model_fn(features, labels, mode, params):
+################################################################################
+# Training loop for the model
+################################################################################
+def model_fn_with_optimizer(features, labels, mode, params):
+  """Wrapper for the model_fn that sets the optimizer for single GPU/CPU.
+  """
+  return model_fn(features, labels, mode, params, tf.train.AdamOptimizer)
```
asimshankar: I'm not sure about this `model_fn_with_optimizer` refactoring. For one, this isn't convenient when trying out different optimizers or configurations as it is effectively defining an unnamed narrow "interface" for optimizer factories. For example, if I want to try adam with a different value of [`beta1` or `beta2`](https://www.tensorflow.org/api_docs/python/tf/train/AdamOptimizer), then this becomes even more involved with something like:

```
return model_fn(features, labels, mode, params, lambda x: tf.train.AdamOptimizer(beta1=0.9, learning_rate=x))
```

Similar, if you want to use `MomentumOptimizer` and providing the value of momentum.

Keeping the single `model_fn` from before seems much easier to follow and change.
All that would change would be that:

```python
if mode == tf.estimator.ModeKeys.TRAIN:
  optimizer = tf.train.AdamOptimizer(learning_rate=1e-4)
  ...
```

becomes:

```python
if mode == tf.estimator.ModeKeys.TRAIN:
  optimizer = tf.train.AdamOptimizer(learning_rate=1e-4)
  if param.has_key('experimetal_multi_gpu'): # Or some other name that is appropriate
    optimizer = tf.contrib.estimator.TowerOptimizer(optimizer)
  ...
```

With the current formulation, comparing single and multi-GPU behaviors on different optimizers means you have to change the optimizer in two places and make sure they are in sync.

I think it will be valuable for this file to be as succinct as possible with as few abstractions as possible so that it is easy to identify a the few lines of code to change (e.g. optimizer configurations, data format etc.)

official/mnist/mnist.py
```diff
@@ -153,55 +165,122 @@ def train_input_fn():
...
+
+
+################################################################################
+# Run this script
+################################################################################
+def main(unused_argv):
+  model_function = model_fn_with_optimizer
+
+  if FLAGS.multi_gpu:
+    # Adjust batch size
```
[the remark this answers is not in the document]
asimshankar: Though: @isaprykin - is that true? _must_ batch size be evenly divisible by the number of GPUs?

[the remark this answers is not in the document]
isaprykin: As of now, yes, unfortunately.  replicate_model_fn could throw an exception if it's not the case, or devise a smarter scheme to do a partial run with only some towers...

official/mnist/mnist.py
```diff
@@ -22,8 +22,11 @@
 import sys
 
 import tensorflow as tf
+from tensorflow.python.client import device_lib
```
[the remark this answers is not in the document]
isaprykin: Could these public https://www.tensorflow.org/versions/master/api_docs/python/tf/contrib/eager/num_gpus   or   https://www.tensorflow.org/versions/master/api_docs/python/tf/Session#list_devices  serve as an alternative?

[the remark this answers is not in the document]
karmel: Ah, interesting, I did not realize that. It seems like reaching into Eager would break some implicit rules here as well, and reaching into the session requires that we have the session available at the time. I am perfectly fine with a function-internal import, especially since we anticipate removing this functionality when it is native to the replication itself.

official/mnist/mnist.py
```diff
@@ -153,55 +165,122 @@ def train_input_fn():
...
+
+  TODO(karmel): This should eventually be handled by replicate_model_fn
+  directly. For now, doing the work here.
+  """
+  remainder = current_size % _get_num_gpu()
+  return current_size - remainder
+
 
+def _get_num_gpu():
+  """TODO(karmel): This should eventually be handled by replicate_model_fn
```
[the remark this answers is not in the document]
karmel: Good point. I feel like we should have some explanation of why this is required here, but I can do that as a comment rather than a TODO, which will be cleaner/less concerning?

official/mnist/mnist.py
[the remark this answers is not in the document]
karmel: That one is a philosophical debate, and I come from the one-function-does-one-thing school of Python ;) More seriously, though, these are going to have to be used for Resnet, etc. as well, so this is a staging ground for extracting and sharing for all, rather than having that one liner spread across multiple files.

official/mnist/mnist.py
```diff
@@ -153,55 +165,122 @@ def train_input_fn():
...
+        model_fn_multi, loss_reduction=tf.losses.Reduction.MEAN)
+
+  main_with_model_fn(FLAGS, unused_argv, model_function)
+
+
+class MNISTArgParser(argparse.ArgumentParser):
+
+  def __init__(self):
+    super(MNISTArgParser, self).__init__()
+    self._set_args()
```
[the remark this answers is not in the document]
karmel: This one again is a staging ground for making an ArgParser with shared functionality across multiple files. I am fine with changing this to a function in theory-- feels like mostly a stylistic preference?

official/mnist/mnist.py
karmel: Another stylistic preference-- buried Magic Numbers can be hard to find and debug, and are prone to copy and paste errors, so I try to avoid them, but can swap back in this case.

official/mnist/mnist.py
karmel: I was going to wrap the grabbing of the AdamOptimizer in a shared function but I figured you would push back on that when there were only two uses :) This becomes useful, however, when looking at Resnet, where the getting of an optimizer is the same for cifar and imagenet, but the containing function is different. So rather than do it in two places there, it makes sense to extract. And then it seems wise to keep the behavior consistent across the models. This has the side effect of keeping the main processing simpler by at least one conditional, which I find useful from the perspective of a learner who wants to understand what is the core of what the main loop is actually doing (versus config details). More than happy to define a get_optimizer that returns the AdamOptimizer, which would then be the one place to pick and choose your optimizer.

official/mnist/mnist.py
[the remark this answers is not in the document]
asimshankar: Hmm...I don't quite follow the value of extracting something like:

```
x = (x // n) * n
``

into a library of shared functions.

In a bunch of these comments you're talking about resnet - so I think there is some mismatch between what the two of us think will be shared between this and resnet. Sorry about the confusion, but let's talk about this offline again.

File: official/mnist/mnist.py (hunk 2 above)
Comment (asimshankar) [reply official/mnist/mnist.py]: I'm confused about what this has to do with ResNet.  Is something about this going to be shared between the two?

File: official/mnist/mnist.py [hunk 7]
```diff
@@ -122,14 +131,14 @@ def model_fn(features, labels, mode, params):
         })
 
 
-def main(unused_argv):
-  data_format = FLAGS.data_format
+def main_with_model_fn(flags, unused_argv, model_function):
```
k-w-w: Is it ok to leave this as part of `main`, or is there a reason for separating this function? Currently, it is only called in `main`.

karmel:
Multi-GPU logic is now split in two and included directly in the model_fn, with the expectation that eventually that will get folded into Estimator or similar directly. CC @asimshankar @k-w-w for another look.

official/mnist/mnist.py
```diff
@@ -122,16 +127,45 @@ def model_fn(features, labels, mode, params):
...
+  remainder = current_size % num_gpus
+  return current_size - remainder
+
+
 def main(unused_argv):
+  model_function = model_fn
+
+  if FLAGS.multi_gpu:
+    # Adjust batch size
+    FLAGS.batch_size = get_batch_size_multi(FLAGS.batch_size)
```
asimshankar: Would it make sense to warn on changes (since it does mean that we may not completely respect the user's batch size).
Options:

1.  Make this:
    ```python
    batch_size = get_batch_size_multi(FLAGS.batch_size)
    if batch_size != FLAGS.batch_size:
      print(<something>)
      FLAGS.batch_size = batch_size
      ```

2. OR, maybe it's better to make this an assertion so that we always respect the user's `--batch_size` and fail when not? So instead of `get_batch_size_multi` we could have an `assert_batch_size_is_multiple_of_number_of_gpus` (I'm sure a better name is possible :)

The more I think about it, I think I prefer the latter (failing instead of silently changing). That way we never have to explain to someone why they don't see any change when they try `--batch_size=64` and `--batch_size=69` :)

official/mnist/mnist.py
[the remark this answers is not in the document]
asimshankar: Optional: Would it make sense to inline the added flags into `__init__`? (Not sure what the `_set_args()` method helps with)

official/mnist/mnist.py
karmel: I agree with warning-- is printing the preferred method in this case? As to halting-- I'm fine if that's generally preferred, but my preference would be that it just does the right thing by default if I don't specify. Having to fully specify everything when all you want to do is run/try certain things can be frustrating. For example-- as a user, I just want to run this as fast as possible on whatever set of GPUs I have available. If we error out and request a batch size, I then have to go figure out how to do that.

official/mnist/mnist.py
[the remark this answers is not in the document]
karmel: There was some reason in the original version; now it does indeed make sense to merge.

official/mnist/mnist.py
asimshankar: The exception message should be something helpful like: `Set --batch_size=128 or --batch_size=136 . --batch_size=129 does not evenly divide amongst the 8 GPUs`, telling the user exactly what they can do.

As for "does the right thing by default" - I agree. The question is whether the "right" thing is to "run however you can, possibly ignoring what the user asked for", or is the "right" thing to "run what the user specified".

karmel: I guess in the ideal we would adjust if the user didn't specify at all and error if they specified something wrong, but that's probably more fine-tuning of configuration than is called for here given that this will eventually be handled by replicate_model_fn anyhow. In which case, an informative error message sounds good to me.

karmel:
Should be all there now, with an error message like so: "ValueError: When running with multiple GPUs, batch size must be a multiple of the number of available GPUs. Found 8 GPUs with a batch size of 100; try --batch_size=96 instead." CC @isaprykin -- we decided in the end not to adjust automatically, but to throw a friendly error.

official/mnist/mnist.py
```diff
@@ -122,16 +127,49 @@ def model_fn(features, labels, mode, params):
...
+  available GPUs.
+
+  Note that this should eventually be handled by replicate_model_fn
+  directly. Multi-GPU support is currently experimental, however,
+  so doing the work here until that feature is in place.
+  """
+  from tensorflow.python.client import device_lib
+
+  local_device_protos = device_lib.list_local_devices()
+  num_gpus = sum([1 for d in local_device_protos if d.device_type == 'GPU'])
```
asimshankar: Perhaps handle the case where num_gpus are 0 (with a return) - otherwise we'll see an unhelpful `ZeroDivisionError`?

karmel: SG-- added below.

karmel:
All right, all-- I think I've addressed everything. I will merge on Monday unless there are any further edits-- thanks for all the review.

nealwu:
Hi all, for some reason tests were never run in this PR (I asked @av8ramit to investigate why), and they are currently failing. See https://github.com/tensorflow/models/pull/3230 for example. The specific error is

```
16:54:05     optimizer = tf.contrib.estimator.TowerOptimizer(optimizer)
16:54:05 AttributeError: 'module' object has no attribute 'TowerOptimizer'
```

because `tf.contrib.estimator.TowerOptimizer` is only available in the master version of TensorFlow.

asimshankar:
Ah, seems like the appropriate fix might be to just disable the multi_gpu test till we move to 1.5/1.6?

nealwu:
I think we likely want to revert this PR until `tf.contrib.estimator.TowerOptimizer` is available in the latest version of TensorFlow. Otherwise the code will fail for any user who attempts to run it on TensorFlow 1.4 (or even 1.5?) with a syntax error.

At the very least, if we still want to keep this in, we need to add a note to the README telling the user they have to install either the nightly or build from source in order to run this model. We [currently tell them](https://github.com/tensorflow/models/blob/master/official/README.md) all models work with TensorFlow 1.4.

isaprykin:
Oops.  It's because 1.5 still has the previous interface: https://github.com/tensorflow/tensorflow/blob/r1.5/tensorflow/contrib/estimator/python/estimator/replicate_model_fn.py.

av8ramit:
This is expected behavior.

@karmel is not a publicly listed member of the TensorFlow, so the bot asks can someone kick these off for us. As per our policy we cannot kick a job if someone we can't verify is internal makes a PR.

Additionally, nobody commented "Jenkins, test this please."

nealwu:
Yep, thanks for checking Amit. We don't want to auto-trigger tests for PRs written by someone external in case of malicious behavior.

The real solution is that PRs should not be allowed to be merged unless the tests have passed. Right now the policy is they can be merged if tests did not fail, which unfortunately includes the case where tests were never run.

I just reconfigured the repo to require tests to pass instead, so once we fix the current failures we should be good to go.

Issue #3103, opened 2018-01-04
from six.moves import xrange


cclauss:
The following files are incompatible with Python 3 at least in part because they use xrange() yet they lack the import:
* __from six.moves import xrange__
```
./research/compression/entropy_coder/dataset/gen_synthetic_dataset.py
./research/compression/entropy_coder/dataset/synthetic_model.py
./research/compression/entropy_coder/lib/blocks_masked_conv2d.py
./research/compression/entropy_coder/lib/blocks_masked_conv2d_test.py
./research/compression/entropy_coder/lib/blocks_std_test.py
./research/delf/delf/python/feature_io.py
./research/differential_privacy/dp_sgd/dp_mnist/dp_mnist.py
./research/differential_privacy/dp_sgd/per_example_gradients/per_example_gradients.py
./research/differential_privacy/multiple_teachers/aggregation.py
./research/differential_privacy/multiple_teachers/deep_cnn.py
./research/differential_privacy/multiple_teachers/input.py
./research/differential_privacy/multiple_teachers/train_student.py
./research/domain_adaptation/domain_separation/dsn_eval.py
./research/gan/cifar/util.py
./research/gan/image_compression/networks_test.py
./research/gan/mnist/util.py
./research/gan/mnist_estimator/train.py
./research/im2txt/im2txt/data/build_mscoco_data.py
./research/learned_optimizer/metaopt.py
./research/learning_to_remember_rare_events/data_utils.py
./research/learning_to_remember_rare_events/memory.py
./research/learning_to_remember_rare_events/train.py
./research/lfads/synth_data/generate_itb_data.py
./research/lfads/synth_data/generate_labeled_rnn_data.py
./research/neural_gpu/data_utils.py
./research/next_frame_prediction/cross_conv/eval.py
./research/next_frame_prediction/cross_conv/example_gen.py
./research/next_frame_prediction/cross_conv/model.py
./research/next_frame_prediction/cross_conv/reader.py
./research/next_frame_prediction/cross_conv/sprites_gen.py
./research/object_detection/dataset_tools/oid_tfrecord_creation.py
./research/object_detection/utils/np_box_list_ops.py
./research/pcl_rl/baseline.py
./research/pcl_rl/controller.py
./research/pcl_rl/env_spec.py
./research/pcl_rl/expert_paths.py
./research/pcl_rl/gym_wrapper.py
./research/pcl_rl/optimizers.py
./research/pcl_rl/replay_buffer.py
[...]

PR #3206, opened 2018-01-20
from six.moves import xrange (en masse)

cclauss:
@nealwu As requested at https://github.com/tensorflow/models/issues/3103#issuecomment-356150374

For all files that use the Python 2-only builtin function __xrange()__, add the line __from six.moves import xrange__ for compatibility with Python 3.

### DIFF
--- a/research/brain_coder/common/config_lib.py
+++ b/research/brain_coder/common/config_lib.py
@@ -10,6 +10,7 @@ else that may be specific to a particular run.
 
 import ast
 import itertools
+from six.moves import xrange
 
 
 class Config(dict):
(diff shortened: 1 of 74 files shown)


panyx0718:
The xrange is imported by not used? Would this cause lint issue?

cclauss:
In which file is __xrange()__ not used?

panyx0718:
I was looking at the diff. They import xrange. But I don't see xrange is used in the diff. I assume they weren't used before, otherwise there would be exceptions before this change.

cclauss:
Each file contains at least one call to __xrange()__.  You can test this via:
```
python3 -m flake8 . --select=F821 | grep xrange
```

panyx0718:
Oh, I see. you changed the existing xrange to use the six xrange

nealwu:
Looks good @cclauss. Thanks!

PR #3207, opened 2018-01-20
old style exceptions --> new style for Python 3

cclauss:
Old style exceptions were removed from Python 3 because that syntax was too ambiguous.

See: http://python-future.org/compatible_idioms.html#catching-exceptions

@nealwu @panyx0718


nealwu:
Closing this as it should be covered by https://github.com/tensorflow/models/pull/3229

PR #3208, opened 2018-01-21
Fix minor bug in resnet_v2

amlankar:
In resnet_v2.py the bottleneck layer should subsample the input after it is passed through bn and relu and not before.


amlankar:
Sorry, I think I am mistaken.

PR #3216, opened 2018-01-22
Refactor Resnet and use class-based model

karmel:
I started trying to add multi-GPU mode to Resnet, but I realized I couldn't do that cleanly without refactoring some of the Cifar10/Imagenet code. But I couldn't really refactor cleanly knowing that the conversion to class-based models was coming, so I attempted to refactor into class-based models here.

### Notes

* I tried not to touch data-processing, input functions, or the layer-assembly functions in this refactor, except to standardize the naming of a few variables. If you note differences... that is probably a mistake.
* This is intended as a true refactor, so no functionality should have changed, though some of the function names did do to the creation of the Model class.
* This appears to work with Cifar10, but Imagenet is still running as of ~2 days ago, so it will take a little while to confirm that that still converges. (See details on performance/accuracy below.)

### Open questions

- [ ] For the ease of review, I kept the functions in resnet_model.py and resnet_shared.py separate, but, given the length/purpose of the files, it might make sense to combine them into a single module, as in MNIST. Thoughts?
- [x] One known difference that I still have to resolve: in the original Imagenet, loss was calculated over all trainable vars [EXCEPT those with 'batch_normalization' in the name](https://github.com/tensorflow/models/blob/master/official/resnet/imagenet_main.py#L189), with a comment saying this was better for accuracy. I have not yet added this in, as I'm not sure the best solution here-- some choices: (1) Remove it for both datasets; (2) add it for both datasets; (3) add a param that indicates whether batch_normalization vars should be excluded; (4) add a more flexible means of excluding arbitrary vars from the loss. Opinions as to which of these is the best path for both these particular models and overall model usability?

### Appendix

Cifar10 trains on a single GPU to comparable accuracy as master, and in comparable time:

```
INFO:tensorflow:learning_rate = 1e-04, cross_entropy = 0.0043777777, train_accuracy = 0.99919873 (10.313 sec)
INFO:tensorflow:loss = 0.15030411, step = 97569 (10.313 sec)
INFO:tensorflow:global_step/sec: 9.71699
INFO:tensorflow:learning_rate = 1e-04, cross_entropy = 0.00943729, train_accuracy = 0.99921876 (10.291 sec)
INFO:tensorflow:loss = 0.15535825, step = 97669 (10.291 sec)
...
INFO:tensorflow:Saving dict for global step 97675: accuracy = 0.9285, global_step = 97675, loss = 0.46341026
{'loss': 0.46341026, 'global_step': 97675, 'accuracy': 0.9285}
```

Imagenet is still training on a single GPU. Current status:

```
INFO:tensorflow:loss = 3.966461, step = 216786 (66.835 sec)
INFO:tensorflow:global_step/sec: 1.49444
INFO:tensorflow:learning_rate = 0.0125, cross_entropy = 2.3299885, train_accuracy = 0.39620537 (66.915 sec)
```

I tested that these run on CPU, but did not wait for them to complete.


official/resnet/resnet_model.py
```diff
@@ -205,162 +205,99 @@ def projection_shortcut(inputs):
...
-        data_format=data_format)
-    inputs = block_layer(
-        inputs=inputs, filters=32, block_fn=building_block, blocks=num_blocks,
-        strides=2, is_training=is_training, name='block_layer2',
-        data_format=data_format)
-    inputs = block_layer(
-        inputs=inputs, filters=64, block_fn=building_block, blocks=num_blocks,
-        strides=2, is_training=is_training, name='block_layer3',
-        data_format=data_format)
+    self.params = self.get_model_params()
```
asimshankar: Would it make sense for all params to be arguments to `__init__` instead of this separate `get_model_params()` method?

official/resnet/resnet_shared.py
```diff
@@ -0,0 +1,175 @@
...
+  tf.summary.scalar('cross_entropy', cross_entropy)
+
+  # Add weight decay to the loss.
+  loss = cross_entropy + params['weight_decay'] * tf.add_n(
+      [tf.nn.l2_loss(v) for v in tf.trainable_variables()])
+
+  if mode == tf.estimator.ModeKeys.TRAIN:
+    # Scale the learning rate linearly with the batch size. When the batch size
+    # is 128, the learning rate should be 0.1.
+    initial_learning_rate = 0.1 * params['batch_size'] / params['batch_denom']
```
asimshankar: Instead of multiple params - `batch_size`, `batch_denorm`, `epochs` etc., my inclination would be to take a single param: `learning_rate`, which is supposed to be a function that returns the learning rate tensor. So, instead of:

```python
tf.estimator.Estimator(model_fn=..., params={
  'batch_size': FLAGS.batch_size,
  'batch_denom': FLAGS.batch_denom,
   ...
})
```

you'd have:

```python
def learning_rate():
   ...

tf.estimator.Estimator(model_fn=..., params={
  'learning_rate' = learning_rate,
})
```

With that, we more cleanly separate out the learning rate decay logic from the rest of the model, make it easier to experiment with for those that are interested, and make for fewer parameters to understand for those who aren't.

Thoughts?

official/resnet/resnet_model.py
karmel: `get_model_params` makes it easy to subclass this for each dataset and just define the params for model setup in the class. Passing through init would work, but makes it harder to run some of the checks that are easy to include in a class. Six of one, half dozen of another at the end of the day, but since we're going class-based, it seems more Pythonic to keep it that way for the specific versions as well.

official/resnet/resnet_shared.py
karmel: I have no issue with that-- I don't have a great sense of what people would want to play with at this stage, so I trust you on that. That would mean each learning_rate would be defined in the dataset-specific file, which separates it from the rest of the training logic, but if you think finding the function is less burdensome in this case than passing in the params, that WFM.

official/resnet/resnet_model.py
asimshankar: Didn't quite follow. I was imagining something like this:

```python
class Model(object):
   def __init__(self,
                       resnet_size,
                       num_classes,
                       num_filters,
                       kernel_size,
                       ...):
     # ...

class Cifar10Model(Model):
   def __init__(self, resnet_size, data_format=None):
         super(ImageNetModel, self).__init__(
                 resnet_size=resnet_size,
                 data_format=data_format,
                 num_classes=10,
                 kernel_size=3,
                 ...):
```

This way there is one clear specification of all the parameters in the `Model` class. With the `get_model_parameters()` method, which returns a dictionary, there is no obvious spec of what the keys in the dictionary should be. One has to comb through the implementation of the base class to figure that out.

I feel having that clear spec will be helpful.

karmel: Oh, got it, I misunderstood. Yes, that makes sense-- will do.

official/resnet/cifar10_main.py
```diff
@@ -180,116 +149,75 @@ def input_fn(is_training, data_dir, batch_size, num_epochs=1):
...
-    with tf.control_dependencies(update_ops):
-      train_op = optimizer.minimize(loss, global_step)
-  else:
-    train_op = None
+  features = tf.reshape(features, [-1, _HEIGHT, _WIDTH, _NUM_CHANNELS])
+
+  learning_rate_fn = resnet_shared.learning_rate_with_decay(
+      batch_size=params['batch_size'], batch_denom=128,
+      train_images=_NUM_IMAGES['train'], epochs=[100, 150, 200],
+      decay_rates=[1, 0.1, 0.01, 1e-3])
```
karmel: @asimshankar -- is this what you had in mind?

official/resnet/resnet_model.py
```diff
@@ -205,162 +205,126 @@ def projection_shortcut(inputs):
...
+      resnet_size: A single integer for the size of the ResNet model.
+      num_classes: The number of classes used as labels.
+      num_filters: The number of filters to use for the model.
+      kernel_size: The kernel size to use for convolution.
+      first_pool_size: Pool size to be used for the first pooling layer.
+        If none, the first pooling layer is skipped.
+      second_pool_size: Pool size to be used for the second pooling layer.
+      block_fn: Which block layer function should be used? Pass in one of
+        the two functions defined above: building_block or bottleneck_block
+      layers: A list containing n values, where n is the bnumber of block
```
asimshankar: `layers` seems like a very ambiguous name (especially since there are other "layers" like the pooling ones and the final layer that aren't affected by this). How about we call out all the "block" affecting parameters together. So `layers` becomes `block_sizes` or `num_blocks` and `stride_sizes` becomes `block_strides` (see comment below).

Then we have three "block" parameters: `block_fn`, `block_sizes`, `block_strides`.

karmel:
That was a very effective review, and I feel like the code is much more readable now. Thanks @asimshankar  for not letting me be lax about documentation. I think I have addressed all the comments accept the original two open questions:

1. I will combine the two files, but in a follow-up PR to prevent Github from losing all of these comments.
2. @nealwu -- thoughts on the second of the two Open Questions in the original PR above? If you don't have a preference, I will just pick one of options (3) or (4) for resolving.

Meanwhile, running the models to make sure things still work.

nealwu:
What I found with including batch normalization in the loss calculations in CIFAR-10 and in ImageNet was that in CIFAR-10, it provided a significant boost to the validation accuracy, whereas in ImageNet, it caused a noticeable drop in the validation accuracy.

Our best theory for why that happens is that since CIFAR-10 is a fairly small dataset, the batch norm variables have the possibility of overfitting the training data, so some regularization helps. Meanwhile in ImageNet, regularization is not necessary and doesn't help but actually hurts. I found the effect in both cases to be decently large, so I would recommend keeping it (either option 3 or 4).

nealwu:
One more thing to consider: unfortunately there is a good chance this PR ends up breaking my [checkpoint](https://github.com/tensorflow/models/tree/master/official/resnet#pre-trained-model). Could you check whether that checkpoint still works, and if not, are you able to generate a new one?

karmel:
All right, batch_normalization now excluded and commented with @nealwu 's notes. Coming in follow-up PRs:

* Combining the two files into one resnet.py
* I will upload new saved model files for these after they finish running. (Just restarted with most recent changes. CIFAR-10 should be done in a few hours, ImageNet will take ... 6.5d at its current rate.)

Anything remaining for this PR? Or are we good at this point?

official/resnet/cifar10_main.py
```diff
@@ -18,63 +18,33 @@
...
-         'provides a performance boost on GPU but is not always compatible '
-         'with CPU. If left unspecified, the data format will be chosen '
-         'automatically based on whether TensorFlow was built for CPU or GPU.')
+import resnet_shared
 
 _HEIGHT = 32
 _WIDTH = 32
-_DEPTH = 3
+_NUM_CHANNELS = 3
+_DEFAULT_IMAGE_SIZE = _HEIGHT * _WIDTH * _NUM_CHANNELS
```
asimshankar: Optional: Perhaps call this `_DEFAULT_IMAGE_BYTES` (to make the unit of "size" clearer)

official/resnet/cifar10_main.py
```diff
@@ -180,116 +149,80 @@ def input_fn(is_training, data_dir, batch_size, num_epochs=1):
...
+      batch_size=params['batch_size'], batch_denom=128,
+      num_images=_NUM_IMAGES['train'], boundary_epochs=[100, 150, 200],
+      decay_rates=[1, 0.1, 0.01, 0.001])
 
-  # Create a tensor named train_accuracy for logging purposes
-  tf.identity(accuracy[1], name='train_accuracy')
-  tf.summary.scalar('train_accuracy', accuracy[1])
+  # We use a weight decay of 0.0002, which performs better
+  # than the 0.0001 that was originally suggested.
+  weight_decay = 2e-4
```
asimshankar: Optional: Inline this into the call to `resnet_shared.resnet_model_fn`?

official/resnet/imagenet_main.py
```diff
@@ -159,120 +126,90 @@ def input_fn(is_training, data_dir, batch_size, num_epochs=1):
...
+class ImagenetModel(resnet_model.Model):
+  # >= this value of resnet_size, use more layers and bottleneck blocks.
+  size_threshold = 50
+
+  def __init__(self, resnet_size, data_format=None):
+    """These are the parameters that work for Imagenet data.
+    """
+
+    # For bigger models, we want to use "bottleneck" layers
+    if resnet_size < self.size_threshold:
```
asimshankar: This seems to be the only place `size_threshold` is used. If so, I would suggest that we remove lines 133 and 134 and inline this to:

```python
if resnet_size < 50:
  ...
```

official/resnet/imagenet_main.py
```diff
@@ -159,120 +126,90 @@ def input_fn(is_training, data_dir, batch_size, num_epochs=1):
...
+        second_pool_size=7,
+        second_pool_stride=1,
+        block_fn=block_fn,
+        block_sizes=_get_imagenet_layers(resnet_size),
+        block_strides=[1, 2, 2, 2],
+        final_size=final_size,
+        data_format=data_format)
+
+
+def _get_imagenet_layers(resnet_size):
```
asimshankar: Rename this (specifically, don't use the word `layers`)? Perhaps `def _block_sizes(resnet_size)` instead?

official/resnet/imagenet_test.py
```diff
@@ -46,11 +47,12 @@ def reshape(shape):
 
     with graph.as_default(), self.test_session(
         use_gpu=with_gpu, force_gpu=with_gpu):
-      model = resnet_model.imagenet_resnet_v2(
-          resnet_size, 456,
+      model = imagenet_main.ImagenetModel(
+          resnet_size,
           data_format='channels_first' if with_gpu else 'channels_last')
+      model.num_classes = 456
```
asimshankar: Changing `num_classes` after the fact seems a bit iffy to me as it isn't obvious that it is safe to do so (for example, does the base-class of `ImagenetModel` set some state based on `num_classes` provided to it?).

I would suggest that instead of changing this after the fact, we make `ImagenetModel.__init__` take a `num_classes` argument (that can default to `_NUM_CLASSES`). So line 50 will become:

```python
model = imagenet_main.ImagenetModel(
  resnet_size,
  data_format=...,
  num_classes=456)
```

official/resnet/cifar10_main.py
karmel: I left this pulled out so it was clear what the comment referred to, since it seems like an important detail that we are changing the original implementation.

official/resnet/imagenet_test.py
karmel: This appeared just to be a relic preserved from the previous test, and not really testing anything in particular. Just used the default class size, rather than introducing a new param for the sake of one test.

karmel:
All right, all in, and everything still runs. Merging this, and will address any remaining concerns that get discovered in follow ups. Thanks, all.

PR #3220, opened 2018-01-22
long was removed in Python 3 (en masse)

cclauss:
__long__ was [removed](https://docs.python.org/3/whatsnew/3.0.html#integers) from Python 3 in favor of __int__.

@nealwu Like we did for __xrange()__ in #3206 but this time for the builtin type __long__.

### DIFF
--- a/research/object_detection/dataset_tools/create_oid_tf_record.py
+++ b/research/object_detection/dataset_tools/create_oid_tf_record.py
@@ -42,11 +42,6 @@ import tensorflow as tf
 from object_detection.dataset_tools import oid_tfrecord_creation
 from object_detection.utils import label_map_util
 
-try:
-    long        # Python 2
-except NameError:
-    long = int  # Python 3
-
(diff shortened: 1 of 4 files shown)


research/object_detection/dataset_tools/create_oid_tf_record.py
```diff
@@ -42,6 +42,11 @@
 from object_detection.dataset_tools import oid_tfrecord_creation
 from object_detection.utils import label_map_util
 
+try:
+    long        # Python 2
+except NameError:
+    long = int  # Python 3
+
```
nealwu: Could we solve this by just using `int` instead of `long` in the code below? I checked [the documentation for `int`](https://docs.python.org/2/library/functions.html#int), and it states, "If the argument is outside the integer range, the function returns a long object instead."

cclauss: Very nice!  You learn something new every day.  I will make it so.

PR #3231, opened 2018-01-24
file() was removed in Python 3 (en masse)

cclauss:
@nealwu The general advise is http://python-future.org/compatible_idioms.html#file but my sense is that just replacing __file()__ with __open()__ is OK for these use cases.  Your thoughts?


nealwu:
Jenkins, test this please.

nealwu:
Jenkins, test this please.

PR #3256, opened 2018-01-26
Use nightly TensorFlow docker image for tests

karmel:
As per new release strategy.


nealwu:
Is there a way we can test both the nightly and the latest stable version? We wouldn't want to stop testing the latest stable version and end up with it breaking.

karmel:
We should be able testing for the release + stable TF, though in theory that should be stable. Probably better to do that in the new testing system directly, rather than in Jenkins.

karmel:
@tensorflow-jenkins: test this please

PR #3257, opened 2018-01-26
MNIST multi-GPU, take 3

karmel:
Reopening this now that tests are against nightly TF.


karmel:
@nealwu @k-w-w This should be the same diff previously merged, but now with the tests passing. Should not require more than a once-over.

Issue #3239, opened 2018-01-24
create_kitti_tf_record.py


PythonImageDeveloper:
hi
when i run this code , convert kitti to tfrecord , i get this error : 
#################################
python3 dataset_tools/create_kitti_tf_record.py \
>         --data_dir=kitti \
>         --output_path=kitti.record
Traceback (most recent call last):
  File "dataset_tools/create_kitti_tf_record.py", line 65, in <module>
    tf.app.flags.DEFINE_list('classes_to_use', ['Car', 'Pedestrian' , 'DontCare'],
AttributeError: module 'tensorflow.python.platform.flags' has no attribute 'DEFINE_list'
#############################################
data_dir=kitti is my dataset , which in this folder is training  and labels . 
what is this error ? 
in folder data : kitti_label_map.pbtxt 
item {
  id: 1
  name: 'Car'
}

item {
  id: 2
  name: 'Pedestrian'
}

item {
  id: 3
  name: 'DontCare'
}

SamDon87:
Change the Line 
tf.app.flags.DEFINE_list to
tf.app.flags.DEFINE_string
in dataset_tools/create_kitti_tf_record.py

tatatodd:
Please provide details about what platform you are using  (operating system, architecture). Also include your TensorFlow version. Also, did you compile from source or install a binary?  Make sure you also include the exact command if possible to produce  the output included in your test case. If you are unclear what to include  see the issue template displayed in  [the Github new issue template](https://github.com/tensorflow/tensorflow/issues/new).

 We ask for this in the issue submission template, because    it is really difficult to help without that information. Thanks!

zeynali:
Ubuntu 16.04
Tensorow 1.4 binary

zeynali:
I also some question :
When i used 3 id in the kitti–label-map.pbtxt, so this be trained only 3 class automatically or be needed to change some modification in some files?

tatatodd:
What is the full stack trace of the error?

zeynali:
thanks for SamDon87 ,
i Change the line , my problem is ok.
tf.app.flags.DEFINE_list to
tf.app.flags.DEFINE_string
in dataset_tools/create_kitti_tf_record.py

i have question also ,
[...]

PR #3271, opened 2018-01-30
There is no DEFINE_list in public tensorflow

kibergus:
Here is a patch to deal with missing DEFINE_list #3239


kibergus:
I signed it!

PR #3278, opened 2018-01-31
Removed unused regularization ops

frreiss:
While using this file as a starting point for some other work, I noticed some redundancy in the regularization code. The fourth argument to `_variable_with_weight_decay()` is supposed to be `None` for variables with no regularization. Instead the script uses the value `0.0`, which results in the creation of some regularization losses that are always zero. This PR changes those `0.0`'s to `None`'s.

After this change, training runs about 2% faster, and the checkpoint files are a tiny bit smaller. The script's behavior is otherwise unchanged.


karmel:
@tensorflow-jenkins: test this please

PR #3301, opened 2018-02-02
Merge Resnet files

karmel:
As per discussion in #3216 , merging resnet_shared and resnet_model into a single module.


karmel:
Thanks-- Cifar does indeed converge, and Imagenet at least runs without trouble. Merging...

PR #3314, opened 2018-02-03
lfads: Adding do_train_readin option allowing fixed readin matrices

djoshea:
Affects research/lfads. @sussillo 

1. For stitched models, the readin matrices and bias vectors are initialized to the
"alignment" matrix and bias specified in each dataset's .h5 file. If do_train_readin is
True, these will be trainable, and if not, they will be fixed.
2. Initializing readout matrices to readin transpose rather than pseudoinverse.
3. Fixes flag parsing issues with ints vs floats

### DIFF
--- a/research/lfads/lfads.py
+++ b/research/lfads/lfads.py
@@ -366,11 +366,11 @@ class LFADS(object):
       if datasets and 'alignment_matrix_cxf' in datasets[name].keys():
         dataset = datasets[name]
         if hps.do_train_readin:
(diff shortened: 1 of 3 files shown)


sussillo:
A few changes to support stitching of multiple sessions.  Primarily, we noticed that matrix transpose was empirically better than pinv() for initializing the output factors matrix from the input factors matrix.

sussillo:
Jenkins, test this please.

PR #3325, opened 2018-02-05
Fix wrong reference in ptb_word_lm.py

yanboliang:


tutorials/rnn/ptb/ptb_word_lm.py
```diff
@@ -225,15 +225,15 @@ def make_cell():
 
     self._initial_state = cell.zero_state(config.batch_size, data_type())
     state = self._initial_state
-    # Simplified version of tensorflow_models/tutorials/rnn/rnn.py's rnn().
```
yanboliang: ```tensorflow_models/tutorials/rnn/rnn.py``` has been removed, and corresponding APIs moved to ```tensorflow/python/ops/rnn.py``` and were exported by ```tf.nn.***```.

tutorials/rnn/ptb/ptb_word_lm.py
```diff
@@ -225,15 +225,15 @@ def make_cell():
...
+    # Simplified version of tf.nn.static_rnn().
     # This builds an unrolled LSTM for tutorial purposes only.
-    # In general, use the rnn() or state_saving_rnn() from rnn.py.
+    # In general, use tf.nn.static_rnn() or tf.nn.static_state_saving_rnn().
```
lukaszkaiser: I think it's in contrib, not in nn, no?

yanboliang: Either can work well, and ```contrib``` only import the API and implementation from ```nn```. So I think ```nn``` should be preferred. Does it reasonable?

WeichenXu123:
Good work 💯

PR #3332, opened 2018-02-06
Refactoring input function and allowing for more parallel processing

karmel:
This is one of what may end up being several changes that come out of the multi-GPU Resnet work. We found in testing that training was blocking on input processing, and the changes here serve to (1) unify the cifar/imagenet input pipelines, (2) conform with the advice from @mrry as to input processing, and (3) allow for more parallel calls during the mapping phase.

An example configuration: 16xCPU, 1xGPU, training Imagenet-- on master runs at ~5.4 steps/sec; with this branch and --parallel_calls=16, runs at ~5.9 steps/sec. (Stats on CIFAR10 don't move much, but that is expected given the size of the dataset.)

Both datasets run; will be convergence testing as we do more multi-GPU work over the next week.

### DIFF
--- a/official/resnet/cifar10_test.py
+++ b/official/resnet/cifar10_test.py
@@ -43,8 +43,10 @@ class BaseTest(tf.test.TestCase):
     data_file.write(fake_data)
     data_file.close()
 
-    fake_dataset = cifar10_main.record_dataset(filename)
-    fake_dataset = fake_dataset.map(cifar10_main.parse_record)
+    fake_dataset = tf.data.FixedLengthRecordDataset(
+        filename, cifar10_main._RECORD_BYTES)
+    fake_dataset = fake_dataset.map(
+        lambda val: cifar10_main.parse_record(val, False))
     image, label = fake_dataset.make_one_shot_iterator().get_next()
 
     self.assertEqual(label.get_shape().as_list(), [10])
@@ -57,7 +59,7 @@ class BaseTest(tf.test.TestCase):
 
       for row in image:
         for pixel in row:
-          self.assertAllEqual(pixel, np.array([0, 1, 2]))
+          self.assertAllClose(pixel, np.array([-1.225, 0., 1.225]), rtol=1e-3)
 
   def input_fn(self):
     features = tf.random_uniform([_BATCH_SIZE, 32, 32, 3])
--- a/official/resnet/resnet.py
+++ b/official/resnet/resnet.py
@@ -46,7 +46,7 @@ _BATCH_NORM_EPSILON = 1e-5
 # Functions for input processing.
 ################################################################################
 def process_record_dataset(dataset, is_training, batch_size, shuffle_buffer,
-                           parse_record_fn, num_epochs=1, parallel_calls=1):
+                           parse_record_fn, num_epochs=1, num_parallel_calls=1):
   """Given a Dataset with raw records, parse each record into images and labels,
   and return an iterator over the records.
   Args:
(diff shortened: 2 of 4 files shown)


official/resnet/resnet.py
```diff
@@ -516,6 +572,10 @@ def __init__(self, resnet_size_choices=None):
         '--data_dir', type=str, default='/tmp/resnet_data',
         help='The directory where the input data is stored.')
 
+    self.add_argument(
+        '--parallel_calls', type=int, default=5,
+        help='The number of input records to process in parallel.')
```
k-w-w: Perhaps mention that this should be equal to the number of CPU cores here as well.

official/resnet/resnet.py
```diff
@@ -516,6 +572,10 @@ def __init__(self, resnet_size_choices=None):
         '--data_dir', type=str, default='/tmp/resnet_data',
         help='The directory where the input data is stored.')
 
+    self.add_argument(
+        '--parallel_calls', type=int, default=5,
```
k-w-w: How about `input_parallel_calls`, so that it's clear that this flag only applies to the # of parallel calls during the input pipeline (and that it's not related to the multi-gpu functionality)

karmel: Looking at the new data processing guide ( https://www.tensorflow.org/versions/master/performance/datasets_performance#parallelize_data_extraction ), I ended up going with num_parallel_calls to be consistent with that. I think with the greater detail in the help comment, it should be pretty clear that it's specifically for input processing?

official/resnet/resnet.py
```diff
@@ -42,6 +42,57 @@
...
+  # load input files as we go through shuffling and processing.
+  dataset = dataset.prefetch(buffer_size=batch_size)
+  if is_training:
+    # Shuffle the records. Note that we shuffle before repeating to ensure
+    # that the shuffling respects epoch boundaries.
+    dataset = dataset.shuffle(buffer_size=shuffle_buffer)
+
+  # If we are training over multiple epochs before evaluating, repeat the
+  # dataset for the appropriate number of epochs.
+  dataset = dataset.repeat(num_epochs)
```
nealwu: We can also consider using [`shuffle_and_repeat`](https://www.tensorflow.org/versions/master/api_docs/python/tf/contrib/data/shuffle_and_repeat) which @jsimsa implemented for another small performance boost -- I vaguely remember this helping by 2-3%, since you no longer have to wait a long time for shuffling at the start of each epoch. The only downside it's still in contrib.

official/resnet/resnet.py
```diff
@@ -42,6 +42,57 @@
...
+  dataset = dataset.map(lambda value: parse_record_fn(value, is_training),
+                        num_parallel_calls=num_parallel_calls)
+
+  dataset = dataset.batch(batch_size)
+
+  # Operations between the final prefetch and the get_next call to the iterator
+  # will happen synchronously during run time. We prefetch here again to
+  # background all of the above processing work and keep it out of the
+  # critical training path.
+  dataset = dataset.prefetch(1)
```
nealwu: Nice. Is this the advice from @mrry you mentioned?

official/resnet/cifar10_test.py
```diff
@@ -57,7 +59,7 @@ def test_dataset_input_fn(self):
 
       for row in image:
         for pixel in row:
-          self.assertAllEqual(pixel, np.array([0, 1, 2]))
+          self.assertAllClose(pixel, np.array([-1.225, 0., 1.225]), rtol=1e-3)
```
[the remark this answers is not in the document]
karmel: Yes, exactly.

official/resnet/resnet.py
karmel: Yes-- but still in contrib :(

official/resnet/resnet.py
karmel: Yes, and his explanation as well.

PR #3343, opened 2018-02-08
Allow users to pass in num_classes to ResNet

nealwu:
This enables users to use our predefined ResNet models for datasets other than CIFAR-10 and ImageNet.

In addition an internal team uses this model, and this change is needed to fix their tests that broke with the latest copybara sync.


nealwu:
Done.

PR #3380, opened 2018-02-14
Internal changes for object detection.

pkulzc:


tushar-dadlani:
@pkulzc @jch1 Can pycocotools be checked in or submoduled? It currently is a dangling dependency.

pkulzc:
@tushar-dadlani You can now follow the [COCO API installation](https://github.com/tensorflow/models/blob/master/research/object_detection/g3doc/installation.md#coco-api-installation) to install it.

PR #3383, opened 2018-02-14
Performance tuning image preprocessing

karmel:
This PR cleans up and tunes some of the VGG preprocessing code. While the changes below don't have any noticeable effect on training with 1 GPU, the difference becomes substantial with multiple GPUs (where preprocessing and IO become bottlenecks). The changes below should not result in any differences in functionality, and we still converge to 75% (tested on 8 GPUs).

@asimshankar @k-w-w -- as @nealwu is out for a few weeks, tagging you two for review.

CC @isaprykin

### DIFF
--- a/official/resnet/imagenet_main.py
+++ b/official/resnet/imagenet_main.py
@@ -82,7 +82,7 @@ def parse_example_proto(example_serialized):
 
   Returns:
     image_buffer: Tensor tf.string containing the contents of a JPEG file.
-    label: Tensor tf.int32 containing the label.
+    label: Tensor tf.int64 containing the label.
   """
   # Dense features in Example proto.
   feature_map = {
@@ -103,15 +103,7 @@ def parse_example_proto(example_serialized):
 
   features = tf.parse_single_example(example_serialized, feature_map)
 
-  image = tf.image.decode_image(tf.reshape(parsed['image/encoded'], shape=[]),
-      _NUM_CHANNELS)
-
-  # Note that tf.image.convert_image_dtype scales the image data to [0, 1).
-  image = tf.image.convert_image_dtype(image, dtype=tf.float32)
-
-  label = tf.cast(features['image/class/label'], dtype=tf.int32)
-
-  return image, label
+  return features['image/encoded'], features['image/class/label']
 
 
 def parse_record(raw_record, is_training):
@@ -122,7 +114,8 @@ def parse_record(raw_record, is_training):
   # Note that the resulting image contains an unknown height and width
   # that is set dynamically by decode_jpeg. In other words, the height
   # and width of image is unknown at compile-time.
-  image = tf.image.decode_jpeg(image_buffer, channels=_NUM_CHANNELS)
+  # Results in a 3-D float Tensor with values ranging from [0, 1).
+  image = tf.image.decode_jpeg(image, channels=_NUM_CHANNELS)
 
   image = vgg_preprocessing.preprocess_image(
       image=image,
@@ -130,7 +123,9 @@ def parse_record(raw_record, is_training):
       output_width=_DEFAULT_IMAGE_SIZE,
       is_training=is_training)
 
-  return image, tf.one_hot(label, _NUM_CLASSES)
+  label = tf.cast(label, dtype=tf.int32)
+
+  return image, label
 
 
 def input_fn(is_training, data_dir, batch_size, num_epochs=1,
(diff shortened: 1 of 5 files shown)


official/resnet/imagenet_main.py
```diff
@@ -54,49 +54,79 @@ def get_filenames(is_training, data_dir):
...
+  # Sparse features in Example proto.
+  feature_map.update(
+      {k: sparse_float32 for k in ['image/object/bbox/xmin',
+                                   'image/object/bbox/ymin',
+                                   'image/object/bbox/xmax',
+                                   'image/object/bbox/ymax']})
+
+  features = tf.parse_single_example(example_serialized, feature_map)
+
+  return features['image/encoded'], features['image/class/label']
```
k-w-w: Does the `feature_map` have to include keys other than `image/encoded` and `image/class/label`? (since those are the only 2 keys used)

official/resnet/vgg_preprocessing.py
```diff
@@ -42,155 +42,62 @@
...
-  """Crops the given list of images.
+  shape = tf.shape(image)
+  return shape[0], shape[1], shape[2]
 
-  The function applies the same crop to each image in the list. This can be
-  effectively applied when there are multiple image inputs of the same
-  dimension such as:
 
-    image, depths, normals = _random_crop([image, depths, normals], 120, 150)
+def _random_crop(image, crop_height, crop_width):
```
k-w-w: Consider renaming to `_random_crop_and_flip` (or similar) to indicate that it also flips the image randomly.

official/resnet/imagenet_main.py
karmel: That's a good question. I left them there in part because they had all been there before; some versions of Resnet we have use the bounding boxes, but this one does not. If no one thinks they are useful/informative, I will remove them.

official/resnet/vgg_preprocessing.py
```diff
@@ -42,155 +42,62 @@
...
-
-    outputs.append(_crop(image, offset_height, offset_width,
-                         crop_height, crop_width))
-  return outputs
+  total_crop_height = (height - crop_height)
+  crop_top = total_crop_height // 2
+  total_crop_width = (width - crop_width)
+  crop_left = total_crop_width // 2
+  return tf.slice(
+      image, [crop_top, crop_left, 0], [crop_height, crop_width, num_channels])
```
asimshankar: FYI: I think you can also do:
```python
return tf.slice(image, [crop_top, crop_left, 0], [crop_height, crop_width, -1])
```

so you don't need `num_channels`.

official/resnet/imagenet_main.py
```diff
@@ -54,49 +54,72 @@ def get_filenames(is_training, data_dir):
...
   image = vgg_preprocessing.preprocess_image(
       image=image,
       output_height=_DEFAULT_IMAGE_SIZE,
       output_width=_DEFAULT_IMAGE_SIZE,
       is_training=is_training)
 
-  label = tf.cast(
-      tf.reshape(parsed['image/class/label'], shape=[]),
-      dtype=tf.int32)
+  label = tf.cast(tf.reshape(label, shape=[]), dtype=tf.int32)
```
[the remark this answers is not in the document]
karmel: Doesn't look like it:

```
TypeError: Value passed to parameter 'dense_defaults' has DataType int32 not in list of allowed values: float32, int64, string
```

official/resnet/vgg_preprocessing.py
```diff
@@ -269,68 +176,13 @@ def _aspect_preserving_resize(image, smallest_side):
   """
   smallest_side = tf.convert_to_tensor(smallest_side, dtype=tf.int32)
 
-  shape = tf.shape(image)
-  height = shape[0]
-  width = shape[1]
+  height, width, _ = _get_h_w_c(image)
```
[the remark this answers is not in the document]
karmel: Interesting. Is that something specific to Dataset.map? Or do operations escape pruning in more cases than that?

karmel:
Thanks, @k-w-w and @asimshankar . Updated accordingly, and merging.

official/resnet/imagenet_main.py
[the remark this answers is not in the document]
asimshankar: This is with changing `tf.FixedLengthFeature(..., dtype=tf.int64)` to `tf.FixedLengthFeature(..., dtype=tf.int32)`?

PR #3389, opened 2018-02-16
bug correction for dataset import

jind11:
The iterator to generate batches for imported dataset is missed.

### DIFF
--- a/official/resnet/resnet.py
+++ b/official/resnet/resnet.py
@@ -14,16 +14,13 @@
 # ==============================================================================
 """Contains definitions for the preactivation form of Residual Networks
 (also known as ResNet v2).
-
 Residual networks (ResNets) were originally proposed in:
 [1] Kaiming He, Xiangyu Zhang, Shaoqing Ren, Jian Sun
     Deep Residual Learning for Image Recognition. arXiv:1512.03385
-
 The full preactivation 'v2' ResNet variant implemented in this module was
 introduced by:
 [2] Kaiming He, Xiangyu Zhang, Shaoqing Ren, Jian Sun
     Identity Mappings in Deep Residual Networks. arXiv: 1603.05027
-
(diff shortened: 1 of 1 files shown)


jind11:
hi, I have signed the CLA. Thanks!

jind11:
I submitted another GLA using my email used for github

jind11:
hi, I have resolved this problem

jind11:
I signed it!

karmel:
Thanks, @jind11 . The current best-practice is to return the dataset directly from the input_fn. That is reflected in the [Dataset performance guide](https://www.tensorflow.org/versions/master/performance/datasets_performance), though I note it is not yet in the [Dataset Quickstart](https://www.tensorflow.org/versions/master/get_started/datasets_quickstart) -- @MarkDaoust , do we plan on updating the sample input_fn in the Dataset docs?

jind11:
I tried running the cifar10_main.py but got an error on the input_fn, so I added these several lines to fix the bug.

MarkDaoust:
> Dataset Quickstart

@karmel  - Thanks for pointing that out I've drafted some clarifications for this section.

@jind11 - Thanks for the PR (the best kind of bug report is one with a fix attached), but in this case the functionality is new in TensorFlow r1.5 I bet you're on an older version try a `pip install --upgrade tensorflow` and it should work without the change.

jind11:
@MarkDaoust I see. Yeah, I am now using tensorflow 1.4. So it would be great that you can clarify this version requirement in the readme page so that we know what could cause the error when we run the code. Thanks a lot for your help!

karmel:
Thanks, @jind11 and @MarkDaoust . I will close this given the successful resolution. Version recommendations for all of the official models can be found here: https://github.com/tensorflow/models/tree/master/official#tensorflow-official-models

PR #2804, opened 2017-11-15
Look for Python syntax errors when testing

cclauss:
http://flake8.readthedocs.io/en/latest/user/error-codes.html#error-violation-codes says:
* _We report __E999__ when we fail to compile a file into an Abstract Syntax Tree_...

PR #3396, opened 2018-02-16
ObjDet: print() is a function in Python 3

cclauss:
The intent of #2804 is to catch these issues automatically.  @pkulzc @lzc5123016

Object Detection Code Owners: @jch1 @tombstone @derekjchow @jesu9 @dreamdragon
* https://github.com/tensorflow/models/commit/7a9934df2afdf95be9405b4e9f1f2480d748dc40#r27482046


cclauss:
These were fixed in other PRs

PR #3399, opened 2018-02-16
Adding dockerfiles that encapsulate this repository.

vishh:
1. These Dockerfiles can be used to generate official docker images that can used for running models in this repo on container based platforms
2. We can leverage google container builder to build images for every branch, tag and/or HEAD.

### DIFF
--- a/official/Dockerfile.cpu
+++ b/official/Dockerfile.cpu
@@ -1,0 +1,17 @@
+# Docker image for running examples in Tensorflow models.
+# base_image depends on whether we are running on GPUs or non-GPUs
+FROM ubuntu:latest
+
+RUN apt-get update && apt-get install -y --no-install-recommends \
+    ca-certificates \
+    build-essential \
+    git \
+    python \
+    python-pip \
+    python-setuptools
+    
+RUN pip install tf-nightly
+
+# Checkout tensorflow/models and pin to a particular version
+RUN git clone https://github.com/tensorflow/models.git /tensorflow_models
+
--- a/official/Dockerfile.gpu
+++ b/official/Dockerfile.gpu
@@ -1,0 +1,18 @@
+# Docker image for running examples in Tensorflow models.
+# base_image depends on whether we are running on GPUs or non-GPUs
+FROM nvidia/cuda:9.0-cudnn7-runtime-ubuntu16.04
+
+RUN apt-get update && apt-get install -y --no-install-recommends \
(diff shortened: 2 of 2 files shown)


Dockerfile.gpu
```diff
@@ -0,0 +1,17 @@
...
+
+RUN apt-get update && apt-get install -y --no-install-recommends \
+    ca-certificates \
+    build-essential \
+    git \
+    python \
+    python-pip \
+    python-setuptools
+    
+RUN pip install tensorflow-gpu
```
karmel: The models now run off of the nightly builds-- tf-nightly and tf-nightly-gpu.

Dockerfile.cpu
```diff
@@ -0,0 +1,17 @@
+# Docker image for running examples in Tensorflow models.
```
karmel: I don't see in Github's interface right now where these files will be placed in the repo-- is that because they are at the top level? If so, can we nest these somewhere under official/testing, maybe with an appropriate subdir there as well?

vishh: Yes, it is at the top level, which is a common practice for defining container images.
Is there a reason you want it to under `testing`? I'd expect users to use these images even outside of testing as long as the model is stable and well tested.

Dockerfile.gpu
vishh: It will be useful to pin specific releases of models repo to tf releases. I'll update this PR to point to nightly.

Dockerfile.cpu
karmel: "As long as the model is stable and well tested" is an unsupported claim in this repo under the Research directory. Which means that, for example, research models may require other builds of TF, or released TF, etc. I would generally prefer to keep these under Official, to avoid the impression that we are supporting these for the research models. As to in the testing dir, more stylistic-- I like things to be contained and orderly ;) Especially if there will be many versions of these (ie, GPU/CPU/multi-GPU, nightly/stable, etc.

vishh:
@karmel PTAL

karmel:
Files themselves look good-- how do you feel about moving under official?

vishh:
@karmel moved to under `official` directory.

karmel:
Thanks! cc @qlzh727 , who is starting to look at convergence testing.

vilmara:
Hi all, is this Dockerfile.gpu approved?, If so, how can I pull its image?

PR #3400, opened 2018-02-16
MNIST: Eager execution.

asimshankar:
Add an example showing how to train the MNIST model with eager execution
enabled.

(This change requires changes to TensorFlow made after
the 1.6 release branch was cut, i.e., will require a build
from source or TensorFlow 1.7+)

FYI @jhseu @guptapriya @fchollet

### DIFF
--- a/official/mnist/mnist.py
+++ b/official/mnist/mnist.py
@@ -26,14 +26,14 @@ import dataset
 
 
 class Model(tf.keras.Model):
-  """Class that defines a graph to recognize digits in the MNIST dataset.
+  """Model to recognize digits in the MNIST dataset.
 
   Network structure is equivalent to:
   https://github.com/tensorflow/tensorflow/blob/r1.5/tensorflow/examples/tutorials/mnist/mnist_deep.py
   and
   https://github.com/tensorflow/models/blob/master/tutorials/image/mnist/convolutional.py
 
-  But written as a tf.keras and tf.layers APIs.
+  But written as a tf.keras.Model using the tf.layers API.
   """
 
   def __init__(self, data_format):
--- a/official/mnist/mnist.py
+++ b/official/mnist/mnist.py
@@ -25,8 +25,16 @@ import tensorflow as tf
 import dataset
 
 
-class Model(object):
-  """Class that defines a graph to recognize digits in the MNIST dataset."""
+class Model(tf.keras.Model):
+  """Class that defines a graph to recognize digits in the MNIST dataset.
+
+  Network structure is equivalent to:
+  https://github.com/tensorflow/tensorflow/blob/r1.5/tensorflow/examples/tutorials/mnist/mnist_deep.py
+  and
+  https://github.com/tensorflow/models/blob/master/tutorials/image/mnist/convolutional.py
+
+  But written as a tf.keras and tf.layers APIs.
+  """
 
   def __init__(self, data_format):
     """Creates a model for classifying a hand-written digit.
@@ -37,6 +45,7 @@ class Model(object):
         typically faster on CPUs. See
         https://www.tensorflow.org/performance/performance_guide#data_formats
     """
+    super(Model, self).__init__()
     if data_format == 'channels_first':
       self._input_shape = [-1, 1, 28, 28]
     else:
--- a/official/mnist/mnist_eager.py
+++ b/official/mnist/mnist_eager.py
@@ -61,6 +61,9 @@ def train(model, optimizer, dataset, log_interval=None):
   start = time.time()
   for (batch, (images, labels)) in enumerate(tfe.Iterator(dataset)):
     with tf.contrib.summary.record_summaries_every_n_global_steps(10):
+      # Record the operations used to compute the loss given the input,
+      # so that the gradient of the loss with respect to the variables
+      # can be computed.
       with tfe.GradientTape() as tape:
         logits = model(images, training=True)
         loss_value = loss(logits, labels)
@@ -111,6 +114,9 @@ def main(_):
   optimizer = tf.train.MomentumOptimizer(FLAGS.lr, FLAGS.momentum)
 
   if FLAGS.output_dir:
+    # Create directories to which summaries will be written
+    # tensorboard --logdir=<output_dir>
+    # can then be used to see the recorded summaries.
     train_dir = os.path.join(FLAGS.output_dir, 'train')
     test_dir = os.path.join(FLAGS.output_dir, 'eval')
     tf.gfile.MakeDirs(FLAGS.output_dir)
--- a/official/mnist/mnist_eager.py
+++ b/official/mnist/mnist_eager.py
@@ -129,6 +129,7 @@ def main(_):
       test_dir, flush_millis=10000, name='test')
   checkpoint_prefix = os.path.join(FLAGS.checkpoint_dir, 'ckpt')
 
+  # Train and evaluate for 11 epochs.
   with tf.device(device):
     for epoch in range(1, 11):
       with tfe.restore_variables_on_create(
--- a/official/mnist/mnist_eager.py
+++ b/official/mnist/mnist_eager.py
@@ -1,0 +1,193 @@
+# Copyright 2018 The TensorFlow Authors. All Rights Reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+# ==============================================================================
+"""MNIST model training with TensorFlow eager execution.
+
+See:
+https://research.googleblog.com/2017/10/eager-execution-imperative-define-by.html
+
+This program demonstrates training of the convolutional neural network model
+defined in mnist.py with eager execution enabled.
+
+If you are not interested in eager execution, you should ignore this file.
+"""
+from __future__ import absolute_import
+from __future__ import division
+from __future__ import print_function
+
+import argparse
+import os
+import sys
+import time
+
+import tensorflow as tf
+import tensorflow.contrib.eager as tfe
+import mnist
+import dataset
+
+FLAGS = None
+
+
+def loss(logits, labels):
+  return tf.reduce_mean(
+      tf.nn.sparse_softmax_cross_entropy_with_logits(
+          logits=logits, labels=labels))
+
+
+def compute_accuracy(logits, labels):
+  predictions = tf.argmax(logits, axis=1, output_type=tf.int64)
+  labels = tf.cast(labels, tf.int64)
+  batch_size = int(logits.shape[0])
+  return tf.reduce_sum(
+      tf.cast(tf.equal(predictions, labels), dtype=tf.float32)) / batch_size
+
+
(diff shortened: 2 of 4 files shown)


official/mnist/mnist.py
```diff
@@ -25,8 +25,16 @@
...
-  """Class that defines a graph to recognize digits in the MNIST dataset."""
+class Model(tf.keras.Model):
+  """Class that defines a graph to recognize digits in the MNIST dataset.
+
+  Network structure is equivalent to:
+  https://github.com/tensorflow/tensorflow/blob/r1.5/tensorflow/examples/tutorials/mnist/mnist_deep.py
+  and
+  https://github.com/tensorflow/models/blob/master/tutorials/image/mnist/convolutional.py
+
+  But written as a tf.keras and tf.layers APIs.
```
allenlavoie: "as a tf.keras.Model and using the tf.layers API"?

official/mnist/mnist.py
```diff
@@ -25,8 +25,16 @@
 import dataset
 
 
-class Model(object):
-  """Class that defines a graph to recognize digits in the MNIST dataset."""
+class Model(tf.keras.Model):
+  """Class that defines a graph to recognize digits in the MNIST dataset.
```
allenlavoie: "that defines a model"?

official/mnist/mnist_eager.py
```diff
@@ -0,0 +1,193 @@
...
+
+  (device, data_format) = ('/gpu:0', 'channels_first')
+  if FLAGS.no_gpu or tfe.num_gpus() <= 0:
+    (device, data_format) = ('/cpu:0', 'channels_last')
+  print('Using device %s, and data format %s.' % (device, data_format))
+
+  # Load the datasets
+  train_ds = dataset.train(FLAGS.data_dir).shuffle(60000).batch(
+      FLAGS.batch_size)
+  test_ds = dataset.test(FLAGS.data_dir).batch(10000)
```
allenlavoie: 10000 seems a bit aggressive for a hard-coded default. Does training take 100x the memory?

official/mnist/mnist.py

official/mnist/mnist.py

official/mnist/mnist_eager.py
asimshankar: Nope, just that the dataset is small enough to put it all in one batch :)
Happy to change to any number you recommend.

asimshankar: Actually, just changed to use FLAGS.batch_size
(I recall someone having trouble running the MNIST graph example on a GPU with less memory, and there is no harm in reducing the batch size)

official/mnist/mnist_eager.py
```diff
@@ -0,0 +1,193 @@
...
+
+def train(model, optimizer, dataset, log_interval=None):
+  """Trains model on `dataset` using `optimizer`."""
+
+  global_step = tf.train.get_or_create_global_step()
+
+  start = time.time()
+  for (batch, (images, labels)) in enumerate(tfe.Iterator(dataset)):
+    with tf.contrib.summary.record_summaries_every_n_global_steps(10):
+      with tfe.GradientTape() as tape:
```
k-w-w: Can you add a comment describing what `tfe.GradientTape()` is used for/why it's needed here?

official/mnist/mnist_eager.py
```diff
@@ -0,0 +1,193 @@
...
+          zip(grads, model.variables), global_step=global_step)
+      if log_interval and batch % log_interval == 0:
+        rate = log_interval / (time.time() - start)
+        print('Step #%d\tLoss: %.6f (%d steps/sec)' % (batch, loss_value, rate))
+        start = time.time()
+
+
+def test(model, dataset):
+  """Perform an evaluation of `model` on the examples from `dataset`."""
+  avg_loss = tfe.metrics.Mean('loss')
```
k-w-w: Just curious, is there a Eager guide for when to use tfe vs tf? (for example, `tfe.metrics` has a tfe version, but for `tf.train.MomentumOptimizer` it's ok to use tf)

official/mnist/mnist_eager.py
```diff
@@ -0,0 +1,193 @@
...
+  # Load the datasets
+  train_ds = dataset.train(FLAGS.data_dir).shuffle(60000).batch(
+      FLAGS.batch_size)
+  test_ds = dataset.test(FLAGS.data_dir).batch(FLAGS.batch_size)
+
+  # Create the model and optimizer
+  model = mnist.Model(data_format)
+  optimizer = tf.train.MomentumOptimizer(FLAGS.lr, FLAGS.momentum)
+
+  if FLAGS.output_dir:
```
k-w-w: A brief comment about this code block and the next (at line 126) would be useful

official/mnist/mnist.py
```diff
@@ -25,8 +25,16 @@
 import dataset
 
 
-class Model(object):
-  """Class that defines a graph to recognize digits in the MNIST dataset."""
+class Model(tf.keras.Model):
```
k-w-w: Should the layers be left as `tf.layers` or changed to `tf.keras.layers`? 

Also, it would be useful to point out somewhere (perhaps in this docstring or in the README) about the advantages of inheriting from `tf.keras.Model`.

official/mnist/mnist_eager.py
asimshankar: Done.

official/mnist/mnist_eager.py
asimshankar: Not really, the general rule is that if there is a `tfe` equivalent, then use that (since it will work for both graph construction and eager execution). But if not, the `tf` symbol will work. A vast majority of the symbols in `tf` work for eager execution.

official/mnist/mnist_eager.py
asimshankar: Done

official/mnist/mnist.py
asimshankar: The two are equivalent for now, so left as `tf.layers` as it is shorter. But yeah, could be `tf.keras.layers` as well. Leaving as is for now, but let me know if you feel strongly (and we might end up changing them later anyway).

The benefits of using `keras.Model` here is that you'll get the ability to introspect the structure of your model (e.g., list all variables used by all layers, get access to all layers etc.). However, not sure if that is a point worth making in the docstring/README (especially if more models start following this pattern)?

k-w-w: I think the fewer characters the better, so lets leave it the way it is now. 

That is a good point the users should know, as the benefits aren't immediately obvious (especially for users new to Keras). You're right that the docstring and README isn't the best way to put it, since more models will use it. Perhaps somewhere in the models/official repo we should have a document describing what tensorflow "best practices" are. The point about `keras.Model` can go there. What do you think? (I can get started on working on this)

asimshankar: Documenting best practices sounds like a good idea, but let's wait till we agree on (and follow) those practices for all the official models first :)

asimshankar:
Thanks for the speedy review.

asimshankar:
(The test failures are because of some infrastructure issues, the tests pass locally (for both Py2 and Py3)).

official/mnist/mnist.py
sbl: Hi, I'm sorry if this is the wrong place to post this question (but it would neither fit a stack overflow question nor a bug ticket…). Please point me into a more appropriate direction if necessary…

I've been looking myself into what could be "best practices" to build models that can be re-used between eager mode and "classic" tensorflow based estimators.

It seems subclassing `tf.keras.Model` is used here and also in the main tensorflow repo https://github.com/tensorflow/tensorflow/blob/master/tensorflow/contrib/eager/python/examples/linear_regression/linear_regression.py#L36 , suggesting this could become the standard or recommended way of doing things. 
From a passing look here, there is no other `tf.keras` specific code in the provided example. Indeed we are actually creating a "vanilla" tensorflow estimator from the provided model.

The differences between the `tf.keras` API and "vanilla" `tf` are becoming less obvious (`tf.layers` provides most of the high level functionality you'd expect to build a net fluently) and the cognitive overhead between switching and searching documentation "how to do things" between the different tensorflow APIs can become quite high. I observe this esp. for newcomers or beginners to the framework.
Having some insights into what is the reasonings between choosing `tf.keras` over "vanilla" tensorflow (esp. that `tf.estimator` is [actively advertised](https://www.tensorflow.org/programmers_guide/estimators) as the way to build things would help us also adjust for future direction in our development practice.

So now the "simple" question: Why is the `Model` living under the `tf.keras` namespace?

official/mnist/mnist.py
```diff
@@ -37,6 +45,7 @@ def __init__(self, data_format):
         typically faster on CPUs. See
         https://www.tensorflow.org/performance/performance_guide#data_formats
     """
+    super(Model, self).__init__()
```
nealwu: Hey @asimshankar, this line causes an error for me when I run the model. Specifically:

```
  File "mnist/mnist.py", line 48, in __init__
    super(Model, self).__init__()
TypeError: __init__() takes at least 3 arguments (1 given)
```

Note I'm currently on TF 1.5. Is this a bug, or is it because this code only works with TF 1.6? If so we should update the README at https://github.com/tensorflow/models/tree/master/official.

official/mnist/mnist.py
asimshankar: I wouldn't make a distinction between "vanilla" and "keras" when using Keras through `tf.keras`; the underlying implementation for both is largely the same. For example, `tf.keras.layers.Dense` and `tf.layers.Dense` are practically interchangeable.

(Ideally, we wouldn't have two symbols for each kind of layer, but maintain both because of our API backward compatibility guarantees.)

For models without control flow, code defining a model/computation should be shareable between eager and graph.  We show how this works in the checked-in eager examples like [MNIST](https://github.com/tensorflow/models/tree/master/official/mnist); note that in [eager](https://github.com/tensorflow/models/blob/master/official/mnist/mnist_eager.py#L112) we just call `mnist.Model()` from the graph version.  The training loops different between eager and graph, however.

With the caveat that things will change a bit before settling down, the best practices for re-using code between eager execution and graph construction could be:

1. Use the object oriented layer APIs (e.g., `tf.keras.layers.Dense` or `tf.layers.Dense`) to define computation.
 
and:

2. Use `tf.keras.Model` to encapsulate models (or otherwise a custom collection of layers). 

If you find you need an Estimator (e.g., for distributed training or convenient checkpointing) then [convert the model to an estimator](https://www.tensorflow.org/programmers_guide/estimators#creating_estimators_from_keras_models).

We sincerely appreciate your input and completely acknowledge that we have more work to do on samples and documentation to continue to clarify this.  We're also working on ways to bring Estimator-like features into eager.

official/mnist/mnist.py
asimshankar: @nealwu : It will infact require version 1.7. Perhaps we should just remove the line

"If you are on a version of TensorFlow earlier than 1.4, please update your installation."
from the README? Since just before that it talks about tagged branches.

nealwu: That line is fine. I was referring to this one specifically:

"We aim to keep them backwards compatible with the latest release when possible (currently TensorFlow 1.5), but we cannot always guarantee compatibility."

If there is also a way we can keep this model backwards-compatible with 1.5/1.6 until 1.7 is the latest release, I would be strongly in favor of that.

official/mnist/mnist.py
sbl: Dear @asimshankar thanks for your response and insights here, very much appreciated. 

Let me just state the hope, that the project will manage to reduce its surface area over time (or takes a jump and introduces an API incompatible `2.x` version) - while tensorflow as a whole does evolve nicely, having so many different - in usage, idiom + spirit - python APIs (`tf.eager, tf.contrib., tf.core, tf.keras`, …) creates quite a complexity I would not underestimate…

Issue #3362, opened 2018-02-12
Documentation Request


Josh-Payne:
Docs request for the official MNIST model: you have some example test data as 2 .pngs and the representative .npy file, but no steps as to how to convert the .pngs to .npy. A script would be super helpful. Thanks so much!

cy89:
@nealwu can you please offer advice?

Josh-Payne:
If anyone else has this problem, I came up with a solution: 
`import numpy as np
from PIL import Image
from array import *

data_image = array('B')
Im = Image.open('image.png').convert('L') # convert to grayscale
pixel = Im.load()
width, height = Im.size
for x in range(0,width):
	for y in range(0,height):
		data_image.append(pixel[y,x]) # use the MNIST format

m = np.array(data_image)
img = np.reshape(m, (1, 28, 28))
img = img/float(255) # scale from [0, 255] to [0, 1]
np.save(imagenpy, img)`

Josh-Payne:
If anyone else has this problem, I came up with a solution:
`import numpy as np
from PIL import Image
from array import *

data_image = array('B')
Im = Image.open('image.png').convert('L') # convert to grayscale
pixel = Im.load()
width, height = Im.size
for x in range(0,width):
	for y in range(0,height):
		data_image.append(pixel[y,x]) # use the MNIST format

m = np.array(data_image)
img = np.reshape(m, (1, 28, 28))
img = img/float(255) # scale from [0, 255] to [0, 1]
np.save(imagenpy, img)`

Josh-Payne:
If anyone else has this problem, [npcon.txt](https://github.com/tensorflow/models/files/1733085/npcon.txt) is a solution

cy89:
Thanks, @Josh-Payne!

PR #3403, opened 2018-02-17
added script requested from #3362

Josh-Payne:
This commit adds the script requested from #3362 for converting images to MNIST-format numpy arrays and includes documentation for use. A user can now call `$ python convert_npy.py --image "example5.png example3.png" --batch True --output examples` to generate a file identical to examples.npy, though the pixel values are reversed because the MNIST numbers are white with black background. Also fixed a typo in the readme.


Josh-Payne:
I signed it!

karmel:
Thanks for the PR. A couple of high-level questions before we review in detail:

1. MNIST is a relatively constrained model, and it seems less likely that people would want to convert their own images for use with MNIST as opposed to, for example, ResNet. Before we provide and maintain a script to do so, it's important to us to assess whether this use case is frequent enough to require a custom solution outside of the guides offered on StackOverflow for converting images to arrays. Have you encountered others having the same problem? It might be better to put it on StackOverflow or a similar place for housing recipes if this isn't a frequent use case.

2. If this is a common use case, it would make sense to have a script that converts images to arrays, but this script seems fairly complex for what it does-- for example, using both Numpy and built-in arrays; iterating through each pixel of the image; having command line options to batch or not batch; etc. Can you simplify some of the code here to just take in an image and convert to a numpy array as efficiently as possible?

Josh-Payne:
Hello @karmel, thanks for the evaluation! As for the use case, my particular motivation for creating this script was to extend the model for doing OCR and recognizing geometric shapes for a UAV competition. This is one of the few models I've found out there that uses the TF.Estimator on a small dataset like the MNIST, so I wanted to use it as a template. 

If you do decide it would be useful, I've trimmed it down considerably and removed some of the command line options. Because the pixels are arranged row-wise in the MNIST, the iterators substitute for having to do a numpy.flip. Thanks again for reviewing this!

karmel:
Thanks @Josh-Payne -- that sounds like a fairly specific use case, so this is probably best housed elsewhere for now. If others end up on this page and it turns out there are many use cases, we can re-open.

As a side note, using built-in numpy functions for array manipulation are generally much faster and more efficient than iterating through an array. Might be a fun exercise to try timing this script with the for loops versus using only numpy operations to see what the delta is.

PR #3410, opened 2018-02-19
Fix PythonPath dependencies

mgruben:
Previously, code block one attempted to import `from object_detection.utils` before the `object_detection` directory was added to the PythonPath via `sys.path.append("..")`.

This generated a `ModuleNotFoundError`.

Accordingly, append `".."` to the PythonPath before attempting to import from `object_detection` as a module, instead of after.


mgruben:
I signed it!

derekjchow:
LGTM. Thank you for the patch!

Issue #3418, opened 2018-02-21
Different accuracy between NCHW-NHWC data format when finetuning from imagenet


chrisrn:
### System information
- **What is the top-level directory of the model you are using**: tensorflow/models
- **Have I written custom code (as opposed to using a stock example script provided in TensorFlow)**: yes
- **OS Platform and Distribution (e.g., Linux Ubuntu 16.04)**: Ubuntu 16.04
- **TensorFlow installed from (source or binary)**: binary
- **TensorFlow version (use command below)**: 1.4.1
- **Bazel version (if compiling from source)**: N/A
- **CUDA/cuDNN version**: 8.0
- **GPU model and memory**: GTX 1080i
- **Exact command to reproduce**: N/A

### Describe the problem
I trained resnet_v2_50 on my data set using the pre-trained model from imageNet. I made some small modifications inside the network to accept NCHW data format. When I train using NHWC order the training accuracy is lower than the NCHW model. So my question is: Are the imageNet models trained with NCHW format? If that's true then it's reasonable to have higher accuracy using NCHW, because I restore the checkpoint correctly. According to this repo, the models are using NHWC. I also opened the checkpoints using `tensorflow/python/tools/inspect_checkpoint.py` and I see that the kernels have the same sizes even using NHWC or NCHW so I can't infer something from the checkpoints.

asimshankar:
See https://www.tensorflow.org/performance/performance_guide#data_formats for some discussion on data formats. Generally, we use NCHW when training on NVIDIA GPUs with cuDNN.  The training script selects the format automatically (see [flags](https://github.com/tensorflow/models/blob/master/official/resnet/resnet.py#L606) and [code](https://github.com/tensorflow/models/blob/master/official/resnet/resnet.py#L314)) depending on whether or not a GPU is present.

If by "Are the imagenet models trained with NCHW format" you mean if the pre-trained models referred to in https://github.com/tensorflow/models/tree/master/official/resnet#pre-trained-model are trained in NCHW format, then yup, they are.

Closing this issue out since I believe your question is answered (and I'll send a PR to make this clearer in the README). Feel free to reopen if I'm mistaken. Thanks.

chrisrn:
Thank you my question is answered but you have to make clear in slim documentation that pre-trained models are in NCHW since slim models are not optimized for this kind of format.

chrisrn:
[...]

PR #3436, opened 2018-02-23
official/resnet: Clarify data format of pre-trained model.

asimshankar:
Motivated by #3418


nealwu:
I believe users should be able to just run the same model code using the checkpoint as the `model_dir` and then add the flag `--data_format=channels_last`.

In fact I just tried this myself with the current master code and current [pre-trained model](https://github.com/tensorflow/models/tree/master/official/resnet#pre-trained-model), and I'm able to continue training just fine with `--data_format=channels_last`, reaching a train accuracy of 75% after a few hundred steps.

tfboyd:
This is a general TensorFlow problem that also exists from TPU to GPU as with GPU to CPU, if you have a saved model and not the original code that was written to support both data formats you are kind of stuck. This is over a year old at this point so I am closing.  

One trick that might work and not related to this is training with NHWC which grappler will change to NCHW to get better perf but I think the saved model still ends up NHWC.  Of course if you wanted to use MKL you may want NCHW for CPU.  It seems like there needs to be a better story and that is beyond this example other than telling people to create model code that supports both formats.  :-(

PR #3440, opened 2018-02-23
Fix get_input_fn() for PREDICT mode

FarooqKhan:
During predict we should not return None tensor, doing so will cause an exception in dataset.map()
Also while extracting data from the dataset iterator we should not expect labels since they do not exist in the dataset


FarooqKhan:
I wanted to update that i was able to successfully test my code with this change.

PR #3453, opened 2018-02-24
Corrected number of shots in research/learning_to_remember_rare_events/train.py (method Trainer.run)

### LINKED ISSUE #2418: Bug in models/learning_to_remember_rare_events/
It seems like the per-shot accuracy calculation in `models/learning_to_remember_rare_events/train.py` is wrong: 
```correct_by_shot = dict((k, []) for k in xrange(self.episode_width + 1))```

Instead of `self.episode_width` , it should be something like `int(self.episode_length/self.episode_width*1.0)-1` since per shot means how many samples have been seen so far. 

The original code does not make sense when `episode_length` is 60 and `episode_width` is 10 in which case we should be seeing accuracies for 0-shot,1-shot,2-shot,3-shot,4-shot,5-shot. With the fix above, this seems to work fine.

MTDzi:
This fixes #2418, as proposed by @himani-arora.

Please let me know (@himani-arora, @ofirnachum, @lukaszkaiser) if this is alright, and if you would like me to add anything more.

### DIFF
--- a/research/learning_to_remember_rare_events/train.py
+++ b/research/learning_to_remember_rare_events/train.py
@@ -197,7 +197,8 @@ class Trainer(object):
 
         # validation
         correct = []
-        correct_by_shot = dict((k, []) for k in xrange(self.episode_width + 1))
+        num_shots = episode_length // episode_width
+        correct_by_shot = dict((k, []) for k in xrange(num_shots))
         for _ in xrange(FLAGS.validation_length):
           x, y = self.sample_episode_batch(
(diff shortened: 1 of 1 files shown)


MTDzi:
I signed the CLA.

MTDzi:
Hey,
I've noticed that my user name and email in git were wrong (and different than the email for my github account), so this seems to be a problem in terms of signing the CLA.

Should I delete this PR and create another one with a matching user name and email? Or is there is a way around it?

PR #3472, opened 2018-02-26
Multi-GPU Resnet

karmel:
Performance on this is still less than expected, but I want to push this in to unblock several other changes. 

Notes:
* I removed the image conversion to a float in the range [0, 1) because it did not affect convergence, but negatively impacted performance. 
* The num_gpus flag is included for @tfboyd's use currently; if we decide there is a better way for his testing purposes, that can be removed (in a follow-up if timing requires).
* This duplicates some of the code in MNIST; we will handle that in follow-up PRs as we introduce some shared functions.

CC @isaprykin @reedwm @tfboyd


official/resnet/imagenet_main.py
```diff
@@ -102,10 +102,9 @@ def parse_record(raw_record, is_training):
   # Note that the resulting image contains an unknown height and width
   # that is set dynamically by decode_jpeg. In other words, the height
   # and width of image is unknown at compile-time.
-  # Results in a 3-D int8 Tensor which we then convert to a float
-  # with values ranging from [0, 1).
+  # Results in a 3-D int8 Tensor. This will be converted to a float later,
+  # during resizing.
   image = tf.image.decode_jpeg(image, channels=_NUM_CHANNELS)
```
k-w-w: Are the numbers scaled to between [0,1] later on?

karmel: As far as I can tell, no, though it doesn't affect convergence in our tests.

k-w-w: Hmm, that's good to hear. I would still consider dividing the image values by 255, because during preprocessing the means subtracted from the image are scaled to [0,1].

official/resnet/imagenet_main.py
```diff
@@ -144,6 +143,12 @@ def input_fn(is_training, data_dir, batch_size, num_epochs=1,
   # Convert to individual records
   dataset = dataset.flat_map(tf.data.TFRecordDataset)
 
+  # Currently, if we are using multiple GPUs, we can't pass in uneven batches.
+  # This ensures that we cut off the uneven images at the end of each epoch
+  # (there are fifteen leftovers), rather than trying and failing to split
```
asimshankar: What does "there are fifteen leftovers" mean? Perhaps say ("e.g. there are 15 leftovers when the batch size is 32").

That said though, some thoughts:

- Why doesn't `cifar10_main.py` need the same truncation logic?

- This silent truncation of the dataset is subtle and may be worth bubbling up to the caller (e.g., by adding an explicit `truncate` argument to `input_fn`).

- (Non-actionable note): `resnet.process_record_dataset` seems to `batch` after `repeat(num_epochs)`, which means that one could have avoided the truncation (or could have truncated less) depending on what `num_epochs` is. For example, for `num_epochs = 15`, no truncation is needed :). I point this out not because I think those 15 examples are particularly important, but just that these interactions are quite subtle and add another source of difference. For example, the comparsion between a batch size of 128 and a batch size of 1024 may also be influenced by the difference in the ~130 examples seen by one and not the other?

official/resnet/resnet.py
```diff
@@ -446,6 +447,9 @@ def resnet_model_fn(features, labels, mode, model_class,
       True if the var should be included in loss calculation, and False
       otherwise. If None, batch_normalization variables will be excluded
       from the loss.
+    multi_gpu: If True, wrap the optimizer in a TowerOptimizer to allow
+      for running on multiple GPUs.
```
asimshankar: Perhaps elaborate on what "allow for running" means, for example:

```
If True, wrap the optimizer in a TowerOptimizer suitable for data-parallel distribution across multiple GPUs.
```

official/resnet/resnet.py
```diff
@@ -608,3 +676,13 @@ def __init__(self, resnet_size_choices=None):
...
              'the data format will be chosen automatically based on '
              'whether TensorFlow was built for CPU or GPU.')
+
+    self.add_argument(
+        '--multi_gpu', action='store_true',
+        help='If set, run across all available GPUs. Note that this is '
+        'superseded by the --num_gpus flag.')
+
+    self.add_argument(
+        '--num_gpus', type=int, default=None,
```
asimshankar: Why have two flags, seems we could do with just one?

official/resnet/imagenet_main.py
karmel: Yeah, this is not my favorite line. Right now, what happens is there are 1281167 training images; if you run with one epoch, you get 1281167/32 batches, which comes to ~40K batches with 15 leftover. Running on one GPU, that doesn't matter. But with the current replicate_model_fn, everything breaks at the end of the epoch when the split process can't divide 15 examples evenly over 4 or 8 GPUs. This is the current solution @isaprykin suggested, pending a decision for replicate_model_fn more generally.

To your specific points--
1. On the one hand, yes, it should be in cifar10 if we want to run that multi-GPU in the same way. However, because cifar happens to have a nicely divisible number of images, it's not practically a likely issue. I therefore wanted to err on the side of not including a hacky line for a case that wasn't likely to arise, but maybe it's better to just have in both.

2. Or maybe just do it for multi-GPU for now? It seems like unnecessary complication to do that for the general case when it is a half-bandaid for now. The reason I didn't do that is that right now that function is not multi-GPU aware, and I hate to introduce a third place where we have to do special things for multi-GPU.

3. Throwing that one back to @isaprykin , who is deciding what the solution should be more generally :)

Thoughts?

official/resnet/imagenet_main.py
karmel: Interesting. That implies that maybe it is scaled later on? Let me go through again and double-check.

official/resnet/resnet.py
karmel: CC @tfboyd who requested the second one. In Toby's version, there num_gpus takes many values representing CPU, num gpu, and all GPU, but I found the multi-state indicator confusing. I thought it better to have a simple flag that people can use to mean "use all," and a expert-use-case flag for specifying the exact number. @tfboyd , thoughts on this? Especially versus using CUDA_VISIBLE_DEVICES?

official/resnet/imagenet_main.py
karmel: It looks like the means were still being scaled, but not the final values-- implying mean subtraction doesn't matter a ton? Is that plausible? I removed the scaling from the means, and am running now to validate convergence.

official/resnet/imagenet_main.py
asimshankar: 2. sounds good.

Thanks for the details. (Perhaps worth capturing this in a comment? Your call)

official/resnet/resnet.py
asimshankar: Yup, CUDA_VISIBLE_DEVICES seems reasonable to me, and then just a single `--multi-gpu=<boolean>` flag.

karmel: There was some reason @tfboyd was mildly opposed to that solution... @tfboyd , thoughts?

official/resnet/resnet.py
```diff
@@ -78,6 +86,18 @@ def process_record_dataset(dataset, is_training, batch_size, shuffle_buffer,
...
+  # must be divisible by 4.) We already ensured this for the batch_size, but
+  # we have to additionally ensure that any "leftover" examples-- the remainder
+  # examples (total examples % batch_size) that get called a batch for the very
+  # last batch of an epoch-- do not raise an error when we try to split them
+  # over the GPUs. This will likely be handled by Estimator during replication
+  # in the future, but for now, we just drop the leftovers here.
+  if multi_gpu:
+    total_examples = num_epochs * examples_per_epoch
+    dataset = dataset.take(batch_size * (total_examples // batch_size))
+
```
karmel: @asimshankar @isaprykin Moved here. Note that there is now a lot of code introduced to make this happen. Igor, give us a better way!

isaprykin: Okay, so my latest idea is to require you to return Dataset from the input_fn (see the design doc).
Then I'm going to call get_next() for every GPU.  There won't be a divisibility problem then.

Sorry!  :-)

official/resnet/resnet.py
karmel: Removed for now as it's not going to be in the final version of replication mode as things currently stand anyways. We can reassess if that causes undue pain for you, @tfboyd

karmel:
This converges. Merging.
screen shot 2018-03-01 at 7 51 02 am

PR #1052, opened 2017-02-22
ENH: Updated autoencoder to be of any given depth

swaroopgj:
Autoencoder was hardcoded to have only one hidden layer. This PR updates the implementation to be of arbitrary depth based on input parameters. Runner is update accordingly.

PR #3477, opened 2018-02-27
autoencoder: bugfix in initialization

swaroopgj:
Bugfix for initialization in autoencoder, thanks to @cclauss for raising the issue in #1052 
Verified it runs.


karmel:
@snurkabill Can you take a look?

snurkabill:
This needs to be merged otherwise the code is not runnable. @karmel

Issue #3476, opened 2018-02-27
when using train.py, getting error TypeError: __init__() got an unexpected keyword argument 'dct_method'


MicrocontrollersAndMore:
What is the top-level directory of the model you are using:
I'm not sure I understand this question entirely, but the file I'm having a problem with is https://github.com/tensorflow/models/blob/master/research/object_detection/data_decoders/tf_example_decoder.py

Have I written custom code (as opposed to using a stock example script provided in TensorFlow):
No, the concern exists using the stock file https://github.com/tensorflow/models/blob/master/research/object_detection/train.py

OS Platform and Distribution:
Window 10

TensorFlow installed from (source or binary):
binary via pip3

TensorFlow version (use command below):
tensorflow (1.5.0)

Bazel version (if compiling from source):
not compiling from source

CUDA/cuDNN version:
using CPU

GPU model and memory:
using CPU

Exact command to reproduce:

If I perform the usual pre-steps to using https://github.com/tensorflow/models/blob/master/research/object_detection/train.py , i.e. set up a pipline config file, download a model (I'm using Inception v2), and then run train.py per the examply usage on line 24 https://github.com/tensorflow/models/blob/master/research/object_detection/train.py#L24 then I get the following error:

`TypeError: __init__() got an unexpected keyword argument 'dct_method'`

Here is a screen shot if that helps:

![error1](https://user-images.githubusercontent.com/5672876/36705830-04abbb3c-1b1c-11e8-8b2e-1e1750b5b9fb.jpg)

This worked well using the same steps a few weeks ago, so I checked the commit history of https://github.com/tensorflow/models/blob/master/research/object_detection/data_decoders/tf_example_decoder.py and found that the dct_method=dct_method in the above screenshot was added recently:

![file_change](https://user-images.githubusercontent.com/5672876/36706050-fb0c93fc-1b1c-11e8-8dfa-b1b97c33f879.jpg)

it seems in https://github.com/tensorflow/models/blob/master/research/object_detection/data_decoders/tf_example_decoder.py this:

        fields.InputDataFields.image: slim_example_decoder.Image(
            image_key='image/encoded', format_key='image/format', channels=3),

was changed to:

        fields.InputDataFields.image:
            slim_example_decoder.Image(
                image_key='image/encoded',
                format_key='image/format',
                channels=3,
                dct_method=dct_method),

And now I'm getting the above error.  If I simply remove the dct_method=dct_method and make no other changes the training process completes well, as it did before:

[...]

PR #3480, opened 2018-02-27
removed dct_method=dct_method from line 110 as that causes an error w…

MicrocontrollersAndMore:
…hen running train.py

In this issue:

https://github.com/tensorflow/models/issues/3476

I've found that the change in this file from

    fields.InputDataFields.image: slim_example_decoder.Image(
        image_key='image/encoded', format_key='image/format', channels=3),

to:

    fields.InputDataFields.image:
        slim_example_decoder.Image(
            image_key='image/encoded',
            format_key='image/format',
            channels=3,
            dct_method=dct_method),

causes an error when running https://github.com/tensorflow/models/blob/master/research/object_detection/train.py

I think this should be changed back as above.  I should mention in issue 3476 another poster verified experiencing the same concern and verified the same fix, thanks!

### DIFF
--- a/research/object_detection/data_decoders/tf_example_decoder.py
+++ b/research/object_detection/data_decoders/tf_example_decoder.py
@@ -106,8 +106,7 @@ class TfExampleDecoder(data_decoder.DataDecoder):
             slim_example_decoder.Image(
                 image_key='image/encoded',
                 format_key='image/format',
-                channels=3,
-                dct_method=dct_method),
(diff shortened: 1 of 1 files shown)


MicrocontrollersAndMore:
I just digitally signed the Individual Contributor License Agreement.  I've never done a pull request on GitHub before on somebody else's repo, please let me know if I didn't do something correctly procedure-wise, thanks!

PR #3484, opened 2018-02-27
Allow the official resnet model to train as resnet_v1 and resnet_v2

robieta:
Previously the official ResNet provided was the "v2" version from He, et al., (2016). However the "v1" described in He, et al., (2015) is also very popular. This PR allows the official ResNet model to be run using either the v1 or v2 conventions.

### DIFF
--- a/official/resnet/cifar10_main.py
+++ b/official/resnet/cifar10_main.py
@@ -140,7 +140,8 @@ def get_synth_input_fn():
 ###############################################################################
 class Cifar10Model(resnet.Model):
 
-  def __init__(self, resnet_size, data_format=None, num_classes=_NUM_CLASSES):
+  def __init__(self, resnet_size, data_format=None, num_classes=_NUM_CLASSES,
+      version=resnet.DEFAULT_VERSION):
     """These are the parameters that work for CIFAR-10 data.
 
     Args:
@@ -149,6 +150,8 @@ class Cifar10Model(resnet.Model):
         data format to use when setting up the model.
       num_classes: The number of output classes needed from the model. This
         enables users to extend the same model to their own datasets.
+      version: Integer representing which version of the ResNet network to use.
+        See README for details. Valid values: [1, 2]
     """
     if resnet_size % 6 != 2:
       raise ValueError('resnet_size must be 6n + 2:', resnet_size)
@@ -157,6 +160,7 @@ class Cifar10Model(resnet.Model):
 
     super(Cifar10Model, self).__init__(
         resnet_size=resnet_size,
+        bottleneck=False,
         num_classes=num_classes,
         num_filters=16,
         kernel_size=3,
@@ -165,10 +169,10 @@ class Cifar10Model(resnet.Model):
         first_pool_stride=None,
         second_pool_size=8,
         second_pool_stride=1,
-        block_fn=resnet.building_block,
         block_sizes=[num_blocks] * 3,
         block_strides=[1, 2, 2],
         final_size=64,
+        version=version,
         data_format=data_format)
 
 
@@ -199,6 +203,7 @@ def cifar10_model_fn(features, labels, mode, params):
                                 learning_rate_fn=learning_rate_fn,
                                 momentum=0.9,
                                 data_format=params['data_format'],
+                                version=params['version'],
                                 loss_filter_fn=loss_filter_fn,
                                 multi_gpu=params['multi_gpu'])
 
--- a/official/resnet/resnet.py
+++ b/official/resnet/resnet.py
@@ -12,15 +12,13 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 # ==============================================================================
-"""Contains definitions for the preactivation form of Residual Networks
-(also known as ResNet v2).
+"""Contains definitions for Residual Networks.
 
-Residual networks (ResNets) were originally proposed in:
+Residual networks ('v1' ResNets) were originally proposed in:
 [1] Kaiming He, Xiangyu Zhang, Shaoqing Ren, Jian Sun
     Deep Residual Learning for Image Recognition. arXiv:1512.03385
 
-The full preactivation 'v2' ResNet variant implemented in this module was
-introduced by:
+The full preactivation 'v2' ResNet variant was introduced by:
 [2] Kaiming He, Xiangyu Zhang, Shaoqing Ren, Jian Sun
     Identity Mappings in Deep Residual Networks. arXiv: 1603.05027
 
@@ -40,6 +38,8 @@ import tensorflow as tf
 
 _BATCH_NORM_DECAY = 0.997
 _BATCH_NORM_EPSILON = 1e-5
+
+DEFAULT_VERSION = 2
 
 
 ################################################################################
@@ -139,18 +139,16 @@ def get_synth_input_fn(height, width, num_channels, num_classes):
 
 
 ################################################################################
-# Functions building the ResNet model.
+# Convenience functions for building the ResNet model.
 ################################################################################
-def batch_norm_relu(inputs, training, data_format):
-  """Performs a batch normalization followed by a ReLU."""
+def batch_norm(inputs, training, data_format):
+  """Performs a batch normalization using a standard set of parameters."""
   # We set fused=True for a significant performance boost. See
   # https://www.tensorflow.org/performance/performance_guide#common_fused_ops
-  inputs = tf.layers.batch_normalization(
+  return tf.layers.batch_normalization(
       inputs=inputs, axis=1 if data_format == 'channels_first' else 3,
       momentum=_BATCH_NORM_DECAY, epsilon=_BATCH_NORM_EPSILON, center=True,
       scale=True, training=training, fused=True)
-  inputs = tf.nn.relu(inputs)
-  return inputs
 
 
 def fixed_padding(inputs, kernel_size, data_format):
@@ -194,9 +192,16 @@ def conv2d_fixed_padding(inputs, filters, kernel_size, strides, data_format):
       data_format=data_format)
 
 
-def building_block(inputs, filters, training, projection_shortcut, strides,
-                   data_format):
-  """Standard building block for residual networks with BN before convolutions.
+################################################################################
+# ResNet block definitions.
+################################################################################
+def _building_block_v1(inputs, filters, training, projection_shortcut, strides,
+    data_format):
+  """
+  Convolution then batch normalization then ReLU as described by:
+    Deep Residual Learning for Image Recognition
+    https://arxiv.org/pdf/1512.03385.pdf
+    by Kaiming He, Xiangyu Zhang, Shaoqing Ren, and Jian Sun, Dec 2015.
 
   Args:
     inputs: A tensor of size [batch, channels, height_in, width_in] or
@@ -214,34 +219,40 @@ def building_block(inputs, filters, training, projection_shortcut, strides,
     The output tensor of the block.
   """
   shortcut = inputs
-  inputs = batch_norm_relu(inputs, training, data_format)
 
-  # The projection shortcut should come after the first batch norm and ReLU
-  # since it performs a 1x1 convolution.
   if projection_shortcut is not None:
     shortcut = projection_shortcut(inputs)
+    shortcut = batch_norm(inputs=shortcut, training=training,
+                          data_format=data_format)
 
   inputs = conv2d_fixed_padding(
       inputs=inputs, filters=filters, kernel_size=3, strides=strides,
       data_format=data_format)
+  inputs = batch_norm(inputs, training, data_format)
+  inputs = tf.nn.relu(inputs)
 
-  inputs = batch_norm_relu(inputs, training, data_format)
   inputs = conv2d_fixed_padding(
       inputs=inputs, filters=filters, kernel_size=3, strides=1,
       data_format=data_format)
+  inputs = batch_norm(inputs, training, data_format)
+  inputs += shortcut
+  inputs = tf.nn.relu(inputs)
 
-  return inputs + shortcut
+  return inputs
 
 
-def bottleneck_block(inputs, filters, training, projection_shortcut,
-                     strides, data_format):
-  """Bottleneck block variant for residual networks with BN before convolutions.
+def _building_block_v2(inputs, filters, training, projection_shortcut, strides,
+    data_format):
+  """
+  Batch normalization then ReLu then convolution as described by:
+    Identity Mappings in Deep Residual Networks
+    https://arxiv.org/pdf/1603.05027.pdf
+    by Kaiming He, Xiangyu Zhang, Shaoqing Ren, and Jian Sun, Jul 2016.
 
   Args:
     inputs: A tensor of size [batch, channels, height_in, width_in] or
       [batch, height_in, width_in, channels] depending on data_format.
-    filters: The number of filters for the first two convolutions. Note
-      that the third and final convolution will use 4 times as many filters.
+    filters: The number of filters for the convolutions.
     training: A Boolean for whether the model is in training or inference
       mode. Needed for batch normalization.
     projection_shortcut: The function to use for projection shortcuts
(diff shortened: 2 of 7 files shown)


official/resnet/cifar10_main.py
```diff
@@ -189,7 +193,8 @@ def loss_filter_fn(name):
                                 learning_rate_fn=learning_rate_fn,
                                 momentum=0.9,
                                 data_format=params['data_format'],
-                                loss_filter_fn=loss_filter_fn)
+                                loss_filter_fn=loss_filter_fn,
+                                version=params['version'])
```
karmel: Word of warning: this is going to conflict with #3472 . Won't be too hard to resolve, I think, but... be prepared.

official/resnet/resnet.py
```diff
@@ -164,10 +168,59 @@ def building_block(inputs, filters, training, projection_shortcut, strides,
     strides: The block's stride. If greater than 1, this block will ultimately
       downsample the input.
     data_format: The input format ('channels_last' or 'channels_first').
+    version: The ordering convention for convolution, batch normalization, and
+             activation.
 
   Returns:
     The output tensor of the block.
   """
+  if version == 1:
```
karmel: I feel like if we are going to have a model class, we may as well use it to store state. Is it feasible to limit the use of version to the Model.__init__? For example, 

```
if version == 1:
  self.building_block_fn = _building_block_v1
elif version == 2:
  self.building_block_fn = _building_block_v2
```
and then just reference self.building_block_fn when the time comes?

official/resnet/resnet.py
```diff
@@ -164,10 +168,59 @@ def building_block(inputs, filters, training, projection_shortcut, strides,
...
+    return _building_block_v1(inputs=inputs, filters=filters, training=training,
+                              projection_shortcut=projection_shortcut,
+                              strides=strides, data_format=data_format)
+  if version == 2:
+    return _building_block_v2(inputs=inputs, filters=filters, training=training,
+                              projection_shortcut=projection_shortcut,
+                              strides=strides, data_format=data_format)
+  raise ValueError("version should be 1 or 2")
```
karmel: THis check can also move to the Model.__init__. May as well be more specific/properly capitalized, too.

official/resnet/resnet.py
```diff
@@ -204,10 +257,54 @@ def bottleneck_block(inputs, filters, training, projection_shortcut,
...
+    return _bottleneck_block_v1(inputs=inputs, filters=filters,
+                                training=training,
+                                projection_shortcut=projection_shortcut,
+                                strides=strides, data_format=data_format)
+  if version == 2:
+    return _bottleneck_block_v2(inputs=inputs, filters=filters,
+                                training=training,
+                                projection_shortcut=projection_shortcut,
+                                strides=strides, data_format=data_format)
+  raise ValueError("version should be 1 or 2")
```
karmel: Ditto the above.

official/resnet/resnet.py
```diff
@@ -279,7 +379,7 @@ class Model(object):
   def __init__(self, resnet_size, num_classes, num_filters, kernel_size,
                conv_stride, first_pool_size, first_pool_stride,
                second_pool_size, second_pool_stride, block_fn, block_sizes,
-               block_strides, final_size, data_format=None):
+               block_strides, final_size, version, data_format=None):
```
karmel: Version may as well take a default value here of _DEFAULT_VERSION

official/resnet/resnet.py
```diff
@@ -264,10 +363,11 @@ def projection_shortcut(inputs):
 
   # Only the first block per block_layer uses projection_shortcut and strides
   inputs = block_fn(inputs, filters, training, projection_shortcut, strides,
-                    data_format)
+                    data_format, version=version)
```
karmel: Again, not sure if it's feasible, but not having to pass the version all the way down would keep things cleaner, it seems.

official/resnet/resnet.py
```diff
@@ -454,7 +555,7 @@ def resnet_model_fn(features, labels, mode, model_class,
   # Generate a summary node for the images
   tf.summary.image('images', features, max_outputs=6)
 
-  model = model_class(resnet_size, data_format)
+  model = model_class(resnet_size, data_format, version=version)
```
karmel: I think the ordering here is inconsistent with the order of args in init? Oh wait, the child classes have data_format and version in a different order than the parent. That seems confusing-- can we standardize the ordering?

official/resnet/resnet.py
```diff
@@ -608,3 +710,9 @@ def __init__(self, resnet_size_choices=None):
              'is not always compatible with CPU. If left unspecified, '
              'the data format will be chosen automatically based on '
              'whether TensorFlow was built for CPU or GPU.')
+
+    self.add_argument(
+        '-v', '--version', type=int, choices=[1, 2], dest="version",
+        metavar="", default=_DEFAULT_VERSION,
```
karmel: What is metavar? Why have it if blank?

official/resnet/resnet.py
```diff
@@ -608,3 +710,9 @@ def __init__(self, resnet_size_choices=None):
...
+        help="Version of resnet. (1 or 2)"
```
karmel: Please describe more fully. Ditto for the comments above-- somewhere, we will need a more full description of what the differences are. Maybe just citations here, and further details in the README?

official/resnet/resnet.py
robieta: metavar is how argparse displays the variable. For whatever reason, when you have multiple spellings of a flag with choices set, rather than display it in -h as:

`-v, --version {1, 2}`

it is displayed as:
`-v {1, 2}, --version {1, 2}`

which is ugly. Setting metavar="" suppresses that behavior.

official/resnet/resnet.py
robieta: If memory serves the order got weird because python doesn't allow args without defaults to come after those with. (I think that's the reason I eventually set the version=None default.) I'll see what I can do to make everything cleaner.

official/resnet/resnet.py
robieta: I moved the block_fn selection logic into the init. It's cleaner, but it will break things that inherit the resnet model. (Not sure if we make any interface promises on these classes.)

official/resnet/resnet.py
robieta: Ditto.

official/resnet/resnet.py
robieta: Dit... you get the idea.

official/resnet/resnet.py
robieta: Done. I made the version non-private so cifar10 and imagenet classes could share the default.

official/resnet/resnet.py
robieta: Indeed, this was a side effect of moving block_fn selection into the init.

official/resnet/resnet.py
```diff
@@ -169,48 +175,122 @@ def building_block(inputs, filters, training, projection_shortcut, strides,
...
-      that the third and final convolution will use 4 times as many filters.
+    filters: The number of filters for the convolutions.
     training: A Boolean for whether the model is in training or inference
       mode. Needed for batch normalization.
     projection_shortcut: The function to use for projection shortcuts
       (typically a 1x1 convolution when downsampling the input).
     strides: The block's stride. If greater than 1, this block will ultimately
       downsample the input.
-    data_format: The input format ('channels_last' or 'channels_first').
+    data_format: The input format ('channels_last' or 'channels_first')..
```
nealwu: Double period here

official/resnet/resnet.py
```diff
@@ -169,48 +175,122 @@ def building_block(inputs, filters, training, projection_shortcut, strides,
     The output tensor of the block.
   """
   shortcut = inputs
-  inputs = batch_norm_relu(inputs, training, data_format)
 
-  # The projection shortcut should come after the first batch norm and ReLU
-  # since it performs a 1x1 convolution.
   if projection_shortcut is not None:
     shortcut = projection_shortcut(inputs)
+    shortcut = batch_norm_relu(shortcut, training, data_format)
```
nealwu: I think you don't want this here. Just add `shortcut` to `inputs` below and then call `batch_norm_relu` afterward on the combined value. In this case you can also remove `intermediate_addition` from `batch_norm_relu` above.

official/resnet/resnet.py
```diff
@@ -169,48 +175,122 @@ def building_block(inputs, filters, training, projection_shortcut, strides,
...
+    Convolution then batch normalization then ReLU as described by:
+      Deep Residual Learning for Image Recognition
+      https://arxiv.org/pdf/1512.03385.pdf
+      by Kaiming He, Xiangyu Zhang, Shaoqing Ren, and Jian Sun, Dec 2015.
+  """
+  shortcut = inputs
+
+  if projection_shortcut is not None:
+    shortcut = projection_shortcut(inputs)
+    shortcut = batch_norm_relu(shortcut, training, data_format)
```
nealwu: Same thing here.

official/resnet/resnet.py
```diff
@@ -169,48 +175,122 @@ def building_block(inputs, filters, training, projection_shortcut, strides,
...
+  inputs = conv2d_fixed_padding(
+      inputs=inputs, filters=4 * filters, kernel_size=1, strides=1,
+      data_format=data_format)
+  inputs = batch_norm_relu(inputs, training, data_format,
+                           intermediate_addition=shortcut)
+
+  return inputs
+
+
+def _bottleneck_block_v2(inputs, filters, training, projection_shortcut,
```
nealwu: Can we actually combine the v1 and v2 functions into a single function? There is a lot of repeated code here, and I think it should also be simpler after the `batch_norm_relu` fix I suggested above.

research/slim/BUILD
```diff
@@ -216,7 +216,7 @@ py_library(
     srcs = ["nets/alexnet.py"],
     srcs_version = "PY2AND3",
     deps = [
-        # "//tensorflow"
+        # "//tensorflow",
```
nealwu: Looks like you pulled in some commits from research/slim into this PR. Can you take those out?

robieta: It turns out this stems from me making a commit on this branch before all of my git setup was complete. Should be resolved now.

official/resnet/resnet.py
robieta: Ok, if you think the BN-ReLU is unnecessary for the projection I'll remove it. As far as removing the `intermediate_addition` arg, the 2016 v2 paper is clear that the shortcut addition goes between the BN and ReLU for v1, so I don't follow how that can be removed.

official/resnet/resnet.py
robieta: I agree, it doesn't seem ideal. As an exercise I put together an ultra concise function that could do v1 and v2 as well as regular and bottleneck.

```
_BOTTLENECK_LAYERS = [(1, 1, False, False), (3, 1, False, True), (1, 4, True, False)]
_REGULAR_LAYERS = [(3, 1, False, True), (3, 1, True, False)]
def _build_block(inputs, filters, training, projection_shortcut, strides,
    data_format, version, bottleneck):
  """
  Docstring removed for easier viewing on GitHub.
  """
  shortcut = inputs
  if version == 2:
    inputs = batch_norm_relu(inputs, training, data_format)

  if projection_shortcut is not None:
    shortcut = projection_shortcut(inputs)

  layers = _BOTTLENECK_LAYERS if bottleneck else _REGULAR_LAYERS
  for kernel_size, filter_num_mult, is_final, use_strides in layers:
    inputs = conv2d_fixed_padding(
        inputs=inputs, filters=filters, kernel_size=kernel_size,
        strides=strides if use_strides else 1,
        data_format=data_format)

    if version == 1:
      inputs = batch_norm_relu(inputs, training, data_format,
                       intermediate_addition=shortcut if is_final else None)
    else:  # version in (1, 2) check is performed by the resnet model
      if is_final:
        inputs += shortcut
      else:
        inputs = batch_norm_relu(inputs, training, data_format)
  return inputs
```

Even though this is < 50 lines (including the full docstring) compared to the ~170 being used now,  it's not easy to read. I think this is one of the rare cases where some code duplication may be good simply to improve readability. (That doesn't necessarily mean what is there now is optimal.) @karmel thoughts?

official/resnet/resnet.py
robieta: Looks like the TPU version of v1 just repeats the ReLU, which I suppose also works.

nealwu: I see; you have the right idea with `intermediate_addition` then. In that case, for the current `batch_norm_relu` after this projection shortcut, I think we actually only want a batch norm, since the relu will come after the shortcut and residual are combined together.

The best thing to do then is likely to add a separate function for just batch norm (without relu), which we can use both here and below, in which case we should no longer need the `intermediate_addition` argument.

nealwu: Looks like this is the TPU version you're referring to? https://github.com/tensorflow/tpu/blob/master/models/official/resnet/resnet_model.py#L171

This doesn't look quite correct to me, since there are too many relus. In fact the relu that comes at the end doesn't do anything, since the two values being added together have both been relu'd themselves. Goes along with what I mentioned of every ResNet implementation having minor differences. :)

official/resnet/resnet.py
```diff
@@ -169,34 +183,38 @@ def building_block(inputs, filters, training, projection_shortcut, strides,
...
       inputs=inputs, filters=filters, kernel_size=3, strides=strides,
       data_format=data_format)
-
   inputs = batch_norm_relu(inputs, training, data_format)
+
   inputs = conv2d_fixed_padding(
       inputs=inputs, filters=filters, kernel_size=3, strides=1,
       data_format=data_format)
+  inputs = batch_norm_relu(inputs, training, data_format,
+                           intermediate_addition=shortcut)
```
nealwu: You can replace this with:
```
inputs = batch_norm(inputs, training, data_format)
inputs = tf.nn.relu(inputs + shortcut)
```
after which `batch_norm_relu` doesn't need the `intermediate_addition` arg anymore (in fact, you can get rid of `batch_norm_relu` entirely if you like).

official/resnet/resnet.py
nealwu: I agree, combining both v1 + v2 and regular + bottleneck is likely too much. I think it's fine to combine just v1 + v2.

official/resnet/resnet.py
```diff
@@ -94,16 +94,23 @@ def process_record_dataset(dataset, is_training, batch_size, shuffle_buffer,
...
   # https://www.tensorflow.org/performance/performance_guide#common_fused_ops
-  inputs = tf.layers.batch_normalization(
+  return tf.layers.batch_normalization(
       inputs=inputs, axis=1 if data_format == 'channels_first' else 3,
       momentum=_BATCH_NORM_DECAY, epsilon=_BATCH_NORM_EPSILON, center=True,
       scale=True, training=training, fused=True)
+
+
+def batch_norm_relu(inputs, training, data_format, intermediate_addition=None):
+  """Performs a batch normalization followed by a ReLU."""
```
karmel: Docstring with Args and Returns here would be nice-- what is intermediate_addition?

official/resnet/resnet.py
```diff
@@ -211,6 +229,78 @@ def bottleneck_block(inputs, filters, training, projection_shortcut,
...
+  """
+  shortcut = inputs
+
+  if projection_shortcut is not None:
+    shortcut = projection_shortcut(inputs)
+    shortcut = batch_norm(inputs=shortcut, training=training,
+                          data_format=data_format)
+
+  inputs = conv2d_fixed_padding(
+      inputs=inputs, filters=filters, kernel_size=1, strides=1,
```
karmel: Just double checking that it is intentional that the conv2d kernel_size and strides are different than in the non-bottleneck version?

official/resnet/resnet.py
karmel: Questions one might use to decide:

1. Are people more likely to want to read through the code for one version, or to compare the two versions? I feel like the answer to this is probably "read through one version," as most users want to run one version of ResNet, rather than use this as the place to learn about the difference between the two. That would imply the current (separated) version is preferable, because it's easier to parse (and extract) each individual model.
2. Are developers more likely to make mistakes with duplicated code or branching code? Both are tricky. Duplicated might be slightly more error-prone, but branching code is pretty bad too.
3. What does the rest of the code do-- duplicate or branch? We use  some of each pattern. We duplicate for imagenet versus cifar, but branch in some of the internals in deciding how many strides, etc. etc. We seem to err on the side of duplication, though.

There doesn't seem to be an obvious answer, but I would say I lean slightly in favor of duplication in the name of clarity given the above. Assuming this is correct WRT what the ops are supposed to be, we can go with this, and reassess if we find that we are changing the code a lot and running into maintenance problems.

official/resnet/resnet.py
```diff
@@ -254,8 +344,10 @@ def block_layer(inputs, filters, block_fn, blocks, strides, training, name,
   Returns:
     The output tensor of the block layer.
   """
+  filters_out = filters
   # Bottleneck blocks end with 4x the number of filters as they start with
-  filters_out = 4 * filters if block_fn is bottleneck_block else filters
+  if block_fn is _bottleneck_block_v1 or block_fn is _bottleneck_block_v2:
```
karmel: This identity check is error-prone. Let's just pass in the bottleneck bool that we already have for this check.

official/resnet/resnet.py
```diff
@@ -278,8 +370,9 @@ class Model(object):
 
   def __init__(self, resnet_size, num_classes, num_filters, kernel_size,
                conv_stride, first_pool_size, first_pool_stride,
-               second_pool_size, second_pool_stride, block_fn, block_sizes,
-               block_strides, final_size, data_format=None):
+               second_pool_size, second_pool_stride, block_sizes, block_strides,
+               final_size, version=DEFAULT_VERSION, bottleneck=False,
```
karmel: nit: can we keep the original ordering of the bottleneck arg? I had originally put them in order of use, so bottleneck would belong in the middle there.

official/resnet/resnet.py
```diff
@@ -304,6 +397,9 @@ def __init__(self, resnet_size, num_classes, num_filters, kernel_size,
       block_strides: List of integers representing the desired stride size for
         each of the sets of block layers. Should be same length as block_sizes.
       final_size: The expected size of the model after the second pooling.
+      version: Integer defining the order of layers. A detailed
+        description can be found in README.md.
+      bottleneck: Use regular blocks or bottleneck blocks.
```
karmel: Make sure to remove unused block_fn arg above.

official/resnet/resnet.py
```diff
@@ -169,9 +176,144 @@ def conv2d_fixed_padding(inputs, filters, kernel_size, strides, data_format):
...
+        inputs = tf.nn.relu(inputs)
+
+      inputs = conv2d_fixed_padding(
+          inputs=inputs,
+          filters=filters * block["filters_mult"],
+          kernel_size=block["kernel_size"],
+          strides=block["strides"],
+          data_format=data_format)
+
+    return inputs + shortcut
```
robieta: @karmel @nealwu I'm putting this forward as a plausible middle ground. By not having much duplication, we can be very liberal about what goes in docstrings and hopefully get clarity that way. But it unfortunately isn't as clear as the 4 function status quo and I think that means that we have to live with the duplicate code method. If you have strong feelings I'm willing to be persuaded otherwise though.

karmel: This is interesting, but I think it really underscores that the readability of the four separate is preferable in this case. And this is coming from someone who goes long distances to get rid of duplicated code under most circumstances.

robieta: Alright, I guess that settles it.

official/resnet/resnet.py
robieta: This is correct.

official/resnet/resnet.py
robieta: Python doesn't allow args without defaults to follow those with, and I do like having defaults on version and bottleneck.

karmel: I would argue with so many other params, and a well-defined set of callers of this, we should just remove the default. Especially since the default run with current command line args actually does use bottleneck blocks.

karmel: For bottleneck. Version we should leave.

robieta: Ok. Changed.

official/resnet/resnet.py
nealwu: OK, I see why you both feel duplication results in less complication for the reader. My main concern here is that our code is getting very long (resnet.py is ~800 lines as of this PR), and that by adding lots of different features we are also adding a lot of extra code paths which can dilute the value of our code to the reader.

I agree now that combining these functions together with extra logic probably doesn't improve the situation much, but my main point is that keeping our code concise is something we should keep in mind.

robieta:
I ran 3 runs of master and 4 of the branch for ResNet_50 on ImageNet under identical conditions. The final accuracies are:

               mean    |    std
master    0.7503  |    0.0012
branch    0.7488  |    0.0011

official/resnet/resnet_test.py
```diff
@@ -0,0 +1,205 @@
...
+# limitations under the License.
+# ==============================================================================
+
+from __future__ import absolute_import
+from __future__ import division
+from __future__ import print_function
+
+
+import numpy as np
+import resnet
```
karmel: I think this will now have to be from official.resnet import resnet. Also, nit, should come below the tf import. That can be in a follow up PR though.

PR #3496, opened 2018-03-01
Fix call to range() to be consistent with 11 epochs

hsm207:


asimshankar:
Yeah, I'd just say modify the comment to say 10 epochs :) 
I think I had the range start from 1 just so that the messages printed out would say "Epoch 1 ..." etc.

I don't have a strong preference though.

k-w-w:
@hsm207 Can you modify the comment, and revert the range? Thanks!

hsm207:
@k-w-w  The edits have been done in #3535 so I am closing this pull request.

PR #3503, opened 2018-03-01
Add synthetic data option to Resnet

karmel:
This adds an option to use fake data (instead of the real images) to Resnet. Using synthetic data is useful in depugging performance, as it allows users to remove the input pipeline from the equation.

CC @tfboyd @reedwm @isaprykin , who may find this useful.

### DIFF
--- a/official/resnet/cifar10_main.py
+++ b/official/resnet/cifar10_main.py
@@ -205,7 +205,7 @@ def cifar10_model_fn(features, labels, mode, params):
 
 def main(unused_argv):
   input_function = FLAGS.use_synthetic_data and get_synth_input_fn() or input_fn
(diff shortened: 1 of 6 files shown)


nealwu:
Hey Karmel, is the idea that we have many external users who are actively interested in running on synthetic data? If on the other hand this is mainly intended for us to use ourselves, might be better not to push to GitHub in order to keep the code smaller.

karmel:
We intend for ResNet to be the example of how to write performant TF code (cf @petermattson ), so in this case, this is likely a helpful example for people debugging their own performance problems as well.

karmel:
Bump on this-- @nealwu , sound reasonable?

Issue #3491, opened 2018-02-28
TypeError: __init__() missing 2 required positional arguments: 'inputs' and ‘outputs'


hhu-luqi:
I want to review the code of tensorflow/models/official/mnist/mnist.py in the Python 3.5.2. But while I was running the code in the Spyder of Anaconde, a mistake, which is typeError: __init__() missing 2 required positional arguments: 'inputs' and ‘outputs' ,occurred.Why?
![default](https://user-images.githubusercontent.com/32589325/36788491-0f788230-1cc9-11e8-94dd-d00a7faa6f9a.png)
That is error which python console hints.
Please give me a favor.

shahparth123:
Facing same error. I tried tensorflow 1.5 and 1.6 (Both CPU and GPU version as well as Python 2.7 and Python 3.5.2)

xuzhiyuan1528:
Following the tutorial models/official/mnist/mnist.py, I meet the same error (python3.5 tensorflow-gpu 1.6), the super class tf.keras.Model needs arguments inputs and outputs.

asimshankar:
The code in the master branch of this repository may not be compatible with TensorFlow 1.5. 
To get the version of code compatible with TensorFlow 1.5, please use the r1.5 branch of the models repository.

Does that work?

(FYI @karmel )

hhu-luqi:
It did work, thanks!

ScottMcCormack:
Checking out the `r1.5` branch suited my version of Tensorflow.
`git checkout r1.5`

asimshankar:
Closing since this seems to be addressed by checking out the appropriate branch of the models repository. Please feel free to reopen if I'm mistaken.

Bolleanman:
> Facing same error.
> I change tensorflow to 1.7 and it work ok ! (I face the same error when my tf is 1.8。)
> The error accur beacause the API only availible when tf is 1.7 or 1.6，so , when chenge tf to 1.7,it work well.

Facing same error.
But my tf is 1.7.0, it doesn't work.

Adrian3652:
Can anyone help me, please? I have this error, but I don't know where am I supposed to add the width and height (I'm a beginner);

ax.add_patch(
        mpatches.Rectangle(
            (caja.x, caja.y),  # (x,y)
            w = (caja.width),          # width
            h = (h),          # height
            facecolor="#00ffff",
            edgecolor="black",
            linewidth=3)

Error:__init__() missing 2 required positional arguments: 'width' and 'height'

PR #3512, opened 2018-03-02
Edits to pass mnist_eager_test.py

hsm207:
These edits will still fail mnist_test.py

### DIFF
--- a/official/mnist/mnist.py
+++ b/official/mnist/mnist.py
@@ -22,10 +22,11 @@ import os
 import sys
 
 import tensorflow as tf
+import tensorflow.contrib.eager as tfe
 import dataset
 
 
-class Model(tf.keras.Model):
+class Model(tfe.Network):
   """Model to recognize digits in the MNIST dataset.
 
   Network structure is equivalent to:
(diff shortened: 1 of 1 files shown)


asimshankar:
Could you elaborate on the problem? Note that the master branch of the models repository may depend on features in the master branch of the TensorFlow repository. So, if you're looking for compatibility with a particular release of TensorFlow, then you should use the corresponding branch in the models repository.

hsm207:
@asimshankar The problem and its source is discussed in #3491.

I stumbled upon this error while reading [this ](https://research.googleblog.com/2017/10/eager-execution-imperative-define-by.html)blog post. It says I we can convert the model to graphs by running the same code without eager execution enabled and pointed readers to [this ](https://github.com/tensorflow/tensorflow/tree/master/tensorflow/contrib/eager/python/examples/mnist)repository which pointed to mnist_eager.py. But mnist_eager,py won't even execute because due to #3491.

asimshankar:
@hsm207 - I've added a comment to #3491 as well. Which version of TensorFlow are you using? The blog post refers to the nightly version of TensorFlow (which will be compatible with this). But if you're using TensorFlow 1.5 or something, then you want to use the r1.5 branch of this repository.

Let us know if that works or not. Thanks.

hsm207:
@asimshankar I am using TensorFlow 1.5. The r1.5 branch does not have mnist_eager.py.

You are right, running mnist_eager_test.py using the nightly version (1.7.0-dev20180302) works.

Thank you for your clarification.

Should I delete this pull request?

asimshankar:
Yes, let's delete it. Thanks.

Katze123456:
pls help

PR #3520, opened 2018-03-05
from six.moves import xrange (en masse) AGAIN

cclauss:
In the six weeks since #3206 was merged, the Python 2-only function __xrange()__ was introduced into 24 files in this repo.  This PR reapplies the same fix as last time: __from six.moves import xrange__.

It would be quite helpful if we could have __flake8 . --select=E901,E999,F821,F822,F823__ added to the test suite to catch these issues compatibility before code is reviewed and merged.  Our current test suite does not flag them.


cclauss:
@nealwu @karmel

PR #3533, opened 2018-03-06
Adding thread args back in, with allow_soft_placement

karmel:
This incorporates @nhasabni 's commits, and adds allow_soft_placement = True, which is necessary for multi-GPU mode.

cc @isaprykin


karmel:
Looks like I confused the CLA bot when cherry-picking @nhasabni 's commit and resolving conflicts. @nhasabni , I will assume you continue to be okay with the CLA given your recent consent on #3502  . If that's not the case, let me know.

nhasabni:
@karmel Yes, that's ok. Thanks.

PR #3558, opened 2018-03-10
More Preprocessing tuning for Resnet

karmel:
These changes bring us to 76+% accuracy, and also improve performance, especially for the 8 GPU case. Notable differences in training:

* Use the bounding boxes supplied in the ImageNet dataset (accuracy)
* Use the sample_distorted_bounding_box op instead of creating a random box for cropping during training (speed)
* Use the fused op for decoding and cropping instead of two separate ops (speed)
* The fused decoding requires that we resize after cropping

Eval is the same as previously, but cleaned up slightly now that the functions are not used for training as well.

Screenshot from a run with 4xGPU is below.

FYI @reedwm @bignamehyp @scott7z @robieta 

screen shot 2018-03-09 at 4 45 15 pm


official/resnet/imagenet_main.py
```diff
@@ -57,29 +57,71 @@ def get_filenames(is_training, data_dir):
 def _parse_example_proto(example_serialized):
   """Parses an Example proto containing a training example of an image.
 
-  The dataset contains serialized Example protocol buffers.
-  The Example proto is expected to contain features named
-  image/encoded (a JPEG-encoded string) and image/class/label (int)
+  The output of the build_image_data.py image preprocessing script is a dataset
+  containing serialized Example protocol buffers. Each Example proto contains
+  the following fields:
```
nealwu: The numbers below are example values, right? Could we specify that in this comment?

nealwu:
Regardless, nice work on 76+% accuracy!

karmel:
As I understand it, there are several names around preprocessing that are used colloquially-- "inception," "resnet," and "vgg." Inception does more color distortion, which can improve accuracy, but slows down processing. Resnet is what we now have here, and VGG is similar, but doing the aspect-preserving-resize and no bounding boxes. We now have what we might call Resnet preprocessing, pulled largely from tf_cnn_benchmarks, but with fewer flags and options for simplicity. The linked inception preprocessing has the aforementioned color distortion, no fused op, etc., so a little more complex. Arguably, we might want to rename the vgg file at this point, but for now, probably okay to leave.

Pulling in the random commits was my attempt to squash a bunch of debugging commits and a master merge, but that ended in confusion and regret, as git rebasing always seems to :/ I think the changes made, however, are limited to the set intended to be made, with only the commit messages weirdly included.

Comment updated-- thanks for the review.

PR #3563, opened 2018-03-11
Allow user to specify allow_growth in training script.

StevenHickson:
Added allow_gpu_growth flag in training script to allow reducing gpu memory usage if requested


StevenHickson:
@sguada I've updated this to be merged. This is useful for those who don't want to use all the gpu memory and allows that with a flag.

PR #3570, opened 2018-03-12
Build ImageNet TFRecord

JoshuaPiinRueyPan:
When building the TFRecord of ILSVRC, I get the following error:
```
file: research/inception/inception/data/build_imagenet_data.py
line: 175: six.binary_type TypeError: str() takes at most 1 argument
```
This probably due to the incompatible of python2 and python3 even if the author uses the six module.
In Python3: 
`class str(object='', encoding='utf-8', ...)
`
however, In Python2:
`class str(object='')`
i.e. It only take one argument.

I fix this bug by catching the exception and assign the string without encoding.


shlens:
I don't know off hand what is going on but the change seems safe and reasonable.

PR #3584, opened 2018-03-13
(WIP) Make TensorBoard graph more readable for ResNet.

robieta:
Current:
screenshot from 2018-03-13 15-08-05

Proposed:
screenshot from 2018-03-13 15-06-46
screenshot from 2018-03-13 15-06-50

### DIFF
--- a/official/resnet/resnet.py
+++ b/official/resnet/resnet.py
@@ -32,6 +32,7 @@ from __future__ import division
 from __future__ import print_function
 
 import argparse
+import functools
 import os
 
 import tensorflow as tf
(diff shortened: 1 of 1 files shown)


official/resnet/resnet.py
```diff
@@ -192,9 +193,22 @@ def conv2d_fixed_padding(inputs, filters, kernel_size, strides, data_format):
       data_format=data_format)
 
 
+def group_block(f):
+  setattr(f, "block_counter", 0)
```
k-w-w: add a docstring describing the function and why/how to use it. It would also be great if you can pass the block type as an argument ("building_v1" or "bottleneck_v2"), and add it to the variable scope name. The current screenshots look great!

robieta:
Abandoning for now. May revisit at a later date.

PR #3585, opened 2018-03-13
Add data schema for the benchmark run in Bigquery.

qlzh727:
The current schema contains the entity information about model
and train data metadata, as well as machine config. Future change
will contain benchmark metric.

The json schema can be used to create bigquery table. A sample
table can be found in
https://bigquery.cloud.google.com/table/tf-benchmark-dashboard:test_benchmark.benchmark_run.

### DIFF
--- a/official/benchmark/datastore/schema/benchmark_run.json
+++ b/official/benchmark/datastore/schema/benchmark_run.json
@@ -18,6 +18,26 @@
     "type": "DATETIME"
   },
   {
+    "description": "The tensorflow version information.",
+    "fields": [
+      {
+        "description": "Version of the tensorflow. E.g. 1.7.0-rc0",
+        "mode": "REQUIRED",
+        "name": "version",
+        "type": "STRING"
+      },
+      {
+        "description": "Git Hash of the tensorflow",
+        "mode": "NULLABLE",
+        "name": "git_hash",
+        "type": "STRING"
+      }
+    ],
+    "mode": "REQUIRED",
+    "name": "tensorflow_version",
+    "type": "RECORD"
+  },
+  {
     "description": "The arbitrary attribute of the model.",
     "fields": [
       {
@@ -38,41 +58,61 @@
     "type": "RECORD"
   },
   {
-    "description": "The list of hyper parameter of the model.",
+    "description": "Enviornment variables when the benchmark run is executed.",
     "fields": [
       {
-        "description": "The name of the hyper parameter.",
+        "description": "The name of the variable.",
         "mode": "REQUIRED",
         "name": "name",
         "type": "STRING"
       },
       {
-        "description": "The string value of the hyper parameter.",
+        "description": "The value of the variable.",
+        "mode": "NULLABLE",
+        "name": "value",
+        "type": "STRING"
+      }
+    ],
+    "mode": "REPEATED",
+    "name": "enviornment_variable",
+    "type": "RECORD"
+  },
+  {
+    "description": "The list of hyperparameters of the model.",
+    "fields": [
+      {
+        "description": "The name of the hyperparameter.",
+        "mode": "REQUIRED",
+        "name": "name",
+        "type": "STRING"
+      },
+      {
+        "description": "The string value of the hyperparameter.",
         "mode": "NULLABLE",
         "name": "string_value",
         "type": "STRING"
       },
       {
-        "description": "The bool value of the hyper parameter.",
+        "description": "The bool value of the hyperparameter.",
         "mode": "NULLABLE",
         "name": "bool_value",
         "type": "STRING"
       },
       {
-        "description": "The int/long value of the hyper parameter.",
+        "description": "The int/long value of the hyperparameter.",
         "mode": "NULLABLE",
         "name": "long_value",
         "type": "INTEGER"
       },
       {
-        "description": "The double/float value of hyper parameter.",
+        "description": "The double/float value of hyperparameter.",
         "mode": "NULLABLE",
         "name": "float_value",
         "type": "FLOAT"
       }
     ],
     "mode": "REPEATED",
-    "name": "hyper_parameter",
+    "name": "hyperparameter",
     "type": "RECORD"
   },
   {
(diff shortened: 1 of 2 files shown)


qlzh727:
Sample tables can be found in https://bigquery.cloud.google.com/dataset/tf-benchmark-dashboard:test_benchmark.

official/benchmark/datastore/schema/benchmark_run.json
```diff
@@ -0,0 +1,245 @@
...
+    "name": "model_id",
+    "type": "STRING"
+  },
+  {
+    "description": "The name of the model, E.g ResNet50, LeNet-5 etc.",
+    "mode": "REQUIRED",
+    "name": "model_name",
+    "type": "STRING"
+  },
+  {
```
karmel: Some way to indicate commit, code path, or other things tied to the specific instance of code run would be good. And the exact command line that was run, and any environment variables.

official/benchmark/datastore/schema/benchmark_run.json
```diff
@@ -0,0 +1,245 @@
...
+        "fields": [
+          {
+            "mode": "NULLABLE",
+            "name": "count",
+            "type": "INTEGER"
+          },
+          {
+            "mode": "NULLABLE",
+            "name": "model",
+            "type": "STRING"
```
karmel: We probably want to indicate in some way how the GPUs are configured:
* Cuda version
* Topology params we care about (cf @tfboyd )?
* Number of hosts will eventually become relevant, if we know it (ie, number of separate boards that all these GPUs live on)
Not sure if we want to anticipate these upfront, or just add as they come.

official/benchmark/datastore/schema/benchmark_run.json
```diff
@@ -0,0 +1,245 @@
...
+            "name": "system",
+            "type": "STRING"
+          },
+          {
+            "description": "Eg: #120-Ubuntu SMP Mon Jan 18 15:59:10 UTC 2016.",
+            "mode": "NULLABLE",
+            "name": "version",
+            "type": "STRING"
+          }
+        ]
```
karmel: Do we want to try to capture cloud info here? ie, running on k8s versus a VM versus metal?

official/benchmark/datastore/schema/benchmark_run.json
```diff
@@ -0,0 +1,245 @@
...
+      {
+        "mode": "NULLABLE",
+        "name": "memory_total",
+        "type": "INTEGER"
+      },
+      {
+        "mode": "NULLABLE",
+        "name": "memory_available",
+        "type": "STRING"
+      }
```
karmel: Capturing env variables relevant to the compute environment would be good-- ie, CUDA_VISIBLE_DEVICES, whether to share GPU memory (sorry, forgetting what that one is right now, but it should suffice to say, there are many).

official/benchmark/datastore/schema/benchmark_run.json
```diff
@@ -0,0 +1,245 @@
...
+              {
+                "mode": "NULLABLE",
+                "name": "size",
+                "type": "INTEGER"
+              }
+            ]
+          }
+        ]
+      },
+      {
```
karmel: I see you went with JSON instead of Proto, which WFM if you find it preferable. But, to make the question more complicated-- what about YAML? We will have a bunch of those for k8s anyhow, and it's much more human-readable without all these brackets. Thoughts?

official/benchmark/datastore/schema/benchmark_run.json
```diff
@@ -0,0 +1,245 @@
...
+            "type": "STRING"
+          }
+        ],
+        "mode": "REPEATED",
+        "name": "attribute",
+        "type": "RECORD"
+      }
+    ]
+  },
+  {
```
karmel: Where should information about parameter server configuration live? I guess that's mostly about how the model itself is run. Maybe we don't need to explicitly store that as long as we capture the command line and code commit that was run.

karmel:
Oh, and another thought: tensorflow build/version should be represented somewhere.

official/benchmark/datastore/schema/benchmark_run.json
```diff
@@ -0,0 +1,245 @@
...
+        "name": "value",
+        "type": "STRING"
+      }
+    ],
+    "mode": "REPEATED",
+    "name": "attribute",
+    "type": "RECORD"
+  },
+  {
+    "description": "The list of hyper parameter of the model.",
```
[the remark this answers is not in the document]
qlzh727: Done.

official/benchmark/datastore/schema/benchmark_run.json
```diff
@@ -0,0 +1,245 @@
...
+      },
+      {
+        "description": "The double/float value of hyper parameter.",
+        "mode": "NULLABLE",
+        "name": "float_value",
+        "type": "FLOAT"
+      }
+    ],
+    "mode": "REPEATED",
+    "name": "hyper_parameter",
```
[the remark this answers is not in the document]
qlzh727: Done.

official/benchmark/datastore/schema/benchmark_run.json
```diff
@@ -0,0 +1,245 @@
...
+        "description": "The value of the attribute.",
+        "mode": "NULLABLE",
+        "name": "value",
+        "type": "STRING"
+      }
+    ],
+    "mode": "REPEATED",
+    "name": "attribute",
+    "type": "RECORD"
+  },
```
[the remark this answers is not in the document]
qlzh727: Good point. Adding tf verison info and environment variables.

official/benchmark/datastore/schema/benchmark_run.json
qlzh727: Ack. I am not worrying that for the moment, we can update the data schema if we want in future.

official/benchmark/datastore/schema/benchmark_run.json
qlzh727: The schema file is used to create bigquery table, and bigquery only accept json as schema format. I don't have other option here.

official/benchmark/datastore/schema/benchmark_run.json
qlzh727: Done. Added a section with minimal cloud info, and a free format key-value pair for the moment.

official/benchmark/datastore/schema/benchmark_run.json
qlzh727: Added cuda_version which is standard, not sure we could capture other info easily or not.

official/benchmark/datastore/schema/benchmark_run.json
qlzh727: Ack, for the moment, we will just dump them into env variables.

qlzh727:
Ping

PR #3586, opened 2018-03-13
Update resnet with logging utils

yhliang2018:


official/resnet/resnet.py
```diff
@@ -746,14 +748,7 @@ def resnet_main(flags, model_function, input_function):
...
   for _ in range(flags.train_epochs // flags.epochs_per_eval):
-    tensors_to_log = {
-        'learning_rate': 'learning_rate',
-        'cross_entropy': 'cross_entropy',
-        'train_accuracy': 'train_accuracy'
-    }
-
-    logging_hook = tf.train.LoggingTensorHook(
-        tensors=tensors_to_log, every_n_iter=100)
+    train_hooks = hooks_helper.get_train_hooks(['LoggingTensorHook'])
```
robieta: I think arguments should still be explicit in the caller. (i.e. still pass every_n_iter)

karmel:
The goal here should be to add this as a command-line arg; can you make Resnet take an arg that allows people to pass in the list of hooks they want? Then this file should just pass those through above. (May want to compare notes with @robieta since he is probably fairly close to pushing the arg parsers change.)

yhliang2018:
@karmel @tfboyd Logging utils are added to resnet with unified argparser from @robieta . Could you help to review it? Thanks!

Now _warm_steps_ for ExamplesPerSecondHook is set as 5 by default. @tfboyd Could you let me know if you have a preferred number?

official/utils/arg_parsers/parsers.py
```diff
@@ -117,6 +118,18 @@ def __init__(self, add_help=False, data_dir=True, model_dir=True,
           help="If set, run across all available GPUs."
       )
 
+    if hooks:
+      self.add_argument(
+          "--hooks", "-hk", nargs="+", default=["LoggingTensorHook"],
+          help="[default: %(default)s] A list of strings to specify the names "
+               "of train hooks. "
+               "Example: --hooks LoggingTensorHook ExamplesPerSecondHook. "
```
robieta: If the number of allowed hooks becomes large it may be worth it to tie this into hooks_helper through the choices arg of add_argument(). Probably overkill for now.

robieta:
LGTM

PR #3598, opened 2018-03-14
Basic end to end test of resnet.

robieta:
Currently arg parsing is not tested for resnet. This aims to resolve such issues by actually running a trivial train.

### DIFF
--- a/official/resnet/imagenet_main.py
+++ b/official/resnet/imagenet_main.py
@@ -281,15 +281,15 @@ def imagenet_model_fn(features, labels, mode, params):
                                          multi_gpu=params['multi_gpu'])
 
 
-def main(unused_argv):
-  input_function = FLAGS.use_synthetic_data and get_synth_input_fn() or input_fn
-  resnet_run_loop.resnet_main(FLAGS, imagenet_model_fn, input_function)
+def main(argv):
+  parser = resnet_run_loop.ResnetArgParser(
+      resnet_size_choices=[18, 34, 50, 101, 152, 200])
+  flags = parser.parse_args(args=argv[1:])
+
+  input_function = flags.use_synthetic_data and get_synth_input_fn() or input_fn
+  resnet_run_loop.resnet_main(flags, imagenet_model_fn, input_function)
 
 
 if __name__ == '__main__':
   tf.logging.set_verbosity(tf.logging.INFO)
-
-  parser = resnet_run_loop.ResnetArgParser(
-      resnet_size_choices=[18, 34, 50, 101, 152, 200])
-  FLAGS = parser.parse_args()
-  tf.app.run(argv=[sys.argv[0]])
+  tf.app.run(argv=sys.argv)
--- a/official/resnet/imagenet_main.py
+++ b/official/resnet/imagenet_main.py
@@ -291,5 +291,5 @@ if __name__ == '__main__':
 
   parser = resnet_run_loop.ResnetArgParser(
       resnet_size_choices=[18, 34, 50, 101, 152, 200])
-  FLAGS, unparsed = parser.parse_known_args()
-  tf.app.run(argv=[sys.argv[0]] + unparsed)
+  FLAGS = parser.parse_args()
+  tf.app.run(argv=[sys.argv[0]])
--- a/official/utils/testing/integration.py
+++ b/official/utils/testing/integration.py
@@ -1,0 +1,71 @@
+# Copyright 2018 The TensorFlow Authors. All Rights Reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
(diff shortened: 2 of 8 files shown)


karmel:
Going to leave this to the other reviewers listed for fine-grained review, but, high level: let's include this in the existing test files if possible (cifar, probably), and include some comments as to what exactly this succeeds in testing.

robieta:
@karmel I was trying to share code since I want this to run with imagenet too, but I realize a much better pattern is to have any shared code in utils and that way these basic "does it run" tests can live in the existing test files. (And it makes it easy to extend to other models.)

official/resnet/end_to_end_test.py
```diff
@@ -0,0 +1,48 @@
...
+import subprocess
+import sys
+import unittest
+
+
+PYTHON_CMD = "python2" if sys.version_info[0] == 2 else "python3"
+ROOT = os.getcwd()
+
+
+def run(args):
```
k-w-w: Agreed with moving this function into utils, and placing test_cifar in the cifar10_test file. Also add docstring for this function and the test.

official/resnet/end_to_end_test.py
```diff
@@ -0,0 +1,48 @@
...
+from __future__ import print_function
+
+import os
+import subprocess
+import sys
+import unittest
+
+
+PYTHON_CMD = "python2" if sys.version_info[0] == 2 else "python3"
+ROOT = os.getcwd()
```
k-w-w: nit: would prefer CWD, as root can also be used to refer to a person's home directory.

official/resnet/end_to_end_test.py
robieta: Added.

official/resnet/end_to_end_test.py
robieta: This is now moot, but good point.

karmel:
I resist the idea of adding a command line arg just for the sake of a single test. I would rather see something that allows us to specify the number of batches, regardless of synthetic data or not, which is useful IRL for perf testing as well.

robieta:
The flag is not longer synthetic specific. Should be good to go unless there are any other issues.

official/utils/testing/integration.py
```diff
@@ -0,0 +1,71 @@
...
+          "--use_synthetic_data",
+          "--max_train_steps", "1"] + extra_flags
+
+  p = subprocess.Popen(args, stdin=subprocess.PIPE, stdout=subprocess.PIPE,
+                       stderr=subprocess.PIPE)
+  output, err = p.communicate()
+  returncode = p.returncode
+
+  if os.path.exists(model_dir):
+    shutil.rmtree(model_dir)
```
qlzh727: I would suggest this to be put into a final block, in case your call to process gets any error and exit early.

official/resnet/imagenet_main.py
```diff
@@ -291,5 +291,5 @@ def main(unused_argv):
 
   parser = resnet_run_loop.ResnetArgParser(
       resnet_size_choices=[18, 34, 50, 101, 152, 200])
-  FLAGS, unparsed = parser.parse_known_args()
-  tf.app.run(argv=[sys.argv[0]] + unparsed)
+  FLAGS = parser.parse_args()
+  tf.app.run(argv=[sys.argv[0]])
```
qlzh727: I think my question is that why the parser need to be in __main__ instead of main() function.

The code in __main__ is only executed when imagenet_main.py is invoked as main target instead of imported as module. For this file, I can hardly foresee it being import as a module for real prod doc, except the case for test. If this assumption is true, then why not push everything here to main(), so that your test code and just import it as module, and avoid calling a new python process.

robieta: Discussed offline and agreed that this is the way to go.

official/utils/testing/integration.py
robieta: Makes sense. Also, TIL a finally block will still run even if sys.exit() is called, which is very useful in this case.

official/utils/testing/integration.py
```diff
@@ -0,0 +1,60 @@
...
+  very limited run is performed using synthetic data.
+
+  Args:
+    main: The primary function used to excercise a code path. Generally this
+      function is "<MODULE>.main(argv)".
+    extra_flags: Additional flags passed by the the caller of this function.
+
+  Raises:
+    OSError: Any error detected in the run causes an OSError. Because the code
+      is run as a subprocess, it is not able to manifest the specific python
```
k-w-w: It doesn't look like the code is still run as a subprocess, so this comment can probably be changed

Ashley-fei:
Dear AI Enthusiast,
I'm Ashley, a team member at AAIA. We're expanding our innovative multimodal AI team, and I thought you might be interested in this exciting opportunity.
Our team is working on:

Developing downstream multimodal large models for practical applications in Education, Healthcare, Embodied intelligence
Collaborations with industry leaders to bring innovative ideas to life
AI solutions that make a real difference in people's lives
Joining us means you'll:

Work on cutting-edge AI challenges
Collaborate with brilliant minds in a dynamic environment
Contribute directly to shaping AI's future
We're looking for passionate, innovative individuals who are eager to push the boundaries of AI.
If this sounds intriguing, I'd be happy to chat more or put you in touch with our hiring team.
Looking forward to potentially welcoming you aboard!

Best regards,
Ashley
AIGC@AAIA
Wechat:18521323869

Ashley-fei:
[Uploading JD_「Senior Video Understanding Algorithm Engineer」.pdf…]()

PR #3599, opened 2018-03-14
Removing dependency on tf.test

karmel:
These can just be normal unittests. Also, changing the names of the functions, which are making glint complain.


official/utils/logging/hooks_helper_test.py
```diff
@@ -19,14 +19,13 @@
 from __future__ import division
 from __future__ import print_function
 
-import tensorflow as tf
+import unittest
 
+import tensorflow as tf
 from official.utils.logging import hooks_helper
```
yhliang2018: Should we add a blank line after tf import?

PR #3609, opened 2018-03-15
fix cityscapes _POSTFIX_MAP

ksnzh:
The annotation image file name is like `aachen_000000_000019_gtFine_labelIds.png` with a postfix `_gtFine_labelIds`.


kr-ish:
The existing code is correct - see https://github.com/tensorflow/models/issues/3631#issuecomment-376271034

PR #3619, opened 2018-03-15
Add benchmark utility functions for metric logging

qlzh727:
Currently it only log the metric, which we could use in any of the hooks.


official/benchmark/logger.py
```diff
@@ -0,0 +1,64 @@
...
+"""Logging utilities for benchmark."""
+from __future__ import absolute_import
+from __future__ import division
+from __future__ import print_function
+
+import datetime
+import json
+import os
+
+from tensorflow.python.platform import gfile
```
karmel: You should be able to use the public API for this, which is preferable-- tf.gfile.GFile, etc. See tensorflow/python/platform/gfile.py for allowed symbols.

official/benchmark/logger.py
```diff
@@ -0,0 +1,64 @@
+# Copyright 2018 The TensorFlow Authors. All Rights Reserved.
```
karmel: Why have this here instead of in utils/logging? I'm not opposed, but we should have reasons, given the similarity of purpose.

karmel: Also, nit: since we have `utils`, and will probably eventually have `models`, and will certainly have multiple benchmarks, how do you feel about naming this package `benchmarks` instead of the singular `benchmark`?

official/benchmark/logger.py
```diff
@@ -0,0 +1,64 @@
...
+
+class BenchmarkLogger(object):
+  """Class to log the benchmark information to local disk."""
+
+  def __init__(self, logging_dir):
+    self._logging_dir = logging_dir
+    if not gfile.IsDirectory(self._logging_dir):
+      gfile.MakeDirs(self._logging_dir)
+
+  def logMetric(self, name, value, unit=None, global_step=None, extras=None):
```
karmel: Shouldn't this be `log_metric`?

official/benchmark/logger.py
```diff
@@ -0,0 +1,64 @@
...
+    """Log the benchmark metric information to local file.
+
+    Currently the logging is done in a synchronized way. This should be updated
+    to log asynchronously.
+
+    Args:
+      name: string, the name of the metric to log.
+      value: number, the value of the metric.
```
karmel: We don't actually do type checking here-- should we? Or maybe relax the stated requirement, and just say this should be json-dumpable (which, IIRC, has all sorts of limitations, but float/str/int should be okay-- though that raises the question of how tensors would get handled here?)

official/benchmark/logger.py
```diff
@@ -0,0 +1,64 @@
...
+      unit: string, the unit of the metric, E.g "image per second".
+      global_step: int, the global_step when the metric is logged.
```
karmel: If you just run tf.train.get_global_step, you get a Tensor. Not sure what will happen if we try to log that directly? Do we have to convert tensor output to raw? We probably want to have a stance and add tests accordingly. I guess it depends on where we are calling this from-- the main loop, or from within the Estimator and related tf code.

official/benchmark/logger.py
```diff
@@ -0,0 +1,64 @@
...
+        os.path.join(self._logging_dir, _METRIC_LOG_FILE_NAME), "a") as f:
+      metric = {
+          "name": name,
+          "value": value,
+          "unit": unit,
+          "global_step": global_step,
+          "timestamp": datetime.datetime.now().strftime(
+              _DATE_TIME_FORMAT_PATTERN),
+          "extras": extras}
+      json.dump(metric, f)
```
karmel: Does json.dump play nicely with the tf file writer? If this is actually happening deferred in a graph, what happens?

official/benchmark/logger_test.py
```diff
@@ -0,0 +1,83 @@
...
+
+
+import json
+import os
+import tempfile
+
+
+from official.benchmark import logger
+import tensorflow as tf
+from tensorflow.python.platform import gfile
```
karmel: Ditto on not about about public API. Also, I know pylint complains, but we have the main TF import first, with the local package imports after. (Note that this is also important for the import-to-google rules to work as they are now.)

official/benchmark/logger_test.py
```diff
@@ -0,0 +1,83 @@
...
+
+
+class BenchmarkLoggerTest(tf.test.TestCase):
+
+  def tearDown(self):
+    gfile.DeleteRecursively(self.get_temp_dir())
+
+  def testCreateLoggingDir(self):
```
karmel: I'm less particular about test names, but I think the rest of our tests are underscored rather than camelCase-- probably best to be consistent.

official/benchmark/logger_test.py
```diff
@@ -0,0 +1,83 @@
...
+    non_exist_temp_dir = os.path.join(self.get_temp_dir(), "unknown_dir")
+    self.assertFalse(gfile.IsDirectory(non_exist_temp_dir))
+
+    logger.BenchmarkLogger(non_exist_temp_dir)
+    self.assertTrue(gfile.IsDirectory(non_exist_temp_dir))
+
+  def testLogMetric(self):
+    log_dir = tempfile.mkdtemp(dir=self.get_temp_dir())
+    log = logger.BenchmarkLogger(log_dir)
+    log.logMetric("accuracy", 0.999, global_step=1e4, extras={"name": "value"})
```
karmel: Stated above, but just to record in the proper place: we should take a stance on Tensor input, and make sure we test accordingly.

official/benchmark/logger.py
qlzh727: My original intention is to put all benchmark related code into official/benchmark, but not the benchmark themselves. I will have other code like upload to bigstore and other libs in future change. I think they are more logically grouped together. 

I could move those code to officials/utils/benchmark if you prefer

official/benchmark/logger.py
```diff
@@ -0,0 +1,64 @@
...
+    Args:
+      name: string, the name of the metric to log.
+      value: number, the value of the metric.
+      unit: string, the unit of the metric, E.g "image per second".
+      global_step: int, the global_step when the metric is logged.
+      extras: map of string:string, the extra information about the metric.
+    """
+    with gfile.GFile(
+        os.path.join(self._logging_dir, _METRIC_LOG_FILE_NAME), "a") as f:
+      metric = {
```
robieta: Can we assign a unique ID to each class instance and include it? That way if a bunch of records get jammed together in the same file (i.e. the default path) it is easy to separate runs later.

official/benchmark/logger.py
qlzh727: Chatted offline, I think we will stick with the simple number value for the moment. When user try to use this in the context of tf run session, they should be able to get the simple value from the run session.

official/benchmark/logger_test.py
qlzh727: Done.

official/benchmark/logger_test.py
qlzh727: addressed in the previous comment.

official/benchmark/logger_test.py
qlzh727: Done.

official/benchmark/logger.py
robieta: On a similar note, we should decide how to handle failures. I'm inclined to not have a logging hook bring down the whole run, but just silently bailing doesn't seem good either. Perhaps salvage what info can be dumped and write that along with an indication that it is not a clean record?

qlzh727: Good point, probably we could add a logger to local warning for this case.

official/benchmark/logger.py
qlzh727: I think the logging_dir is the flag that caller should tweak. Similar as the log_dir flag for summary writer, the caller should make sure to specify different log dir for different run.

qlzh727: Chatted offline, we actually have the UUID field in the design for record keeping. Will have the ID field in future change.

official/benchmark/logger.py
qlzh727: Done. Added a warning instead of throwing valueError here.

karmel:
Poor Git. Try: 
```
git merge origin master
git diff > ~/benchmark.patch # Open that patch. Should look like your set of changes and only your set.
git push # Not necessary, but just in case
# Grab the commit hash for current master HEAD
git reset --hard <commit hash for master HEAD>
git apply ~/benchmark.patch
git diff # Make sure this looks right
git push
```

qlzh727:
See https://github.com/qlzh727/models/commit/6ad9e0b84da7791784e46ae56253b2177a899494 as a demo for metric logger

qlzh727:
Btw, the metric log from the https://github.com/qlzh727/models/commit/6ad9e0b84da7791784e46ae56253b2177a899494 looks like below:

{"name": "train_accuracy", "timestamp": "2018-03-19T11:36:57.969879Z", "value": 0.9988235235214233, "extras": null, "unit": null, "global_step": 25604}
{"name": "loss", "timestamp": "2018-03-19T11:37:07.436055Z", "value": 1.953561877598986e-05, "extras": null, "unit": null, "global_step": 25704}
{"name": "train_accuracy", "timestamp": "2018-03-19T11:37:07.436379Z", "value": 0.9988889098167419, "extras": null, "unit": null, "global_step": 25704}
{"name": "loss", "timestamp": "2018-03-19T11:37:16.885552Z", "value": 0.01344460342079401, "extras": null, "unit": null, "global_step": 25804}
{"name": "train_accuracy", "timestamp": "2018-03-19T11:37:16.886348Z", "value": 0.9984210729598999, "extras": null, "unit": null, "global_step": 25804}
{"name": "loss", "timestamp": "2018-03-19T11:37:26.488251Z", "value": 0.00038322151522152126, "extras": null, "unit": null, "global_step": 25904}
{"name": "train_accuracy", "timestamp": "2018-03-19T11:37:26.488567Z", "value": 0.9984999895095825, "extras": null, "unit": null, "global_step": 25904}
{"name": "loss", "timestamp": "2018-03-19T11:37:36.103425Z", "value": 8.149698260240257e-05, "extras": null, "unit": null, "global_step": 26004}
{"name": "train_accuracy", "timestamp": "2018-03-19T11:37:36.103754Z", "value": 0.9985714554786682, "extras": null, "unit": null, "global_step": 26004}
{"name": "loss", "timestamp": "2018-03-19T11:37:45.991382Z", "value": 0.0013781175948679447, "extras": null, "unit": null, "global_step": 26104}
{"name": "train_accuracy", "timestamp": "2018-03-19T11:37:45.991713Z", "value": 0.9986363649368286, "extras": null, "unit": null, "global_step": 26104}
{"name": "loss", "timestamp": "2018-03-19T11:37:55.977899Z", "value": 0.0007561420788988471, "extras": null, "unit": null, "global_step": 26204}
{"name": "train_accuracy", "timestamp": "2018-03-19T11:37:55.978276Z", "value": 0.9986956715583801, "extras": null, "unit": null, "global_step": 26204}
{"name": "loss", "timestamp": "2018-03-19T11:38:05.737698Z", "value": 0.008610671386122704, "extras": null, "unit": null, "global_step": 26304}

karmel:
That is great-- thanks. I do feel like the example makes the case stronger that the logger should live in utils, since one might want to use to to write arbitrary JSON logs separate from benchmarking. But very nice regardless.

qlzh727:
Done. Moved to offical/utils/logging.

karmel:
Thanks- LGTM, though we should still try to clean up the git record so that the CLA is clear.

PR #3621, opened 2018-03-15
Partition resnet code

robieta:
Currently the official resnet model definition file is quite long. This PR separates code that defines the model layers and supporting code to make the model more readable.

### DIFF
--- a/official/resnet/cifar10_main.py
+++ b/official/resnet/cifar10_main.py
@@ -23,7 +23,8 @@ import sys
 
 import tensorflow as tf
 
-from official.resnet import resnet
+from official.resnet import resnet_model
+from official.resnet import resnet_run_loop
 
 _HEIGHT = 32
 _WIDTH = 32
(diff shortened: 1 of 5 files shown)


official/resnet/cifar10_main.py
```diff
@@ -23,7 +23,8 @@
 
 import tensorflow as tf
 
-from official.resnet import resnet
+from official.resnet import resnet_model
+from official.resnet import resnet_util
```
karmel: resnet_util seems unclear; maybe resnet_run? resnet_loop? I don't think there is a good name for "the stuff that lives around the model," unfortunately.

official/resnet/resnet_model.py
```diff
@@ -454,7 +353,7 @@ def __init__(self, resnet_size, bottleneck, num_classes, num_filters,
 
     if not data_format:
       data_format = (
-          'channels_first' if tf.test.is_built_with_cuda() else 'channels_last')
+        'channels_first' if tf.test.is_built_with_cuda() else 'channels_last')
```
karmel: glint me!

official/resnet/cifar10_main.py
robieta: Perhaps it is better to make the support code just resnet.py as before, and the other file something like layer_definitions.py?

PR #3650, opened 2018-03-19
Use util functions hooks_helper and parser in mnist and wide_deep

k-w-w:


official/wide_deep/wide_deep.py
```diff
@@ -218,5 +203,16 @@ def main(unused_argv):
...
   tf.logging.set_verbosity(tf.logging.INFO)
+
+  parser = parsers.BaseParser()
+  parser.set_defaults(data_dir='/tmp/census_data',
+                      model_dir='/tmp/census_model',
+                      train_epochs=40,
+                      epochs_per_eval=2,
+                      batch_size=40)
+  parser.add_argument(
+      '--model_type', type=str, default='wide_deep',
```
robieta: May as well add an abbreviation (is -mt available?), choices=['wide', 'deep', 'wide_deep'], and metavar="\<MT>" for consistency.

official/wide_deep/wide_deep.py
```diff
@@ -218,5 +203,16 @@ def main(unused_argv):
...
+      help="Valid model types: {'wide', 'deep', 'wide_deep'}.")
```
robieta: Prepend "[default %(default)s] " to the help string.

official/mnist/mnist.py
```diff
@@ -220,48 +231,23 @@ def eval_input_fn():
...
-        choices=['channels_first', 'channels_last'],
-        help='A flag to override the data format used in the model. '
-        'channels_first provides a performance boost on GPU but is not always '
-        'compatible with CPU. If left unspecified, the data format will be '
-        'chosen automatically based on whether TensorFlow was built for CPU or '
-        'GPU.')
     self.add_argument(
         '--export_dir',
         type=str,
         help='The directory where the exported SavedModel will be stored.')
```
robieta: If you prepend "[default: %(default)s] " to the help string argparse will populate it when -h is called. (I've been doing this even when the default is None.) Probably also a good idea to indicate in the help string what the effect of not setting is. (Is the save file just in the model_dir or is it not created?)

official/mnist/mnist.py
```diff
@@ -185,30 +193,33 @@ def main(unused_argv):
...
+  # Set up hook that outputs training logs every 100 steps.
+  train_hooks = hooks_helper.get_train_hooks(
+      FLAGS.hooks, batch_size=FLAGS.batch_size)
+
+  # Train and evaluate model.
+  for n in range(FLAGS.train_epochs // FLAGS.epochs_per_eval):
+    mnist_classifier.train(input_fn=train_input_fn, hooks=train_hooks)
+    eval_results = mnist_classifier.evaluate(input_fn=eval_input_fn)
+    print()
+    print('Evaluation results at epoch %d:\n\t%s\n' % eval_results)
```
robieta: nit: Can we prepend a "\n" and drop the prior print()?

official/wide_deep/wide_deep.py
```diff
@@ -218,5 +203,16 @@ def main(unused_argv):
...
+
```
karmel: For consistency, can we make this a class like the others?

official/mnist/mnist.py
k-w-w: Done

official/mnist/mnist.py
```diff
@@ -220,48 +231,23 @@ def eval_input_fn():
 
 
 class MNISTArgParser(argparse.ArgumentParser):
+  """Argument parser for running MNIST model."""
+  def __init__(self, **kwargs):
+    super(MNISTArgParser, self).__init__(parents=[
+      parsers.BaseParser(**kwargs),
```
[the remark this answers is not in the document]
k-w-w: Whoops, I had begun to make other changes and accidentally left this in. Removed

official/mnist/mnist.py
```diff
@@ -220,48 +231,23 @@ def eval_input_fn():
...
-        '--data_format',
-        type=str,
-        default=None,
-        choices=['channels_first', 'channels_last'],
-        help='A flag to override the data format used in the model. '
-        'channels_first provides a performance boost on GPU but is not always '
-        'compatible with CPU. If left unspecified, the data format will be '
-        'chosen automatically based on whether TensorFlow was built for CPU or '
-        'GPU.')
     self.add_argument(
```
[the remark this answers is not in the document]
k-w-w: if `--export_dir` is set, a SavedModel is saved. Perhaps renaming it to `--export_savedmodel` would be clearer?

official/mnist/mnist.py
k-w-w: Thanks for the suggestions, I've added to the help string

official/mnist/mnist.py
```diff
@@ -220,48 +231,23 @@ def eval_input_fn():
...
         '--export_dir',
         type=str,
         help='The directory where the exported SavedModel will be stored.')
 
+    self.set_defaults(
+        data_dir='/tmp/mnist_data',
+        model_dir='--model_dir',
```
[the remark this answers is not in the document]
k-w-w: Whoooops good catch!

official/wide_deep/wide_deep.py
k-w-w: Done

official/wide_deep/wide_deep.py
k-w-w: Done

official/wide_deep/wide_deep.py
k-w-w: Done

official/mnist/mnist.py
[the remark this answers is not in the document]
robieta: Is there a reason it isn't just automatically saved in the model_dir? (size? wall time for bench marking?) I think it's pretty reasonable to just automatically snapshot since one generally wants the model afterward.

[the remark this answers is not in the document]
karmel: A savedmodel is different than the checkpoint, and most people don't need it. It is used for serving the model with TF Serving or similar.

karmel:
Can we do the MNIST eager parser as well? TPU we can leave as-is for now.

official/mnist/mnist.py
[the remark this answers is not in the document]
robieta: Oh, I see. Thanks for clarifying.

k-w-w:
@karmel Eager argparser added. Some flags in MNIST aren't used in eager, so I wrote MNISTEagerArgParser to inherit directly from parsers.BaseParser and parsers.ImageModelParser, instead of MNISTArgParser.

official/mnist/mnist_eager.py
```diff
@@ -98,20 +100,25 @@ def test(model, dataset):
...
     (device, data_format) = ('/cpu:0', 'channels_last')
+  # If data_format is defined in FLAGS, overwrite automatically set value.
+  if FLAGS.data_format is not None:
+    data_format = data_format
   print('Using device %s, and data format %s.' % (device, data_format))
 
   # Load the datasets
   train_ds = dataset.train(FLAGS.data_dir).shuffle(60000).batch(
-      FLAGS.batch_size)
+      FLAGS.batch_size).repeat(FLAGS.epochs_per_eval)
```
asimshankar: Out of curiosity: Why is there an "epochs_per_eval"? Why does it make sense to go through the eval dataset more than once?

official/mnist/mnist_eager.py
```diff
@@ -148,54 +158,52 @@ def main(_):
...
+        metavar='N',
+        help='[default: %(default)s] batches between logging training status')
+    self.add_argument(
+        '--output_dir', '-od',
+        type=str,
+        default=None,
+        metavar='<OD>',
+        help='[default: %(default)s] Directory to write TensorBoard summaries')
+    self.add_argument(
+        '--lr', '-1r',
```
karmel: Am I seeing things, or is this a one instead of an el?

official/mnist/mnist_eager.py
```diff
@@ -148,54 +158,52 @@ def main(_):
...
-      default=False,
-      help='disables GPU usage even if a GPU is available')
+class MNISTEagerArgParser(argparse.ArgumentParser):
+  """Argument parser for running MNIST model with eager trainng loop."""
+  def __init__(self):
+    super(MNISTEagerArgParser, self).__init__(parents=[
+      parsers.BaseParser(multi_gpu=False, hooks=False),
+      parsers.ImageModelParser()])
+
+    self.set_defaults(
```
karmel: nit: I think we do new args then set_defaults in the other parsers

official/mnist/mnist_eager.py
```diff
@@ -33,8 +33,10 @@
 
 import tensorflow as tf
 import tensorflow.contrib.eager as tfe
+
 from official.mnist import mnist
 from official.mnist import dataset
+from official.utils.arg_parsers import parsers
```
karmel: I think we'll need to add this to the build file as well now.

official/mnist/mnist_eager.py
k-w-w: This is the number of epochs to train between evaluations. 

In retrospect, the naming isn't the clearest. Would `train_eval_interval` make more sense (open to ideas)?

official/mnist/mnist_eager.py
k-w-w: You're right! I'm not sure how that happened... fixed :)

official/mnist/mnist_eager.py
asimshankar: Ah, sorry, somehow missed that this was `train_ds` and not `test_ds`. Anyway, yeah, is there a need to have this be larger than 1? Could we do away with the flag altogether? If not, maybe `--epochs_between_evals` or something?

k-w-w: The flag can be disabled. The flag is mainly there for estimator training, where constantly switching between eval and train requires reloading/saving the last checkpoint is rather slow. The eager training loop doesn't have this issue, so I'll remove the flag for this file. 

I like the name `--epochs_between_eval`!

PR #3651, opened 2018-03-19
Improve directory treatment in ResNet end-to-end tests.

robieta:
Currently /tmp is hard coded in as a location for temp files such as model snapshots and event files. This changes the behavior to use self.get_temp_dir() and tempfile.mkdtemp() which should be more robust.

### DIFF
--- a/official/utils/testing/integration.py
+++ b/official/utils/testing/integration.py
@@ -23,10 +23,10 @@ from __future__ import print_function
 import os
 import shutil
 import sys
-import time
+import tempfile
 
 
-def run_synthetic(main, extra_flags=None):
+def run_synthetic(main, tmp_root, extra_flags=None):
   """Performs a minimal run of a model.
 
     This function is intended to test for syntax errors throughout a model. A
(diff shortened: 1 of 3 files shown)


karmel:
Thanks, merging.

official/utils/testing/integration.py
```diff
@@ -35,14 +35,13 @@ def run_synthetic(main, extra_flags=None):
...
 
   extra_flags = [] if extra_flags is None else extra_flags
 
-  model_dir = "/tmp/model_test_{}".format(hash(time.time()))
-  if os.path.exists(model_dir):
-    shutil.rmtree(model_dir)
+  model_dir = tempfile.mkdtemp(dir=tmp_root)
 
   args = [sys.argv[0], "--model_dir", model_dir, "--train_epochs", "1",
           "--epochs_per_eval", "1", "--use_synthetic_data",
```
karmel: Post-merge realization: git collapsed a tree removal bit below. Do we still want/need that? Seems to not cause problems, but is it redundant?

PR #3653, opened 2018-03-19
ResNet argparse fixes.

robieta:
As noted, imagenet can no longer just use the default number of train_epochs. This PR explicitly sets it back to 100, and also fixes a bug related to "-h" being inadvertently caught by tf.app.run().

### DIFF
--- a/official/resnet/imagenet_main.py
+++ b/official/resnet/imagenet_main.py
@@ -297,6 +297,11 @@ def imagenet_model_fn(features, labels, mode, params):
 def main(argv):
   parser = resnet_run_loop.ResnetArgParser(
       resnet_size_choices=[18, 34, 50, 101, 152, 200])
+
+  parser.set_defaults(
+      train_epochs=100
+  )
+
   flags = parser.parse_args(args=argv[1:])
 
   input_function = flags.use_synthetic_data and get_synth_input_fn() or input_fn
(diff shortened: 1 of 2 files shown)


karmel:
(1) The default should be set for Cifar too.
(2) This is a lot of extra boilerplate to add, just for the sake of running a test, especially considering that most people looking at this code will never see the tests at all. I think either having a parse_args function that gets called before tf.run, or better still just have the test create the class arg parser itself and call main, makes more sense. Remember, most users will just want to read once and run, not understand all of our testing and infra.

robieta:
@karmel The more I look at it I'm inclined to just drop tf.app.run() altogether in favor of just:

```
if __name__ == "__main__":
  main()
```
It appears to be primarily for arg parsing, so we basically have 2 arg parsers stepping on each other's toes. From what I've read it's not a necessity.

robieta:
Removing tf.app.run significantly simplified the amount of "counter magic" needed.

official/resnet/imagenet_main.py
```diff
@@ -281,9 +281,16 @@ def imagenet_model_fn(features, labels, mode, params):
                                          multi_gpu=params['multi_gpu'])
 
 
-def main(argv):
+def main(argv=None):
+  argv = sys.argv if argv is None else argv
```
karmel: Do we still need this at all if we're calling main directly?

robieta: The end to end test passes its own args instead of using sys.argv.

karmel: I think it's more conventional then to have the `if name ==` section pass in sys.argv, rather than having this extra line here.

robieta:
The argv call is now in the ==main section.

PR #3654, opened 2018-03-19
Glint everything

karmel:
So much linting. Please don't introduce any new lint errors after this. Tests will be updated to check as well.


karmel:
In the interest of sharing the pain-- everyone look over the files you touched recently.

official/utils/logging/hooks_helper.py
```diff
@@ -40,7 +40,7 @@ def get_train_hooks(name_list, **kwargs):
     name_list: a list of strings to name desired hook classes. Allowed:
       LoggingTensorHook, ProfilerHook, ExamplesPerSecondHook, which are defined
       as keys in HOOKS
-    kwargs: a dictionary of arguments to the hooks.
+    **kwargs: a dictionary of arguments to the hooks.
```
karmel: This seems silly in the Args, but glint complains otherwise.

official/resnet/imagenet_preprocessing.py
```diff
@@ -204,8 +204,10 @@ def _aspect_preserving_resize(image, resize_min):
 
 
 def _resize_image(image, height, width):
-  """Simple wrapper around tf.resize_images to make sure we use the same
-  `method` and other details each time.
+  """Simple wrapper around tf.resize_images.
+
+  This is primarily to make sure we use the same `method` and other details
```
k-w-w: nit: replace `method` with `ResizeMethod`

official/resnet/resnet_model.py
```diff
@@ -190,13 +192,29 @@ def _building_block_v2(inputs, filters, training, projection_shortcut, strides,
...
+    training: A Boolean for whether the model is in training or inference
+      mode. Needed for batch normalization.
+    projection_shortcut: The function to use for projection shortcuts
+      (typically a 1x1 convolution when downsampling the input).
+    strides: The block's stride. If greater than 1, this block will ultimately
+      downsample the input.
+    data_format: The input format ('channels_last' or 'channels_first').
+
+  Returns:
+    The output tensor of the block.
```
k-w-w: + comment that the output shape should be the same as the inputs

official/mnist/dataset.py
```diff
@@ -36,7 +36,7 @@ def check_image_file_header(filename):
   """Validate that filename corresponds to images for the MNIST dataset."""
   with tf.gfile.Open(filename, 'rb') as f:
     magic = read32(f)
-    num_images = read32(f)
+    num_images = read32(f)  # pylint: disable=unused-variable
```
asimshankar: Would it be better to write this as:

```python
read32(f) # num_images, ignored
```

And same for `num_items` below?

official/utils/logging/hooks_helper.py
yhliang2018: Yeah, I got the same warning when glinting it . Will follow this style later. LGTM, thanks.

official/mnist/dataset.py
```diff
@@ -52,7 +52,7 @@ def check_labels_file_header(filename):
   """Validate that filename corresponds to labels for the MNIST dataset."""
   with tf.gfile.Open(filename, 'rb') as f:
     magic = read32(f)
-    num_items = read32(f)
+    num_items = read32(f)  # pylint: disable=unused-variable
```
qlzh727: I think instead of put a suppression of lint here, you can just rename the variable as unused_num_items and I think it will make the code more readable, and also make the lint warning go away.

official/mnist/mnist.py
```diff
@@ -137,27 +137,32 @@ def model_fn(features, labels, mode, params):
...
   so doing the work here until that feature is in place.
+
+  Args:
+    batch_size: the number of examples processed in each training batch.
+
+  Raises:
+    ValueError: if no GPUs are found, or selected batch_size is invalid.
   """
-  from tensorflow.python.client import device_lib
+  from tensorflow.python.client import device_lib  # pylint: disable=g-import-not-at-top
```
qlzh727: Is there any reason that we have to do a local import and not put it on top?

official/mnist/mnist_eager.py
```diff
@@ -31,10 +31,11 @@
...
 import time
 
-import tensorflow as tf
-import tensorflow.contrib.eager as tfe
+import tensorflow as tf  # pylint: disable=g-bad-import-order
+import tensorflow.contrib.eager as tfe  # pylint: disable=g-bad-import-order
+
+from official.mnist import dataset as mnist_dataset
 from official.mnist import mnist
-from official.mnist import dataset
```
qlzh727: I think glint will give u a snippet to paste for incorrect order, should we do that instead of put a lint suppression here?

official/mnist/mnist.py
karmel: We want to hide this code, as it should not be exposed to the user, and we want to remove it as soon as it's done by Estimator directly. So, for now, leaving bundled.

official/mnist/mnist_eager.py
karmel: The problem is TF versus official; alphabetically TF should come after official, and to Google, their both the same, but for us, TF is third party.

official/mnist/dataset.py
karmel: Solved slightly differently, as per @asimshankar 's comments above.

official/wide_deep/wide_deep.py
```diff
@@ -176,17 +176,25 @@ def main(_):
...
+
+  train_hooks = hooks_helper.get_train_hooks(
+      FLAGS.hooks, batch_size=FLAGS.batch_size,
+      tensors_to_log={'average_loss': 'head/truediv',
+                      'loss': 'head/weighted_loss/Sum'})
 
-  for n in range(FLAGS.train_epochs // FLAGS.epochs_per_eval):
-    model.train(input_fn=train_input_fn)
+  # Train and evaluate the model every `FLAGS.epochs_between_evals` epochs.
+  for n in range(FLAGS.train_epochs // FLAGS.epochs_between_evals):
```
k-w-w: nit: remove extra space

karmel:
I added an rcfile that we will be able to run with pylint:

```
RCFILE="utils/testing/pylint.rcfile"

for file in `find . -name '*.py' ! -name '*test.py' -print`
do
  echo "Linting ${file}"
  pylint --rcfile=$RCFILE $file
done

# More lenient for test files.
for file in `find . -name '*test.py' -print`
do
  echo "Linting ${file}"
  pylint --rcfile=$RCFILE --disable=missing-docstring,protected-access $file
done
```

PR #3669, opened 2018-03-20
syntax error : download_and_convert_VOC2012.sh

sm176357:
Probable syntax error in line 43. Remove "function" for execution.


sm176357:
I signed it!

YknZhu:
I believe function keyword should work fine for bash?

PR #3672, opened 2018-03-20
Add session hook for benchmark metric logging.

qlzh727:
Current hook is very similar as the LoggingTensorHook. Some of the
function are directly copied since the original one was not
exposed for import. We should seek to eventually move this code to
core when it is mature enough.

### DIFF
--- a/official/utils/logging/metric_hook.py
+++ b/official/utils/logging/metric_hook.py
@@ -83,7 +83,7 @@ class LoggingMetricHook(tf.train.LoggingTensorHook):
           "Global step should be created to use LoggingMetricHook.")
     if self._global_step_tensor.name not in self._current_tensors:
       self._current_tensors[self._global_step_tensor.name] = (
-        self._global_step_tensor)
+          self._global_step_tensor)
 
   def after_run(self, unused_run_context, run_values):
     # should_trigger is a internal state that populated at before_run, and it is
--- a/official/utils/logging/metric_hook.py
+++ b/official/utils/logging/metric_hook.py
@@ -19,17 +19,18 @@ from __future__ import division
 from __future__ import print_function
 
 import tensorflow as tf
+from tensorflow.python.framework import ops
 
 from official.utils.logging import logger
-
-GLOBAL_STEP_TENSOR_NAME = "global_step"
 
 
 class LoggingMetricHook(tf.train.LoggingTensorHook):
   """Hook to log benchmark metric information.
 
   This hook is very similar as tf.train.LoggingTensorHook, which logs given
-  tensors every N local steps, every N seconds, or at the end.
+  tensors every N local steps, every N seconds, or at the end. The metric
+  information will be logged to given log_dir or via metric_logger in JSON
+  format, which can be consumed by data analysis pipeline later.
 
   Note that if `at_end` is True, `tensors` should not include any tensor
   whose evaluation produces a side effect such as consuming additional inputs.
@@ -43,9 +44,9 @@ class LoggingMetricHook(tf.train.LoggingTensorHook):
       tensors: `dict` that maps string-valued tags to tensors/tensor names,
           or `iterable` of tensors/tensor names.
       log_dir: `string`, directory path that metric hook should write log to.
-      metric_logger: `BenchmarkLogger`, the benchmark logger that hook should
-          use to write the log. Exactly one of the `log_dir` and `metric_logger`
-          should be provided.
+      metric_logger: instance of `BenchmarkLogger`, the benchmark logger that
+          hook should use to write the log. Exactly one of the `log_dir` and
+          `metric_logger` should be provided.
       every_n_iter: `int`, print the values of `tensors` once every N local
           steps taken on the current worker.
       every_n_secs: `int` or `float`, print the values of `tensors` once every N
@@ -77,12 +78,16 @@ class LoggingMetricHook(tf.train.LoggingTensorHook):
 
   def begin(self):
     super(LoggingMetricHook, self).begin()
-    if tf.train.get_global_step() is None:
+    global_step_tensor = tf.train.get_global_step()
+    if global_step_tensor is None:
       raise RuntimeError(
           "Global step should be created to use LoggingMetricHook.")
-    self._current_tensors[GLOBAL_STEP_TENSOR_NAME] = tf.train.get_global_step()
+    if not self._current_tensors.has_key(ops.GraphKeys.GLOBAL_STEP):
+      self._current_tensors[ops.GraphKeys.GLOBAL_STEP] = global_step_tensor
 
   def after_run(self, unused_run_context, run_values):
+    # should_trigger is a internal state that populated at before_run, and it is
+    # using self_timer to determine whether it should trigger.
     if self._should_trigger:
       self._log_metric(run_values.results)
 
@@ -95,6 +100,7 @@ class LoggingMetricHook(tf.train.LoggingTensorHook):
 
   def _log_metric(self, tensor_values):
     self._timer.update_last_triggered_step(self._iter_count)
-    global_step = tensor_values[GLOBAL_STEP_TENSOR_NAME]
+    global_step = tensor_values[ops.GraphKeys.GLOBAL_STEP]
+    # self._tag_order is populated during the init of LoggingTensorHook
     for tag in self._tag_order:
       self._logger.log_metric(tag, tensor_values[tag], global_step=global_step)
--- a/official/utils/logging/metric_hook.py
+++ b/official/utils/logging/metric_hook.py
@@ -19,7 +19,6 @@ from __future__ import division
 from __future__ import print_function
 
 import tensorflow as tf
-from tensorflow.python.framework import ops
 
 from official.utils.logging import logger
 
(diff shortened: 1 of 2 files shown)


official/utils/logging/metric_hook.py
```diff
@@ -0,0 +1,183 @@
...
+# See the License for the specific language governing permissions and
+# limitations under the License.
+# ==============================================================================
+"""Session hook for logging benchmark metric."""
+
+from __future__ import absolute_import
+from __future__ import division
+from __future__ import print_function
+
+from official.utils.logging import logger
```
karmel: nit: local (official) imports should go below the third party.

qlzh727:
Good point. Update to inherit LoggingTensorHook as parent.

official/utils/logging/metric_hook.py
```diff
@@ -0,0 +1,100 @@
...
+
+GLOBAL_STEP_TENSOR_NAME = "global_step"
+
+
+class LoggingMetricHook(tf.train.LoggingTensorHook):
+  """Hook to log benchmark metric information.
+
+  This hook is very similar as tf.train.LoggingTensorHook, which logs given
+  tensors every N local steps, every N seconds, or at the end.
+
```
karmel: Some details on how it is different/what it is used for?

official/utils/logging/metric_hook.py
```diff
@@ -0,0 +1,100 @@
...
+  def __init__(self, tensors, log_dir=None, metric_logger=None,
+               every_n_iter=None, every_n_secs=None, at_end=False):
+    """Initializer for LoggingMetricHook.
+
+    Args:
+      tensors: `dict` that maps string-valued tags to tensors/tensor names,
+          or `iterable` of tensors/tensor names.
+      log_dir: `string`, directory path that metric hook should write log to.
+      metric_logger: `BenchmarkLogger`, the benchmark logger that hook should
```
karmel: An instance of BL or a class?

official/utils/logging/metric_hook.py
```diff
@@ -0,0 +1,100 @@
...
+          "exactly one of log_dir and metric_logger should be provided.")
+
+    if log_dir is not None:
+      self._logger = logger.BenchmarkLogger(log_dir)
+    else:
+      self._logger = metric_logger
+
+  def begin(self):
+    super(LoggingMetricHook, self).begin()
+    if tf.train.get_global_step() is None:
```
karmel: I suspect the graph can optimize this out, but, for clarity-- maybe call get_global_step once, assigned to a var, and then check and use that var below?

official/utils/logging/metric_hook.py
```diff
@@ -0,0 +1,100 @@
...
+      raise RuntimeError(
+          "Global step should be created to use LoggingMetricHook.")
+    self._current_tensors[GLOBAL_STEP_TENSOR_NAME] = tf.train.get_global_step()
+
+  def after_run(self, unused_run_context, run_values):
+    if self._should_trigger:
```
karmel: Comments for all these methods would be helpful, since most people won't know what the parent class does.

official/utils/logging/metric_hook.py
```diff
@@ -0,0 +1,100 @@
...
+
+  def end(self, session):
+    if self._log_at_end:
+      values = session.run(self._current_tensors)
+      self._log_metric(values)
+
+  def _log_metric(self, tensor_values):
+    self._timer.update_last_triggered_step(self._iter_count)
+    global_step = tensor_values[GLOBAL_STEP_TENSOR_NAME]
+    for tag in self._tag_order:
```
karmel: Probably worth noting that this comes from LoggingTensorHook, which captures the keys of `tensors` during init.

official/utils/logging/metric_hook.py
```diff
@@ -0,0 +1,100 @@
...
+      self._logger.log_metric(tag, tensor_values[tag], global_step=global_step)
```
karmel: Performance-wise, is it worth considering having the BenchmarkLogger do a batch-write of all the tensors in question rather than one at a time? Or does that not make a difference?

official/utils/logging/metric_hook_test.py
```diff
@@ -0,0 +1,204 @@
...
+
+from __future__ import absolute_import
+from __future__ import division
+from __future__ import print_function
+
+import tempfile
+import time
+
+from official.utils.logging import metric_hook
+import tensorflow as tf
```
karmel: import order

official/utils/logging/metric_hook_test.py
```diff
@@ -0,0 +1,204 @@
...
+from tensorflow.python.client import session as session_lib
+from tensorflow.python.framework import constant_op
+from tensorflow.python.framework import ops
+from tensorflow.python.ops import variables as variables_lib
+from tensorflow.python.training import monitored_session
```
karmel: I haven't validated this for all, but certainly many of the classes/functions used below that I see is available from the top-level `tf` import. Let's stick with that unless absolutely necessary.

official/utils/logging/metric_hook_test.py
```diff
@@ -0,0 +1,204 @@
...
+      tf.train.get_or_create_global_step()
+      t = constant_op.constant(42.0, name='foo')
+      train_op = constant_op.constant(3)
+      hook = metric_hook.LoggingMetricHook(
+          tensors=[t.name], at_end=True, metric_logger=self._logger)
+      hook.begin()
+      mon_sess = monitored_session._HookedSession(sess, [hook])
+      sess.run(variables_lib.global_variables_initializer())
+
+      # metric_log = os.path.join(self.log_dir, "metric.log")
```
karmel: nit: remove debugging lines

official/utils/logging/metric_hook.py
```diff
@@ -0,0 +1,100 @@
...
+      self._logger = logger.BenchmarkLogger(log_dir)
+    else:
+      self._logger = metric_logger
+
+  def begin(self):
+    super(LoggingMetricHook, self).begin()
+    if tf.train.get_global_step() is None:
+      raise RuntimeError(
+          "Global step should be created to use LoggingMetricHook.")
+    self._current_tensors[GLOBAL_STEP_TENSOR_NAME] = tf.train.get_global_step()
```
karmel: Will this create a problem if someone happens to pass in the global step tensor, or a tensor by the same name? TF is often more cautious about placeholder vars that it adds in-- see, for example, https://github.com/tensorflow/tensorflow/blob/r1.6/tensorflow/python/estimator/inputs/numpy_io.py#L51 .

official/utils/logging/metric_hook_test.py
```diff
@@ -0,0 +1,204 @@
...
+      self._validate_print_every_n_steps(sess, at_end=True)
+      # Verify proper reset.
+      self._validate_print_every_n_steps(sess, at_end=True)
+
+  def _validate_print_every_n_secs(self, sess, at_end):
+    t = constant_op.constant(42.0, name='foo')
+    train_op = constant_op.constant(3)
+
+    hook = metric_hook.LoggingMetricHook(
+        tensors=[t.name], every_n_secs=1.0, at_end=at_end,
```
karmel: Can we make sure to test the case of actually having multiple tensors passed in, and also passed in as a dict? Also, we use tensor names in all of these tests; let's make sure to test also with the tensors themselves, which are allowed according to the docstring.

official/utils/logging/metric_hook.py
qlzh727: Chatted offline. we will see if we hit any usage problem in future. I will leave this as is for the moment.

official/utils/logging/metric_hook_test.py
qlzh727: Done. Thanks for the suggestion.

official/utils/logging/metric_hook_test.py
qlzh727: Done. Added test with multi tensors.

official/utils/logging/metric_hook.py
```diff
@@ -0,0 +1,106 @@
...
+      self._log_metric(run_values.results)
+
+    self._iter_count += 1
+
+  def end(self, session):
+    if self._log_at_end:
+      values = session.run(self._current_tensors)
+      self._log_metric(values)
+
+  def _log_metric(self, tensor_values):
```
robieta: It would be nice to sync timestamps for different tensors in the same measurement. Right now the times are slightly off which could be annoying later.

```
{"name": "train_accuracy",  "timestamp": "2018-03-21T12:40:03.460442Z", ... "global_step": 33}
{"name": "learning_rate",   "timestamp": "2018-03-21T12:40:03.460687Z", ... "global_step": 33}
```

official/utils/logging/metric_hook.py
```diff
@@ -0,0 +1,106 @@
...
+      self._logger = metric_logger
+
+  def begin(self):
+    super(LoggingMetricHook, self).begin()
+    global_step_tensor = tf.train.get_global_step()
+    if global_step_tensor is None:
+      raise RuntimeError(
+          "Global step should be created to use LoggingMetricHook.")
+    if not self._current_tensors.has_key(ops.GraphKeys.GLOBAL_STEP):
+      self._current_tensors[ops.GraphKeys.GLOBAL_STEP] = global_step_tensor
```
karmel: I think you should be able to just say global_step_tensor.name here, which also allows us to remove the dependency on ops.

karmel:
Ah, looks like some py3 errors as well; should be easy to resolve.

official/utils/logging/metric_hook.py
qlzh727: Done.

official/utils/logging/metric_hook.py
```diff
@@ -0,0 +1,106 @@
...
+    else:
+      self._logger = metric_logger
+
+  def begin(self):
+    super(LoggingMetricHook, self).begin()
+    global_step_tensor = tf.train.get_global_step()
+    if global_step_tensor is None:
+      raise RuntimeError(
+          "Global step should be created to use LoggingMetricHook.")
+    if not self._current_tensors.has_key(ops.GraphKeys.GLOBAL_STEP):
```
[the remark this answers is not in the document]
qlzh727: Done.

official/utils/logging/metric_hook.py
qlzh727: Understood, I think Karmel also had similar comment about bulk logging metrics. Currently we still can align them via global_step. Will address this when in future change.

PR #3674, opened 2018-03-20
fix shell function declaration

kr-ish:
Fixes broken shell function declaration. See https://stackoverflow.com/a/6212408/6457747


kr-ish:
duplicate of #3669

PR #3675, opened 2018-03-20
Use reduce_mean instead of Average Pooling in Resnet

karmel:
This gives us an extra 0.5% - 1.5% examples/second, depending on the environment, and should not in theory change the model architecture itself. Final accuracy for ImageNet was unchanged at 76.38%, CIFAR-10 92.5%.

CC @bignamehyp @reedwm


k-w-w:
Nice!
Can you rename `final_avg_pool` to `final_reduce_mean` in [imagenet_test.py](https://github.com/tensorflow/models/blob/master/official/resnet/imagenet_test.py#L67) as well?

karmel:
Nice catch. Tests fixed. (Linting being fixed in #3676 .)

PR #3676, opened 2018-03-20
A few linting fixes

karmel:
Not sure why, but Kokoro's pylint is complaining about a few method names that were okay locally. Fixing the rcfile here.


qlzh727:
Can u point me to the kokoro run that complains about the lint error?

PR #3679, opened 2018-03-21
Update Installation.md

harshul1610:
Cython is must installed package for building cocoapi. else the build will not initiate.


harshul1610:
I signed it!

PR #3684, opened 2018-03-21
Arg parsing cleanup for MNIST and Wide-Deep

robieta:
This PR replaces tf.app.run() with a simple main(argv) following what has already been done with resnet. It also replaces the parser.parse_known_args() with parser.parse_args() calls so that mistakes in typing args raise errors.

### DIFF
--- a/official/mnist/mnist.py
+++ b/official/mnist/mnist.py
@@ -175,11 +175,14 @@ def validate_batch_size_for_multi_gpu(batch_size):
     raise ValueError(err)
 
 
-def main(_):
+def main(argv):
+  parser = MNISTArgParser()
+  flags = parser.parse_args(args=argv[1:])
+
   model_function = model_fn
 
-  if FLAGS.multi_gpu:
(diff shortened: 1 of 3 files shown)


karmel:
@qlzh727 @k-w-w Can you take a look? Removing myself as reviewer here.

official/mnist/mnist.py
```diff
@@ -261,6 +264,4 @@ def __init__(self):
 
 if __name__ == '__main__':
   tf.logging.set_verbosity(tf.logging.INFO)
-  parser = MNISTArgParser()
-  FLAGS, unparsed = parser.parse_known_args()
-  tf.app.run(main=main, argv=[sys.argv[0]] + unparsed)
```
qlzh727: I am bit hesitate to get rid of tf.app.run(). Usually app.run contains most of the init code that is common to all the models, and it makes the lib owner easily to add a common dependencies without modifying all the client code.

robieta: Discussed offline.

PR #3687, opened 2018-03-21
Add fp16 support to official ResNet.

robieta:
This PR adds fp16 support to official/resnet. The vast majority of the work was already done by Reed a month ago (including wonderful comments), and this just rolls those changes into the current master.

Currently training is I/O bound; however synthetic data runs confirm the fp16 accelleration to ~4000 images/sec during training.

I will update when I have run results.

### DIFF
--- a/official/resnet/cifar10_main.py
+++ b/official/resnet/cifar10_main.py
@@ -145,7 +145,8 @@ class Cifar10Model(resnet_model.Model):
   """Model class with appropriate defaults for CIFAR-10 data."""
 
   def __init__(self, resnet_size, data_format=None, num_classes=_NUM_CLASSES,
-               version=resnet_model.DEFAULT_VERSION, dtype=None):
+               version=resnet_model.DEFAULT_VERSION,
+               dtype=resnet_model.DEFAULT_DTYPE):
     """These are the parameters that work for CIFAR-10 data.
 
     Args:
@@ -208,8 +209,10 @@ def cifar10_model_fn(features, labels, mode, params):
     return True
 
   return resnet_run_loop.resnet_model_fn(
-      dtype=params['dtype'],
-      features=features, labels=labels, mode=mode, model_class=Cifar10Model,
+      features=features,
+      labels=labels,
+      mode=mode,
+      model_class=Cifar10Model,
       resnet_size=params['resnet_size'],
       weight_decay=weight_decay,
       learning_rate_fn=learning_rate_fn,
@@ -218,7 +221,8 @@ def cifar10_model_fn(features, labels, mode, params):
       version=params['version'],
       loss_scale=params['loss_scale'],
       loss_filter_fn=loss_filter_fn,
-      multi_gpu=params['multi_gpu']
+      multi_gpu=params['multi_gpu'],
+      dtype=params['dtype']
   )
 
 
--- a/official/resnet/cifar10_main.py
+++ b/official/resnet/cifar10_main.py
@@ -145,7 +145,7 @@ class Cifar10Model(resnet_model.Model):
   """Model class with appropriate defaults for CIFAR-10 data."""
 
   def __init__(self, resnet_size, data_format=None, num_classes=_NUM_CLASSES,
-               version=resnet_model.DEFAULT_VERSION):
+               version=resnet_model.DEFAULT_VERSION, dtype=None):
     """These are the parameters that work for CIFAR-10 data.
 
     Args:
@@ -156,6 +156,7 @@ class Cifar10Model(resnet_model.Model):
         enables users to extend the same model to their own datasets.
       version: Integer representing which version of the ResNet network to use.
         See README for details. Valid values: [1, 2]
+      dtype: The TensorFlow dtype to use for calculations.
 
     Raises:
       ValueError: if invalid resnet_size is chosen
@@ -180,7 +181,9 @@ class Cifar10Model(resnet_model.Model):
         block_strides=[1, 2, 2],
         final_size=64,
         version=version,
-        data_format=data_format)
+        data_format=data_format,
+        dtype=dtype
+    )
 
 
 def cifar10_model_fn(features, labels, mode, params):
@@ -204,15 +207,19 @@ def cifar10_model_fn(features, labels, mode, params):
   def loss_filter_fn(_):
     return True
 
-  return resnet_run_loop.resnet_model_fn(features, labels, mode, Cifar10Model,
-                                         resnet_size=params['resnet_size'],
-                                         weight_decay=weight_decay,
-                                         learning_rate_fn=learning_rate_fn,
-                                         momentum=0.9,
-                                         data_format=params['data_format'],
-                                         version=params['version'],
-                                         loss_filter_fn=loss_filter_fn,
-                                         multi_gpu=params['multi_gpu'])
+  return resnet_run_loop.resnet_model_fn(
+      dtype=params['dtype'],
+      features=features, labels=labels, mode=mode, model_class=Cifar10Model,
+      resnet_size=params['resnet_size'],
+      weight_decay=weight_decay,
+      learning_rate_fn=learning_rate_fn,
+      momentum=0.9,
+      data_format=params['data_format'],
+      version=params['version'],
+      loss_scale=params['loss_scale'],
+      loss_filter_fn=loss_filter_fn,
+      multi_gpu=params['multi_gpu']
+  )
 
 
 def main(argv):
--- a/official/resnet/resnet_model.py
+++ b/official/resnet/resnet_model.py
@@ -36,6 +36,8 @@ import tensorflow as tf
 _BATCH_NORM_DECAY = 0.997
 _BATCH_NORM_EPSILON = 1e-5
 DEFAULT_VERSION = 2
+DEFAULT_DTYPE = tf.float32
+CASTABLE_TYPES = (tf.float16,)
 
 
 ################################################################################
@@ -352,7 +354,7 @@ class Model(object):
                conv_stride, first_pool_size, first_pool_stride,
                second_pool_size, second_pool_stride, block_sizes, block_strides,
                final_size, version=DEFAULT_VERSION, data_format=None,
-               dtype=None):
+               dtype=DEFAULT_DTYPE):
     """Creates a model for classifying an image.
 
     Args:
@@ -421,13 +423,13 @@ class Model(object):
     self.block_sizes = block_sizes
     self.block_strides = block_strides
     self.final_size = final_size
-    self.dtype = dtype or tf.float32
+    self.dtype = dtype
 
-  def _custom_getter(self, getter, name, shape=None, dtype=tf.float32,
-                     *args, **kwargs):
+  def _custom_dtype_getter(self, getter, name, shape=None, dtype=DEFAULT_DTYPE,
+                           *args, **kwargs):
     """Creates variables in fp32, then casts to fp16 if necessary.
 
-      This function is a custom getter. A custom getter is a function with the
+    This function is a custom getter. A custom getter is a function with the
     same signature as tf.get_variable, except it has an additional getter
     parameter. Custom getters can be passed as the `custom_getter` parameter of
     tf.variable_scope. Then, tf.get_variable will call the custom getter,
@@ -436,20 +438,21 @@ class Model(object):
     The `getter` parameter is the underlying variable getter, that would have
     been called if no custom getter was used. Custom getters typically get a
     variable with `getter`, then modify it in some way.
-    This custom getter will create an fp32 variable if an fp16 variable was
-    requested. It will then cast the variable to fp16 and return the result.
-    The reason we do not directly create variables in fp16 is that applying
-    small gradients to fp16 variables may cause the variable not to change. This
-    is because fp16 variables have very low precision.
+
+    This custom getter will create an fp32 variable. If an low precision
+    (e.g. float16) variable was requested it will then cast the variable to the
+    requested dtype. The reason we do not directly create variables in low
+    precision dtypes is that applying small gradients to such variables may
+    cause the variable not to change.
 
     Args:
       getter: The underlying variable getter, that has the same signature as
         tf.get_variable and returns a variable.
       name: The name of the variable to get.
       shape: The shape of the variable to get.
-      dtype: The dtype of the variable to get. Note that if this if tf.float16,
-        the variable will be created as a tf.float32 variable, then casted to
-        tf.float16
+      dtype: The dtype of the variable to get. Note that if this is a low
+        precision dtype, the variable will be created as a tf.float32 variable,
+        then cast to the appropriate dtype
       *args: Additional arguments to pass unmodified to getter.
       **kwargs: Additional keyword arguments to pass unmodified to getter.
 
@@ -457,24 +460,24 @@ class Model(object):
       A variable which is cast to fp16 if necessary.
     """
 
-    if dtype == tf.float16:
+    if dtype in CASTABLE_TYPES:
       var = getter(name, shape, tf.float32, *args, **kwargs)
-      return tf.cast(var, tf.float16, name=name + '_casted')
+      return tf.cast(var, dtype=dtype, name=name + '_cast')
     else:
       return getter(name, shape, dtype, *args, **kwargs)
 
   def _model_variable_scope(self):
     """Returns a variable scope that the model should be created under.
 
-    If self.use_fp16 is True, model variable will be created in fp32 then casted
-    to fp16 before being used.
+    If self.dtype is a castable type, model variable will be created in fp32
+    then cast to self.dtype before being used.
 
     Returns:
       A variable scope for the model.
     """
 
-    custom_getter = self._custom_getter if self.dtype == tf.float16 else None
-    return tf.variable_scope('resnet_model', custom_getter=custom_getter)
+    return tf.variable_scope('resnet_model',
+                             custom_getter=self._custom_dtype_getter)
 
   def __call__(self, inputs, training):
     """Add operations to classify a batch of input images.
--- a/official/resnet/resnet_model.py
+++ b/official/resnet/resnet_model.py
@@ -413,7 +413,7 @@ class Model(object):
         self.block_fn = _building_block_v2
 
     if dtype not in ALLOWED_TYPES:
-      raise ValueError("dtype must be one of: {}".format(LEGAL_TYPES))
+      raise ValueError('dtype must be one of: {}'.format(ALLOWED_TYPES))
 
     self.data_format = data_format
     self.num_classes = num_classes
--- a/official/resnet/resnet_model.py
+++ b/official/resnet/resnet_model.py
@@ -38,6 +38,7 @@ _BATCH_NORM_EPSILON = 1e-5
 DEFAULT_VERSION = 2
 DEFAULT_DTYPE = tf.float32
 CASTABLE_TYPES = (tf.float16,)
+ALLOWED_TYPES = (tf.float32,) + CASTABLE_TYPES
 
 
 ################################################################################
@@ -410,6 +411,9 @@ class Model(object):
         self.block_fn = _building_block_v1
       else:
         self.block_fn = _building_block_v2
+
+    if dtype not in ALLOWED_TYPES:
+      raise ValueError("dtype must be one of: {}".format(LEGAL_TYPES))
 
     self.data_format = data_format
     self.num_classes = num_classes
--- a/official/resnet/resnet_model.py
+++ b/official/resnet/resnet_model.py
@@ -38,7 +38,7 @@ _BATCH_NORM_EPSILON = 1e-5
 DEFAULT_VERSION = 2
 DEFAULT_DTYPE = tf.float32
 CASTABLE_TYPES = (tf.float16,)
-ALLOWED_TYPES = (tf.float32,) + CASTABLE_TYPES
+ALLOWED_TYPES = (DEFAULT_DTYPE,) + CASTABLE_TYPES
 
 
 ################################################################################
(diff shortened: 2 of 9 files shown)


karmel:
Quick thought before a more full review: there is a future in which we support a number of different quantization options, at least for some models. Is it feasible generalize this to a dtype being passed around rather than fixing to only fp16? You don't need to handle multiple dtypes for now, but at least not fix all the params and names to "fp16".

tfboyd:
yes, that is how the other platforms handle it.  I doubt we would handle
int8 but you never know and maybe.  passing the dtype is popular.

robieta:
My understanding is that once automatic fp16 scaling is ready all of this will be ripped out, and at that point passing dtype will be natural. If you're worried about the interface changing it's no trouble to have dtype as the CLI flag. The only thing that might be weird is loss_scaling, as that would still have to be an fp16 CLI flag.

karmel:
fp16 today, int8 tomorrow. Let's generalize to dtype. Then allow for a loss_scale to be passed in for the selected dtype. If not passed in, select from a dict of defaults based on the selected dtype? Open to pros and cons on that, but I think we should anticipate more dtypes rather than fixing on the one currently ready.

tfboyd:
+1 to karmel's idea.  That matches or is better than what I have seen from
other platforms.

robieta:
The interface now uses dtype and loss_scale instead of fp16 and fp16_loss_scale.

I did a training run on the V100's, and got 75.87%. I also did 3 runs with master (I had to use a smaller batch size due to OOM's on P100's) and got a mean of 75.75% and a std of 0.28%. So the use of fp16 does not seem to affect accuracy.

Also @karmel, this PR conflicts with your checkpoint PR because it changes the namespace with the custom getter, so we will need to coordinate those two PR's.

official/resnet/resnet_model.py
```diff
@@ -418,6 +421,60 @@ def __init__(self, resnet_size, bottleneck, num_classes, num_filters,
...
+      *args: Additional arguments to pass unmodified to getter.
+      **kwargs: Additional keyword arguments to pass unmodified to getter.
+
+    Returns:
+      A variable which is cast to fp16 if necessary.
+    """
+
+    if dtype == tf.float16:
+      var = getter(name, shape, tf.float32, *args, **kwargs)
+      return tf.cast(var, tf.float16, name=name + '_casted')
```
karmel: nit: use dtype in the cast rather than tf.float16 for greater reusability.

official/resnet/resnet_model.py
```diff
@@ -431,46 +488,46 @@ def __call__(self, inputs, training):
...
+      # ResNet does an Average Pooling layer over pool_size,
+      # but that is the same as doing a reduce_mean. We do a reduce_mean
+      # here because it performs better than AveragePooling2D.
+      axes = [2, 3] if self.data_format == 'channels_first' else [1, 2]
+      inputs = tf.reduce_mean(inputs, axes, keepdims=True)
+      inputs = tf.identity(inputs, 'final_reduce_mean')
+
+      inputs = tf.reshape(inputs, [-1, self.final_size])
+      inputs = tf.layers.dense(inputs=inputs, units=self.num_classes)
+      inputs = tf.identity(inputs, 'final_dense')
```
karmel: Just to confirm given Github... the only change in this block is the addition of the variable scope and the indent, correct?

official/resnet/resnet_run_loop.py
```diff
@@ -181,6 +182,7 @@ def resnet_model_fn(features, labels, mode, model_class,
   a train op, but with the necessary parameters for the given mode.
 
   Args:
+    dtype: the TensorFlow dtype to use for calculations.
```
karmel: As above: dtype at the end for the sake of convention?

official/resnet/resnet_run_loop.py
```diff
@@ -196,6 +198,7 @@ def resnet_model_fn(features, labels, mode, model_class,
       If set to None, the format is dependent on whether a GPU is available.
     version: Integer representing which version of the ResNet network to use.
       See README for details. Valid values: [1, 2]
+    loss_scale: The factor to scale the loss for numerical stability.
```
karmel: Please describe further, even if just to say, see the arg parser.

official/resnet/resnet_run_loop.py
```diff
@@ -211,9 +214,17 @@ def resnet_model_fn(features, labels, mode, model_class,
   # Generate a summary node for the images
   tf.summary.image('images', features, max_outputs=6)
 
-  model = model_class(resnet_size, data_format, version=version)
+  if dtype == tf.float16:
+    features = tf.cast(features, tf.float16)
```
karmel: As above: I think we can get away with just doing the cast to dtype? Check speed and check with @reedwm , but my guess is a no-op cast is fine, and saves us from conditionals. If that's the case, make sure to add a comment about why we are casting to avoid someone in the future thinking, gee, this is a no-op cast, because the features and dtype are both fp32.

official/resnet/resnet_run_loop.py
```diff
@@ -211,9 +214,17 @@ def resnet_model_fn(features, labels, mode, model_class,
...
+
+  model = model_class(resnet_size, data_format, version=version, dtype=dtype)
+
   logits = model(features, mode == tf.estimator.ModeKeys.TRAIN)
 
+  # This acts as a no-op if the logits are already in fp32. If dtype is
+  # tf.float16, logits must be casted to fp32 for numerical stability.
```
karmel: nit: must be cast to...

official/resnet/resnet_run_loop.py
```diff
@@ -244,7 +255,7 @@ def exclude_batch_norm(name):
 
   # Add weight decay to the loss.
   l2_loss = weight_decay * tf.add_n(
-      [tf.nn.l2_loss(v) for v in tf.trainable_variables()
+      [tf.nn.l2_loss(tf.cast(v, tf.float32)) for v in tf.trainable_variables()
```
karmel: Comment as to why we must explicitly cast here and not elsewhere?

official/resnet/resnet_run_loop.py
```diff
@@ -266,8 +277,19 @@ def exclude_batch_norm(name):
...
       optimizer = tf.contrib.estimator.TowerOptimizer(optimizer)
 
+    if loss_scale != 1:
+      # When computing fp16 gradients, often intermediate tensor values are
+      # so small, they underflow to 0. To avoid this, we multiply the loss by
+      # loss_scale to make these tensor values loss_scales times bigger.
+      scaled_grad_vars = optimizer.compute_gradients(loss * loss_scale)
+      unscaled_grad_vars = [(grad / loss_scale, var)
+                            for grad, var in scaled_grad_vars]
+      minimize_op = optimizer.apply_gradients(unscaled_grad_vars, global_step)
```
karmel: Extract to separate function and test? This seems like a good candidate for something the optimizer should handle itself. Consider adding to core? I would suggest you add a bug, self-assign, and see if this is something that is generalizable and of general interest.

official/resnet/resnet_run_loop.py
```diff
@@ -360,16 +382,18 @@ def resnet_main(flags, model_function, input_function, shape=None):
   classifier = tf.estimator.Estimator(
       model_fn=model_function, model_dir=flags.model_dir, config=run_config,
       params={
+          'dtype': flags.dtype,
```
karmel: nit: at the end.

official/resnet/resnet_run_loop.py
```diff
@@ -266,8 +277,19 @@ def exclude_batch_norm(name):
     if multi_gpu:
       optimizer = tf.contrib.estimator.TowerOptimizer(optimizer)
 
+    if loss_scale != 1:
+      # When computing fp16 gradients, often intermediate tensor values are
+      # so small, they underflow to 0. To avoid this, we multiply the loss by
+      # loss_scale to make these tensor values loss_scales times bigger.
```
karmel: nit: loss_scale times bigger

official/resnet/resnet_run_loop.py
```diff
@@ -266,8 +277,19 @@ def exclude_batch_norm(name):
...
+      scaled_grad_vars = optimizer.compute_gradients(loss * loss_scale)
+      unscaled_grad_vars = [(grad / loss_scale, var)
```
karmel: Can you comment to explain the second step here too?

official/utils/arg_parsers/parsers.py
```diff
@@ -58,9 +58,33 @@ def __init__(self):
...
 from __future__ import print_function
 
-
 import argparse
 
+import tensorflow as tf
+
+
+def parse_dtype_info(flags):
+  """Convert dtype string to tf dtype, and set loss_scale default as needed.
```
karmel: Tests, please.

official/utils/arg_parsers/parsers.py
```diff
@@ -201,6 +226,31 @@ def __init__(self, add_help=False, num_parallel_calls=True, inter_op=True,
           metavar="<MTS>"
       )
 
+    if dtype:
+      self.add_argument(
+          "--dtype", "-dt",
+          default="fp32",
+          choices=["fp16", "float16", "fp32", "float32"],
```
karmel: This is nice, but also 2x the maintenance. Should we just enforce one nomenclature or the other?

official/utils/arg_parsers/parsers.py
```diff
@@ -201,6 +226,31 @@ def __init__(self, add_help=False, num_parallel_calls=True, inter_op=True,
...
+          help="[default: %(default)s] The amount to scale the loss by when "
+               "the model is run. Before gradients are computed, the loss is "
+               "multiplied by the loss scale, making all gradients loss_scale "
+               "times larger. To adjust for this, gradients are divided by the "
+               "loss scale before being applied to variables. This is "
+               "mathematically equivalent to training without a loss scale, "
+               "but the loss scale helps avoid some intermediate gradients "
+               "from underflowing to zero. If not provided the default for "
+               "fp16 is 128 and 1 for all other dtypes.",
+      )
```
karmel: Happy face.

official/resnet/resnet_model.py
```diff
@@ -431,46 +488,46 @@ def __call__(self, inputs, training):
...
-    axes = [2, 3] if self.data_format == 'channels_first' else [1, 2]
-    inputs = tf.reduce_mean(inputs, axes, keepdims=True)
-    inputs = tf.identity(inputs, 'final_reduce_mean')
-
-    inputs = tf.reshape(inputs, [-1, self.final_size])
-    inputs = tf.layers.dense(inputs=inputs, units=self.num_classes)
-    inputs = tf.identity(inputs, 'final_dense')
-
-    return inputs
+    with self._model_variable_scope():
```
karmel: Aw, snap. Can you generated new checkpoints and SavedModels for this?

robieta:
@karmel I'll address comments specifically shortly, but two points of note:

1. tf.cast is sometimes a literal no-op. So if `layer` is a tf.Tensor, then `layer is tf.cast(layer, tf.float32)` is True. However if it is a tf.SparseTensor of float32's then `layer is tf.cast(layer, tf.float32)` is False. The reason seems to be that SparseTensors allow heterogeneous dtypes. So we take the tf.cast's out of conditionals if we are sure we won't use SparseTensors. (We could also request tf.cast change it's behavior if ALL elements of a SparseTensor are the requested dtype.)
2. Rolling a lot of the dtype logic into a util would be cleaner and more generalizable. However I think it's better if that sort of code is provided by TensorFlow. So we need to determine if this is just a short term stopgap so we can do some V100 testing, or if the automated mixed precision code is far enough out to warrant a formal utility in the mean time.

robieta:
Discussed offline.
1) Just unconditionally cast, and note in a comment that for SparseTensor of fp32's it may still not be a no-op.
2) Leave mixed precision code in official/resnet in the hope that we can soon rip it out and replace it with mixed precision management from tf proper. If we have a second model where we want fp16 and the tf proper version isn't ready, we can reevaluate moving that code to official/utils

official/resnet/cifar10_main.py
```diff
@@ -204,15 +207,19 @@ def cifar10_model_fn(features, labels, mode, params):
...
-                                         learning_rate_fn=learning_rate_fn,
-                                         momentum=0.9,
-                                         data_format=params['data_format'],
-                                         version=params['version'],
-                                         loss_filter_fn=loss_filter_fn,
-                                         multi_gpu=params['multi_gpu'])
+  return resnet_run_loop.resnet_model_fn(
+      dtype=params['dtype'],
+      features=features, labels=labels, mode=mode, model_class=Cifar10Model,
+      resnet_size=params['resnet_size'],
```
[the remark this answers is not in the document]
robieta: Done.

official/resnet/cifar10_main.py
```diff
@@ -204,15 +207,19 @@ def cifar10_model_fn(features, labels, mode, params):
...
-                                         resnet_size=params['resnet_size'],
-                                         weight_decay=weight_decay,
-                                         learning_rate_fn=learning_rate_fn,
-                                         momentum=0.9,
-                                         data_format=params['data_format'],
-                                         version=params['version'],
-                                         loss_filter_fn=loss_filter_fn,
-                                         multi_gpu=params['multi_gpu'])
+  return resnet_run_loop.resnet_model_fn(
+      dtype=params['dtype'],
```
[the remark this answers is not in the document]
robieta: Done.

official/resnet/resnet_model.py
```diff
@@ -418,6 +421,60 @@ def __init__(self, resnet_size, bottleneck, num_classes, num_filters,
...
+
+    If self.use_fp16 is True, model variable will be created in fp32 then casted
+    to fp16 before being used.
+
+    Returns:
+      A variable scope for the model.
+    """
+
+    custom_getter = self._custom_getter if self.dtype == tf.float16 else None
+    return tf.variable_scope('resnet_model', custom_getter=custom_getter)
```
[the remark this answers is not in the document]
robieta: That makes sense. Changed.

official/resnet/resnet_model.py
robieta: Correct.

official/resnet/resnet_run_loop.py
robieta: Done.

official/resnet/resnet_run_loop.py
robieta: Discussed offline.

official/resnet/resnet_run_loop.py
robieta: Fixed. (Turns out casted is the old English version)

official/resnet/resnet_run_loop.py
robieta: Done.

official/resnet/resnet_run_loop.py
robieta: Since we are not formalizing the mixed precision code at this time, my inclination is to leave it as is and consider doing this if it gets moved to a util. Seem reasonable?

official/resnet/resnet_run_loop.py
robieta: Done.

official/resnet/resnet_run_loop.py
robieta: Fixed.

official/resnet/resnet_run_loop.py
robieta: Done.

official/utils/arg_parsers/parsers.py
robieta: Added.

official/utils/arg_parsers/parsers.py
robieta: I think it's fine simply because the first thing we do is convert them out of strings. But then again I'd write a 1000 page opus in defense of shaving off a single character from a CLI arg. I'm going to leave it for now, but if you decide to decree I won't fight it.

official/resnet/resnet_model.py
robieta: Sure.

robieta:
@karmel I have addressed your comments.

official/resnet/resnet_model.py
```diff
@@ -36,6 +36,9 @@
 _BATCH_NORM_DECAY = 0.997
 _BATCH_NORM_EPSILON = 1e-5
 DEFAULT_VERSION = 2
+DEFAULT_DTYPE = tf.float32
+CASTABLE_TYPES = (tf.float16,)
+ALLOWED_TYPES = (tf.float32,) + CASTABLE_TYPES
```
karmel: nit: (DEFAULT_DTYPE, ) +...

official/resnet/resnet_model.py
```diff
@@ -418,6 +427,61 @@ def __init__(self, resnet_size, bottleneck, num_classes, num_filters,
...
+    same signature as tf.get_variable, except it has an additional getter
+    parameter. Custom getters can be passed as the `custom_getter` parameter of
+    tf.variable_scope. Then, tf.get_variable will call the custom getter,
+    instead of directly getting a variable itself. This can be used to change
+    the types of variables that are retrieved with tf.get_variable.
+    The `getter` parameter is the underlying variable getter, that would have
+    been called if no custom getter was used. Custom getters typically get a
+    variable with `getter`, then modify it in some way.
+
+    This custom getter will create an fp32 variable. If an low precision
```
karmel: nit: a low precision

official/utils/arg_parsers/parsers.py
```diff
@@ -58,9 +58,36 @@ def __init__(self):
...
+
+  Raises:
+    ValueError: If an invalid dtype is provided.
+  """
+  flags.dtype = {
+      "fp16": tf.float16,
+      "float16": tf.float16,
+      "fp32": tf.float32,
+      "float32": tf.float32,
+  }.get(flags.dtype, flags.dtype)
```
karmel: On second read, I think we should stick with just two options-- fp32, fp16.

Also, let's move this to a module dict, then you can just get DTYPE_MAP.keys() for choices below.

official/utils/arg_parsers/parsers.py
```diff
@@ -58,9 +58,36 @@ def __init__(self):
...
+
+  if flags.dtype is None or isinstance(flags.dtype, str):
+    raise ValueError("Invalid dtype: {}".format(flags.dtype))
```
karmel: Why not just try to do DTYPE_MAP[flags.dtype] and catch the KeyError? Fewer lines, no isinstance check.

official/utils/arg_parsers/parsers.py
```diff
@@ -58,9 +58,36 @@ def __init__(self):
...
+
+  if flags.loss_scale is None:
+    flags.loss_scale = {
+        "float16": 128,
+        "float32": 1,
+    }[flags.dtype.name]
```
karmel: Also a constant? Perhaps in the same DTYPE_MAP. Ideally keyed on the same args, rather than a different string drawn from the TF name, which is not in our control.

official/utils/arg_parsers/parsers.py
```diff
@@ -201,6 +229,31 @@ def __init__(self, add_help=False, num_parallel_calls=True, inter_op=True,
           metavar="<MTS>"
       )
 
+    if dtype:
+      self.add_argument(
+          "--dtype", "-dt",
+          default="fp32",
+          choices=["fp16", "float16", "fp32", "float32", "int8"],
```
karmel: int8 is not actually a choice currently, and will fail with a ValueError above, correct? In any case, a call to .keys() on a constant will keep us in sync.

official/utils/arg_parsers/parsers_test.py
```diff
@@ -83,6 +84,21 @@ def test_booleans(self):
...
+      args = parser.parse_args(["--dtype", dtype_str])
+      parsers.parse_dtype_info(args)
+
+      assert args.dtype == tf_dtype
+      assert args.loss_scale == loss_scale
+
+      args = parser.parse_args(["--dtype", dtype_str, "--loss_scale", "5"])
+      parsers.parse_dtype_info(args)
+
+      assert args.loss_scale == 5
```
karmel: Test that int8/invalid types raise errors, given the discrepancy currently in choices versus this function.

robieta:
I made the changes. Was able to refactor parse_dtype_info() to still be idempotent, but much cleaner and with everything keyed off of DTYPE_MAP.

official/utils/arg_parsers/parsers.py
```diff
@@ -58,9 +58,38 @@ def __init__(self):
...
+  """Convert dtype string to tf dtype, and set loss_scale default as needed.
+
+  Args:
+    flags: namespace object returned by arg parser.
+
+  Raises:
+    ValueError: If an invalid dtype is provided.
+  """
+  if not (flags.dtype is None or isinstance(flags.dtype, str)):
+    return  # Make function idempotent
```
karmel: Shouldn't this instead be if dtype is in the set of allowed tf dtypes? Otherwise, you could pass in an int and get through, right? Although maybe that's being too defensive if we assume this only gets called with flags generated by the argparser, which requires a string.

official/utils/arg_parsers/parsers.py
```diff
@@ -58,9 +58,38 @@ def __init__(self):
...
+
+  try:
+    flags.dtype, default_loss_scale = DTYPE_MAP[flags.dtype]
+  except KeyError:
+    raise ValueError("Invalid dtype: {}".format(flags.dtype))
+
+  flags.loss_scale = (flags.loss_scale if flags.loss_scale else
+                      default_loss_scale)
+
```
karmel: nit: this can just be `= flags.loss_scale or default_loss_scale`

jonasrauber:
@robieta Great work! I have a couple of questions regarding the performance:

> Currently training is I/O bound; however synthetic data runs confirm the fp16 accelleration to ~4000 images/sec during training.

Currently? Is there any hope that this will not be I/O-bound in the near future?

That’s 4000 images/sec on ImageNet, right? Which GPU? 4000 compared to fp32 with the same setup and on the same GPU?

PR #3702, opened 2018-03-22
from six.moves import xrange (en masse) YET AGAIN

cclauss:
14 new instances in the 17 days since #3520 which was a clean up after #3206 

@a-dai @nealwu @karmel


nealwu:
Thanks @cclauss

PR #3704, opened 2018-03-22
Resnet benchmark logging

qlzh727:
Adding benchmark logging for resnet training and evaluation.

Tested on GCP with GPU for cifar10. Run following command to enable benchmark logging:
# only log the evaluation result.
python official/resnet/cifar10_main.py --benchmark_log_dir /tmp/metric_dir
# also log the training metric
python official/resnet/cifar10_main.py --hooks LoggingTensorHook LoggingMetricHook --benchmark_log_dir /tmp/metric_dir


official/resnet/resnet_run_loop.py
```diff
@@ -344,10 +345,14 @@ def resnet_main(flags, model_function, input_function):
...
       })
+  benchmark_logger = (logger.BenchmarkLogger(flags.benchmark_log_dir)
+                      if flags.benchmark_log_dir is not None else None)
 
   for _ in range(flags.train_epochs // flags.epochs_between_evals):
     train_hooks = hooks_helper.get_train_hooks(
-        flags.hooks, batch_size=flags.batch_size)
+        flags.hooks,
+        batch_size=flags.batch_size,
+        benchmark_log_dir=flags.benchmark_log_dir)
```
karmel: Is this the same logger as above? Or passing in the log dir will create a separate logging instance?

official/resnet/resnet_run_loop.py
```diff
@@ -374,6 +379,16 @@ def input_fn_eval():
     eval_results = classifier.evaluate(input_fn=input_fn_eval,
                                        steps=flags.max_train_steps)
     print(eval_results)
+    if benchmark_logger is not None:
+      # The eval_results result contains 3 entries:
```
karmel: Since the logger is not used before this, may as well instantiate it here under if flags.benchmark_log_dir, and not have two separate conditionals.

official/resnet/resnet_run_loop.py
```diff
@@ -374,6 +379,16 @@ def input_fn_eval():
...
+      # 1. loss, which is default output for Evaluator.evaluate
+      # 2. global_step, which is default output for Evaluator.evaluate
+      # 3. accuracy, which is defined in eval_metric_ops
+      global_step = eval_results[tf.GraphKeys.GLOBAL_STEP]
+      for key in eval_results:
```
karmel: I feel like this should be a function of the logger if every classifier is going to return global step-- a wrapper around log_metric that says, log_dict_with_global_step or something.

official/resnet/resnet_run_loop.py
robieta: Rather than require benchmark_log_dir, I think it would be convenient if it was:
`benchmark_log_dir if benchmark_log_dir is not None else model_dir`
as this makes manual runs with profiling easier.

robieta: Rather than require benchmark_log_dir, I think it would be convenient if it was:
`benchmark_log_dir if benchmark_log_dir is not None else model_dir`
as this makes manual runs with profiling easier.

official/utils/arg_parsers/parsers.py
```diff
@@ -224,3 +224,21 @@ def __init__(self, add_help=False, data_format=True):
                "was built for CPU or GPU.",
           metavar="<CF>"
       )
+
+
+class BenchmarkParser(argparse.ArgumentParser):
```
robieta: Do we need a new parser? Can this live under the auspices of PerformanceParser?

official/resnet/resnet_run_loop.py
qlzh727: It will create a separate logger instance. Since the logger instance does not keep any internal state, It is fine to create many instances. Alternatively, I could check the benchmark logger and get _logger instance from it, which is not ideal since it is trying to access an private field.

official/resnet/resnet_run_loop.py
qlzh727: Done.

official/resnet/resnet_run_loop.py
karmel: SGTM.

karmel: Will that result in unintended behavior? The logger should have to be turned on explicitly, and saving things to the model dir without explicit instruction might seem like an odd side effect.

official/utils/arg_parsers/parsers.py
```diff
@@ -201,6 +202,13 @@ def __init__(self, add_help=False, num_parallel_calls=True, inter_op=True,
           metavar="<MTS>"
       )
 
+    if benchmark_log_dir:
+      self.add_argument(
+          "--benchmark_log_dir", "-bld", default=None,
+          help="[default: %(default)s] The location of the benchmark logging.",
+          metavar="<BLD>"
+      )
+
```
karmel: I think this makes more sense not as part of Performance; we will want all (or nearly all) models to be benchmarkable, but only a subset will be actively perf optimized. Not opposed to having as separate, or to including with the hooks above; that will force us to make sure all models do this correctly more quickly.

qlzh727:
A sample output for multi_gpu on cifar10 is like below, it contains the hook logging at training time and also evaluation results.

{"name": "learning_rate", "timestamp": "2018-03-23T17:58:42.859234Z", "value": 9.999999747378752e-05, "extras": null, "unit": null, "global_step": 207082}
{"name": "cross_entropy", "timestamp": "2018-03-23T17:58:42.859778Z", "value": 0.0005431846948340535, "extras": null, "unit": null, "global_step": 207082}
{"name": "train_accuracy", "timestamp": "2018-03-23T17:58:42.859947Z", "value": 1.0, "extras": null, "unit": null, "global_step": 207082}
{"name": "loss", "timestamp": "2018-03-23T18:00:51.178905Z", "value": 0.4610125720500946, "extras": null, "unit": null, "global_step": 210988}
{"name": "accuracy", "timestamp": "2018-03-23T18:00:51.179588Z", "value": 0.9282852411270142, "extras": null, "unit": null, "global_step": 210988}
{"name": "learning_rate", "timestamp": "2018-03-23T18:01:02.111248Z", "value": 9.999999747378752e-05, "extras": null, "unit": null, "global_step": 210988}
{"name": "cross_entropy", "timestamp": "2018-03-23T18:01:02.111555Z", "value": 0.0015897309640422463, "extras": null, "unit": null, "global_step": 210988}
{"name": "train_accuracy", "timestamp": "2018-03-23T18:01:02.111742Z", "value": 1.0, "extras": null, "unit": null, "global_step": 210988}
{"name": "loss", "timestamp": "2018-03-23T18:03:13.375434Z", "value": 0.46126365661621094, "extras": null, "unit": null, "global_step": 214894}
{"name": "accuracy", "timestamp": "2018-03-23T18:03:13.375702Z", "value": 0.9283854365348816, "extras": null, "unit": null, "global_step": 214894}
{"name": "learning_rate", "timestamp": "2018-03-23T18:03:24.382612Z", "value": 9.999999747378752e-05, "extras": null, "unit": null, "global_step": 214894}
{"name": "cross_entropy", "timestamp": "2018-03-23T18:03:24.382916Z", "value": 0.0037553715519607067, "extras": null, "unit": null, "global_step": 214894}
{"name": "train_accuracy", "timestamp": "2018-03-23T18:03:24.383084Z", "value": 1.0, "extras": null, "unit": null, "global_step": 214894}

official/resnet/resnet_run_loop.py
robieta: Specifically what I am suggesting is for the case when `benchmarklogginghook` is included but --benchmark_log_dir is not.

qlzh727: In that case, it will spit out a error message when init the hook and ask user to config the flag.

robieta: Yes. What I'm suggesting is that rather than spit out an error it uses model dir. That way in the case that you want the timeline with the rest of the info it saves a command line arg.

qlzh727: Since the benchmarklogginghook is not there by default, when user actually specifying that in the hook list, I think its fair to ask them specifying where they want to log the benchmark.

robieta: Alright. I'll leave the final say to you.

qlzh727:
@karmel, please do a final owner approval for the PR.

PR #3708, opened 2018-03-22
Benchmark run info logging

qlzh727:
Update for benchmark logger to log most of the local runtime env information.

Some of the code for CPU and GPU info are borrowed from tf.tools.test, which is not exposed for import. I guess we can dedup this when we port our code to tf core. 

Also the code is hard to do unit test since the lib to gather CPU and GPU info is not easy to mock. I manually run the test with cifar10 and it works. Let me know if you need more test coverage.


official/utils/logging/logger.py
```diff
@@ -19,13 +19,25 @@
...
 import datetime
+import glob
 import json
+import multiprocessing
 import numbers
 import os
+import re
+
+# pylint: disable=g-bad-import-order
+# Note: cpuinfo and psutil are not installed in the TensorFlow OSS tree.
```
karmel: Presumably this will become a requirement then of any machine or person trying to run the benchmarks. We can include that in a README when it exists; for now, might be worth noting again in the module docstring.

official/utils/logging/logger.py
```diff
@@ -72,3 +84,106 @@ def log_metric(self, name, value, unit=None, global_step=None, extras=None):
...
+    """
+    run_info = {"model_name": model_name}
+    _collect_tensorflow_info(run_info)
+    _collect_environment_variable(run_info)
+    _collect_cpu_info(run_info)
+    _collect_gpu_info(run_info)
+    _collect_memory_info(run_info)
+
+    with tf.gfile.GFile(os.path.join(
+        self._logging_dir, _BENCHMARK_RUN_LOG_FILE_NAME), "w") as f:
```
karmel: Does this assume this is the first thing that gets run? Why not append mode here?

official/utils/logging/logger.py
```diff
@@ -72,3 +84,106 @@ def log_metric(self, name, value, unit=None, global_step=None, extras=None):
...
+      except (TypeError, ValueError) as e:
+        tf.logging.warning("Failed to dump benchmark run info to log file: %s",
+                           e)
+
+def _collect_tensorflow_info(run_info):
+  run_info["tensorflow_version"] = {
+      "version": tf.VERSION, "git_hash": tf.GIT_VERSION}
+
+def _collect_environment_variable(run_info):
+  run_info["environment_variable"] = {
```
karmel: nit: pluralize?

karmel: And potentially specify tf_environment_variables? Unless we intend to add other types here later.

official/utils/logging/logger.py
```diff
@@ -72,3 +84,106 @@ def log_metric(self, name, value, unit=None, global_step=None, extras=None):
...
+# The following code is mirrored from tensorflow/tools/test/system_info_lib
+# which is not exposed for import.
+def _collect_cpu_info(run_info):
+  cpu_info = {}
+
+  cpu_info["num_cores"] = multiprocessing.cpu_count()
+  # Gather num_cores_allowed
+  try:
+    with tf.gfile.GFile("/proc/self/status", "rb") as fh:
+      nc = re.search(r"(?m)^Cpus_allowed:\s*(.*)$", fh.read())
```
karmel: Oof. This is scary. What's happening here? Is there a better way? What exactly is this regex doing, and is it dangerous to read in an arbitrary file? Does this limit us to particular OSes?

official/utils/logging/logger.py
```diff
@@ -72,3 +84,106 @@ def log_metric(self, name, value, unit=None, global_step=None, extras=None):
...
+    if nc:  # e.g. "ff" => 8, "fff" => 12
+      cpu_info["num_cores_allowed"] = (
+        bin(int(nc.group(1).replace(",", ""), 16)).count("1"))
```
karmel: We should minimally comment this, even if (especially if?) c&ped from TF.

official/utils/logging/logger.py
```diff
@@ -72,3 +84,106 @@ def log_metric(self, name, value, unit=None, global_step=None, extras=None):
...
+  except tf.OpError:
+    pass
+  finally:
+    if "num_cores_allowed" not in cpu_info:
+      cpu_info["num_cores_allowed"]= cpu_info["num_cores"]
```
karmel: nit: missing space. Make sure your IDE is linting :)

official/utils/logging/logger.py
```diff
@@ -72,3 +84,106 @@ def log_metric(self, name, value, unit=None, global_step=None, extras=None):
...
+
+  info = cpuinfo.get_cpu_info()
+  cpu_info["cpu_info"] = info["brand"]
+  cpu_info["num_cores"] = info["count"]
+  cpu_info["mhz_per_cpu"] = info["hz_advertised_raw"][0] / 1.0e6
+  l2_cache_size = re.match(r"(\d+)", str(info.get("l2_cache_size", "")))
+  if l2_cache_size:
+    # If a value is returned, it"s in KB
```
karmel: nit: it's

official/utils/logging/logger.py
```diff
@@ -72,3 +84,106 @@ def log_metric(self, name, value, unit=None, global_step=None, extras=None):
...
+    _collect_memory_info(run_info)
+
+    with tf.gfile.GFile(os.path.join(
+        self._logging_dir, _BENCHMARK_RUN_LOG_FILE_NAME), "w") as f:
+      try:
+        json.dump(run_info, f)
+        f.write("\n")
+      except (TypeError, ValueError) as e:
+        tf.logging.warning("Failed to dump benchmark run info to log file: %s",
+                           e)
```
karmel: Seems like we might want to catch errors for each of the above logging functions; I can imagine successfully getting TF info and failing out on CPU-- we would still want to get whatever we could, right?

official/utils/logging/logger.py
```diff
@@ -72,3 +84,106 @@ def log_metric(self, name, value, unit=None, global_step=None, extras=None):
...
+      for f in glob.glob(
+          "/sys/devices/system/cpu/cpu*/cpufreq/scaling_governor")
+    ])
+    if cpu_governors:
+      if len(cpu_governors) > 1:
+        cpu_info["cpu_governor"] = "mixed"
+      else:
+        cpu_info["cpu_governor"] = list(cpu_governors)[0]
+  except tf.OpError:
+    pass
```
karmel: How frequently does something like this fail? My internal sirens are going off with all the hardcoded paths and regexes...

official/utils/logging/logger.py
qlzh727: The metric info is logged to a different file, which gets appended, whereas the run info is a big blob of JSON, which is stored in a single file. Each run should have one run info, and append run info kind of confuse the uploading part in future.

official/utils/logging/logger.py
qlzh727: Good point, I will update the function above to not fail but warning.

official/utils/logging/logger.py

official/utils/logging/logger.py
qlzh727: It is trying to read the local os state to get CPU allowed, You can find the format and content from https://www.kernel.org/doc/Documentation/filesystems/proc.txt and it is not an arbitrary file. I think it works on Linux/Mac system, and Windows will probably fail.

The regex is trying to din the line with Cpus_allowed as start and find the matching group in the bracket.

official/utils/logging/logger.py

official/utils/logging/logger.py

official/utils/logging/logger.py

official/utils/logging/logger.py
qlzh727: I think it works for linux system mostly and should be quite stable. Windows will definitely break.

official/utils/logging/logger.py
karmel: What about something like this instead? https://stackoverflow.com/questions/1006289/how-to-find-out-the-number-of-cpus-using-python ? Or https://docs.python.org/3/library/platform.html#cross-platform . I am inclined to think that the set of information we want should be limited to the set accessible from the default Python libs, rather than adding extra dependencies or very complex regexes to get platform-specific info that we're not sure we need yet. Especially since we will control the hardware we are running on. Can we start conservative, and then add the additional complexity if we find we need it down the line?

qlzh727: Done. Removed most of the regex parse to eliminate potential error.

qlzh727:
Done. Simplified the CPU metric logic and getting rid of the regex parse.

official/utils/logging/logger.py
```diff
@@ -13,19 +13,37 @@
...
+packages need be installed. Run the following commands for dependency packages:
+  > pip install --upgrade py-cpuinfo
+  > pip install --upgrade psutil
+"""
 from __future__ import absolute_import
 from __future__ import division
 from __future__ import print_function
 
 import datetime
+import glob
```
robieta: Why are glob and re imported? I don't see them used anywhere.

PR #3725, opened 2018-03-23
Fix random order problem in benchmark logging.

qlzh727:


qlzh727:
Tested several runs on my local and it works fine.

PR #3740, opened 2018-03-25
Update build_data.py

grayskripko:
I've got a lot of type errors here when was launching https://github.com/tensorflow/models/blob/master/research/deeplab/g3doc/pascal.md in this line
https://github.com/tensorflow/models/blob/master/research/deeplab/datasets/build_voc2012_data.py#L128


grayskripko:
It was I've year ago. I don't remember. You can come the issue

Issue #3578, opened 2018-03-13
Add purely negative images gets error when using object detection API to create tf records?


liangxiao05:
I met this problem when I tested my SSD model trained by object detection API.The model preforms well until it meets some images that contain the objects very simillar to the objective classfications.
I don't want to add another class to these wrong detected objects ,so I decide to include these images in my training datasets.
Refered to https://github.com/tensorflow/models/issues/2544，
''
You need to set min_negatives_per_image to a non zero number in the config for the model to sample boxes from purely negative images: 
''
After reviewed the code
https://github.com/tensorflow/models/blob/master/research/object_detection/core/losses.py#L433,
it tells me setting  'min_negatives_per_image ' to a positive number allows sampling negatives in an image without any positive anchors ,

However，if the image has no positive anchors , it contains no 'object' in its xml file ,then the script 
https://github.com/tensorflow/models/blob/master/research/object_detection/dataset_tools/create_pascal_tf_record.py#L107   will get this error when I creat the record :
for obj in data['object']:
   Key Error : 'object' 

I think many people are puzzled by this, please help us  to solve it .Hope for the replys ,thanks!

twtygqyy:
Hi @liangxiao05 Could you please share your solution?

marcoromelli:
@liangxiao05 can you share the solution you've found?

liangxiao05:
@marcoromelli，sure!
First you can  set min_negatives_per_image to a non zero number in the config.
Then when you generate xmls , for purely negative images you can label their xmls only to contain the path,filename, width or height informations (for example ,if you generate the xml with labelImg,you don't have to draw a bbox on it)
then when creating tf records , take a 'if' judgement on it like this:
 -----
if data.has_key('object'):
   for obj in data['object']:
----
That's all ~~~~

saiaman:
Do you need to change the other config variables in hard_example_miner ?

leccyril:
I am sorry but i have same problem.. But i could not add negative picture because i could use algorithm for any picture or video,

I am trying to find only one object.. Then i test my detection on similar object, for example trained model with samsung gear s3 and detect in appel watch.. It says to me apple watch is a samsung watch... 

This is really annoying

There is no simple way to only detect object when is it found and not trying to detect object everywhere ? Even if not present at all ?

jackweiwang:
Hi @liangxiao05  May I see your config file?

AeroWRX:
> @marcoromelli，sure!
[...]

PR #3744, opened 2018-03-26
Update create_pascal_tf_record.py

liangxiao05:
When creat a tf record using purely negative images which have no 'object' in their xmls ,the script gets a key error : for obj in data['object']:  Key Error : 'object'
In order to fix this , add a 'if ' judegement before it .

### DIFF
--- a/research/object_detection/dataset_tools/create_pascal_tf_record.py
+++ b/research/object_detection/dataset_tools/create_pascal_tf_record.py
@@ -104,21 +104,23 @@ def dict_to_tf_example(data,
   truncated = []
   poses = []
   difficult_obj = []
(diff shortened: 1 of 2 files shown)


liangxiao05:
please refer to this
https://github.com/tensorflow/models/issues/3578

liangxiao05:
hi,@pkulzc,thank you .
same question in the script ,so you will have to fix this script too~
[https://github.com/tensorflow/models/blob/master/research/object_detection/dataset_tools/create_pet_tf_record.py](url)

pkulzc:
Sure, we will fix that in the next release. Thanks!

PR #3760, opened 2018-03-27
Add requirements.txt to official.

robieta:
Now that we have dependencies beyond what TensorFlow already pulls in, we may as well put a bit of formalism to help people out.

### DIFF
--- a/official/requirements.txt
+++ b/official/requirements.txt
@@ -1,0 +1,2 @@
+psutil>=5.4.3
+py-cpuinfo>=3.3.0
--- a/official/utils/logging/logger.py
+++ b/official/utils/logging/logger.py
@@ -16,9 +16,7 @@
 """Logging utilities for benchmark.
 
 For collecting local environment metrics like CPU and memory, certain python
-packages need be installed. Run the following commands for dependency packages:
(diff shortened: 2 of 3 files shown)


official/utils/logging/logger.py
```diff
@@ -16,9 +16,7 @@
 """Logging utilities for benchmark.
 
 For collecting local environment metrics like CPU and memory, certain python
-packages need be installed. Run the following commands for dependency packages:
-  > pip install --upgrade py-cpuinfo
-  > pip install --upgrade psutil
+packages need be installed. See README for details.
```
karmel: No reason to make comments less explicit here.

official/requirements.txt
```diff
@@ -0,0 +1,2 @@
+psutil>=5.4.3
+py-cpuinfo>=3.3.0
```
karmel: Should tf-nightly be listed here too? I don't know if there's a way to do a conditional on gpu.

karmel:
P.S. I assume you tested that this works, as I did not.

robieta:
Yes. Knowing me that is a perfectly reasonable question.

PR #3766, opened 2018-03-27
vis.py: use basename of image when saving raw annotations to prevent file-not-found error

walkerlala:
when setting `--also_save_raw_predictions=True`, a file-not-found error is report because we are saving into the `--vis_logdir` using the fullpath of that image.

### DIFF
--- a/research/deeplab/vis.py
+++ b/research/deeplab/vis.py
@@ -173,7 +173,7 @@ def _process_batch(sess, original_images, semantic_predictions, image_names,
         colormap_type=FLAGS.colormap_type)
 
     if FLAGS.also_save_raw_predictions:
-      image_filename = image_names[i]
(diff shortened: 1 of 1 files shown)


walkerlala:
I signed it

YknZhu:
Thanks! A more proper fix would be using the absolute path in dataset generation process?

walkerlala:
@YknZhu The problem is  we are saving into the `--vis_logdir` using the fullpath of that image. For example if `--vis_logdir` is `/home/someuser/vislog` and the path of image is `/home/someuser/evaldata/image1.jpg`, and then when `--also_save_raw_predictions=True`, we are saving the result into `/home/someuser/vislog/home/someuser/evaldata/image1.jpg`, which will raise a file-not-found error.

PR #3767, opened 2018-03-27
Update faster_rcnn_inception_resnet_v2_atrous_oid.config

tanyarags:
from_detection_checkpoint check needs to be true to using pre-trained models.
(otherwise no variables found error thrown while training)

### DIFF
--- a/research/object_detection/samples/configs/faster_rcnn_inception_resnet_v2_atrous_oid.config
+++ b/research/object_detection/samples/configs/faster_rcnn_inception_resnet_v2_atrous_oid.config
@@ -106,6 +106,7 @@ train_config: {
   }
   gradient_clipping_by_norm: 10.0
   fine_tune_checkpoint: "PATH_TO_BE_CONFIGURED/model.ckpt"
+  from_detection_checkpoint: true
   # Note: The below line limits the training process to 800K steps, which we
   # empirically found to be sufficient enough to train the Open Images dataset.
   # This effectively bypasses the learning rate schedule (the learning rate will


research/object_detection/samples/configs/faster_rcnn_inception_resnet_v2_atrous_oid.config
```diff
@@ -106,6 +106,7 @@ train_config: {
   }
   gradient_clipping_by_norm: 10.0
   fine_tune_checkpoint: "PATH_TO_BE_CONFIGURED/model.ckpt"
+  from_detection_checkpoint: true
```
pkulzc: By default we assume the checkpoint used is object classification ckpt. So this may not be necessary.
Please refer to [this](https://github.com/pkulzc/models/blob/master/research/object_detection/g3doc/configuring_jobs.md#model-parameter-initialization) for more details.

tanyarags: From what I understand, by setting fine_tune_checkpoint and from_detection_checkpoint helps us use pre-trained model. If from_detection_checkpoint is false, it will pick up what ever available variables from the checkpoint(fine_tune_checkpoint) and use them as initialization for training the model. 
Please correct me if I am wrong.

tanyarags: Due to code restructuring, this will not be required now.
Closing the request.

PR #3771, opened 2018-03-27
更新fork的tensorflow

jackgurong:


pkulzc:
This PR is largely duplicated with existing files and looks very informal. Do you really want to send it for review?

PR #3776, opened 2018-03-27
Add benchmark upload util to Bigquery.

qlzh727:
Also update the benchmark logger and bigquery schema for the
errors found during the integration test.


official/utils/logging/benchmark_uploader.py
```diff
@@ -0,0 +1,141 @@
...
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+# ==============================================================================
+
+"""Library to upload benchmark generated by BenchmarkLogger to remote repo.
+
+This library require google cloud bigquery lib as dependency, which can be
+installed with:
+  > pip install --upgrade google-cloud-bigquery
```
robieta: Also add to requirements.txt

official/utils/logging/benchmark_uploader.py
```diff
@@ -0,0 +1,141 @@
...
+  """Upload the benchmark and metric info to BigQuery."""
+
+  def __init__(self, logging_dir, gcp_project=None, credentials=None):
+    """Initialized BigQueryUploader with proper setting.
+
+    Args:
+      logging_dir: string, logging directory that contains the benchmark log.
+      gcp_project: string, the name of the GCP project that the log will be
+        uploaded to. The default project name will be detected from local
+        environment if no value is provide.
```
robieta: µnit: provided.

official/utils/logging/benchmark_uploader.py
```diff
@@ -0,0 +1,141 @@
...
+      credentials: google.auth.credentials. The credential to access the
+        BigQuery service. The default service account credential will be
+        detected from local environment if no value is provided. Please use
+        google.oauth2.service_account.Credentials to load credential from local
```
robieta: µnit: space before "Credentials"

official/utils/logging/benchmark_uploader.py
```diff
@@ -0,0 +1,141 @@
...
+  def upload_metric(self, dataset_name, table_name, run_id):
+    """Upload metric information to Bigquery."""
+    expected_file = os.path.join(
+        self._logging_dir, logger.METRIC_LOG_FILE_NAME)
+    with tf.gfile.GFile(expected_file) as f:
+      lines = f.readlines()
+      metrics = []
+      for l in lines:
+        if not l.strip(): continue
+        metric = json.loads(l)
```
robieta: I would be inclined to wrap this in a try-except (and log error). That way a single malformed line doesn't ruin all entries.

official/utils/logging/benchmark_uploader.py
```diff
@@ -0,0 +1,141 @@
...
+
+import tensorflow as tf # pylint: disable=g-bad-import-order
+
+from official.utils.logging import logger
+
+
+FLAGS = None
+
+
+class BigQueryUploader(object):
```
robieta: I have a conceptual proposal:
  1) Pull dataset_name and run_id into `__init__()`. In effect make all local information part of the class state.
  2) Support intermediate uploading. So if I call upload_metric() with 500 lines, and then call it again with 1000 the second call will upload lines 501 to 1000.

The reason is that is would allow intermediate uploading during the training loop, rather than at the very end. Depending on the sort of monitoring we want to build on top of BigTable that could be desirable. Discuss.

official/utils/logging/benchmark_uploader.py
```diff
@@ -0,0 +1,141 @@
...
+        detected from local environment if no value is provided. Please use
+        google.oauth2.service_account.Credentials to load credential from local
+        file for the case that the test is run out side of GCP.
+    """
+    self._logging_dir = logging_dir
+    self._bq_client = bigquery.Client(
+        project=gcp_project, credentials=credentials)
+
+  def upload_benchmark_run(self, dataset_name, table_name, run_id):
+    """Upload benchmark run information to Bigquery."""
```
robieta: Can you give docstrings for inputs?

official/utils/logging/benchmark_uploader.py
```diff
@@ -0,0 +1,141 @@
...
+      lines = f.readlines()
+      metrics = []
+      for l in lines:
+        if not l.strip(): continue
+        metric = json.loads(l)
+        metric["run_id"] = run_id
+        metrics.append(metric)
+      table_ref = self._bq_client.dataset(dataset_name).table(table_name)
+      errors = self._bq_client.insert_rows_json(table_ref, metrics)
+      if errors:
```
robieta: Should we have retry logic here? (Or does the BigTable API already support it.) Once we're containerized we should expect these files will disappear after the test is done, so it would be a shame to lose a run due to a DB connection hiccup.

official/utils/logging/benchmark_uploader.py
```diff
@@ -0,0 +1,141 @@
...
+        tf.logging.error(
+            "Failed to upload benchmark info to bigquery: {}".format(errors))
+
+
+def main(unused_argv):
+  if not FLAGS.benchmark_log_dir:
```
robieta: I prefer to move away from global FLAGS to having the arg parser defined in main().

official/utils/logging/benchmark_uploader.py
```diff
@@ -0,0 +1,141 @@
...
+    print("Usage: benchmark_uploader.py --benchmark_log_dir=/some/dir")
+    sys.exit(1)
```
robieta: Is there a reason to use sys.exit() instead of raising?

official/utils/logging/benchmark_uploader.py
```diff
@@ -0,0 +1,141 @@
...
+      gcp_project=FLAGS.gcp_project)
+  run_id = str(uuid.uuid4())
+  uploader.upload_benchmark_run(
+      FLAGS.bigquery_data_set, FLAGS.bigquery_run_table, run_id)
+  uploader.upload_metric(
+      FLAGS.bigquery_data_set, FLAGS.bigquery_metric_table, run_id)
+
+
+if __name__ == "__main__":
+  parser = argparse.ArgumentParser()
```
robieta: Can we make this a class?

official/utils/logging/benchmark_uploader.py
```diff
@@ -0,0 +1,141 @@
...
+      metavar="<BLD>"
+  )
+  parser.add_argument(
+      "--gcp_project", "-gp", default=None,
+      help="The GCP project name where the benchmark will be uploaded.",
+      metavar="<GP>"
+  )
+  parser.add_argument(
+      "--bigquery_data_set", "-bds", default="test_benchmark",
+      help="The Bigquery dataset name where the benchmark will be uploaded.",
```
robieta: How come there are no "[default: %(default)s]" for the rest of these?

official/utils/logging/benchmark_uploader.py
```diff
@@ -0,0 +1,141 @@
...
+           " uploaded.",
+      metavar="<BRT>"
+  )
+  parser.add_argument(
+      "--bigquery_metric_table", "-bmt", default="benchmark_metric",
+      help="The Bigquery table name where the benchmark metric information will"
+           " be uploaded.",
+      metavar="<BMT>"
+  )
+  FLAGS, unparsed = parser.parse_known_args()
```
robieta: Use parse_args() instead of parse_known_args() so it blows up with unrecognized args.

official/utils/logging/logger.py
```diff
@@ -72,7 +72,7 @@ def log_estimator_evaluation_result(self, eval_results):
       if key != tf.GraphKeys.GLOBAL_STEP:
         self.log_metric(key, eval_results[key], global_step=global_step)
 
-  def log_metric(self, name, value, unit=None, global_step=None, extras=None):
+  def log_metric(self, name, value, unit=None, global_step=None, extras={}):
```
robieta: Mutable types should not be default args.

official/utils/logging/logger_test.py
```diff
@@ -25,16 +25,22 @@
...
-from tensorflow.python.client import device_lib
 
 from official.utils.logging import logger
 
 
 class BenchmarkLoggerTest(tf.test.TestCase):
 
+  def setUp(self):
+    super(BenchmarkLoggerTest, self).setUp()
+    self.original_environ = dict(os.environ)
```
robieta: Could you put a comment about why this was necessary and when others would need to worry about environment variables?

official/utils/logging/benchmark_uploader.py
qlzh727: Since this is in the main function, i guess both of them does not make big difference. 

Raise an exception give the caller the chance to catch it, whereas the sys.exit() is a strong signal when something is really wrong and no point to moving forward. In this case, I think I would prefer exit.

official/utils/logging/benchmark_uploader.py
qlzh727: k, moving this arg_parser

official/utils/logging/benchmark_uploader.py
qlzh727: Done.

official/utils/logging/benchmark_uploader.py
qlzh727: Done. Moved to use arg_parser

official/utils/logging/logger.py

official/utils/logging/benchmark_uploader.py
robieta: I don't remember the exact failure modes, but because JSON is a fragile container (1 bad byte can corrupt the entire entry), I've always seen code that skips invalid entries.

official/utils/logging/benchmark_uploader.py
robieta: Seems reasonable.

official/utils/logging/logger_test.py
qlzh727: Done.

PR #3781, opened 2018-03-27
Update the importing logic for cpuinfo and psutil.

qlzh727:
Those two libs are usually not installed by default, and we should
not force people to install them if they just want to run resnet.

### DIFF
--- a/official/utils/logging/logger.py
+++ b/official/utils/logging/logger.py
@@ -144,7 +144,7 @@ def _collect_cpu_info(run_info):
 
   # Note: cpuinfo is not installed in the TensorFlow OSS tree.
   # It is installable via pip.
-  import cpuinfo
+  import cpuinfo    # pylint: disable=g-import-not-at-top
 
   info = cpuinfo.get_cpu_info()
   cpu_info["cpu_info"] = info["brand"]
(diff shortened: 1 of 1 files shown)


official/utils/logging/logger.py
```diff
@@ -149,6 +142,10 @@ def _collect_cpu_info(run_info):
 
   cpu_info["num_cores"] = multiprocessing.cpu_count()
 
+  # Note: cpuinfo is not installed in the TensorFlow OSS tree.
+  # It is installable via pip.
+  import cpuinfo
+
```
karmel: You will probably need a `  # pylint: disable=g-import-not-at-top` to make glint clear these.

official/utils/logging/logger.py
```diff
@@ -149,6 +142,10 @@ def _collect_cpu_info(run_info):
 
   cpu_info["num_cores"] = multiprocessing.cpu_count()
 
+  # Note: cpuinfo is not installed in the TensorFlow OSS tree.
+  # It is installable via pip.
+  import cpuinfo
```
karmel: You will probably need a `  # pylint: disable=g-import-not-at-top` to make glint clear these.

official/utils/logging/logger.py

official/utils/logging/logger.py

PR #3797, opened 2018-03-28
Add testing script for local lint and python test.

qlzh727:


official/utils/testing/scripts/presubmit.sh
```diff
@@ -0,0 +1,43 @@
...
+
+# Assume the pwd is under models.
+cd official
+
+exit_code=0
+
+# Check lint:
+RC_FILE="utils/testing/pylint.rcfile"
+
+echo "===========Running lint test============"
```
robieta: `git diff --name-only origin/master ./official/` has the nice property that you only have to care about stuff you touched. (And it makes linting faster.) The one caveat is it includes deleted files so you would have to do an existence test.

official/utils/testing/scripts/presubmit.sh
```diff
@@ -0,0 +1,43 @@
...
+do
+  echo "Linting ${file}"
+  pylint --rcfile="${RC_FILE}" --disable=missing-docstring,protected-access "${file}" || exit_code=$?
+done
+
+# Run Python test, the python version depends on the local configured defaults.
+echo "===========Running Python test============"
+
+for test_file in `find . -name '*test.py' -print`
+do
```
robieta: May as well run in parallel. I looks like xargs can do it; otherwise it's pretty trivial to do in python with subprocess. The reason being...

official/utils/testing/scripts/presubmit.sh
```diff
@@ -0,0 +1,43 @@
...
+  echo "Testing ${test_file}"
+  python "${test_file}" || exit_code=$?
```
robieta: ... this should probably do py2 and py3.

official/utils/testing/scripts/presubmit.sh
```diff
@@ -0,0 +1,43 @@
+#!/bin/bash
```
robieta: It would be nice to provide a command to run this from anywhere in the repo. For instance:
`alias lint_and_test=$(git rev-parse --show-toplevel)/official/utils/testing/scripts/presubmit.sh`
Either in the README or in this file.

official/utils/testing/scripts/presubmit.sh
```diff
@@ -0,0 +1,43 @@
...
+exit_code=0
+
+# Check lint:
+RC_FILE="utils/testing/pylint.rcfile"
+
+echo "===========Running lint test============"
+for file in `find . -name '*.py' ! -name '*test.py' -print`
+do
+  echo "Linting ${file}"
+  pylint --rcfile="${RC_FILE}" "${file}" || exit_code=$?
```
robieta: If you change the `cd` at the top, you should also inject the absolute path into the rcfile path. It has trouble finding it otherwise. (Not sure if there is some weird `cd` related bash behavior.)

official/utils/testing/scripts/presubmit.sh
```diff
@@ -0,0 +1,43 @@
+#!/bin/bash
+
+# Presubmit script that run tests and lint under local environment.
+# Make sure that tensorflow and pylint is installed.
+# usage: models >: ./official/utils/testing/scripts/presubmit.sh
+set +x
+
+# Assume the pwd is under models.
+cd official
```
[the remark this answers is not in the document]
karmel: I am wary of adding git magic into the script. Maybe instructions above with the alias lint_and_test, along with setting $TF_MODELS_ROOT or something to that effect, which, if empty, will mean you cd into just `official` as now, but, if set, the right place.

official/utils/testing/scripts/presubmit.sh
karmel: If there are errors in things you didn't touch, the tests on Kokoro will fail as well. Printing them all will make us fix them faster. One hopes.

official/utils/testing/scripts/presubmit.sh
karmel: It's worth noting that even our unit tests take a long time now. How long will this take locally? In my experience, once testing takes longer than ~5 min, people stop doing it. That said, I don't want to add a ton of logic here that has to be separately maintained from the Kokoro files. In fact, I already feel nervous about this duplication. It's easy to imagine the two sets of files diverging. I don't have a solution there, but... maybe the better way to handle this is a private wrapper around the kokoro files. That way, we can include git magic, whatever you want. Runs from the internal files, on external files.

robieta: In total it's something like 7-8 min. (So ~15 if you do py2 and py3) I agree that likelihood of being run goes down with run time.

As far as code duplication, what if the heavy lifting is in a public script and Kokoro's build.sh just finds and runs that script? (At least for testing and linting.) I *think* this is the same as what Karmel is suggesting?

official/utils/testing/scripts/presubmit.sh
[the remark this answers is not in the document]
qlzh727: Done. Changed to check the file location of the script and cd into the relative directory of the git repository root.

official/utils/testing/scripts/presubmit.sh
qlzh727: Yea, currently running all the tests rather than the touched one. Since the file depend on each other, you cannot tell whether the test will break or not even its not touched.

official/utils/testing/scripts/presubmit.sh
qlzh727: The updated version of script will the one for all. Kokoro tests will just call it.

official/utils/testing/scripts/presubmit.sh
qlzh727: Done.

official/utils/testing/scripts/presubmit.sh
qlzh727: Have usage example on the top of the script.

official/utils/testing/scripts/presubmit.sh
qlzh727: Ack.

qlzh727:
Done. The script has been updated so that it can be run both on Kokoro and local. We can update Kokoro scripts once this PR is committed.

robieta:
The tests are still serial though, right?

qlzh727:
Correct. Running tests in parallel might generate confusing error logs since they are mixed together. On the other hand, Bazel is helpful for those case, since it generate a summary at the end the run.

robieta:
Discussed offline. I am fine with the plan of "serial for now, Bazel eventually" rather than have a somewhat complicated python script that we'll eventually get rid of anyway.

PR #3798, opened 2018-03-28
Add End-to-end tests for wide deep, and fix "wide" and "deep" configurations.

robieta:
There is already static test data for wide-deep, so I simply used that rather than adding synthetic data.


robieta:
@karmel Ready for review.

PR #3802, opened 2018-03-29
Add config options to change default saving behavior in TrainConfig

maclauss:
* checkpoint_save_interval_secs is used to change how often the checkpoints are serialized (default 600 seconds)

* save_summaries_secs is used to change how often summaries are saved (default 120 seconds)

* max_checkpoints_to_keep is used to change how many checkpoints are kept (default 5 checkpoints)

### DIFF
--- a/research/object_detection/trainer.py
+++ b/research/object_detection/trainer.py
@@ -366,7 +366,7 @@ def train(create_tensor_dict_fn, create_model_fn, train_config, master, task,
         summary_op=summary_op,
         number_of_steps=(
             train_config.num_steps if train_config.num_steps else None),
-        save_summaries_secs=save_interval_secs=train_config.save_summaries_secs,
+        save_summaries_secs=train_config.save_summaries_secs,
         save_interval_secs=train_config.checkpoint_save_interval_secs,
         sync_optimizer=sync_optimizer,
         saver=saver)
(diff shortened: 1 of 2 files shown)


research/object_detection/trainer.py
```diff
@@ -365,6 +366,7 @@ def initializer_fn(sess):
         summary_op=summary_op,
         number_of_steps=(
             train_config.num_steps if train_config.num_steps else None),
-        save_summaries_secs=120,
+        save_summaries_secs=save_interval_secs=train_config.save_summaries_secs,
```
pkulzc: typo?

maclauss: Indeed thanks, fixed in d005fa6

maclauss:
@pkulzc Can I get your approval if you agree with these changes?

pkulzc:
Please resolve the conflicts.

denmoroz:
Any updates here? Is there any option to separately store the best model on validation?

yangsiyu007:
@denmoroz as a workaround I've been setting the option `keep_checkpoint_every_n_hours` in `train_config` to 1 or 2 so that checkpoints are saved every hour or every two hours.

YijinLiu:
I like this change. @maclauss Is it still your intention to commit it?

maclauss:
@YijinLiu Sorry I should've closed this a long time ago, I moved away from the object detection API for other reasons. Feel free to fix the conflicts and merge it.

liuchang8am:
It seems that trainer.py is deprecated in current version of api? So I assume theses changes are not useable anymore?

tarun-ssharma:
I see a `checkpoint_every_n` in https://github.com/tensorflow/models/blob/master/research/object_detection/model_main_tf2.py in the current version.

vitorbds:
Someone find some solution for it ?

PR #3813, opened 2018-03-29
Adding new checkpoints for Resnets

karmel:


karmel:
The files here seem to be giving the wrong accuracy; closing this while I sort out what happened.

karmel:
Reopening, as the checkpoints seem to work with master + tf-nightly.

karmel:
@robieta , in the interest of minimizing the time during which we are wrong, how do you feel about merging this, and then updating with the checkpoints/savedmodels from #3687 when they are ready and that is in? You can ditch the Frozen Graphs at that point.

cc @qlzh727 - it would be nice if the benchmarks stuck their checkpoints and savedmodels somewhere were we could just link to a directory, and tell people to take the most recent ones, rather than manually updating.

robieta:
Agreed. That seems like the most sensible approach.

PR #3821, opened 2018-03-30
Fix so images render on github.

lamberta:
And fix a typo.


lamberta:
please test on colab as well.

PR #3828, opened 2018-03-31
First commit of learning_unsupervised_learning

lukemetz:
This is code for our paper: "Learning of Unsupervised Update Rules".
Currently there is a placeholder link in the README that will be updated when the paper is live.


lukemetz:
Paper is now live, and link is updated.

PR #3844, opened 2018-04-02
tensorrt: long was removed in Python 3

cclauss:
__long__ (used on line 603) was removed in Python 3 in favor of __int__.

### DIFF
--- a/research/tensorrt/tensorrt.py
+++ b/research/tensorrt/tensorrt.py
@@ -35,6 +35,12 @@ import tensorflow.contrib.tensorrt as trt
 
 from official.resnet import imagenet_preprocessing  # pylint: disable=g-bad-import-order
 
+try:
+    long        # Python 2
+except NameError:
+    long = int  # Python 3
+
+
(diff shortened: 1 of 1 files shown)


karmel:
CC @samikama -- is this safe for TensorRT?

samikama:
@karmel , It looks like it is OK but using six instead of try catch might be better. Or even better workspace size can be converted to int and be in units of megabytes. I let @aaroey  to decide.

cclauss:
@samikama The [usage of __long__](https://github.com/tensorflow/models/blob/master/research/tensorrt/tensorrt.py#L592) is `type=long` and I was not sure if passing a tuple like `type=six.integer_types` would work.

samikama:
@cclauss you are right at changing long. What I am saying is it might be more sensible to use int there and change the units to KB or MB. Passing an allocation size down to the last byte is unlikely to be ever used.

aaroey:
Thanks @cclauss for the fix. I agree with @samikama that using MB may be a more suitable solution. Would you help to change that?

PR #3845, opened 2018-04-03
Allow users passing master argument to avoid instantiating a TPUClusterResolver

dlibenzi:


nealwu:
Thanks Davide. Re-assigning to @saeta and @frankchn for review.

PR #3858, opened 2018-04-03
Fix the testing script exit code for python test.

qlzh727:


official/utils/testing/scripts/presubmit.sh
```diff
@@ -53,12 +53,14 @@ py_test() {
 
 py2_test() {
   local PY_BINARY=$(which python2)
-  return $(py_test "${PY_BINARY}")
+  py_test "$PY_BINARY"
+  return $?
```
karmel: Can you add comments describing what this means, and why it is necessary, so that we avoid the next person "helpfully" removing the extra line?

qlzh727: sure. Will do in a separate PR.

PR #3863, opened 2018-04-03
Adding stop threshold logic

karmel:
We want to be able to stop training at a specified accuracy or other model-specific metric. Helper function added here, along with run loop logic for existing models.

@qlzh727 -- when running time-to-converge tests, we will likely want to run with the flag --stop_threshold=<relevant number for the given model>

@yhliang2018 -- when adding new models, they should include stopping logic in the run loop.


karmel:
The linting test is reporting a Kokoro infrastructure error right now. I will force-retry in a bit to see if that clears things.

official/utils/misc/model_helpers.py
```diff
@@ -0,0 +1,55 @@
...
+from __future__ import absolute_import
+from __future__ import division
+from __future__ import print_function
+
+import numbers
+
+import tensorflow as tf
+
+
+def past_stop_threshold(stop_threshold, eval_metric):
```
qlzh727: As discussed offline, probably should have a similar function which stop when a metric is smaller that the threshold.

karmel: It's pretty simple to add a flag in theory, but in practice, I don't know that we have a use-case yet for a lower-is-better metric. @petermattson , is any target metric decreasing?

karmel:
@petermattson -- there are a couple questions for you in the comments; can you address?

official/utils/misc/model_helpers.py
petermattson: For performance purposes, crossing the line is acceptable, since all the benchmark implementations play by the same rules. The only question is timing instability, and we're addressing that through multiple runs. We can revisit if that's insufficient.

petermattson:
Increasing only seems fine for now. We can always rephrase the metric.
Thanks!

karmel:
Pushing this now, and then will copy over manually to add build files.

petermattson:
Thanks Karmel! I appreciate you putting this in. :-)

PR #3867, opened 2018-04-04
ADD inference.py for simple inference

aggresss:
ADD inference.py for simple inference


antoinebrl:
The name of the file, `inference.py`, might cause interferences (:smiley:) with the package `object_detection.inference`.

pkulzc:
This code is largely duplicated with the notebook, so we prefer not doing this. You can have it with your own repo.

PR #3878, opened 2018-04-04
Add time to train logging

karmel:
This writes the end time of training as well as elapsed time to benchmark_run.log. If used as in the resnet loop, this will be equivalent to time-to-converge in cases where we run from start to convergence.

Note that this does not handle run starting and stopping; this is intentional for now.
Note also that the user bears the burden of calling the log_run_end method; this is also intentional, as we want to maintain flexibility in the face of different model architectures and rule sets.

@yhliang2018 - For new models added, we would want to add the logging logic as in Resnet.
@qlzh727 - This will require an update to the BigQuery Table, which, once approved, I can do.

Sample output (emphasis added):

{**"end_date": "2018-04-04T20:49:41.972510Z"**, "run_date": "2018-04-04T20:15:49.432197Z", "machine_config": {"cpu_info": {"cpu_info": "Intel(R) Xeon(R) CPU @ 2.30GHz", "mhz_per_cpu": 2300.0, "num_cores": 32}, "gpu_info": {"count": 4, "model": "Tesla P100-PCIE-16GB"}, "memory_total": 126739050496, "memory_available": 121141805056}, "tensorflow_version": {"git_hash": "v1.7.0-rc1-1091-gc7a04561fb", "version": "1.8.0-dev20180402"}, **"elapsed_time": "0:33:52.540313"**, **"elapsed_seconds": 2032.540313**, "tensorflow_environment_variables": [{"name": "TF_ENABLE_WINOGRAD_NONFUSED", "value": "1"}], "model_name": "resnet"}

### DIFF
--- a/official/resnet/resnet_run_loop.py
+++ b/official/resnet/resnet_run_loop.py
@@ -369,7 +369,7 @@ def resnet_main(flags, model_function, input_function, shape=None):
 
   if flags.benchmark_log_dir is not None:
     benchmark_logger = logger.BenchmarkLogger(flags.benchmark_log_dir)
-    benchmark_logger.log_run_info("resnet")
+    benchmark_logger.log_run_info('resnet')
   else:
     benchmark_logger = None
 
@@ -407,6 +407,9 @@ def resnet_main(flags, model_function, input_function, shape=None):
 
     if benchmark_logger:
       benchmark_logger.log_estimator_evaluation_result(eval_results)
+
+  if benchmark_logger:
+    benchmark_logger.log_run_end()
 
   if flags.export_dir is not None:
     warn_on_multi_gpu_export(flags.multi_gpu)
--- a/official/utils/logs/logger.py
+++ b/official/utils/logs/logger.py
@@ -44,6 +44,8 @@ class BenchmarkLogger(object):
     if not tf.gfile.IsDirectory(self._logging_dir):
       tf.gfile.MakeDirs(self._logging_dir)
 
+    self.run_date = datetime.datetime.now()
+
   def log_estimator_evaluation_result(self, eval_results):
     """Log the evaluation result for a estimator.
 
@@ -92,8 +94,7 @@ class BenchmarkLogger(object):
           "value": float(value),
           "unit": unit,
           "global_step": global_step,
-          "timestamp": datetime.datetime.now().strftime(
-              _DATE_TIME_FORMAT_PATTERN),
+          "timestamp": _formatted_date(),
           "extras": extras}
       try:
         json.dump(metric, f)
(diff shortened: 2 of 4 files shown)


official/resnet/resnet_run_loop.py
```diff
@@ -408,6 +408,9 @@ def input_fn_eval():
     if benchmark_logger:
       benchmark_logger.log_estimator_evaluation_result(eval_results)
 
+  if benchmark_logger:
```
karmel: This feels very inelegant to me, but I haven't come up with a better solution, at least assuming that explicit >> implicit.

official/utils/logs/logger.py
```diff
@@ -44,6 +44,8 @@ def __init__(self, logging_dir):
     if not tf.gfile.IsDirectory(self._logging_dir):
       tf.gfile.MakeDirs(self._logging_dir)
 
+    self.run_date = datetime.datetime.now()
```
qlzh727: I think its better to have a new method to log the run_date, since its not necessary true that the training start when the logger instance is created.

official/utils/logs/logger.py
```diff
@@ -113,21 +114,56 @@ def log_run_info(self, model_name):
...
         json.dump(run_info, f)
         f.write("\n")
       except (TypeError, ValueError) as e:
-        tf.logging.warning("Failed to dump benchmark run info to log file: %s",
-                           e)
+        tf.logging.warning(
+            "Failed to dump benchmark run info to log file: %s", e)
+
+
+def _formatted_date(date=None):
```
qlzh727: I think its more intuitive to not have the default date, and force caller to provide the value.

official/utils/logs/logger.py
qlzh727: Also, i think its better to keep as fewer internal field as possible. You can get the start timestamp by reading the log file.

official/utils/logs/logger.py
```diff
@@ -113,21 +114,56 @@ def log_run_info(self, model_name):
...
+    self._log_run_info(run_info)
+
+  def log_run_end(self):
+    """Log timing information at the end of a run."""
+    end_date = datetime.datetime.now()
+    delta = end_date - self.run_date
+    end_info = {
+        "end_date": _formatted_date(end_date),
+        "elapsed_time": str(delta),
+        "elapsed_seconds": delta.total_seconds()
```
qlzh727: Since those two field are just derived data, can we generate them on the repository side?

official/utils/logs/logger.py
yhliang2018: Will reading from the log file take longer?

official/utils/logs/logger.py
```diff
@@ -44,6 +44,8 @@ def __init__(self, logging_dir):
     if not tf.gfile.IsDirectory(self._logging_dir):
       tf.gfile.MakeDirs(self._logging_dir)
 
+    self.run_date = datetime.datetime.now()
+
   def log_estimator_evaluation_result(self, eval_results):
     """Log the evaluation result for a estimator.
```
yhliang2018: nit: an estimator

PR #3887, opened 2018-04-05
Resnet distribution strategies

robieta:
This PR pulls the work of the distribution strategies team back into official/resnet. Specifically it removes replicate model function and tower optimizer, incorporates some changes to the dataset pipeline, and uses various distribution strategies in the estimators.

@guptapriya Would you be so kind as to check that this is a faithful port of your code?

So far I have only performed one run: ResNet_50_v2 on ImageNet. It converged to 76% in 2 days. I will be performing a full battery of runs in the coming days.

### DIFF
--- a/official/resnet/cifar10_main.py
+++ b/official/resnet/cifar10_main.py
@@ -104,7 +104,8 @@ def preprocess_image(image, is_training):
 
 
 def input_fn(is_training, data_dir, batch_size, num_epochs=1,
-             num_parallel_calls=1, multi_gpu=False):
+             num_parallel_calls=1, use_distribution_strategy=False,
+             gpus_for_distribution_strategy=1):
   """Input_fn using the tf.data input pipeline for CIFAR-10 dataset.
 
   Args:
@@ -130,7 +131,10 @@ def input_fn(is_training, data_dir, batch_size, num_epochs=1,
   return resnet_run_loop.process_record_dataset(
       dataset, is_training, batch_size, _NUM_IMAGES['train'],
       parse_record, num_epochs, num_parallel_calls,
-      examples_per_epoch=num_images, multi_gpu=multi_gpu)
+      examples_per_epoch=num_images,
+      use_distribution_strategy=use_distribution_strategy,
+      gpus_for_distribution_strategy=gpus_for_distribution_strategy
+  )
 
 
 def get_synth_input_fn():
@@ -211,8 +215,7 @@ def cifar10_model_fn(features, labels, mode, params):
                                          momentum=0.9,
                                          data_format=params['data_format'],
                                          version=params['version'],
-                                         loss_filter_fn=loss_filter_fn,
-                                         multi_gpu=params['multi_gpu'])
+                                         loss_filter_fn=loss_filter_fn)
 
 
 def main(argv):
--- a/official/resnet/cifar10_main.py
+++ b/official/resnet/cifar10_main.py
@@ -104,7 +104,7 @@ def preprocess_image(image, is_training):
 
 
 def input_fn(is_training, data_dir, batch_size, num_epochs=1,
-             num_parallel_calls=1, use_distribution_strategy=False,
+             use_distribution_strategy=False,
              gpus_for_distribution_strategy=1):
   """Input_fn using the tf.data input pipeline for CIFAR-10 dataset.
 
@@ -113,12 +113,9 @@ def input_fn(is_training, data_dir, batch_size, num_epochs=1,
     data_dir: The directory containing the input data.
     batch_size: The number of samples per batch.
     num_epochs: The number of epochs to repeat the dataset.
-    num_parallel_calls: The number of records that are processed in parallel.
-      This can be optimized per data set but for generally homogeneous data
-      sets, should be approximately the number of available CPU cores.
-    multi_gpu: Whether this is run multi-GPU. Note that this is only required
-      currently to handle the batch leftovers, and can be removed
-      when that is handled directly by Estimator.
+    use_distribution_strategy: Whether DistributionStrategies API is used.
+    gpus_for_distribution_strategy: How many GPUs are used with
+      DistributionStrategies.
 
   Returns:
     A dataset that can be used for iteration.
@@ -130,7 +127,7 @@ def input_fn(is_training, data_dir, batch_size, num_epochs=1,
 
   return resnet_run_loop.process_record_dataset(
       dataset, is_training, batch_size, _NUM_IMAGES['train'],
-      parse_record, num_epochs, num_parallel_calls,
+      parse_record, num_epochs,
       examples_per_epoch=num_images,
       use_distribution_strategy=use_distribution_strategy,
       gpus_for_distribution_strategy=gpus_for_distribution_strategy
--- a/official/resnet/resnet_run_loop.py
+++ b/official/resnet/resnet_run_loop.py
@@ -26,7 +26,8 @@ from __future__ import print_function
 import argparse
 import os
 
-import tensorflow as tf  # pylint: disable=g-bad-import-order
+# pylint: disable=g-bad-import-order
+import tensorflow as tf
 from tensorflow.contrib.distribute.python import mirrored_strategy
 from tensorflow.contrib.distribute.python import one_device_strategy
 from tensorflow.python.client import device_lib
@@ -37,14 +38,14 @@ from official.utils.arg_parsers import parsers
 from official.utils.export import export
 from official.utils.logs import hooks_helper
 from official.utils.logs import logger
+# pylint: enable=g-bad-import-order
 
 
 ################################################################################
 # Functions for input processing.
 ################################################################################
 def process_record_dataset(dataset, is_training, batch_size, shuffle_buffer,
-                           parse_record_fn, num_epochs=1, num_parallel_calls=1,
-                           examples_per_epoch=0,
+                           parse_record_fn, num_epochs=1, examples_per_epoch=0,
                            use_distribution_strategy=False,
                            gpus_for_distribution_strategy=1):
   """Given a Dataset with raw records, return an iterator over the records.
@@ -59,9 +60,6 @@ def process_record_dataset(dataset, is_training, batch_size, shuffle_buffer,
     parse_record_fn: A function that takes a raw record and returns the
       corresponding (image, label) pair.
     num_epochs: The number of epochs to repeat the dataset.
-    num_parallel_calls: The number of records that are processed in parallel.
-      This can be optimized per data set but for generally homogeneous data
-      sets, should be approximately the number of available CPU cores.
     examples_per_epoch: The number of examples in the current set that
       are processed each epoch. Note that this is only used for multi-GPU mode,
       and only to handle what will eventually be handled inside of Estimator.
@@ -138,10 +136,10 @@ def get_synth_input_fn(height, width, num_channels, num_classes):
     An input_fn that can be used in place of a real one to return a dataset
     that can be used for iteration.
   """
-  def input_fn(is_training, data_dir, batch_size,
-      use_distribution_strategy=False, gpus_for_distribution_strategy=1,
-      *args, **kwargs):  # pylint: disable=unused-argument
-    # TODO(taylorrobie@) remove when DistributionStrategies uses global batch size
+  def input_fn(is_training, data_dir, batch_size,  # pylint: disable=unused-argument
+               use_distribution_strategy=False,
+               gpus_for_distribution_strategy=1, *args, **kwargs):  # pylint: disable=unused-argument
+    # TODO(taylorrobie@) cull DistributionStrategies uses global batch size
     per_device_batch_size = compute_per_device_batch_size(
         batch_size=batch_size,
         use_distribution_strategy=use_distribution_strategy,
@@ -320,7 +318,17 @@ def compute_per_device_batch_size(batch_size, use_distribution_strategy,
   directly. Multi-GPU support is currently experimental, however,
   so doing the work here until that feature is in place.
 
+  Args:
+    batch_size: Global batch size to be divided among devices.
+    use_distribution_strategy: Whether DistributionStrategies API is used.
+    gpus_for_distribution_strategy: How many GPUs are used with
+      DistributionStrategies.
 
+  Returns:
+    Batch size per device.
+
+  Raises:
+    ValueError: if batch_size is not divisible by number of devices
   """
   if use_distribution_strategy and gpus_for_distribution_strategy > 1:
     remainder = batch_size % gpus_for_distribution_strategy
@@ -328,8 +336,8 @@ def compute_per_device_batch_size(batch_size, use_distribution_strategy,
       err = ('When running with multiple GPUs, batch size '
              'must be a multiple of the number of available GPUs. Found {} '
              'GPUs with a batch size of {}; try --batch_size={} instead.'
-             ).format(gpus_for_distribution_strategy, batch_size,
-                      batch_size - remainder)
+            ).format(gpus_for_distribution_strategy, batch_size,
+                     batch_size - remainder)
       raise ValueError(err)
     return int(batch_size / gpus_for_distribution_strategy)
   return batch_size
@@ -350,13 +358,13 @@ def assign_multi_gpu(flags):
   if (flags.use_distribution_strategy and
       flags.gpus_for_distribution_strategy > 1):
     if flags.gpus_for_distribution_strategy > num_gpus:
-      raise ValueError("{} GPUs specified, only {} detected.".format(
+      raise ValueError('{} GPUs specified, only {} detected.'.format(
           flags.gpus_for_distribution_strategy,
           num_gpus
       ))
-    vars(flags)["multi_gpu"] = True
+    vars(flags)['multi_gpu'] = True
   else:
-    vars(flags)["multi_gpu"] = False
+    vars(flags)['multi_gpu'] = False
 
 
 def resnet_main(flags, model_function, input_function, shape=None):
@@ -418,7 +426,7 @@ def resnet_main(flags, model_function, input_function, shape=None):
 
   if flags.benchmark_log_dir is not None:
     benchmark_logger = logger.BenchmarkLogger(flags.benchmark_log_dir)
-    benchmark_logger.log_run_info("resnet")
+    benchmark_logger.log_run_info('resnet')
   else:
     benchmark_logger = None
 
@@ -493,7 +501,7 @@ class ResnetArgParser(argparse.ArgumentParser):
     super(ResnetArgParser, self).__init__(parents=[
         parsers.BaseParser(multi_gpu=False),
         parsers.DistributionStrategiesParser(),
-        parsers.PerformanceParser(),
+        parsers.PerformanceParser(num_parallel_calls=False),
         parsers.ImageModelParser(),
         parsers.ExportParser(),
         parsers.BenchmarkParser(),
--- a/official/resnet/resnet_run_loop.py
+++ b/official/resnet/resnet_run_loop.py
@@ -27,6 +27,10 @@ import argparse
 import os
 
 import tensorflow as tf  # pylint: disable=g-bad-import-order
+from tensorflow.contrib.distribute.python import mirrored_strategy
+from tensorflow.contrib.distribute.python import one_device_strategy
+from tensorflow.python.client import device_lib
+from tensorflow.python.training import distribute as distribute_lib
 
 from official.resnet import resnet_model
 from official.utils.arg_parsers import parsers
@@ -40,7 +44,9 @@ from official.utils.logs import logger
 ################################################################################
 def process_record_dataset(dataset, is_training, batch_size, shuffle_buffer,
                            parse_record_fn, num_epochs=1, num_parallel_calls=1,
-                           examples_per_epoch=0, multi_gpu=False):
+                           examples_per_epoch=0,
+                           use_distribution_strategy=False,
+                           gpus_for_distribution_strategy=1):
   """Given a Dataset with raw records, return an iterator over the records.
 
   Args:
@@ -59,9 +65,9 @@ def process_record_dataset(dataset, is_training, batch_size, shuffle_buffer,
     examples_per_epoch: The number of examples in the current set that
       are processed each epoch. Note that this is only used for multi-GPU mode,
       and only to handle what will eventually be handled inside of Estimator.
-    multi_gpu: Whether this is run multi-GPU. Note that this is only required
-      currently to handle the batch leftovers (see below), and can be removed
-      when that is handled directly by Estimator.
+    use_distribution_strategy: Whether DistributionStrategies API is used.
+    gpus_for_distribution_strategy: How many GPUs are used with
+      DistributionStrategies.
 
   Returns:
     Dataset of (image, label) pairs ready for iteration.
@@ -86,21 +92,26 @@ def process_record_dataset(dataset, is_training, batch_size, shuffle_buffer,
   # last batch of an epoch-- do not raise an error when we try to split them
   # over the GPUs. This will likely be handled by Estimator during replication
   # in the future, but for now, we just drop the leftovers here.
-  if multi_gpu:
+  if use_distribution_strategy:
     total_examples = num_epochs * examples_per_epoch
     dataset = dataset.take(batch_size * (total_examples // batch_size))
 
   # Parse the raw records into images and labels
-  dataset = dataset.map(lambda value: parse_record_fn(value, is_training),
-                        num_parallel_calls=num_parallel_calls)
-
-  dataset = dataset.batch(batch_size)
+  dataset = dataset.apply(
+      tf.contrib.data.map_and_batch(
+          lambda value: parse_record_fn(value, is_training),
+          batch_size=batch_size,
+          num_parallel_batches=1))
 
   # Operations between the final prefetch and the get_next call to the iterator
   # will happen synchronously during run time. We prefetch here again to
   # background all of the above processing work and keep it out of the
-  # critical training path.
-  dataset = dataset.prefetch(1)
+  # critical training path. In the future Estimator will likely scale prefetch
+  # for multi-GPU automatically.
+  if use_distribution_strategy:
+    dataset = dataset.prefetch(gpus_for_distribution_strategy)
+  else:
+    dataset = dataset.prefetch(1)
 
   return dataset
 
@@ -122,7 +133,7 @@ def get_synth_input_fn(height, width, num_channels, num_classes):
     An input_fn that can be used in place of a real one to return a dataset
     that can be used for iteration.
   """
-  def input_fn(is_training, data_dir, batch_size, *args):  # pylint: disable=unused-argument
+  def input_fn(is_training, data_dir, batch_size, *args, **kwargs):  # pylint: disable=unused-argument
     images = tf.zeros((batch_size, height, width, num_channels), tf.float32)
     labels = tf.zeros((batch_size, num_classes), tf.int32)
     return tf.data.Dataset.from_tensors((images, labels)).repeat()
@@ -170,7 +181,7 @@ def learning_rate_with_decay(
 
 def resnet_model_fn(features, labels, mode, model_class,
                     resnet_size, weight_decay, learning_rate_fn, momentum,
-                    data_format, version, loss_filter_fn=None, multi_gpu=False):
+                    data_format, version, loss_filter_fn=None):
   """Shared functionality for different resnet model_fns.
 
   Initializes the ResnetModel representing the model layers
@@ -200,8 +211,6 @@ def resnet_model_fn(features, labels, mode, model_class,
       True if the var should be included in loss calculation, and False
       otherwise. If None, batch_normalization variables will be excluded
       from the loss.
-    multi_gpu: If True, wrap the optimizer in a TowerOptimizer suitable for
-      data-parallel distribution across multiple GPUs.
 
   Returns:
     EstimatorSpec parameterized according to the input params and the
@@ -260,19 +269,21 @@ def resnet_model_fn(features, labels, mode, model_class,
 
     optimizer = tf.train.MomentumOptimizer(
         learning_rate=learning_rate,
-        momentum=momentum)
-
-    # If we are running multi-GPU, we need to wrap the optimizer.
-    if multi_gpu:
-      optimizer = tf.contrib.estimator.TowerOptimizer(optimizer)
+        momentum=momentum
+    )
 
     update_ops = tf.get_collection(tf.GraphKeys.UPDATE_OPS)
     train_op = tf.group(optimizer.minimize(loss, global_step), update_ops)
   else:
     train_op = None
 
-  accuracy = tf.metrics.accuracy(
-      tf.argmax(labels, axis=1), predictions['classes'])
+  if not distribute_lib.has_distribution_strategy():
+    accuracy = tf.metrics.accuracy(
+        tf.argmax(labels, axis=1), predictions['classes'])
+  else:
+    # Metrics are currently no compatible with distribution strategies
+    accuracy = (tf.no_op(), tf.constant(0))
+
   metrics = {'accuracy': accuracy}
 
   # Create a tensor named train_accuracy for logging purposes
@@ -300,7 +311,6 @@ def validate_batch_size_for_multi_gpu(batch_size):
   Raises:
     ValueError: if no GPUs are found, or selected batch_size is invalid.
   """
-  from tensorflow.python.client import device_lib  # pylint: disable=g-import-not-at-top
 
   local_device_protos = device_lib.list_local_devices()
   num_gpus = sum([1 for d in local_device_protos if d.device_type == 'GPU'])
@@ -355,8 +365,23 @@ def resnet_main(flags, model_function, input_function, shape=None):
       allow_soft_placement=True)
 
   # Set up a RunConfig to save checkpoint and set session config.
-  run_config = tf.estimator.RunConfig().replace(save_checkpoints_secs=1e9,
-                                                session_config=session_config)
+  if not flags.use_distribution_strategy:
+    run_config = tf.estimator.RunConfig().replace(
+        save_checkpoints_secs=1e9, session_config=session_config)
+  else:
+    if flags.gpus_for_distribution_strategy == 0:
+      distribution = one_device_strategy.OneDeviceStrategy(
+          'device:CPU:0')
+    elif flags.gpus_for_distribution_strategy == 1:
+      distribution = one_device_strategy.OneDeviceStrategy(
+          'device:GPU:0')
+    else:
+      distribution = mirrored_strategy.MirroredStrategy(
+          num_gpus=flags.gpus_for_distribution_strategy
+      )
+    run_config = tf.estimator.RunConfig(distribute=distribution).replace(
+        save_checkpoints_secs=1e9, session_config=session_config)
+
   classifier = tf.estimator.Estimator(
       model_fn=model_function, model_dir=flags.model_dir, config=run_config,
       params={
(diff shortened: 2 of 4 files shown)


official/utils/arg_parsers/parsers.py
```diff
@@ -118,6 +118,7 @@ def __init__(self, add_help=False, data_dir=True, model_dir=True,
           metavar="<BS>"
       )
 
+    # TODO(taylorrobie@): depricate and only use DistributionStrategies
```
k-w-w: nit: (sp) deprecate

official/utils/arg_parsers/parsers.py
```diff
@@ -151,6 +152,7 @@ def __init__(self, add_help=False, num_parallel_calls=True, inter_op=True,
                intra_op=True, use_synthetic_data=True, max_train_steps=True):
     super(PerformanceParser, self).__init__(add_help=add_help)
 
+    # TODO(taylorrobie@): depricate and only use DistributionStrategies
```
k-w-w: nit: (sp) deprecate

official/resnet/resnet_run_loop.py
```diff
@@ -86,21 +98,23 @@ def process_record_dataset(dataset, is_training, batch_size, shuffle_buffer,
...
   # Parse the raw records into images and labels
-  dataset = dataset.map(lambda value: parse_record_fn(value, is_training),
-                        num_parallel_calls=num_parallel_calls)
-
-  dataset = dataset.batch(batch_size)
+  dataset = dataset.apply(
+      tf.contrib.data.map_and_batch(
+          lambda value: parse_record_fn(value, is_training),
+          batch_size=per_device_batch_size,
+          num_parallel_batches=1))
```
k-w-w: Why is num_parallel_batches set to 1 (wouldn't more improve performance)?

official/resnet/resnet_run_loop.py
```diff
@@ -260,19 +281,21 @@ def exclude_batch_norm(name):
...
   else:
     train_op = None
 
-  accuracy = tf.metrics.accuracy(
-      tf.argmax(labels, axis=1), predictions['classes'])
+  if not distribute_lib.has_distribution_strategy():
+    accuracy = tf.metrics.accuracy(
+        tf.argmax(labels, axis=1), predictions['classes'])
+  else:
+    # Metrics are currently no compatible with distribution strategies
```
k-w-w: nit: not compatible

official/resnet/resnet_run_loop.py
```diff
@@ -335,15 +385,8 @@ def resnet_main(flags, model_function, input_function, shape=None):
...
-    validate_batch_size_for_multi_gpu(flags.batch_size)
-
-    # There are two steps required if using multi-GPU: (1) wrap the model_fn,
-    # and (2) wrap the optimizer. The first happens here, and (2) happens
-    # in the model_fn itself when the optimizer is defined.
-    model_function = tf.contrib.estimator.replicate_model_fn(
-        model_function,
-        loss_reduction=tf.losses.Reduction.MEAN)
+  # TODO(taylorrobie@): remove when per_device is no longer needed.
+  assign_multi_gpu(flags)
```
k-w-w: How does this relate to per_device? 

Also, it appears flags.multi_gpu is only used in the line `   warn_on_multi_gpu_export(flags.multi_gpu)`. This can be changed to use flags.use_distribution_strategy, and the multi_gpu flag can be completely removed.

official/resnet/resnet_run_loop.py
robieta: It still parallelizes the lambda call. So we probably only need >1 if there are stragglers, n_cores > batch_size.

official/resnet/resnet_run_loop.py
robieta: fixed.

official/resnet/resnet_run_loop.py
robieta: It turns out we can probably remove multi_gpu altogether. Karmel just has to check that it doesn't break saved models.

official/utils/arg_parsers/parsers.py
robieta: fixed.

official/utils/arg_parsers/parsers.py
robieta: fixed.

official/resnet/cifar10_test.py
```diff
@@ -110,14 +109,6 @@ def test_cifar10_model_fn_train_mode_v1(self):
   def test_cifar10_model_fn_trainmode__v2(self):
     self.cifar10_model_fn_helper(tf.estimator.ModeKeys.TRAIN, version=2)
 
-  def test_cifar10_model_fn_train_mode_multi_gpu_v1(self):
```
guptapriya: could we instead change these tests to run distributed strategy version on multiple GPUs?

official/resnet/resnet_run_loop.py
```diff
@@ -26,21 +26,27 @@
 import argparse
 import os
 
-import tensorflow as tf  # pylint: disable=g-bad-import-order
+# pylint: disable=g-bad-import-order
+import tensorflow as tf
+from tensorflow.contrib.distribute.python import mirrored_strategy
```
guptapriya: actually you shouldn't need this. you can directly use them as 
`tf.contrib.distribute.MirroredStrategy`
`tf.contrib.distribute.OneDeviceStrategy`

official/resnet/resnet_run_loop.py
guptapriya: if you want to make this more generic, the ideal thing is num_parallel_batches = num_cores / batch_size
Since usually batch_size > num_cores, i used 1. But yeah you could do that division and take the cieling or something?

official/resnet/resnet_run_loop.py
```diff
@@ -86,21 +97,23 @@ def process_record_dataset(dataset, is_training, batch_size, shuffle_buffer,
...
+          lambda value: parse_record_fn(value, is_training),
+          batch_size=per_device_batch_size,
+          num_parallel_batches=1))
 
   # Operations between the final prefetch and the get_next call to the iterator
   # will happen synchronously during run time. We prefetch here again to
   # background all of the above processing work and keep it out of the
   # critical training path.
-  dataset = dataset.prefetch(1)
+  dataset.prefetch(buffer_size=tf.contrib.data.AUTOTUNE)
```
guptapriya: thanks!

official/resnet/resnet_run_loop.py
```diff
@@ -260,19 +280,21 @@ def exclude_batch_norm(name):
...
+    )
 
     update_ops = tf.get_collection(tf.GraphKeys.UPDATE_OPS)
     train_op = tf.group(optimizer.minimize(loss, global_step), update_ops)
   else:
     train_op = None
 
-  accuracy = tf.metrics.accuracy(
-      tf.argmax(labels, axis=1), predictions['classes'])
+  if not distribute_lib.has_distribution_strategy():
```
guptapriya: this can be `tf.contrib.distribute.has_distribution_strategy()`

official/resnet/resnet_run_loop.py
```diff
@@ -287,34 +309,37 @@ def exclude_batch_norm(name):
...
+    Batch size per device.
 
   Raises:
-    ValueError: if no GPUs are found, or selected batch_size is invalid.
+    ValueError: if batch_size is not divisible by number of devices
   """
-  from tensorflow.python.client import device_lib  # pylint: disable=g-import-not-at-top
-
-  local_device_protos = device_lib.list_local_devices()
-  num_gpus = sum([1 for d in local_device_protos if d.device_type == 'GPU'])
```
guptapriya: i wonder if this check is still useful in some form... for e.g. should we check that actual number of gpus available >= gpus_for_distribution_strategy

official/utils/arg_parsers/parsers.py
```diff
@@ -202,6 +204,33 @@ def __init__(self, add_help=False, num_parallel_calls=True, inter_op=True,
       )
 
 
+class DistributionStrategiesParser(argparse.ArgumentParser):
+  """Parser to add flags need for distribution strategies.
+
+  Args:
+    add_help: Create the "--help" flag. False if class instance is a parent.
+    batch_size: Create a flag to specify the batch size. (Instead of the one
```
guptapriya: what is this batch size here?

official/utils/arg_parsers/parsers.py
```diff
@@ -202,6 +204,33 @@ def __init__(self, add_help=False, num_parallel_calls=True, inter_op=True,
...
+    super(DistributionStrategiesParser, self).__init__(add_help=add_help)
+
+    self.add_argument(
+        "--use_distribution_strategy", "-uds",
+        action="store_true",
+        help="If set, use the DistributionStrategies API."
+    )
+
+    self.add_argument(
+        "--gpus_for_distribution_strategy", "-gds",
```
guptapriya: wonder if we should just rename `gpus_for_distribution_strategy` -> `num_gpus` everywhere?

official/resnet/resnet_run_loop.py
```diff
@@ -26,21 +26,27 @@
...
+from tensorflow.contrib.distribute.python import mirrored_strategy
+from tensorflow.contrib.distribute.python import one_device_strategy
+from tensorflow.python.training import distribute as distribute_lib
 
 from official.resnet import resnet_model
 from official.utils.arg_parsers import parsers
 from official.utils.export import export
 from official.utils.logs import hooks_helper
 from official.utils.logs import logger
+# pylint: enable=g-bad-import-order
```
karmel: Once we move to top-level imports as @guptapriya notes above, we can just switch back to the single-line import order.

official/resnet/cifar10_main.py
```diff
@@ -104,20 +104,18 @@ def preprocess_image(image, is_training):
 
 
 def input_fn(is_training, data_dir, batch_size, num_epochs=1,
-             num_parallel_calls=1, multi_gpu=False):
+             use_distribution_strategy=False,
+             gpus_for_distribution_strategy=1):
```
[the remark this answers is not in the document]
karmel: (Ditto throughout.)

official/resnet/resnet_run_loop.py
```diff
@@ -53,22 +59,27 @@ def process_record_dataset(dataset, is_training, batch_size, shuffle_buffer,
...
-      when that is handled directly by Estimator.
+    use_distribution_strategy: Whether DistributionStrategies API is used.
+    gpus_for_distribution_strategy: How many GPUs are used with
+      DistributionStrategies.
 
   Returns:
     Dataset of (image, label) pairs ready for iteration.
   """
+
+  # TODO(taylorrobie@) remove when DistributionStrategies uses global batch size
```
karmel: Let's not leave TODOs in the public code; a comment explaining that this is only necessary for a short period because etc etc is sufficient. Also, nit, for future reference: TODO(taylorrobie), and, this is public code, so, robieta.

official/resnet/resnet_run_loop.py
```diff
@@ -86,21 +97,23 @@ def process_record_dataset(dataset, is_training, batch_size, shuffle_buffer,
   # last batch of an epoch-- do not raise an error when we try to split them
   # over the GPUs. This will likely be handled by Estimator during replication
   # in the future, but for now, we just drop the leftovers here.
-  if multi_gpu:
+  if use_distribution_strategy:
     total_examples = num_epochs * examples_per_epoch
-    dataset = dataset.take(batch_size * (total_examples // batch_size))
+    dataset = dataset.take(
+        per_device_batch_size * (total_examples // per_device_batch_size))
```
karmel: Is this still necessary? I would imagine not? This was originally a fix for the fact that replicate_model_fn would error out, as noted in the comment (which should also be updated in the unexpected case that this is still relevant). It would be great to remove this, because then we can not pass multi-gpu knowledge this far in, and compute batch size closer to the top of the processing, pass in the desired batch_size here without this func caring whether it's global or per device.

official/resnet/resnet_run_loop.py
karmel: Interestingly, when I had tried this with replicate_model_fn alone, it actually slowed performance. Same for tf_cnn. See https://github.com/tensorflow/benchmarks/issues/137 , and the abandoned branch in which I attempted- https://github.com/tensorflow/models/compare/feat/contrib-data . This is JFYI assuming that DistStrat gets better performance, but not a bad idea to take in num_cpus in any case, as Transformer uses that as well, will be useful elsewhere.

official/resnet/resnet_run_loop.py
karmel: Please add instructional comments-- what is autotune? What does it do?

official/resnet/resnet_run_loop.py
```diff
@@ -122,9 +135,18 @@ def get_synth_input_fn(height, width, num_channels, num_classes):
     An input_fn that can be used in place of a real one to return a dataset
     that can be used for iteration.
   """
-  def input_fn(is_training, data_dir, batch_size, *args):  # pylint: disable=unused-argument
-    images = tf.zeros((batch_size, height, width, num_channels), tf.float32)
-    labels = tf.zeros((batch_size, num_classes), tf.int32)
+  def input_fn(is_training, data_dir, batch_size,  # pylint: disable=unused-argument, missing-docstring
+               use_distribution_strategy=False,
+               gpus_for_distribution_strategy=1, *args, **kwargs):  # pylint: disable=unused-argument
+    # TODO(taylorrobie@) cull DistributionStrategies uses global batch size
```
karmel: Ditto on TODOs.

karmel: Also, ditto on I think we can move this higher up, do once, and just have this function be blind to where the batch_size is coming from.

official/resnet/resnet_run_loop.py
```diff
@@ -260,19 +280,21 @@ def exclude_batch_norm(name):
 
     optimizer = tf.train.MomentumOptimizer(
         learning_rate=learning_rate,
-        momentum=momentum)
-
-    # If we are running multi-GPU, we need to wrap the optimizer.
-    if multi_gpu:
-      optimizer = tf.contrib.estimator.TowerOptimizer(optimizer)
+        momentum=momentum
```
karmel: Yay.

official/resnet/resnet_run_loop.py
karmel: If we can check in code whether we have a dist strat via a tf function, why do we need to pass around use_distribution_strategies?

official/resnet/resnet_run_loop.py
```diff
@@ -260,19 +280,21 @@ def exclude_batch_norm(name):
...
     train_op = None
 
-  accuracy = tf.metrics.accuracy(
-      tf.argmax(labels, axis=1), predictions['classes'])
+  if not distribute_lib.has_distribution_strategy():
+    accuracy = tf.metrics.accuracy(
+        tf.argmax(labels, axis=1), predictions['classes'])
+  else:
+    # Metrics are currently not compatible with distribution strategies
+    accuracy = (tf.no_op(), tf.constant(0))
```
karmel: Please comment noting that this is just for training metrics, and has not been found to affect performance at least for this Resnet model.

official/resnet/resnet_run_loop.py
```diff
@@ -287,34 +309,37 @@ def exclude_batch_norm(name):
...
   """For multi-gpu, batch-size must be a multiple of the number of GPUs.
 
-  Note that this should eventually be handled by replicate_model_fn
+  Note that this should eventually be handled by DistributionStrategies
   directly. Multi-GPU support is currently experimental, however,
   so doing the work here until that feature is in place.
 
   Args:
-    batch_size: the number of examples processed in each training batch.
+    batch_size: Global batch size to be divided among devices.
```
karmel: Maybe add something like, This should be equal to num_gpus times the single-GPU batch_size.

official/resnet/resnet_run_loop.py
```diff
@@ -287,34 +309,37 @@ def exclude_batch_norm(name):
...
+    use_distribution_strategy: Whether DistributionStrategies API is used.
+    gpus_for_distribution_strategy: How many GPUs are used with
+      DistributionStrategies.
+
+  Returns:
+    Batch size per device.
```
karmel: We keep switching between calling them GPUs and devices, which is a little unclear. Until we are actually talking about anything other than GPUs, let's stick with the more externally-readable "GPU."

official/resnet/resnet_run_loop.py
```diff
@@ -287,34 +309,37 @@ def exclude_batch_norm(name):
...
-                     'were found. To use CPU, run without --multi_gpu.')
-
-  remainder = batch_size % num_gpus
-  if remainder:
-    err = ('When running with multiple GPUs, batch size '
-           'must be a multiple of the number of available GPUs. '
-           'Found {} GPUs with a batch size of {}; try --batch_size={} instead.'
-          ).format(num_gpus, batch_size, batch_size - remainder)
-    raise ValueError(err)
+  if use_distribution_strategy and gpus_for_distribution_strategy > 1:
```
karmel: As above, I don't see the value of the dual check here. If num_gpus > 1 should be sufficient, since we don't allow dist strat with num_gpu <= 1, and we don't have another way of doing multi-GPU.

official/resnet/resnet_run_loop.py
```diff
@@ -287,34 +309,37 @@ def exclude_batch_norm(name):
...
+    remainder = batch_size % gpus_for_distribution_strategy
+    if remainder:
+      err = ('When running with multiple GPUs, batch size '
+             'must be a multiple of the number of available GPUs. Found {} '
+             'GPUs with a batch size of {}; try --batch_size={} instead.'
+            ).format(gpus_for_distribution_strategy, batch_size,
+                     batch_size - remainder)
+      raise ValueError(err)
+    return int(batch_size / gpus_for_distribution_strategy)
+  return batch_size
```
karmel: If we're going to have two returns here, may as well un-indent and just return batch_size immediately if num_gpus <= 1. In general, I disprefer early returns, but in this implementation we still have an early return. Alternatively, just overwrite batch_size and return once.

official/resnet/resnet_run_loop.py
karmel: I feel like that check should live under the hood in DistStrat-- we had to break our own rules and import from the private API here. Is it feasible to do that on the DistStrat side? Or just ignore for now and let the error bubble up when it is hit...

official/resnet/resnet_run_loop.py
```diff
@@ -355,21 +370,35 @@ def resnet_main(flags, model_function, input_function, shape=None):
       allow_soft_placement=True)
```
karmel: The allow_soft_placement was only relevant for replicate_model_fn; @guptapriya -- still necessary to set this explicitly? I doubt it?

official/resnet/resnet_run_loop.py
```diff
@@ -355,21 +370,35 @@ def resnet_main(flags, model_function, input_function, shape=None):
...
 
   # Set up a RunConfig to save checkpoint and set session config.
-  run_config = tf.estimator.RunConfig().replace(save_checkpoints_secs=1e9,
-                                                session_config=session_config)
+  if not flags.use_distribution_strategy:
+    run_config = tf.estimator.RunConfig().replace(
+        save_checkpoints_secs=1e9, session_config=session_config)
```
karmel: I think the save_checkpoints_secs is a relic that may even be undesireable at this point; the default is 600 secs, which might be more reasonable than...never. In fact, now that I think about it, this explains why the checkpoints collected during training are only ~once per epoch, which seemed too few. (cf https://www.tensorflow.org/api_docs/python/tf/estimator/RunConfig ) Can we remove? Also, I think we can just initialize the RunConfig directly, without .replace. All of that would allow us to simplify this-- distribution=None unless we have these flags set, and we only init the RunConfig once.

official/resnet/resnet_run_loop.py
```diff
@@ -355,21 +370,35 @@ def resnet_main(flags, model_function, input_function, shape=None):
...
+      distribution = one_device_strategy.OneDeviceStrategy(
+          'device:CPU:0')
+    elif flags.gpus_for_distribution_strategy == 1:
+      distribution = one_device_strategy.OneDeviceStrategy(
+          'device:GPU:0')
+    else:
+      distribution = mirrored_strategy.MirroredStrategy(
+          num_gpus=flags.gpus_for_distribution_strategy
+      )
+    run_config = tf.estimator.RunConfig(distribute=distribution).replace(
```
karmel: Current TF master implies the kwarg is now train_distribute; @guptapriya , what is the current format?

official/resnet/resnet_run_loop.py
```diff
@@ -355,21 +370,35 @@ def resnet_main(flags, model_function, input_function, shape=None):
...
+        save_checkpoints_secs=1e9, session_config=session_config)
```
karmel: @guptapriya What is the effect of passing inter_ and intra_op parallelism with a dist strategy? Are they still respected?

official/resnet/resnet_run_loop.py
```diff
@@ -355,21 +370,35 @@ def resnet_main(flags, model_function, input_function, shape=None):
       allow_soft_placement=True)
 
   # Set up a RunConfig to save checkpoint and set session config.
-  run_config = tf.estimator.RunConfig().replace(save_checkpoints_secs=1e9,
-                                                session_config=session_config)
+  if not flags.use_distribution_strategy:
```
karmel: Proposition: We should just use DistStrat in all cases in this model now. Opinions? @guptapriya and @robieta for thoughts.

official/resnet/resnet_run_loop.py
```diff
@@ -355,21 +370,35 @@ def resnet_main(flags, model_function, input_function, shape=None):
...
           'data_format': flags.data_format,
           'batch_size': flags.batch_size,
-          'multi_gpu': flags.multi_gpu,
           'version': flags.version,
       })
 
   if flags.benchmark_log_dir is not None:
     benchmark_logger = logger.BenchmarkLogger(flags.benchmark_log_dir)
-    benchmark_logger.log_run_info("resnet")
+    benchmark_logger.log_run_info('resnet')
```
karmel: I swear I've changed this in about 5 branches now.

official/utils/arg_parsers/parsers.py
```diff
@@ -118,6 +118,7 @@ def __init__(self, add_help=False, data_dir=True, model_dir=True,
           metavar="<BS>"
       )
 
+    # TODO(taylorrobie@): deprecate and only use DistributionStrategies
```
karmel: Ditto on Todos.

official/utils/arg_parsers/parsers.py
```diff
@@ -151,6 +152,7 @@ def __init__(self, add_help=False, num_parallel_calls=True, inter_op=True,
                intra_op=True, use_synthetic_data=True, max_train_steps=True):
     super(PerformanceParser, self).__init__(add_help=add_help)
 
+    # TODO(taylorrobie@): deprecate and only use DistributionStrategies
```
karmel: Right now, only MNIST uses, correct? Can you follow up with making that use DistStrat as well? And, in theory, WideDeep should be easy, because it's just estimators... is that true? Can you check and update as well if so?

official/utils/arg_parsers/parsers.py
karmel: Yes please. And I think we can also plausibly use num_gpus=0, num_gpus=1 to represent the fact that we want diststrat even in single-device cases. That would allow us to ditch the second arg, and roll this back into the main parser, where we already have multi_gpu.

official/utils/arg_parsers/parsers.py
```diff
@@ -202,6 +204,33 @@ def __init__(self, add_help=False, num_parallel_calls=True, inter_op=True,
...
+        "--use_distribution_strategy", "-uds",
+        action="store_true",
+        help="If set, use the DistributionStrategies API."
+    )
+
+    self.add_argument(
+        "--gpus_for_distribution_strategy", "-gds",
+        type=int, default=2,
+        help="[default: %(default)s] How many GPUs to use with the "
+             "DistributionStrategies API.",
```
karmel: We need links to the docs somewhere, and this seems like a good place. Maybe also in the comment noting that multi-GPU is experimental.

karmel:
I just checked with a saved_model exported from this PR, and it does indeed appear that this fixes the problem with replicate_model_fn where you had to export the savedmodel on single gpu even if training on multi. CC @isaprykin and @k-w-w , who are interested parties. @robieta -- can you remove warn_on_multi_gpu as well?

official/resnet/cifar10_main.py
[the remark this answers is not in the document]
guptapriya: +1 
Think of the params etc that we had as something done in crunch time :) so it would be great to change it to whatever makes most sense for a user.

official/resnet/resnet_run_loop.py
guptapriya: Yes, that's definitely interesting. We saw a performance bump both in OneDeviceStrategy and MirroredStrategy with this approach.. agreed that it might still be worthwhile to take in the num cores and use that to get the num_parallel_batches

official/resnet/resnet_run_loop.py
guptapriya: You're absolutely right @karmel. Anytime one is in the distribution strategy scope, the `has_distribution_strategy` check should work. I believe in our earlier code, we didn't have the input function running inside the scope, so we had to pass around this boolean. But Since then we've moved input processing to be under distribution scope as well and can now use `has_distribution_strategy()` almost everywhere. 

I just checked where all it's used. I think it can be replaced with `has_distribution_strategy` in all places except this one in `resnet_main` . And you already suggested removing that check from there entirely. So I think we can get rid of this flag entirely!

official/resnet/resnet_run_loop.py
guptapriya: +1

official/resnet/resnet_run_loop.py
guptapriya: I think that's a good idea. When we originally added these flags, we were not removing the previous multi gpu approach. So we didn't want to enable if by default. But I think now it makes sense. `OneDeviceStrategy` should pretty much do what having no distribution strategy does.

official/resnet/resnet_run_loop.py
guptapriya: ah yes, sorry we just changed it after a discussion. you're right, it's called `train_distribute` now. thanks for catching!

official/resnet/resnet_run_loop.py
guptapriya: Hm, I am not a 100% sure. We did test with setting those as env variables during performance tuning but did not find significant benefits. But I don't believe we changed anything in dist strategy itself to support/not support them explicitly. @isaprykin perhaps can shed more light?

official/utils/arg_parsers/parsers.py
guptapriya: +1

official/utils/arg_parsers/parsers.py
guptapriya: we have a reasonable README here that we can link to? 
https://github.com/tensorflow/tensorflow/blob/master/tensorflow/contrib/distribute/README.md

official/resnet/resnet_run_loop.py
robieta: Fine by me.

official/resnet/resnet_run_loop.py
robieta: Perils of working from contrib. ^^

official/resnet/resnet_run_loop.py
robieta: No kidding.

official/utils/arg_parsers/parsers.py
robieta: Done.

official/resnet/resnet_run_loop.py
robieta: I just tested this on a trivial imagenet resnet (so any limit is the input pipeline), and found no difference with num_parallel_batches. (This is 32 cores, 4xP100, batch_size=512.) I'm hard pressed to think of a case where num_cores < batch_size makes sense, so I'm inclined to leave it at 1 to keep the code simple.

karmel: That sounds good to me if you add a comment describing the other approach. CC @mrry in case you have thoughts.

official/utils/arg_parsers/parsers.py
robieta: Global. I will make that more clear.

official/resnet/imagenet_test.py
```diff
@@ -214,14 +213,6 @@ def test_resnet_model_fn_train_mode_v1(self):
   def test_resnet_model_fn_train_mode_v2(self):
     self.resnet_model_fn_helper(tf.estimator.ModeKeys.TRAIN, version=2)
 
-  def test_resnet_model_fn_train_mode_multi_gpu_v1(self):
-    self.resnet_model_fn_helper(tf.estimator.ModeKeys.TRAIN, version=1,
-                                multi_gpu=True)
-
-  def test_resnet_model_fn_train_mode_multi_gpu_v2(self):
-    self.resnet_model_fn_helper(tf.estimator.ModeKeys.TRAIN, version=2,
-                                multi_gpu=True)
```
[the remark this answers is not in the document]
robieta: The thing is model_fn used to be device aware because of replicate_model_fn, and now it isn't. That's why I can safely remove that test. And we don't currently have the infrastructure for me to set up a multi-gpu end-to-end test.

official/resnet/resnet_run_loop.py
robieta: There is currently an issue with TensorFlow not respecting thread limits (https://github.com/tensorflow/tensorflow/issues/18300), so we'll have to wait until that is resolved to test this.

robieta:
It appears that some git tomfoolery has occurred. I will sort it out, and apologies to those of you who got dragged in as owners.

official/resnet/resnet_run_loop.py
mrry: Setting it to 1 seems like a fine way to go. Any larger would probably lead to congestion on the threadpool queues. The ideal number for a batch size of 512 is probably between 0 and 1, and there's an outstanding bug to support more precise control of the parallelism here. /cc @jsimsa

robieta:
@karmel I think we're ready for you to take another pass.

official/resnet/resnet_run_loop.py
isaprykin: Hi!  We didn't change anything regarding inter/intra op options.  We indeed tried them out, but they weren't super impactful that time.  I also remember Reed saying that these options have reasonable defaults.

official/utils/arg_parsers/parsers.py
```diff
@@ -99,14 +99,14 @@ class BaseParser(argparse.ArgumentParser):
     model_dir: Create a flag for specifying the model file directory.
     train_epochs: Create a flag to specify the number of training epochs.
     epochs_between_evals: Create a flag to specify the frequency of testing.
-    batch_size: Create a flag to specify the batch size.
+    batch_size: Create a flag to specify the global batch size.
     multi_gpu: Create a flag to allow the use of all available GPUs.
```
qlzh727: Missing doc for num_gpu

official/resnet/resnet_run_loop.py
```diff
@@ -78,29 +69,22 @@ def process_record_dataset(dataset, is_training, batch_size, shuffle_buffer,
...
 
   # Operations between the final prefetch and the get_next call to the iterator
   # will happen synchronously during run time. We prefetch here again to
   # background all of the above processing work and keep it out of the
-  # critical training path.
-  dataset = dataset.prefetch(1)
+  # critical training path. Setting buffer_size to tf.contrib.data.AUTOTUNE
+  # allows DistributionStrategies to adjust how many batches to fetch based
+  # on how many devices are present.
+  dataset.prefetch(buffer_size=tf.contrib.data.AUTOTUNE)
```
karmel: http://knowyourmeme.com/memes/auto-tune

official/resnet/resnet_run_loop.py
```diff
@@ -122,7 +106,7 @@ def get_synth_input_fn(height, width, num_channels, num_classes):
     An input_fn that can be used in place of a real one to return a dataset
     that can be used for iteration.
   """
-  def input_fn(is_training, data_dir, batch_size, *args):  # pylint: disable=unused-argument
+  def input_fn(is_training, data_dir, batch_size, *args, **kwargs):  # pylint: disable=unused-argument,missing-docstring
```
karmel: This function should be short enough to not require a docstring (< 10 lines); are you sure this extra disable is necessary?

official/resnet/resnet_run_loop.py
karmel: Tagging @isaprykin on this one too-- do we need allow_soft_placement still?

official/utils/arg_parsers/parsers.py
```diff
@@ -99,14 +99,14 @@ class BaseParser(argparse.ArgumentParser):
...
     hooks: Create a flag to specify hooks for logging.
   """
 
   def __init__(self, add_help=False, data_dir=True, model_dir=True,
                train_epochs=True, epochs_between_evals=True, batch_size=True,
-               multi_gpu=True, hooks=True):
+               multi_gpu=True, num_gpu=True, hooks=True):
```
karmel: I know it will be ripped out in a coming PR, but, for now, let's set the default for multi_gpu=False

official/utils/arg_parsers/parsers.py
```diff
@@ -142,16 +142,29 @@ def __init__(self, add_help=False, data_dir=True, model_dir=True,
...
+
     if multi_gpu:
       self.add_argument(
           "--multi_gpu", action="store_true",
           help="If set, run across all available GPUs."
       )
 
+    if num_gpu:
+      self.add_argument(
+          "--num_gpus", "-ng",
```
karmel: I feel like we are reaching some upper bound of abbreviation strings, but, then again, I am the type that always prefers the fully explicit versions.

official/utils/arg_parsers/parsers.py
```diff
@@ -142,16 +142,29 @@ def __init__(self, add_help=False, data_dir=True, model_dir=True,
...
+          type=int,
+          default=1 if tf.test.is_built_with_cuda() else 0,
```
karmel: Can you add a test to make sure this default gets set correctly in the arg parsers test? I think there's a way to force one mode or the other... if not, have you at least confirmed that this does work correctly on GPU/CPU?

official/utils/arg_parsers/parsers.py
```diff
@@ -142,16 +142,29 @@ def __init__(self, add_help=False, data_dir=True, model_dir=True,
...
+          help="[default: %(default)s] How many GPUs to use with the "
+               "DistributionStrategies API.",
```
karmel: Can you add a note here that reflects the details in the readme? Specifically, that 0==CPU, 1==GPU, default is what you built TF with.

official/resnet/resnet_run_loop.py
isaprykin: Sigh, yes, I think so.   Remember, this got added because Estimator wants allow_soft_placement=true and it sets it when session_config is not specified.  But when the session config is specified, then Estimator doesn't add allow_soft_placement.  Without allow_soft_placement, there device placement failures with no GPU kernels for some Ops...       I just looked at the Estimator code and I think this behavior is still in place.

official/resnet/resnet_run_loop.py
robieta: It was erroring on the Kokoro lint, but looks like it isn't now. It will forever remain a mystery.

official/utils/arg_parsers/parsers.py
robieta: U wil pry my CLI abr frm my cld, ded hnds.

official/utils/arg_parsers/parsers.py
robieta: I tested manually and confirmed. Probably not worth the effort to make a formal test simply because at that point it's more of a test of tf.test than the model garden.

PR #3890, opened 2018-04-06
Allow mnist data to be writen to remote file system.

qlzh727:
For b/77657597. Btw, I don't have a TPU instance to validate this change, I think the tf.gfile should do the trick with internal env.


robieta:
Is this sufficient? You're still relying on urllib to be able to write the archive to the file system.

qlzh727:
@robieta, the urllib is only used to read the input file object, which does not change within this PR. The tf.gfile.Open should support different file system name space (eg, gs:/ in this case), whereas the default python open() only support local file system. You can read internal doc about gfile API to get more details.

PR #3915, opened 2018-04-09
Adds max_to_keep as a possible train pipeline parameter

perdasilva:
I have the need to keep more than just 5 checkpoints during training. So, I've added tf.train.Saver's max_to_keep as a configurable parameter in the training pipeline file.


perdasilva:
Np, ty.

PR #3917, opened 2018-04-09
Adding cast to fp32 for TensorRT no image case

karmel:
Numpy defaults to creating fp64, which causes TF/tensorRT to reject the random data. This fixes that.

Just for fun-- what does random data look like to the classifier? This particular bit of randomness came out as: 
Precision:  native [u'nematode, nematode worm, roundworm', u'panpipe, pandean pipe, syrinx', u'revolver, six-gun, six-shooter', u'bolo tie, bolo, bola tie, bola', u'hatchet']


karmel:
Also, realized this was not in CODEOWNERS. Adding.

Issue #3885, opened 2018-04-05
[deeplab] UnicodeDecodeError


hhwxxx:
### System information
- **What is the top-level directory of the model you are using**: /deeplab
- **Have I written custom code (as opposed to using a stock example script provided in TensorFlow)**: no
- **OS Platform and Distribution (e.g., Linux Ubuntu 16.04)**: Linux Ubuntu 16.04
- **TensorFlow installed from (source or binary)**: binary
- **TensorFlow version (use command below)**: ('v1.6.0-0-gd2e24b6039', '1.6.0')
- **Bazel version (if compiling from source)**:
- **CUDA/cuDNN version**: 9.0/7.0
- **GPU model and memory**: 4 * 1080 Ti / 11GB
- **Exact command to reproduce**: bash local_test.sh

### Describe the problem
Hello. I am trying to run the script /deeplab/local_test.sh, but I got the error:
UnicodeDecodeError: 'ascii' codec can't decode byte 0xff in position 0: ordinal not in range(128)
when it begin to run build_voc2012_data.py. 
I have no idea how to solve this probelm. 
Can anyone give me some advice. Thanks for your help.

### Source code / logs
Converting PASCAL VOC 2012 dataset...
2018-04-05 23:16:40.948383: I tensorflow/core/platform/cpu_feature_guard.cc:140] Your CPU supports instructions that this TensorFlow binary was not compiled to use: AVX2 FMA
2018-04-05 23:16:41.469332: I tensorflow/core/common_runtime/gpu/gpu_device.cc:1212] Found device 0 with properties: 
name: GeForce GTX 1080 Ti major: 6 minor: 1 memoryClockRate(GHz): 1.582
pciBusID: 0000:02:00.0
totalMemory: 10.91GiB freeMemory: 10.71GiB
2018-04-05 23:16:41.849896: I tensorflow/core/common_runtime/gpu/gpu_device.cc:1212] Found device 1 with properties: 
name: GeForce GTX 1080 Ti major: 6 minor: 1 memoryClockRate(GHz): 1.582
pciBusID: 0000:03:00.0
totalMemory: 10.91GiB freeMemory: 10.75GiB
2018-04-05 23:16:42.222259: I tensorflow/core/common_runtime/gpu/gpu_device.cc:1212] Found device 2 with properties: 
name: GeForce GTX 1080 Ti major: 6 minor: 1 memoryClockRate(GHz): 1.582
pciBusID: 0000:82:00.0
[...]

PR #3924, opened 2018-04-10
Fix string to bytes conversion in deeplab data input.

qlzh727:
The current does not work in py2 since the input string could contain
unicode string, and default encoding in ASCII in py2. Change the
method to only do encode() which convert string to byte array when
running python3.


qlzh727:
Somehow https://github.com/tensorflow/models/pull/3910 was closed by accident. Recreating PR again.

qlzh727:
This PR is trying to fix #3885

PR #3942, opened 2018-04-10
official/mnist: Use tf.keras.Sequential

asimshankar:


official/mnist/mnist.py
```diff
@@ -30,68 +30,49 @@
...
-    y = self.conv2(y)
-    y = self.max_pool2d(y)
-    y = tf.layers.flatten(y)
-    y = self.fc1(y)
-    y = self.dropout(y, training=training)
-    return self.fc2(y)
+  Returns:
+    A tf.keras.Model.
+  """
+  input_shape = None
```
qlzh727: Probably can skip this here since it will be declared by the if-else below anyway.

official/mnist/mnist.py
```diff
@@ -30,68 +30,49 @@
...
+  if data_format == 'channels_first':
+    input_shape = [1, 28, 28]
+  else:
+    assert data_format == 'channels_last'
+    input_shape = [28, 28, 1]
+
+  L = tf.keras.layers
+  max_pool = L.MaxPooling2D((2, 2), (2, 2), padding='same', data_format=data_format)
+  return tf.keras.Sequential([
```
qlzh727: This is much more readable than the previous version. Thanks.

qlzh727:
Please also fix the lint error.

official/mnist/mnist.py
```diff
@@ -30,68 +30,59 @@
...
+      (2, 2), (2, 2), padding='same', data_format=data_format)
+  return tf.keras.Sequential(
+      [
+          l.Reshape(input_shape),
+          l.Conv2D(
+              32,
+              5,
+              padding='same',
+              data_format=data_format,
+              activation=tf.nn.relu), max_pool,
```
karmel: nit: max_pool gets lost at the end of the other function call. Can you move to its own line?

official/mnist/mnist.py
```diff
@@ -30,68 +30,59 @@
...
 
-  But written as a tf.keras.Model using the tf.layers API.
-  """
+  But uses the tf.keras API.
+
+  Args:
+    data_format: Either 'channels_first' or 'channels_last'. 'channels_first' is
+      typically faster on GPUs while 'channels_last' is typically faster on
+      CPUs. See
+      https://www.tensorflow.org/performance/performance_guide#data_formats
```
karmel: This is neat. Anticipating that people are going to ask how this is different than the Resnet/__call__ style, and why we use tf.keras.layers here, can you add comments explaining what Sequential does? And, for our reference-- should we be using tf.keras.layers at this point? Sequential in the cases where it makes sense?

official/mnist/mnist.py
```diff
@@ -30,68 +30,59 @@
...
-    y = self.dropout(y, training=training)
-    return self.fc2(y)
+  Returns:
+    A tf.keras.Model.
+  """
+  if data_format == 'channels_first':
+    input_shape = [1, 28, 28]
+  else:
+    assert data_format == 'channels_last'
+    input_shape = [28, 28, 1]
```
karmel: Might be worth commenting that data_format is different between tf.layers and tf.keras.layers. I would have missed that and been confused if trying this+other models.

official/mnist/mnist.py
asimshankar: `Sequential` is a sub-class of `Model` so this isn't really different, it's just simpler. If your network is really just a chain of layers, `Sequential` saves some boilerplate.

I'm a bit hesitant to add a comment here explaining that since that comment would either have to pre-suppose that model-subclassing is the most typical usage (which it isn't), or first explain that there is another more-verbose way to write this network and then explain how `Sequential` is simpler :). But perhaps I'm overthinking, so if you have a suggestion of what the comment should be - happy to add it.

Yes, in other samples we should use `Sequential` wherever it fits, and use `tf.keras.layers`.

karmel: "A subclass of tf.keras.Model, tf.keras.Sequential, is returned here, as the MNIST model is just a chain of layers." Or something to that effect.

official/mnist/mnist.py
karmel: Ah, I see. SG.

official/mnist/mnist.py

Issue #1916, opened 2017-07-11
object detection export inference graph error when RPN_only 


shzygmyx:
When I set first_stage_only = true in my config file, it trains well. However, when I try to use export_inference_graph.py to export the trained model for inference, there is a bug:
> Traceback (most recent call last):
>   File "export_inference_graph.py", line 101, in <module>
>     tf.app.run()
>   File "/home/linhb/anaconda2/envs/tensorflow/lib/python2.7/site-packages/tensorflow/python/platform/app.py", line 48, in run
>     _sys.exit(main(_sys.argv[:1] + flags_passthrough))
>   File "export_inference_graph.py", line 97, in main
>     FLAGS.export_as_saved_model)
>   File "/home/linhb/horse/models/object_detection/exporter.py", line 339, in export_inference_graph
>     export_as_saved_model)
>   File "/home/linhb/horse/models/object_detection/exporter.py", line 310, in _export_inference_graph
>     outputs = _add_output_tensor_nodes(postprocessed_tensors)
>   File "/home/linhb/horse/models/object_detection/exporter.py", line 184, in _add_output_tensor_nodes
>     classes = postprocessed_tensors.get('detection_classes') + label_id_offset
> TypeError: unsupported operand type(s) for +: 'NoneType' and 'int'

That's due to the output detection_classes is None when using first_stage_only. I change the code in `exporter.py`, line184, from,
`classes = postprocessed_tensors.get('detection_classes')+label_id_offset`
to,
```
classes = postprocessed_tensors.get('detection_classes') 
if classes: 
    classes += label_id_offset 
else:
    classes = tf.constant(1, tf.int32, scores.shape)
```
then the program can run correctly.

PierreBoyeau:
Hey guys, 

Any news on this subject? Personnaly, I also face issues when I try to run `eval.py`.
I get the following error 
```
Traceback (most recent call last):
  File "object_detection/eval.py", line 161, in <module>
    tf.app.run()
  File "/usr/local/lib/python2.7/dist-packages/tensorflow/python/platform/app.py", line 48, in run
    _sys.exit(main(_sys.argv[:1] + flags_passthrough))
  File "object_detection/eval.py", line 157, in main
    FLAGS.checkpoint_dir, FLAGS.eval_dir)
  File "/home/ubuntu/models-tf/object_detection/evaluator.py", line 132, in evaluate
    ignore_groundtruth=eval_config.ignore_groundtruth)
[...]

PR #3947, opened 2018-04-11
Fix bug exporting FasterRCNN with only one stage

dsuess:
Fixes issue #1916


dseuss:
I signed it!

danyfang:
I am surprised why this has not been reviewed until now because there is clearly a bug and the fix worked good.

danyfang:
Ran into problems when trying to export the whole model, the fix is still buggy. It should be:
`if classes is not None:
    classes += label_id_offset
  else:
    classes = tf.constant(1, tf.int32, scores.shape)`

PR #3964, opened 2018-04-12
make MNIST shape more explicit

robieta:
The recent change to make MNIST use Sequential is not compatible with TensorFlow 1.7. This is an issue because 1.7 is the latest publicly available cloud TPU version. The fix is a very minor additional shape specification.

### DIFF
--- a/official/mnist/mnist.py
+++ b/official/mnist/mnist.py
@@ -49,11 +49,12 @@ def create_model(data_format):
   Returns:
     A tf.keras.Model.
   """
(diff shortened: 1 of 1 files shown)


asimshankar:
My understanding was that the versioned branches are supposed to be compatible with the versioned TensorFlows releases (https://github.com/tensorflow/models/blob/master/official/README.md).

Are we required to keep the master branch of the models repository compatible with the latest release?

Issue #3959, opened 2018-04-12
wide and deep example error: KeyError: "The name 'head/truediv:0' refers to a Tensor which does not exist. The operation, 'head/truediv', does not exist in the graph."


CanoeFZH:
with same error from:
https://stackoverflow.com/questions/49579096/tensorflow-wide-and-deep-example-error
I make some code removing and it works:

```
--- a/official/wide_deep/wide_deep.py
+++ b/official/wide_deep/wide_deep.py
@@ -195,14 +195,9 @@ def main(argv):
     return input_fn(test_file, 1, False, flags.batch_size)
 
   loss_prefix = LOSS_PREFIX.get(flags.model_type, '')
-  train_hooks = hooks_helper.get_train_hooks(
-      flags.hooks, batch_size=flags.batch_size,
-      tensors_to_log={'average_loss': loss_prefix + 'head/truediv',
-                      'loss': loss_prefix + 'head/weighted_loss/Sum'})
-
   # Train and evaluate the model every `flags.epochs_between_evals` epochs.
   for n in range(flags.train_epochs // flags.epochs_between_evals):
-    model.train(input_fn=train_input_fn, hooks=train_hooks)
+    model.train(input_fn=train_input_fn)
     results = model.evaluate(input_fn=eval_input_fn)
 
     # Display evaluation metrics
```

yhliang2018:
@k-w-w Could you help on this issue?

k-w-w:
Thanks for submitting this issue. It appears that the graph produced by canned estimators changes depending on the version of tensorflow. In that case, we should remove the logging tensor hook from the wide deep model to lower maintenance cost.

Submitting a PR shortly.

karmel:
@ispirmustafa - The linked PR seems like a non-ideal solution. What's the recommended way of logging tensors of interest from CannedEstimators?

ispirmustafa:
best way is using `tf.train.LoggingTensorHook`. You need to provide the name of the tensor.

k-w-w:
@CanoeFZH The master version of the repo works with the tensorflow nightly build. If you are running an older version of tensorflow, I would recommend checking out the [r1.5 branch](https://github.com/tensorflow/models/tree/r1.5/official) or one of the releases: [https://github.com/tensorflow/models/releases](https://github.com/tensorflow/models/releases)

xmutizabal:
I have the same problem in wide and deep tutorial

in ops.py
line 3365, in _as_graph_element_locked
[...]

PR #3965, opened 2018-04-13
remove logging hook from wide_deep

k-w-w:
Quick bug fix for #3959. It's probably best not to have graph-dependent logging in the model.

### DIFF
--- a/official/wide_deep/wide_deep.py
+++ b/official/wide_deep/wide_deep.py
@@ -44,9 +44,6 @@ _NUM_EXAMPLES = {
 }
 
 
-LOSS_PREFIX = {'wide': 'linear/', 'deep': 'dnn/'}
-
-
 def build_model_columns():
   """Builds a set of wide and deep feature columns."""
   # Continuous columns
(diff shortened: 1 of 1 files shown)


robieta:
My 2¢. I'd like to re-float the idea of version checks. I don't think it is practical to try to support a variety of releases; however there is no real way for someone who hits something like https://github.com/tensorflow/models/issues/3959 to deduce that it's caused by a stale TensorFlow version. At least if we hard-fail for version we can limit our exposure to backward incompatible changes in TensorFlow proper.

karmel:
Just spoke to @ispirmustafa , and he said changes like this should be infrequent. Let's count this as strike 1, and just update to match the current names. If we run into this problem again, we can reconsider.

k-w-w:
Going with @karmel's suggestion. The errors come from running Tensorflow 1.5 (nightly runs fine), so for now we can point people running 1.5 to the r1.5 branch. If there appears to be many users running tensorflow 1.5, then version checks would work.

PR #3971, opened 2018-04-13
Add interface to pass the number of threads for inter/intra-op parall…

opencici2006:
…elism

Signed-off-by: shaohua <shaohua.zhang@intel.com>


opencici2006:
I signed it!

PR #3973, opened 2018-04-13
Object detection internal changes

pkulzc:


research/object_detection/meta_architectures/faster_rcnn_meta_arch_test.py
```diff
@@ -15,14 +15,16 @@
 
 """Tests for object_detection.meta_architectures.faster_rcnn_meta_arch."""
 
+from absl.testing import parameterized
```
derekjchow: This looks like a new dependency. Should we update the installation docs?

research/object_detection/model_main.py
```diff
@@ -0,0 +1,76 @@
+# Copyright 2017 The TensorFlow Authors. All Rights Reserved.
```
derekjchow: Copyright 2018

research/object_detection/meta_architectures/faster_rcnn_meta_arch_test.py
pkulzc: Actually I'm able to run this test externally without installing anything else, so I guess it should be fine.

research/object_detection/model_main.py
pkulzc: Done, nice catch :)

derekjchow:
LGTM

PR #3986, opened 2018-04-15
Update distributions.py

gautam1858:
posterior post_zs should start from 0

Example
```
>>> a = [1,5,23,45]
>>> a[1:]
[5, 23, 45]
>>> a[0:]
[1, 5, 23, 45]
```


research/lfads/distributions.py
```diff
@@ -480,7 +480,7 @@ def __init__(self, post_zs, prior_z_process):
     logq_bxu = post_zs[0].logp(z0_bxu)
     logp_bxu = prior_z_process.logp_t(z0_bxu)
```
DavidWiesner: @gautam1858 the loglikelihood of the first element is already calculated and added to the sum in this line and in the line above

PR #4005, opened 2018-04-17
Adjust number of warmup loops

karmel:
Five seems like a nice round number? Is two really sufficient?


karmel:
CC @samikama as well

aaroey:
Thanks @karmel for the change, I think it's sufficient. Based on my understanding, only the first run will be slower as it needs to set up the running environment (e.g. executors).

PR #4033, opened 2018-04-19
Revert "Resnet distribution strategies (#3887)"

robieta:
This reverts commit 32aa656373dd483f9a6a8aa45c51b595699e06e7.

There is a bad interaction between logging and distribution strategies. Rolling back distribution strategies to keep master in  a reasonable state.


robieta:
It is indeed a strict rollback, and I have confirmed that resnet does start.

PR #4034, opened 2018-04-19
Benchmark update

qlzh727:
1. Create global instance of benchmark logger, which default log to
tf.logging.info
2. Allow user to config the logging location.
3. Fix nits in code and comment.

### DIFF
--- a/official/resnet/resnet_run_loop.py
+++ b/official/resnet/resnet_run_loop.py
@@ -381,11 +381,9 @@ def resnet_main(flags, model_function, input_function, shape=None):
           'dtype': flags.dtype
       })
 
-  if flags.benchmark_log_dir is not None:
-    benchmark_logger = logger.BenchmarkLogger(flags.benchmark_log_dir)
-    benchmark_logger.log_run_info('resnet')
-  else:
-    benchmark_logger = None
+  logger.config_benchmark_logger(flags.benchmark_log_dir)
+  benchmark_logger = logger.get_benchmark_logger()
+  benchmark_logger.log_run_info('resnet')
 
   for _ in range(flags.train_epochs // flags.epochs_between_evals):
     train_hooks = hooks_helper.get_train_hooks(
@@ -426,8 +424,7 @@ def resnet_main(flags, model_function, input_function, shape=None):
                                        steps=flags.max_train_steps)
     print(eval_results)
 
-    if benchmark_logger:
-      benchmark_logger.log_estimator_evaluation_result(eval_results)
+    benchmark_logger.log_evaluation_result(eval_results)
 
     if model_helpers.past_stop_threshold(
         flags.stop_threshold, eval_results['accuracy']):
--- a/official/resnet/resnet_run_loop.py
+++ b/official/resnet/resnet_run_loop.py
@@ -421,7 +421,6 @@ def resnet_main(flags, model_function, input_function, shape=None):
     # global_step count.
     eval_results = classifier.evaluate(input_fn=input_fn_eval,
                                        steps=flags.max_train_steps)
-    print(eval_results)
 
     benchmark_logger.log_evaluation_result(eval_results)
 
--- a/official/resnet/resnet_run_loop.py
+++ b/official/resnet/resnet_run_loop.py
@@ -381,8 +381,7 @@ def resnet_main(flags, model_function, input_function, shape=None):
           'dtype': flags.dtype
       })
 
-  logger.config_benchmark_logger(flags.benchmark_log_dir)
-  benchmark_logger = logger.get_benchmark_logger()
+  benchmark_logger = logger.config_benchmark_logger(flags.benchmark_log_dir)
   benchmark_logger.log_run_info('resnet')
 
   for _ in range(flags.train_epochs // flags.epochs_between_evals):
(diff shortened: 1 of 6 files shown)


official/resnet/resnet_run_loop.py
```diff
@@ -381,11 +381,9 @@ def resnet_main(flags, model_function, input_function, shape=None):
           'dtype': flags.dtype
       })
 
-  if flags.benchmark_log_dir is not None:
-    benchmark_logger = logger.BenchmarkLogger(flags.benchmark_log_dir)
-    benchmark_logger.log_run_info('resnet')
-  else:
-    benchmark_logger = None
+  logger.config_benchmark_logger(flags.benchmark_log_dir)
+  benchmark_logger = logger.get_benchmark_logger()
```
karmel: Why not have this rolled into one line? Config and return the logger in one call, with separate interfaces as necessary.

official/utils/logs/logger.py
```diff
@@ -36,33 +37,96 @@
...
-    if not tf.gfile.IsDirectory(self._logging_dir):
-      tf.gfile.MakeDirs(self._logging_dir)
 
-  def log_estimator_evaluation_result(self, eval_results):
-    """Log the evaluation result for a estimator.
+def config_benchmark_logger(logging_dir):
+  """Config the global benchmark logger"""
+  _logger_lock.acquire()
+  try:
+    global _benchmark_logger
```
karmel: Why do we have to declare this as global?

official/utils/logs/logger.py
```diff
@@ -36,33 +37,96 @@
...
+      extras: map of string:string, the extra information about the metric.
+    """
+    if not isinstance(value, numbers.Number):
+      tf.logging.warning(
+          "Metric value to log should be a number. Got %s", type(value))
+      return
+    if extras:
+      extras = [{"name": k, "value": v} for k, v in sorted(extras.items())]
+    else:
+      extras = []
```
karmel: Rather than have this duplicated, pull out as a function that both can call.

official/utils/logs/logger.py
```diff
@@ -36,33 +37,96 @@
...
+
+    tf.logging.info("Benchmark metric: "
+                    "Name %s, value %d, unit %s, global_step %d, extras %s",
+                    name, value, unit, global_step, extras)
```
karmel: This is going to end up seeming strangely redundant with the LoggingTensorHook, right? What does it actually look like from the user perspective?

karmel: If it's redundant, maybe it's better to make it log to debug?

official/utils/logs/logger.py
```diff
@@ -36,33 +37,96 @@
...
+    config_benchmark_logger(None)
+
+  return _benchmark_logger
+
+
+class BaseBenchmarkLogger(object):
+  """Class to log the benchmark information to STDOUT."""
+
+  def __init__(self):
+    pass
```
karmel: Just leave this undefined. Python will fill in the init for you.

official/utils/logs/logger_test.py
```diff
@@ -31,16 +31,58 @@
...
+
+  def test_config_base_benchmark_logger(self):
+    logger.config_benchmark_logger("")
+    self.assertIsInstance(logger.get_benchmark_logger(),
+                          logger.BaseBenchmarkLogger)
+
+  def test_config_benchmark_file_logger(self):
+    logger.config_benchmark_logger("/tmp/abc")
+    self.assertIsInstance(logger.get_benchmark_logger(),
+                          logger.BenchmarkFileLogger)
```
karmel: Can we test that the lock is acquired as well? Don't know how feasible that is in a unit test, but would be nice to make sure we can't end up with two loggers.

official/resnet/resnet_run_loop.py
qlzh727: Done.

official/utils/logs/logger.py
qlzh727: You have to put the global here to force python realize that you are trying to modify the instance above, rather than declare a local variable.

official/utils/logs/logger.py
qlzh727: Done.

official/utils/logs/logger.py
qlzh727: The log will be similar. I think it is not the problem for the logger, but the loggingMetricHook with default config. Do u want me to disable the loggingMetricHook when the log_dir is not configured?

official/utils/logs/logger.py
qlzh727: Done

official/utils/logs/logger_test.py
qlzh727: locking is hard to test, which usually involve some fuzzy test, like create instance thousands of times, and hope can catch a error, which usually hard to do in a unit test.

official/utils/logs/logger.py
karmel: Can you at least comment that here? Along with any side effects we should be aware of. (ie, what happens if I have a file with a local var _benchmark_logger?)

qlzh727: The global scope is within this file/module, so this should be safe enough. If you want to access this field from other file, you usually just do logger. _benchmark_logger

PR #4040, opened 2018-04-20
Update the timestamp to default to UTC.

qlzh727:
Previously the timestamp we pushed to bigquery was PDT, but the
timezone spec was set to 'Z' (UTC). This is provide incorrect
value to bigquery, and might affect analysis down the road.

### DIFF
--- a/official/utils/logs/logger.py
+++ b/official/utils/logs/logger.py
@@ -149,7 +149,7 @@ class BenchmarkFileLogger(BaseBenchmarkLogger):
           "value": float(value),
           "unit": unit,
           "global_step": global_step,
-          "timestamp": datetime.datetime.now().strftime(
+          "timestamp": datetime.datetime.utcnow().strftime(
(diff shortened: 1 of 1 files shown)


official/utils/logs/logger.py
```diff
@@ -184,7 +184,8 @@ def _gather_run_info(model_name):
   run_info = {
       "model_name": model_name,
       "machine_config": {},
-      "run_date": datetime.datetime.now().strftime(_DATE_TIME_FORMAT_PATTERN)}
+      "run_date": datetime.datetime.utcnow().strftime(
+          _DATE_TIME_FORMAT_PATTERN)}
```
karmel: Might be useful to add a get_timestamp() function to avoid having these diverge over time? Two statements of this I can handle, though, so feel free to decide.

qlzh727: I think for now I will leave it as is since even it will just be a one line wrap function. I will extract it when it becomes more complicated.

PR #4049, opened 2018-04-22
Add run individual step only option

yhliang2018:
Add more flags to allow run each individual step only. Arguments/flags are not finalized. Feel free to add comments/suggestions, thanks.
Will update readme with more detailed commands.

### DIFF
--- a/research/minigo/README.md
+++ b/research/minigo/README.md
@@ -1,19 +1,24 @@
 # MiniGo
 This is a simplified implementation of MiniGo based on the code provided by the authors: [MiniGo](https://github.com/tensorflow/minigo).
 
-MiniGo is a minimalist Go engine modeled after AlphaGo Zero, built on MuGo. The current implementation consists of three main modules: the DualNet model, the Monte Carlo Tree Search (MCTS), and Go domain knowledge. Currently the **model** part is our focus.
+MiniGo is a minimalist Go engine modeled after AlphaGo Zero, ["Mastering the Game of Go without Human
+Knowledge"](https://www.nature.com/articles/nature24270). An useful one-diagram overview of Alphago Zero can be found in the [cheat sheet](https://medium.com/applied-data-science/alphago-zero-explained-in-one-diagram-365f5abf67e0).
 
-This implementation maintains the features of model training and validation, and also provides evaluation of two Go models.
+The implementation of MiniGo consists of three main components: the DualNet model, the Monte Carlo Tree Search (MCTS), and Go domain knowledge. Currently, the **DualNet model** is our focus.
 
 
-## DualNet Model
+## DualNet Architecture
+DualNet is the neural network used in MiniGo. It's based on residual blocks with two heads output. Following is a brief overview of the DualNet architecture.
+
+### Input Features
 The input to the neural network is a [board_size * board_size * 17] image stack
 comprising 17 binary feature planes. 8 feature planes consist of binary values
 indicating the presence of the current player's stones; A further 8 feature
 planes represent the corresponding features for the opponent's stones; The final
 feature plane represents the color to play, and has a constant value of either 1
-if black is to play or 0 if white to play. Check `features.py` for more details.
+if black is to play or 0 if white to play. Check [features.py](features.py) for more details.
 
+### Neural Network Structure
 In MiniGo implementation, the input features are processed by a residual tower
 that consists of a single convolutional block followed by either 9 or 19
 residual blocks.
@@ -31,8 +36,9 @@ Each residual block applies the following modules sequentially to its input:
   6. A skip connection that adds the input to the block
   7. A rectifier non-linearity
 
-Note: num_filter is 128 for 19 x 19 board size, and 32 for 9 x 9 board size.
+Note: num_filter is 128 for 19 x 19 board size, and 32 for 9 x 9 board size in MiniGo implementation.
 
+### Dual Heads Output
 The output of the residual tower is passed into two separate "heads" for
 computing the policy and value respectively. The policy head applies the
 following modules:
@@ -51,7 +57,7 @@ The value head applies the following modules:
   6. A fully connected linear layer to a scalar
   7. A tanh non-linearity outputting a scalar in the range [-1, 1]
 
-The overall network depth, in the 10 or 20 block network, is 19 or 39
+In MiniGo, the overall network depth, in the 10 or 20 block network, is 19 or 39
 parameterized layers respectively for the residual tower, plus an additional 2
 layers for the policy head and 3 layers for the value head.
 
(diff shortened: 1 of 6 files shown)


research/minigo/README.md
```diff
@@ -88,27 +87,26 @@ One iteration of reinforcement learning consists of the following steps:
  The `--evaluation` argument is to invoke the evaluation between the latest model and the current best model.
 
 ## Testing Pipeline
-As the whole RL pipeline may takes hours to train even for a 9x9 board size, we provide a dummy model with a `--debug` mode for testing purpose.
+As the whole RL pipeline may takes hours to train even for a 9x9 board size, we provide a dummy model with a `--dummy` mode for testing purpose.
```
qlzh727: I think "dummy" provide less context or carry less meaningful information for user. Its hard for user to guess the meaning of the flag without look into the doc. If the flag is for testing/validation purpose, then "--test" or "--validate" is more meaningful.

research/minigo/minigo.py
```diff
@@ -79,41 +79,17 @@ def bootstrap(estimator_model_dir, trained_models_dir, params):
   dualnet.export_model(estimator_model_dir, bootstrap_model_path)
 
 
-def selfplay(model_name, trained_models_dir, selfplay_dir, holdout_dir, sgf_dir,
-             params):
+def selfplay(selfplay_dirs, selfplay_model, params):
```
qlzh727: Should selfplay be one word or two? @karmel?

research/minigo/minigo.py
```diff
@@ -79,41 +79,17 @@ def bootstrap(estimator_model_dir, trained_models_dir, params):
...
 
   Args:
-    model_name: The name of the model used for selfplay.
-    trained_models_dir: The path to the model files.
-    selfplay_dir: Where to write the games. Set as 'base_dir/data/selfplay/'.
-    holdout_dir: Where to write the holdout data. Set as
-      'base_dir/data/holdout/'.
-    sgf_dir: Where to write the sgf (Smart Game Format) files. Set as
-      'base_dir/sgf/'.
+    selfplay_dirs: A dict to specify the directories used in selfplay.
```
qlzh727: Can u give some example of the keys you expecting here? Sometimes explicit parameters are more appropriate, and provide more strong contract between caller and callee.

research/minigo/minigo.py
```diff
@@ -79,41 +79,17 @@ def bootstrap(estimator_model_dir, trained_models_dir, params):
...
+    selfplay_model: The actual Dualnet runner for selfplay.
     params: An object of hyperparameters for the model.
```
qlzh727: Using object as type is not ideal for a data entity. If there are specific fields, then we should create a data entity class for it.

research/minigo/README.md
karmel: nits: may take hours; mode for testing purposes.

research/minigo/minigo.py
karmel: Please list out the set of expected directories.

research/minigo/minigo.py
```diff
@@ -259,20 +244,21 @@ def validate(trained_models_dir, holdout_dir, estimator_model_dir, params):
...
-def evaluate(trained_models_dir, black_model_name, white_model_name,
+def evaluate(black_model_name, black_net, white_model_name, white_net,
              evaluate_dir, params):
   """Evaluate with two models.
 
-  With the model name, construct two DualNetRunners to play as black and white
-  in a Go match. Two models play several names, and the model that wins by a
-  margin of 55% will be the winner.
+  With two DualNetRunners to play as black and white in a Go match. Two models
+  play several names, and the model that wins by a margin of 55% will be the
```
karmel: nit: several games

research/minigo/minigo.py
```diff
@@ -259,20 +244,21 @@ def validate(trained_models_dir, holdout_dir, estimator_model_dir, params):
...
+  winner.
 
   Args:
-    trained_models_dir: Directories where the completed generations/models are.
     black_model_name: The name of the model playing black.
+    black_net: The actual DualNetRunner model for black
```
karmel: nit: "actual" here makes sense when you know it used to be a name, but doesn't make sense if you read this for the first time. That word can be removed.

research/minigo/minigo.py
```diff
@@ -305,38 +278,189 @@ def evaluate(trained_models_dir, black_model_name, white_model_name,
   return black_model_name if winner == go.BLACK_NAME else white_model_name
 
 
-def _set_params_from_board_size(board_size):
+def _set_params():
   """Set hyperparameters from board size."""
   params = model_params.MiniGoParams()
-  k = utils.round_power_of_two(board_size ** 2 / 3)
+  k = utils.round_power_of_two(FLAGS.board_size ** 2 / 3)
```
karmel: Taking board size and other required params is preferable to relying on global flags here.

research/minigo/minigo.py
```diff
@@ -305,38 +278,189 @@ def evaluate(trained_models_dir, black_model_name, white_model_name,
...
+  Args:
+    params: An object of hyperparameters for the model.
+    dirs: An object of directories used in each step.
+
+  Raises:
+    ValueError: when neither of the black model and the white model is provided
+    for evaluation.
+  """
+  # bootstrap only
+  if FLAGS.bootstrap_only:
```
karmel: Why give the option to run one only, and not just the ability to run multiple selected parts? Ie, have skip_bootstrap or skip_selfplay etc. so that if I want to bootstrap and selfplay, I can. That is no extra work or code, but gives the user more flexibility in running modes.

research/minigo/minigo.py
```diff
@@ -196,31 +174,33 @@ def gather(selfplay_dir, training_chunk_dir, params):
...
-def train(trained_models_dir, estimator_model_dir, training_chunk_dir, params):
+def train(trained_models_dir, estimator_model_dir, training_chunk_dir,
+          generation_num, params):
   """Train the latest model from gathered data.
 
   Args:
     trained_models_dir: Where to export the completed generation.
     estimator_model_dir: tf.estimator model directory.
     training_chunk_dir: Directory where gathered training chunks are.
+    generation_num: Which generation you are training.
```
qlzh727: I feel "generation" is enough for indicating it is a number and integer.

research/minigo/minigo.py
karmel: Looking further down below, we are introducing a ton of flags with interactions and dependencies that are not clear, and this seems like a bad idea. Not to mention that there is now essentially a duplicated main function. Why aren't there just booleans like flags.evaluation and flags.validation below? We should simply skip a step if it is not requested, not change the path of execution. As to the extra flags like model_names, do we actually need that? Or is the directory structure sufficient? If we don't have specific use-cases, we should leave out, and if we do, we should consider requiring a config file, as this is a ton of flags.

research/minigo/minigo.py
```diff
@@ -251,6 +231,11 @@ def validate(trained_models_dir, holdout_dir, estimator_model_dir, params):
         tf_records.extend(
             tf.gfile.Glob(os.path.join(record_dir, '*'+_TF_RECORD_SUFFIX)))
 
+  if len(tf_records) is 0:
```
qlzh727: I think the quickest way is "if not tf_records".

research/minigo/minigo.py
```diff
@@ -251,6 +231,11 @@ def validate(trained_models_dir, holdout_dir, estimator_model_dir, params):
...
+    print('No holdout dataset for validation! ' +
```
qlzh727: I think we should change all those print statement to tf.logging in future.

research/minigo/minigo.py
qlzh727: I think this function is long enough and with several layer of if-else. I think we should extract the body of each top level if and put them into function calls. Eg.

if FLAGS.bootstrap_only:
  _handle_bootstrap()
  return
if FLAGS.selfplay_only:
  _handle_selfplay()
....

research/minigo/minigo.py
```diff
@@ -305,38 +278,189 @@ def evaluate(trained_models_dir, black_model_name, white_model_name,
...
+    return
+
+  # self-play only, for parallelization
+  if FLAGS.selfplay_only:
+    if FLAGS.selfplay_model_name is not None:
+      selfplay_model_name = FLAGS.selfplay_model_name
+    else:
+      #  use the latest model for self-play
+      _, best_model_so_far = utils.get_latest_model(dirs.trained_models_dir)
+      selfplay_model_name = best_model_so_far
```
qlzh727: I think you can just do the following since you don't care best_model_so_far
_, selfplay_model_name = utils.get_latest_model(dirs.trained_models_dir)

research/minigo/minigo.py
yhliang2018: params is an instance of class MiniGoParams, like params = model_params.MiniGoParams(). Not quite sure about "a data entity class" you mentioned. Could you explain more? Thanks!

research/minigo/minigo.py
yhliang2018: Sure. Will update it.

research/minigo/minigo.py
yhliang2018: Will do.

research/minigo/minigo.py
yhliang2018: Yeah, thanks!

research/minigo/minigo.py
yhliang2018: Thanks for the these comments, Karmel! This is exactly something that needs your inputs. 
1. I think skipping some specific step is okay. But according to the discussions with Victor last week, my understanding is that he cares more about parallelization of selfplay and train steps with multiple GPU, as those two steps take long. As for other steps, like bootstrap, we need it only once; gather is executed once after selfplay; validation is supposed to be called once to validate the model-just-trained; Not sure about evaluation, it may be called multiple times for model comparison. I think how to set flags mainly depends on how the users use it. @bitfort As you are the main user now, could you give some suggestions on it?
2. For model_names with selfplay, if users always want the latest model to generate selfplay data, the directory structure is totally enough; Otherwise, user has to specify a model for selfplay. Like you said, how we design these flags is up to the use-cases. When working on this part, I also feel like we have too many flags which may make users confused. Any advice from user's perspective @bitfort ?

yhliang2018: Sure. Will do the refactoring.

research/minigo/minigo.py
qlzh727: Then the type of params is not just a object, but MiniGoParams, the comment should be updated accordingly.

research/minigo/README.md
yhliang2018: change it to --test. Thanks!

research/minigo/gtp_wrapper.py
```diff
@@ -49,8 +49,8 @@ def __init__(self, board_size):
   def set_size(self, n):
     if n != self.board_size:
       raise ValueError((
-          '''Can't handle boardsize {n}!Restart with env var BOARD_SIZE={n}'''
-          ).format(n=n))
+          '''Can't handle boardsize {}! Please check the board size.'''
```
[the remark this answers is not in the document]
yhliang2018: Hahh, I guess it tries to avoid the handle of the single quote in Can't. Should we use double quotes or the escape character "\" in this case?

research/minigo/minigo.py
```diff
@@ -79,41 +79,23 @@ def bootstrap(estimator_model_dir, trained_models_dir, params):
...
   Args:
-    model_name: The name of the model used for selfplay.
-    trained_models_dir: The path to the model files.
-    selfplay_dir: Where to write the games. Set as 'base_dir/data/selfplay/'.
-    holdout_dir: Where to write the holdout data. Set as
-      'base_dir/data/holdout/'.
-    sgf_dir: Where to write the sgf (Smart Game Format) files. Set as
-      'base_dir/sgf/'.
-    params: An object of hyperparameters for the model.
+    selfplay_dirs: A dict to specify the directories used in selfplay.
```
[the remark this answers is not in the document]
yhliang2018: The original implementation is to specify each one individually. Here as selfplay may be called individually, or in the RL pipeline, we use this dict to avoid duplicate codes.

research/minigo/gtp_wrapper.py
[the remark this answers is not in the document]
yhliang2018: Use triple quotes here based on offline discussion.

research/minigo/minigo.py
```diff
@@ -124,19 +106,21 @@ def _write_sgf_data(dir_sgf, use_comments):
...
-  _write_sgf_data(clean_sgf, use_comments=False)
-  _write_sgf_data(full_sgf, use_comments=True)
+  _write_sgf_data(selfplay_dirs['clean_sgf'], use_comments=False)
+  _write_sgf_data(selfplay_dirs['full_sgf'], use_comments=True)
 
   game_data = player.extract_data()
   tf_examples = preprocessing.make_dataset_from_selfplay(game_data, params)
 
   # Hold out 5% of games for evaluation.
   if random.random() < params.holdout_pct:
```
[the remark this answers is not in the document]
yhliang2018: This is a good solution.

yhliang2018:
Thanks for all the comments! New commit has been submitted to the PR to
address the comments.

Best,
Yanhui

research/minigo/minigo.py
```diff
@@ -305,38 +282,128 @@ def evaluate(trained_models_dir, black_model_name, white_model_name,
...
+    base_dir = FLAGS.base_dir + str(FLAGS.board_size) + '_size/'
 
-  # Set directories for models and datasets
-  base_dir = FLAGS.base_dir + str(FLAGS.board_size) + '_board_size/'
   dirs = utils.MiniGoDirectory(base_dir)
+
+  # Run selfplay only if user specifies the argument.
+  if FLAGS.selfplay_only:
+    if FLAGS.selfplay_model_name is not None:
+      selfplay_model_name = FLAGS.selfplay_model_name
```
karmel: You can make this much shorter:

```
if FLAGS.selfplay_only:
  selfplay_model_name = FLAGS.selfplay_model_name or utils.get_latest_model(dirs.trained_models_dir)[1]
  max_games = FLAGS.selfplay_max_games or params.max_games_per_generation
  run_selfplay(selfplay_model_name, max_games, dirs, params)
  return

```

research/minigo/gtp_wrapper.py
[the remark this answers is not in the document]
qlzh727: I think you can just put can\'t to escape, and get rid of triple quotes.

research/minigo/README.md
```diff
@@ -51,64 +57,82 @@ The value head applies the following modules:
...
+
+To test the RL pipeline with a dummy model, issue the following command:
+```
+ python minigo.py --base_dir=$HOME/minigo/ --board_size=9 --batch_size=256 --test
+```
+
+## Running Self-play Only
+Self-play only option is provided to run selfplay step individually to generate training data in parallel. Issue the following command to run selfplay only with the latest trained model:
+```
+ python minigo.py --selfplay_only
```
qlzh727: usually --selfplay is clear enough as a flag name. Word "only" is bit redundant.

research/minigo/minigo.py
```diff
@@ -124,19 +106,21 @@ def _write_sgf_data(dir_sgf, use_comments):
...
 
   game_data = player.extract_data()
   tf_examples = preprocessing.make_dataset_from_selfplay(game_data, params)
 
   # Hold out 5% of games for evaluation.
   if random.random() < params.holdout_pct:
     fname = os.path.join(
-        holdout_dir, ('{}'+_TF_RECORD_SUFFIX).format(output_name))
+        selfplay_dirs['holdout_dir'], ('{}' + _TF_RECORD_SUFFIX).format(
+            output_name))
```
qlzh727: Seems to be a simple string concat, let's get rid of format().

research/minigo/README.md
yhliang2018: Sure. Will update it.

research/minigo/gtp_wrapper.py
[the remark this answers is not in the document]
yhliang2018: Hah, looks like people have different opinions. I discussed with Taylor, and he likes triple quotes better than escape. Do we have any code style/convention we can refer to? @robieta @qlzh727

research/minigo/minigo.py
yhliang2018: Makes sense.

yhliang2018:
Thank you all for the comments!

PR #4074, opened 2018-04-24
Adds boosted_trees to the official models

yk5:
Hello @karmel and @yhliang2018 

We are adding a boosted_trees model for the higgs dataset.
Please review.

Thanks!

FYI @ispirmustafa


yk5:
@karmel Do I need to write a test as well? Then should it data_download also, or just train model file?

official/boosted_trees/train_higgs.py
```diff
@@ -0,0 +1,184 @@
...
+  print('## data loaded; train: {}{}, eval: {}{}'.format(
+      train_data.dtype, train_data.shape, eval_data.dtype, eval_data.shape))
+  # data consists of one label column and 28 feature columns following.
+  train_input_fn, feature_columns = make_inputs_from_np_arrays(
+      features_np=train_data[:, 1:], label_np=train_data[:, 0:1])
+  eval_input_fn, _ = make_inputs_from_np_arrays(
+      features_np=eval_data[:, 1:], label_np=eval_data[:, 0:1])
+  print('## features prepared. training starts..')
+
+  # TODO(youngheek): consider remove the normalization factor part from l2.
```
yk5: @ispirmustafa Mustafa, I'm weighing between keeping TODO and removing l2 from the example (so that we could make some changes on the scale later).

WDYT?

official/boosted_trees/README.md
```diff
@@ -0,0 +1,56 @@
...
+
+We use Gradient Boosted Trees algorithm to distinguish the two classes.
+
+---
+
+The code sample in this directory uses the high level `tf.estimator.Estimator` API. This API is great for fast iteration and quickly adapting models to your own datasets without major code overhauls. It allows you to move from single-worker training to distributed training, and it makes it easy to export model binaries for prediction. Here, for further simplicity and faster execution, we use a utility function `tf.contrib.estimator.boosted_trees_classifier_train_in_memory`. This utility function is especially effective when the input is provided as in-memory data sets like numpy arrays.
+
+An input function for the `Estimator` typically uses `tf.data.Dataset` API, which can handle various data control like streaming, batching, transform and shuffling. However `boosted_trees_classifier_train_in_memory()` utility function requires that the entire data is provided as a single batch (i.e. without using `batch()` API). Thus in this practice, simply `Dataset.from_tensors()` is used to convert numpy arrays into structured tensors, and `Dataset.zip()` is used to put features and label together.
```
robieta: You mention that the cleaned dataset is about 1.3 GB; is that also approximately the amount of memory needed to do in-memory training?

robieta: I think some of this might be too detailed for a README, and may make more sense as code comments.

official/boosted_trees/README.md
```diff
@@ -0,0 +1,56 @@
...
+
+```
+python train_higgs.py
+```
+
+The model is saved to `/tmp/higgs_model` by default, which can be changed using the `--model_dir` flag. To train a new model after one, provide a new `--model_dir` value, or remove the dir set by `--model_dir` before launching.
+
+Model parameters can be adjusted by flags, like `--n_trees`, `--max_depth`, `--learning_rate`.
+
+The final accuacy will be around 74% and loss will be around 0.5165 over the eval set, when trained with the default parameters.
```
robieta: nit: do we have that many significant digits on the loss?

official/boosted_trees/README.md
```diff
@@ -0,0 +1,56 @@
...
+
+By default, the first 1 million examples among 11 millions are used for training, and the last 1 million examples are used for evaluation.
```
robieta: This is a curious convention. Why do we ignore 9MM examples by default?

official/boosted_trees/README.md
```diff
@@ -0,0 +1,56 @@
...
+
+Run TensorBoard to inspect the details about the graph and training progression.
+
+```
+tensorboard --logdir=/tmp/higgs_model
+```
+
+## Additional Links
+
+If you are interested in distributed training, take a look at [Distributed TensorFlow](https://www.tensorflow.org/deploy/distributed).
```
robieta: For these sorts of non-local platforms, I think we should either provide explicit instructions or not mention them at all. (I'm inclined towards the latter, simply because I think they may be beyond the current scope of the repo.) @karmel thoughts?

official/boosted_trees/data_download.py
```diff
@@ -0,0 +1,78 @@
...
+  if not tf.gfile.Exists(FLAGS.data_dir):
+    tf.gfile.MkDir(FLAGS.data_dir)
+  # 2.8 GB to download.
+  try:
+    print('Data downloading..')
+    temp_filename, _ = urllib.request.urlretrieve(input_url)
+
+    # Reading and parsing 11 million csv lines takes 2~3 minutes.
+    print('Data processing.. taking multiple minutes..')
+    data = pd.read_csv(
```
robieta: TensorFlow datasets support CSV naively. Is it necessary to go through Pandas?

official/boosted_trees/train_higgs.py
```diff
@@ -0,0 +1,184 @@
...
+        data = npz['data']
+  except Exception as e:
+    raise RuntimeError(
+        'Error loading data; use data_download.py to prepare the data:\n{}: {}'
+        .format(type(e).__name__, e))
+  return (data[FLAGS.train_start:FLAGS.train_start+FLAGS.train_count],
+          data[FLAGS.eval_start:FLAGS.eval_start+FLAGS.eval_count])
+
+
+def make_inputs_from_np_arrays(features_np, label_np):
```
robieta: Could you comment on why so much of the data pipeline is done outside of tf.data.Dataset? I vaguely recall there being some nuances around boosted trees and datasets, but I don't recall the details.

robieta:
@yk5 With respect to unit tests we do not test the download scripts.

official/boosted_trees/train_higgs.py
ispirmustafa: agreed. let's remove l2. let's not put TODO. you can update the example later.

official/boosted_trees/train_higgs.py
yk5: There are several reasons; mostly to utilize benefits of numpy arrays.
1) bucketized_column requires static bucket boundaries, which are generated from numpy arrays in this example.
2) data.Dataset is streaming data and could be slow. The focus of this example is to train a model with data in memory. (data streaming use case might be a separate example.)
3) we've seen examples of gradient boosted trees algorithms use various data formats (like svm text formats or many custom formats), which Tensorflow doesn't support, and not likely support any time soon. OTOH, there are many libraries like pandas or sklearn that support various data sources and eventually return numpy matrices . So it's worth to show a data pipeline that starts with numpy array, not from raw files.

official/boosted_trees/data_download.py
yk5: Answered below. We want to show a Dataset pipeline starting from numpy arrays for some reasons.

official/boosted_trees/README.md
yk5: Not for a big reason. Just because it's faster.. and yet gives reasonably good result. It's controllable anyway, and algorithms often compare between 1M training and 10M training.

official/boosted_trees/README.md
yk5: Yes, the algorithm is deterministic, so it's very consistent.

official/boosted_trees/README.md
yk5: Yes. Memory needed to keep the input data. Of course additional memory is needed for Tensorflow execution.

official/boosted_trees/train_higgs.py
yk5: Got it. thanks.

official/boosted_trees/README.md
```diff
@@ -0,0 +1,56 @@
...
+
+
+### Training
+You can run the code locally as follows:
+
+```
+python train_higgs.py
+```
+
+The model is saved to `/tmp/higgs_model` by default, which can be changed using the `--model_dir` flag. To train a new model after one, provide a new `--model_dir` value, or remove the dir set by `--model_dir` before launching.
```
yhliang2018: What does the "a new model" mean here? Do we allow the "new model" to load checkpoints from previously trained model to continue the training?

official/boosted_trees/README.md
```diff
@@ -0,0 +1,56 @@
...
+
+Model parameters can be adjusted by flags, like `--n_trees`, `--max_depth`, `--learning_rate`.
```
yhliang2018: A detailed explanation of other parameters would be helpful, either here or in the code comments.

official/boosted_trees/README.md
```diff
@@ -0,0 +1,56 @@
...
+
+The final accuacy will be around 74% and loss will be around 0.5165 over the eval set, when trained with the default parameters.
+
+By default, the first 1 million examples among 11 millions are used for training, and the last 1 million examples are used for evaluation.
+The training/evaluation data can be selected as index ranges by flags `--train_start`, `--train_count`, `--eval_start`, `--eval_count`, etc.
```
yhliang2018: Ditto as above. An example or more detailed info would help.

official/boosted_trees/README.md
```diff
@@ -0,0 +1,56 @@
...
+
+### TensorBoard
+
+Run TensorBoard to inspect the details about the graph and training progression.
+
+```
+tensorboard --logdir=/tmp/higgs_model
```
yhliang2018: --logdir should be the model_dir (either the default one here, or user specified one)?

official/boosted_trees/README.md
```diff
@@ -0,0 +1,56 @@
...
+
+### Setup
+The [HIGGS Data Set](https://archive.ics.uci.edu/ml/datasets/HIGGS) that this sample uses for training is hosted by the [UC Irvine Machine Learning Repository](https://archive.ics.uci.edu/ml/datasets/). We have provided a script that downloads and cleans the necessary files.
+
+```
+python data_download.py
+```
+
+This will download a file and store the processed file under the directory designated by `--data_dir` (defaults to `/tmp/higgs_data/`). To change the target directory, set the `--data_dir` flag. The directory could be network storages that Tensorflow supports (like Google Cloud Storage, `gs://<bucket>/<path>/`).
+The file downloaded to the local temporary folder is about 2.8 GB, and the processed file is about 1.3G, so there should be enough storage to handle them.
```
yhliang2018: nit:1.3 GB

official/boosted_trees/train_higgs.py
```diff
@@ -0,0 +1,184 @@
...
+
+Note that BoostedTreesClassifier is available since Tensorflow 1.8.0.
+So you need to install recent enough version of Tensorflow to use this example.
+
+The training data is by default the first million examples out of 11M examples,
+and eval data is by default the last million examples.
+They are controlled by --train_start, --train_count, --eval_start, --eval_count.
+e.g. to train over the first 10 million examples instead of 1 million:
+$ python train_higgs.py --n_trees=100 --max_depth=6 --learning_rate=0.1 \
+    --model_dir=/tmp/higgs_model --train_count=10000000
```
yhliang2018: Nice example! Maybe you can refer users to check here with the brief readme info.

official/boosted_trees/README.md
yk5: I copied many parts from wide_deep, and these two paragraphs are copied and modified.
But I admit it might have got too much complicated. Will simplify.

official/boosted_trees/README.md
karmel: I'm fine with a little bit of cross-product promotion. If we get lots of issues submitted about it, we can reconsider, but okay to leave for now.

official/boosted_trees/data_download.py
karmel: Can we add to the set of dependencies as well in that case? https://github.com/tensorflow/models/blob/master/official/requirements.txt

official/boosted_trees/train_higgs.py
yk5: Just for the reference, 
- the data_download.py tool downloads the csv.gz file and parses using pandas to save into a numpy compressed file.
- numpy save/load is pretty light weight - both in several seconds for the float matrix of size (10 million * 29).
- pandas.read_csv() takes 2~3 mins to create ndarray from the csv.gz file in 6-core workstation (bogomips 7000 per core).
- numpy.loadtext() or genfromtxt() takes ~10 mins.
- tf.data pipeline to read through and parse the entire csv file takes ~ 9 mins with TextLineDataset 
 and map.

The numpy file can be generated once and reused multiple times without overhead. 
Since most users would run training many times with different parameters,
we wanted to reduce the total cost (and better user experience).

official/boosted_trees/README.md
```diff
@@ -0,0 +1,56 @@
+# Classifying Higgs boson processes in the HIGGS Data Set
+## Overview
+The [HIGGS Data Set](https://archive.ics.uci.edu/ml/datasets/HIGGS) contains 11 million samples with 28 features, and is for the classification problem to distinguish between a signal process which produces Higgs bosons and a background process which does not.
+
+We use Gradient Boosted Trees algorithm to distinguish the two classes.
+
+---
+
+The code sample in this directory uses the high level `tf.estimator.Estimator` API. This API is great for fast iteration and quickly adapting models to your own datasets without major code overhauls. It allows you to move from single-worker training to distributed training, and it makes it easy to export model binaries for prediction. Here, for further simplicity and faster execution, we use a utility function `tf.contrib.estimator.boosted_trees_classifier_train_in_memory`. This utility function is especially effective when the input is provided as in-memory data sets like numpy arrays.
```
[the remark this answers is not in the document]
yk5: done.

official/boosted_trees/README.md
yk5: That's the default behavior of Estimator. If you provide the existing model_dir, Estimator loads the checkpoint and starts training from there, which would confuse users new to Tensorflow.

I rephrased a bit.

official/boosted_trees/README.md
yk5: Each FLAG has its own description in the code. Isn't it enough?

official/boosted_trees/README.md
yk5: Good point. Added a comment to clarify it.

official/boosted_trees/README.md
yk5: Done.

robieta:
Hi. We changed from argparse to absl flag parsing. I'm sorry that you went through the trouble of learning the old way.

yk5:
Ah, okay. Thanks for letting me know. Let me make changes.

official/boosted_trees/data_download.py
yk5: Done.

yk5:
Changed the flags to use absl. PTAL @robieta

yk5:
I've switched to absl. PTAL

official/boosted_trees/train_higgs.py
```diff
@@ -0,0 +1,221 @@
...
+
+
+def define_train_higgs_flags():
+  """Add tree related flags as well as training/eval configuration."""
+  flags_core.define_base(stop_threshold=False, batch_size=False, num_gpu=False)
+  flags.adopt_module_key_flags(flags_core)
+
+  flags.DEFINE_integer(
+      name='train_start', default=0,
+      help='Start index of train examples within the data.')
```
robieta: Can you wrap your help strings with `flags_core.help_wrap()`?

yk5: Done.

yk5:
Thanks for review! Comment addressed. PTAL.. @robieta 

Btw, do you squash when you commit, or should I beforehand? Please let me know.

Also, I have been using single quotes, but found many files under official folder use double quotes (haven't checked all though). Should I convert them into doubles?

yk5:
Thank you!

PR #4094, opened 2018-04-25
Update TensorRT unit for workspace size

karmel:
Fixes #3844 . cc @samikama


samikama:
@karmel  Looks good to me

PR #4098, opened 2018-04-26
Fix 2 typos in eager.ipynb

augustjanse:
- "*more* helpful" instead of "*more helpful*"
- `matplotlib` instead of `mathplotlib`


MarkDaoust:
Thanks.

PR #4133, opened 2018-04-30
Add new import paths to mnist_tpu.py

frankchn:
This adds new import paths to mnist_tpu.py so individuals can check out the git repository and run `python mnist_tpu.py` without finagling with python import search paths.


robieta:
Why is the parent path needed? Shouldn't it only be grandparent?

frankchn:
@robieta Good point, removed.

PR #4136, opened 2018-05-01
[samples/core/get_starter/eager]: Update with API simplifications in 1.8

asimshankar:
@MarkDaoust : Is there some tests I should be running as well? And once this is in, does https://www.tensorflow.org/get_started/eager need to be regenerated?

### DIFF
--- a/samples/core/get_started/eager.ipynb
+++ b/samples/core/get_started/eager.ipynb
@@ -114,7 +114,7 @@
       "source": [
         "### Install the latest version of TensorFlow\n",
         "\n",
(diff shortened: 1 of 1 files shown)


MarkDaoust:
> Is there some tests I should be running as well?

There's nothing setup at the moment. I'll go talk to the tensorflow/models and tensorflow/testing people to see what the best way to handle these is. 

I've manually tested this PR.

> Once this is in, does https://www.tensorflow.org/get_started/eager need to be regenerated?

Yes. I will do that.

tapanprakasht:
@asimshankar @MarkDaoust Changes are not reflected in the documentation in the following link https://www.tensorflow.org/get_started/eager
Still it is referring to older api as of now. Please see the screenshot.
image

MarkDaoust:
Thanks, I'm on it.

PR #4153, opened 2018-05-02
Export @195097388.

terryykoo:
Releases the DRAGNN runtime, along with a large batch of other updates.  Notably, upgrades the TF submodule to r1.8.


terryykoo:
PTAL, I did "rm -r dragnn/runtime/" and uploaded a commit.

PR #4163, opened 2018-05-03
Add official flag-parsing and benchmarking logging utils to Transformer

k-w-w:


official/transformer/data_download.py
```diff
@@ -416,4 +415,4 @@ def main(unused_argv):
            "closest to the target size (%d)." % _TARGET_VOCAB_SIZE)
 
   FLAGS, unparsed = parser.parse_known_args()
-  main(sys.argv)
+  tf.app.run()
```
qlzh727: Since this is a standalone library that download and process the data, tf.app.run() does not do much here. If we want to use all the tf lib here, probably we should change the flag parse part also to absl flags, so that its consistent across the module.

official/transformer/transformer_main.py
```diff
@@ -249,141 +265,185 @@ def train_schedule(
...
+
+      # Write bleu scores using summary writer and benchmark logger
+      global_step = get_global_step(estimator)
+      summary = tf.Summary(value=[
+          tf.Summary.Value(tag="bleu/uncased", simple_value=uncased_score),
+          tf.Summary.Value(tag="bleu/cased", simple_value=cased_score),
+      ])
+      bleu_writer.add_summary(summary, global_step)
+      bleu_writer.flush()
+      benchmark_logger.log_metric(
```
qlzh727: What's the difference between the eval_results and the value here?

official/transformer/transformer_main.py
```diff
@@ -249,141 +265,185 @@ def train_schedule(
...
+    if not tf.gfile.Exists(flags_obj.bleu_ref):
+      raise ValueError(
+          "BLEU source file %s does not exist" % flags_obj.bleu_ref)
+    # Make sure that vocab file is in the data directory defined
+    vocab_file_path = os.path.join(flags_obj.data_dir, flags_obj.vocab_file)
+    if not tf.gfile.Exists(vocab_file_path):
+      raise ValueError("Vocabulary file %s does not exist" % vocab_file_path)
+
+
+def main(_):
```
qlzh727: I think you can follow taylor's change in imagenet, which creates a wrapper of main function, so that flag object can be injected

official/transformer/transformer_main.py
```diff
@@ -249,141 +265,185 @@ def train_schedule(
...
+    params.batch_size = flags_obj.batch_size
+
+  # Create hooks that log information about the training and metric values
+  train_hooks = hooks_helper.get_train_hooks(
+      flags_obj.hooks,
+      tensors_to_log=TENSORS_TO_LOG,  # used for logging hooks
+      batch_size=params.batch_size  # for ExamplesPerSecondHook
+  )
+  benchmark_logger = logger.config_benchmark_logger(flags_obj.benchmark_log_dir)
+  benchmark_logger.log_run_info("transformer")
```
qlzh727: The interface is updated with some extra params, please rebase and add value if needed.

official/transformer/transformer_main.py
```diff
@@ -249,141 +265,185 @@ def train_schedule(
...
-    single_iteration_train_epochs = FLAGS.epochs_between_eval
+    single_iteration_train_epochs = flags_obj.epochs_between_evals
 
-  # Make sure that the BLEU source and ref files if set
-  if FLAGS.bleu_source is not None and FLAGS.bleu_ref is not None:
-    if not tf.gfile.Exists(FLAGS.bleu_source):
-      raise ValueError("BLEU source file %s does not exist" % FLAGS.bleu_source)
-    if not tf.gfile.Exists(FLAGS.bleu_ref):
-      raise ValueError("BLEU source file %s does not exist" % FLAGS.bleu_ref)
+  # Define parameters based on flags
```
robieta: I really like this pattern of packaging hyperparameters. Very clean.

official/transformer/transformer_main.py
```diff
@@ -249,141 +265,185 @@ def train_schedule(
...
   # Add flag-defined parameters to params object
-  params.data_dir = FLAGS.data_dir
-  params.num_cpu_cores = FLAGS.num_cpu_cores
-  params.epochs_between_eval = FLAGS.epochs_between_eval
+  params.data_dir = flags_obj.data_dir
+  params.num_parallel_calls = flags_obj.num_parallel_calls
+  params.epochs_between_evals = flags_obj.epochs_between_evals
   params.repeat_dataset = single_iteration_train_epochs
-
+  if flags_obj.batch_size is not None:
```
robieta: `params.batch_size = flags_obj.batch_size or params.batch_size`? (Since we don't need to respect 0 as a legitimate batch size)

official/transformer/transformer_main.py
```diff
@@ -249,141 +265,185 @@ def train_schedule(
...
+      batch_size=params.batch_size  # for ExamplesPerSecondHook
+  )
+  benchmark_logger = logger.config_benchmark_logger(flags_obj.benchmark_log_dir)
+  benchmark_logger.log_run_info("transformer")
+
+  # Train and evaluate transformer model
   estimator = tf.estimator.Estimator(
-      model_fn=model_fn, model_dir=FLAGS.model_dir, params=params)
+      model_fn=model_fn, model_dir=flags_obj.model_dir, params=params)
   train_schedule(
```
robieta: As long as these are already nice and broken out, may I request that they be kwargs?

official/transformer/transformer_main.py
robieta: Indeed.

official/transformer/transformer_main.py
k-w-w: The results from estimator.evaluate() are based on an approximate translation (long story short, approx. translations are heavily based on the target golden values provided).
The function `evaluate_and_log_bleu` uses the estimator.predict() path to compute the translations, where the golden values are not provided. The translations are compared to reference file to get the actual bleu score.

official/transformer/compute_bleu.py
```diff
@@ -126,4 +126,4 @@ def main(unused_argv):
       metavar="<BV>")
 
   FLAGS, unparsed = parser.parse_known_args()
-  main(sys.argv)
+  tf.app.run()
```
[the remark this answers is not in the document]
k-w-w: Changed to use absl

k-w-w:
Whoops, pushed a bunch of unrelated commits. Rebased them away.

@qlzh727 @robieta Thanks for the initial reviews, I've made the requested changes. PTAL

official/transformer/transformer_main.py
```diff
@@ -378,15 +378,16 @@ def _check_train_limits(flag_dict):
...
+    """Validate files when bleu_source and bleu_ref are defined."""
     if flags_dict["bleu_source"] is None or flags_dict["bleu_ref"] is None:
       return True
     # Ensure that bleu_source, bleu_ref, and vocab files exist.
     vocab_file_path = os.path.join(
         flags_dict["data_dir"], flags_dict["vocab_file"])
-    return tf.gfile.Exists(flags_dict["bleu_source"]) and (
-        tf.gfile.Exists(flags_dict["bleu_ref"])) and (
-        tf.gfile.Exists(vocab_file_path))
+    return (
```
k-w-w: Does anyone know how to make this look better? It is difficult to please the lint overlords.

robieta:
```
    return all([
        tf.gfile.Exists(flags_dict["bleu_source"]),
        tf.gfile.Exists(flags_dict["bleu_ref"]),
        tf.gfile.Exists(vocab_file_path),
    ])
```

k-w-w: thank you

official/transformer/compute_bleu.py
```diff
@@ -109,21 +111,22 @@ def main(unused_argv):
...
-      "--bleu_variant", "-bv", type=str, choices=["uncased", "cased"],
-      nargs="*", default=None,
-      help="Specify one or more BLEU variants to calculate (both are "
-           "calculated by default. Variants: \"cased\" or \"uncased\".",
-      metavar="<BV>")
-
-  FLAGS, unparsed = parser.parse_known_args()
-  main(sys.argv)
+  flags.DEFINE_string(
+      name="translation", short_name="t", default=None,
```
karmel: The collision of abbreviations is inevitable at this point. Maybe we should make a rule that we don't have abbreviations for flags defined within model modules. Else, we might add a flag to the base set, conflict, and not realize it until someone tries to run an inheriting model and it either errors out at arg load time, or treats flags in strange ways. Thoughts, @robieta ?

official/transformer/transformer_main.py
karmel: Can this be added in a comment?

official/transformer/transformer_main.py
```diff
@@ -245,145 +267,190 @@ def train_schedule(
...
+
+  # Create hooks that log information about the training and metric values
+  train_hooks = hooks_helper.get_train_hooks(
+      flags_obj.hooks,
+      tensors_to_log=TENSORS_TO_LOG,  # used for logging hooks
+      batch_size=params.batch_size  # for ExamplesPerSecondHook
+  )
+  benchmark_logger = logger.config_benchmark_logger(flags_obj.benchmark_log_dir)
+  benchmark_logger.log_run_info(
+      "transformer", "wmt_translate_ende", params.__dict__)
```
karmel: maybe include kwarg names here for clarity?

official/transformer/compute_bleu.py
robieta: I think at this point one letter abbreviations are too dangerous to add, because they may collide not only with our own flags but flags defined who knows where. Two letters seems reasonably safe. 

If we're doing an end-to-end test it will blow up for collisions, so I'm less worried that we will accidentally have internal collisions.

robieta: Right now mnist is missing it. It requires synthetic to be hermetic which I think is why mnist doesn't have it yet.

I definitely think it is something we want for every model.

k-w-w:
@karmel Thanks for the review. I'll add the savedmodel export in a separate PR, to keep changes relatively related to flag parsing + logging.

k-w-w:
Bump, PTAL

official/transformer/compute_bleu.py
```diff
@@ -109,21 +111,22 @@ def main(unused_argv):
...
+      name="translation", default=None,
+      help=flags_core.help_wrap("File containing translated text."))
+  flags.mark_flag_as_required("translation")
+
+  flags.DEFINE_string(
+      name="reference", default=None,
+      help=flags_core.help_wrap("File containing reference translation."))
+  flags.mark_flag_as_required("reference")
+
+  flags.DEFINE_multi_enum(
```
robieta: Can you add a both option and propagate that through? (and make this just a DEFINE_enum) with this one would have to input 
`--bleu_variant cased --bleu_variant uncased`, and may be confused when 
`--bleu_variant cased uncased` doesn't compute uncased. Unfortunately absl doesn't have a great way to define a list of enumerables.

official/transformer/compute_bleu.py
```diff
@@ -109,21 +111,22 @@ def main(unused_argv):
...
+      name="bleu_variant", short_name="bv", default=["uncased", "cased"],
+      enum_values=["uncased", "cased"], case_sensitive=False,
+      help=flags_core.help_wrap(
+          "Specify one or more BLEU variants to calculate. Variants: \"cased\" "
+          "or \"uncased\"."))
+
+  FLAGS = flags.FLAGS
```
robieta: Can you encapsulate everything the same way you did in transformer_main?

official/transformer/transformer_main.py
```diff
@@ -245,145 +267,198 @@ def train_schedule(
...
+  if flags_obj.train_steps is not None:
+    train_eval_iterations = (
+        flags_obj.train_steps // flags_obj.steps_between_evals)
+    single_iteration_train_steps = flags_obj.steps_between_evals
     single_iteration_train_epochs = None
   else:
-    if FLAGS.train_epochs is None:
-      FLAGS.train_epochs = DEFAULT_TRAIN_EPOCHS
-    train_eval_iterations = FLAGS.train_epochs // FLAGS.epochs_between_eval
+    if flags_obj.train_epochs is None:
```
robieta: We have adopted a convention of not modifying flags objects, and instead calling getter functions to retrieve values. See `official.utils.flags._performance.get_loss_scale()` as an example.

official/transformer/data_download.py
```diff
@@ -18,19 +18,20 @@
 from __future__ import division
 from __future__ import print_function
 
-import argparse
 import os
 import random
-import sys
 import tarfile
 import urllib
```
robieta: This is breaking python3. Need to use `six.moves.urllib`

official/transformer/transformer_main.py
k-w-w: Yes, it makes sense not to alter the value of the flags after. It would great if these conventions can be listed (maybe as a sub-checkbox in karmel's list above).

official/transformer/compute_bleu.py
k-w-w: What is the reason for encapsulating everything?

official/transformer/transformer_main.py
robieta: Would you mind adding an "immutability" section to flags/README.md?

official/transformer/compute_bleu.py
robieta: In part it is conceptually easier if functions are pure, and in part because we may well want to use this in a way other than calling the file directly.

k-w-w: Oooh I see. Thanks for the explanation.

k-w-w:
@robieta Thanks, for the comments. 

I edited compute_bleu, but I'm not sure I prefer this version to how it was before. Currently there isn't really a use case for calling the `compute_bleu.main` (or the added `compute_bleu.compute_bleu`) outside of running `compute_bleu.py`. The function `bleu_wrapper` does all of the work in computing the bleu score. The main function just prints it out. 

In general, I don't think we should require a encapsulating function for scripts like `compute_bleu.py` or `data_download.py`, unless there is a foreseeable use case.

official/transformer/compute_bleu.py
```diff
@@ -98,32 +100,42 @@ def bleu_wrapper(ref_filename, hyp_filename, case_sensitive=False):
...
 
 
-def main(unused_argv):
-  if FLAGS.bleu_variant is None or "uncased" in FLAGS.bleu_variant:
-    score = bleu_wrapper(FLAGS.reference, FLAGS.translation, False)
+def compute_bleu(flags_obj):
+  """Print out the BLEU scores calculated from the files defined in flags."""
+  if flags_obj.bleu_variant in ("both", "uncased"):
+    score = bleu_wrapper(flags_obj.reference, flags_obj.translation, False)
     print("Case-insensitive results:", score)
```
qlzh727: I think a following change will probably needed to change all the print to tf.logging for consistency.

official/transformer/data_download.py
```diff
@@ -156,7 +157,7 @@ def download_from_url(path, url):
     filename = os.path.join(path, filename)
     tf.logging.info("Downloading from %s to %s." % (url, filename))
     inprogress_filepath = filename + ".incomplete"
-    inprogress_filepath, _ = urllib.urlretrieve(
+    inprogress_filepath, _ = urllib.request.urlretrieve(
         url, inprogress_filepath, reporthook=download_report_hook)
     # Print newline to clear the carriage return from the download progress.
     print()
```
qlzh727: Let's grep and replace all the print() into tf.logging.info

k-w-w:
@robieta Thanks
@qlzh727 Do the changes look good to you?

official/transformer/transformer_main.py
```diff
@@ -245,145 +267,199 @@ def train_schedule(
...
+  flags_core.define_benchmark()
+
+  # Set flags from the flags_core module as "key flags" so they're listed when
+  # the '-h' flag is used. Without this line, the flags defined above are
+  # only shown in the full `--helpful` help text.
+  flags.adopt_module_key_flags(flags_core)
+
+  # Add transformer-specific flags
+  flags.DEFINE_enum(
+      name="params", short_name="mp", default="big",
```
qlzh727: Name a parameter to "params" is quite evil since it does not provide much context. How about rename this into param_set or param_template so that its more explicit.

official/transformer/transformer_main.py
```diff
@@ -245,145 +267,199 @@ def train_schedule(
...
+      enum_values=["base", "big"],
+      help=flags_core.help_wrap(
+          "Parameter set to use when creating and training the model."))
```
qlzh727: I think its also worth mentioning all the individual param values under this umbrella as well.

official/transformer/transformer_main.py
```diff
@@ -245,145 +267,199 @@ def train_schedule(
...
+  if flags_obj.train_steps is not None:
+    train_eval_iterations = (
+        flags_obj.train_steps // flags_obj.steps_between_evals)
+    single_iteration_train_steps = flags_obj.steps_between_evals
     single_iteration_train_epochs = None
   else:
-    if FLAGS.train_epochs is None:
-      FLAGS.train_epochs = DEFAULT_TRAIN_EPOCHS
-    train_eval_iterations = FLAGS.train_epochs // FLAGS.epochs_between_eval
+    if flags_obj.train_epochs is not None:
```
qlzh727: i think this if-else can be combined into:

train_epochs = flags_obj.train_epochs or DEFAULT_TRAIN_EPOCHS

official/transformer/transformer_main.py
```diff
@@ -245,145 +267,199 @@ def train_schedule(
...
+  train_hooks = hooks_helper.get_train_hooks(
+      flags_obj.hooks,
+      tensors_to_log=TENSORS_TO_LOG,  # used for logging hooks
+      batch_size=params.batch_size  # for ExamplesPerSecondHook
+  )
+  benchmark_logger = logger.config_benchmark_logger(flags_obj.benchmark_log_dir)
+  benchmark_logger.log_run_info(
+      model_name="transformer",
+      dataset_name="wmt_translate_ende",
+      run_params=params.__dict__)
```
qlzh727: I think its better to not log all the params, since it includes noises like data_dir. Maybe a more explicit set of param is better here.

official/transformer/data_download.py
k-w-w: The download progress rewrites the line to update the progress. I don't think it works well with tf.logging.info.

official/transformer/transformer_main.py
k-w-w: Evil is a bit harsh, don't you think?

I do agree that the argument name is vague. I'll change it to param_set.

official/transformer/transformer_main.py
k-w-w: I left it out because the absl `-h` option shows the possible enum values. 

This is what the help text looks like:

      -mp,--params: <base|big>:
        Parameter set to use when creating and training the model.
        (default: 'big')

official/transformer/transformer_main.py
k-w-w: That's much cleaner, thanks

official/transformer/data_download.py
qlzh727: ah, sorry for missing that.

official/transformer/transformer_main.py
qlzh727: Sorry for being too harsh. param_set SGTM.

official/transformer/transformer_main.py
qlzh727: What i am trying to say is param_set=base will populate other a, b and c param, which is not showing up the help txt.

k-w-w: My bad, I completely misread that. Yes, I think it would be good to see the individual param values. Maybe not all of them, but at least the ones that change between the big and base parameter sets.

robieta: Something like `setting params_set=big increases the default batch size, as well as the hidden_size, filter_size, and num_heads topology hyperparameters.\nSee transformer/model/model_params.py for details.`?

k-w-w: Yes, like that. I've pushed the change that updates the help text

k-w-w:
@qlzh727 Thanks again for all of the comments. They were very helpful. I've pushed the changes requested.

qlzh727:
Please wait for the comments from karmel@ if there is any.

PR #4205, opened 2018-05-08
Fixed Bug regarding tfrecord shuffling in object_detection

vladpaunescu:
In current implementation in `models/research/object_detection/utils/dataset_util.py` at `line 137` `records_dataset.shuffle(config.shuffle_buffer_size)` is not assigned to variable, and it should be `
records_dataset = records_dataset.shuffle(config.shuffle_buffer_size)`


vladpaunescu:
Can you take a look please? Shuffling is also done in such a way that it doesn't generate random permatation. That is because file dataset is being repeated instead of records inside file. This impacts smaller datasets with no sharding.

PR #4207, opened 2018-05-08
Forbid ResNet v1 from running with fp16

robieta:
This PR adds a flag validator to prevent this particular numerically unstable configuration.

### DIFF
--- a/official/resnet/resnet_run_loop.py
+++ b/official/resnet/resnet_run_loop.py
@@ -475,7 +475,8 @@ def define_resnet_flags(resnet_size_choices=None):
 
   # The current implementation of ResNet v1 is numerically unstable when run
   # with fp16 and will produce NaN errors soon after training begins.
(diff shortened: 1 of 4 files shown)


official/resnet/resnet_run_loop.py
```diff
@@ -471,3 +472,11 @@ def define_resnet_flags(resnet_size_choices=None):
     flags.DEFINE_string(**choice_kwargs)
   else:
     flags.DEFINE_enum(enum_values=resnet_size_choices, **choice_kwargs)
+
+  # The current implementation of ResNet v1 is numerically unstable when run
+  # with fp16 and will produce NaN errors soon after training begins.
+  msg = 'ResNet version 1 does not currently support fp16.'
```
karmel: 'ResNet version 1 is not currently supported with fp16. Please use version 2 instead.'

robieta:
I changed the wording and added an entry to the unittests.

PR #4209, opened 2018-05-08
Add benchmark logger that does stream upload to bigquery.

qlzh727:
1. All three type of benchmark loggers are behind the control flag:
benchmark_logger_type.

2. The benchmark_uploader is spilt into lib and main module.

3. Added test cases with mock for bigquery related classes.


karmel:
Hmm, it is hard to review this because of the move; what changed in the moved file? For future reference, there is a git mv that I think helps git track changes a little more clearly.

samples/core/get_started/eager.ipynb
```diff
@@ -8,6 +8,9 @@
       "views": {},
```
karmel: Looks like a whole bunch of changes got pulled in... I will hold off reviewing until the changes are resolved, just in case the solution is, "delete this PR and start over."

qlzh727:
Closing this PR and created new one #4210

PR #4210, opened 2018-05-09
Add benchmark logger that does stream upload to bigquery.

qlzh727:
1. All three type of benchmark loggers are behind the control flag:
benchmark_logger_type.

2. The benchmark_uploader is spilt into lib and main module.

3. Added test cases with mock for bigquery related classes.

4. The git mv does not seems to preserve the file history, even I try to just move the file to new location in the first commit.


qlzh727:
FYI, if compare the first commit and second one, you will be able to see the diff between for the moved file.

robieta:
I think a README would be quite helpful here. It's not clear to me how all of this fits together, or why multiple flavors of uploaders are necessary in the first place.

qlzh727:
Sure. Will do in a following commit.

official/benchmark/benchmark_uploader.py
```diff
@@ -72,19 +61,17 @@ def upload_benchmark_run(self, dataset_name, table_name, run_id):
         the data will be uploaded.
       run_id: string, a unique ID that will be attached to the data, usually
         this is a UUID4 format.
+      run_json: dict, the JSON data that contains the benchmark run info.
```
karmel: The naming of this is confusing-- is it JSON or a python dict? If the latter, why is it called json?

official/benchmark/benchmark_uploader.py
```diff
@@ -72,19 +61,17 @@ def upload_benchmark_run(self, dataset_name, table_name, run_id):
...
-      benchmark_json["model_id"] = run_id
-      table_ref = self._bq_client.dataset(dataset_name).table(table_name)
-      errors = self._bq_client.insert_rows_json(table_ref, [benchmark_json])
-      if errors:
-        tf.logging.error(
-            "Failed to upload benchmark info to bigquery: {}".format(errors))
-
-  def upload_metric(self, dataset_name, table_name, run_id):
+    run_json["model_id"] = run_id
+    table_ref = self._bq_client.dataset(dataset_name).table(table_name)
```
karmel: Comment me-- what is happening here?

official/benchmark/benchmark_uploader.py
```diff
@@ -95,39 +82,54 @@ def upload_metric(self, dataset_name, table_name, run_id):
...
         this is a UUID4 format. This should be the same as the benchmark run_id.
+      metric_json_list: list, a list of JSON object that record the metric info.
+    """
+    for m in metric_json_list:
+      m["run_id"] = run_id
+    table_ref = self._bq_client.dataset(dataset_name).table(table_name)
+    errors = self._bq_client.insert_rows_json(table_ref, metric_json_list)
+    if errors:
+      tf.logging.error(
+          "Failed to upload benchmark info to bigquery: {}".format(errors))
```
karmel: This is the same as above; why not just have the above method call this method with the wrapped json object?

official/benchmark/benchmark_uploader.py
```diff
@@ -95,39 +82,54 @@ def upload_metric(self, dataset_name, table_name, run_id):
...
+      run_id: string, a unique ID that will be attached to the data, usually
+        this is a UUID4 format. This should be the same as the benchmark run_id.
+      metric_json_file: string, the file path that contains the metric JSON
+        data.
     """
-    expected_file = os.path.join(
-        self._logging_dir, logger.METRIC_LOG_FILE_NAME)
-    with tf.gfile.GFile(expected_file) as f:
+    with tf.gfile.GFile(metric_json_file) as f:
       lines = f.readlines()
```
karmel: It's easy to imagine this consuming all of the memory for a really big file. Why not do this as a generator? (That is, for line in f.read(), or whatever the correct syntax is, rather than reading in the whole file upfront.)

official/utils/flags/_benchmark.py
```diff
@@ -36,6 +36,14 @@ def define_benchmark(benchmark_log_dir=True, bigquery_uploader=True):
 
   key_flags = []
 
+  flags.DEFINE_enum(
+      name="benchmark_logger_type", default="BaseBenchmarkLogger",
+      enum_values=["BaseBenchmarkLogger", "BenchmarkFileLogger",
+                   "BenchmarkBigQueryLogger"],
+      help=help_wrap("The type of benchmark logger to use. Default to use "
```
karmel: nit: Defaults to using

official/utils/flags/_benchmark.py
```diff
@@ -36,6 +36,14 @@ def define_benchmark(benchmark_log_dir=True, bigquery_uploader=True):
...
+                     "BaseBenchmarkLogger which logs to STDOUT. Different "
+                     "loggers will require other flags to be able to work."))
```
karmel: Can we add some fancy validators for those flags?

official/utils/logs/logger.py
```diff
@@ -28,38 +28,57 @@
...
-    if logging_dir:
-      _benchmark_logger = BenchmarkFileLogger(logging_dir)
-    else:
+    if not flag_obj:
+      flag_obj = FLAGS
+
+    if (not hasattr(flag_obj, 'benchmark_logger_type') or
+        flag_obj.benchmark_logger_type == 'BaseBenchmarkLogger'):
       _benchmark_logger = BaseBenchmarkLogger()
+    elif flag_obj.benchmark_logger_type == 'BenchmarkFileLogger':
```
karmel: Do we still need the filelogger as an option? If we are not going to use it, why complicate this with a third option that we have to maintain? That seems like it would allow us to strip half the code here and above.

official/utils/logs/logger.py
```diff
@@ -183,6 +202,68 @@ def log_run_info(self, model_name, dataset_name, run_params):
...
+      value: number, the value of the metric. The value will not be logged if it
+        is not a number type.
+      unit: string, the unit of the metric, E.g "image per second".
+      global_step: int, the global_step when the metric is logged.
+      extras: map of string:string, the extra information about the metric.
+    """
+    if not isinstance(value, numbers.Number):
+      tf.logging.warning(
+          "Metric value to log should be a number. Got %s", type(value))
+      return
```
karmel: If we need to maintain the File version, I would rather not have this check in multiple places-- extract to function?

official/utils/logs/logger.py
```diff
@@ -183,6 +202,68 @@ def log_run_info(self, model_name, dataset_name, run_params):
...
+    extras = _convert_to_json_dict(extras)
+
+    metric = [{
+        "name": name,
+        "value": float(value),
+        "unit": unit,
+        "global_step": global_step,
+        "timestamp": datetime.datetime.utcnow().strftime(
+            _DATE_TIME_FORMAT_PATTERN),
+        "extras": extras}]
```
karmel: Or, better still, looks like this is all shared code; extract or make these inherit from a shared object, rather than repeat.

karmel:
Also, just to confirm: this should be non-blocking, correct? That is, it shouldn't affect the speed of training at all?

official/utils/flags/_benchmark.py
robieta: I see that I'm winning hearts and minds.

official/benchmark/benchmark_uploader.py
qlzh727: The side effect of that is we will have one API call per line. I could implement some buffering and limit the max number of lines we load for each upload, if this is what u want. At the moment, we don't see a memory issue since usually we only have O(100) lines of metric record for small model. For big model like imagenet, it will be covered by the streaming upload.

official/utils/flags/_benchmark.py
qlzh727: Done.

official/utils/flags/_benchmark.py
qlzh727: Done. Added validation for flag combination.

official/utils/logs/logger.py
qlzh727: yes. The file logger is still useful for small model so that we can save the bigquery upload quota. The streaming one will be useful for those long running benchmark.

official/utils/logs/logger.py
qlzh727: Done.

official/utils/logs/logger.py
qlzh727: Done.

qlzh727:
I think the hooks are executed on a separate thread during training time, if I understand it correctly, so the training performance shouldn't be affected. During eval stage, this will add some overhead because of the bigquery upload for the eval result.

official/benchmark/benchmark_uploader.py
robieta: I think we should absolutely not refer to dicts as JSON. Some dicts can be converted to JSON, but I tend to think that json in the variable names implies that the variable refers to the serialized string.

qlzh727:
And I was lying about the hook behavior. It seems that the hook does block on the after_run method, so computation running in hook will have a potential performance impact. 

Let me update the metric logging to be execute on a different thread.

qlzh727:
Updated to use new thread for bigquery upload.

official/benchmark/benchmark_uploader.py
karmel: But isn't this exactly the same as:

```
for line in f:
  if line.strip():
    metrics.append(json.loads(line))
```
Except that readlines requires all lines read into memory, and the above creates a generator that reads line by line?

official/utils/logs/logger.py
karmel: Why not always stream?

official/utils/logs/logger.py
```diff
@@ -183,6 +185,65 @@ def log_run_info(self, model_name, dataset_name, run_params):
...
+    Args:
+      model_name: string, the name of the model.
+      dataset_name: string, the name of dataset for training and evaluation.
+      run_params: dict, the dictionary of parameters for the run, it could
+        include hyperparameters or other params that are important for the run.
+    """
+    run_info = _gather_run_info(model_name, dataset_name, run_params)
+    # Starting new thread for bigquery upload in case it might take long time
+    # and impact the benchmark and performance measurement.
+    thread.start_new_thread(
```
karmel: Q1: If we spawn a thread to do this, what happens to errors raised? They will not interrupt the run, correct? What will happen in that scenario? 
Q2: Under normal circumstances, we allow up to num_cpus for input processing, whereas this will require one separate CPU thread. In practice, it's really unlikely that this will matter, but it's important to be aware of possible heisenbug situations, where we are somehow slowing down runs by trying to observe them, especially when we get to CPU benchmarks. Thoughts? I guess we can compare the speed of non-streaming to streaming runs periodically to check for such an effect, but can this be done in a more automated way?

official/benchmark/benchmark_uploader.py
qlzh727: Done. Thanks for the suggestion.

official/utils/logs/logger.py
qlzh727: Bigquery API support to have a usage quota, even we currently have small load, I think its still worthwhile to not spam the service. For small dataset like cifar10 which take 40 mins to finish, I would still prefer to upload the data after the run.

official/utils/logs/logger.py
qlzh727: 1. We currently log the error and does not stop the training process.
2. CPU benchmark might be harder since all the process running on the machine will have the potential impact of the performance. We can setup a separate run for non-streaming benchmark when we start testing on CPU.

karmel: Okay-- let's just make sure to record here in the comments that users should be aware of this potential performance implication, especially for CPU training, and run tests accordingly.

qlzh727: Done.

PR #4213, opened 2018-05-09
Make codes more pythonic.

Kongsea:
Make some code parts more pythonic.


robieta:
It looks like you missed an indent. (I'm in particular looking at train_image_classifier.py) If you did intend that clause to indented to be in the same if as `if var.op.name.startswith(exclusion)`, then the logic is different rather than just an aestetic change.

Kongsea:
Thanks.
Now I have restored the indents.
Besides, I think the code logic is the same as before.

research/slim/train_image_classifier.py
```diff
@@ -340,12 +340,10 @@ def _get_init_fn():
   # TODO(sguada) variables.filter_variables()
   variables_to_restore = []
   for var in slim.get_model_variables():
-    excluded = False
     for exclusion in exclusions:
       if var.op.name.startswith(exclusion):
-        excluded = True
         break
-    if not excluded:
+    else:
```
robieta: This is the unmatched `else` I was referring to.

Kongsea: No, it matches with **for**. It's a for...else... structure.

robieta: Huh. TIL. That does indeed appear to be equivalent.

Let me get back to you tomorrow.

Kongsea: Look forward to you.

PR #4224, opened 2018-05-10
Fix bugs in object detection docs

xiaoyongzhu:
According to code here:
https://github.com/tensorflow/models/blob/master/research/object_detection/meta_architectures/ssd_meta_arch.py#L780

The interface that DetectionModels should implement to load a checkpoint into the Tensorflow graph should be `restore_map` rather than `restore`.


xiaoyongzhu:
I signed it!

PR #4230, opened 2018-05-10
Fix model layer for resnet v1.

qlzh727:
The final BN and ReLU layer is only need for v2 model since it was
doing preactivation in each block.

### DIFF
--- a/official/resnet/resnet_model.py
+++ b/official/resnet/resnet_model.py
@@ -424,6 +424,7 @@ class Model(object):
     self.block_strides = block_strides
     self.final_size = final_size
     self.dtype = dtype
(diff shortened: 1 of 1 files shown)


robieta:
Sadly this does not resolve the divergent fp16 v1 training.

qlzh727:
@karmel, the extra ReLU shouldn't impact the correctness of the model. The BN will affect the number to let the final number scale down further. 

The fp16 problem in v1 is because of overflow of the number on the top few layer of the model. Removing the final BN and ReLU won't fix it.

HiKapok:
have the pretrained model of resnet-v1 been replaced ?

PR #4246, opened 2018-05-11
Update the wide_deep and transformer code for latest benchmark config.

qlzh727:


k-w-w:
Can you apply this to [transformer_main.py](https://github.com/tensorflow/models/blob/master/official/transformer/transformer_main.py) as well?

PR #4270, opened 2018-05-15
Fix transformer loss

k-w-w:
Thanks for finding this @robieta

### DIFF
--- a/official/transformer/transformer_main.py
+++ b/official/transformer/transformer_main.py
@@ -83,7 +83,7 @@ def model_fn(features, labels, mode, params):
     # Calculate model loss.
     xentropy, weights = metrics.padded_cross_entropy_loss(
         logits, targets, params.label_smoothing, params.vocab_size)
-    loss = tf.reduce_sum(xentropy * weights) / tf.reduce_sum(weights)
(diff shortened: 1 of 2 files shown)


official/transformer/transformer_main.py
```diff
@@ -83,7 +83,7 @@ def model_fn(features, labels, mode, params):
     # Calculate model loss.
     xentropy, weights = metrics.padded_cross_entropy_loss(
         logits, targets, params.label_smoothing, params.vocab_size)
-    loss = tf.reduce_sum(xentropy * weights) / tf.reduce_sum(weights)
+    loss = tf.reduce_sum(xentropy) / tf.reduce_sum(weights)
 
     # Save loss as named tensor that will be logged with the logging hook.
     tf.identity(loss, "cross_entropy")
```
robieta: As long as you're in here could you disambiguate xentropy and loss? this tf.identity name is quite confusing.

k-w-w: I've added comments describing xentropy and loss. Loss represents the overall cross entropy loss of the model, so the tf.identity name should be accurate.

k-w-w:
Thanks!

PR #4281, opened 2018-05-16
Export Transformer saved model, and clarify vocab file flag.

k-w-w:
Also added instructions for using the SavedModel to the README


official/transformer/transformer_main.py
```diff
@@ -388,20 +400,35 @@ def _check_train_limits(flag_dict):
...
+      message="Please ensure that the --vocab_file exists in --data_dir. If it "
+              "exists, please make sure the contents match that of the vocab "
+              "file in --model_dir.")
+  def _check_vocab_file(flags_dict):
+    data_vocab = os.path.join(
+        flags_dict["data_dir"], flags_dict["vocab_file"])
+    model_vocab = os.path.join(flags_dict["model_dir"], VOCAB_FILE)
+
+    # If both exist, then make sure the vocab files are the same.
+    if tf.gfile.Exists(data_vocab) and tf.gfile.Exists(model_vocab):
```
qlzh727: Please use filecmp.cmp to diff the file content.

official/transformer/transformer_main.py
```diff
@@ -388,20 +400,35 @@ def _check_train_limits(flag_dict):
...
-        flags_dict["data_dir"], flags_dict["vocab_file"])
     return all([
         tf.gfile.Exists(flags_dict["bleu_source"]),
-        tf.gfile.Exists(flags_dict["bleu_ref"]),
-        tf.gfile.Exists(vocab_file_path)])
+        tf.gfile.Exists(flags_dict["bleu_ref"])])
+
+  @flags.multi_flags_validator(
+      ["data_dir", "model_dir", "vocab_file"],
+      message="Please ensure that the --vocab_file exists in --data_dir. If it "
```
qlzh727: I think the error message should be more explicit about at least one of them should exist. The message is clear about the content need to be same when both of the exist.

official/transformer/transformer_main.py
```diff
@@ -388,20 +400,35 @@ def _check_train_limits(flag_dict):
...
+              "exists, please make sure the contents match that of the vocab "
+              "file in --model_dir.")
```
qlzh727: I also think the error message should provide action items to user to unblock them or fix the issue. In this case, probably a command to remove the vocab file in model is more appropriate.

k-w-w:
@qlzh727 @karmel Thanks for the super helpful suggestions, I've added vocab_file flag back in and removed the file copying.

official/transformer/transformer_main.py
```diff
@@ -457,7 +463,15 @@ def run_transformer(flags_obj):
       bleu_source=flags_obj.bleu_source,
       bleu_ref=flags_obj.bleu_ref,
       bleu_threshold=flags_obj.stop_threshold,
-      vocab_file_path=os.path.join(flags_obj.data_dir, flags_obj.vocab_file))
+      vocab_file=flags_obj.vocab_file)
+
+  if flags_obj.export_dir:
+    serving_input_fn = tf.estimator.export.build_raw_serving_input_receiver_fn({
+        "inputs": tf.placeholder(tf.int64, [None, None])
```
karmel: You can use a TensorServingInputReceiver, and then you don't need to unwrap the dict above.

official/transformer/transformer_main.py
```diff
@@ -457,7 +463,15 @@ def run_transformer(flags_obj):
...
+    })
+    estimator.export_savedmodel(
+        flags_obj.export_dir, serving_input_fn, as_text=True,
```
karmel: as_text should be false

official/transformer/transformer_main.py
```diff
@@ -64,7 +63,8 @@
 def model_fn(features, labels, mode, params):
   """Defines how to train, evaluate and predict from the transformer model."""
   with tf.variable_scope("model"):
-    inputs, targets = features, labels
+    inputs = features['inputs'] if isinstance(features, dict) else features
```
karmel: Should be unnecessary; see below.

k-w-w:
@karmel PTAL (I forgot about this PR until I took a look at the branches...)

PR #4295, opened 2018-05-17
fix read mode of  build_ade20k_data.py

tanutarou:
related : https://github.com/tensorflow/models/pull/3741  
build_ade20k_data.py has a same problem.
The FastGFile(.., 'b') does not work for Python 3


tanutarou:
I signed it!

PR #4296, opened 2018-05-17
First pass at a TPU loop for Transformer

robieta:
This is very much a first attempt. Performance is not good (~15% TPU utilization), but it's a start.
EDIT: Utilization is now at 40%. Still not great, but better.

1) Static batching:
Transformer tries to group sequences of shorter lengths and run with larger batches when it does so. So one batch may have shape [30, 68], while the next might have shape [20, 100]. This is completely anathema to XLA's shape requirements. In order to run on TPUs sequences must be padded to the max allowable length which ensures static shape. (Thanks a lot @k-w-w for explaining and implementing this.)

2) Feed forward padding:
Similarly the final dense layers of transformer can save  work by ignoring the padded elements. However, this confuses XLA because of the presence of tf.where, and so must be disable for TPU use.

3) params dict instead of params object:
Transformer stores parameters in an object which results in very clean code, but TPUEstimator demands params be passed as a dict because it does checking and strategic value insertion. params are still initially stored in an object for readability and inheritance, and then simply converted to a dict before being passed to the estimator. As a result there are lots of changes in this PR that simply change `params.foo` to `params["foo"]`.

4) Schedule manager:
The control flow logic in Transformer was already non-trivial as it supports both a step basis and an epoch basis, and TPUs further complicate things because even when epochs are specified the estimator still has to use steps under the hood. (As well as other such complications) I elected to pull all of this logic into a controller class rather than have it interspersed throughout the run loop.

5) Host call:
Summary ops are complicated with TPUs. I more or less copied the host call (and lovely docstring) from the cloud TPU implementation of ResNet.

6) Hooks and eval metrics:
Hooks are completely disabled with TPUs for now. Hopefully they can be brought back soon. 3 of the eval metrics are disabled for TPUs because they use tf.py_func which is not supported by XLA. They may or may not ever support TPUs.

### DIFF
--- a/official/transformer/model/embedding_layer.py
+++ b/official/transformer/model/embedding_layer.py
@@ -21,15 +21,31 @@ from __future__ import print_function
 import tensorflow as tf  # pylint: disable=g-bad-import-order
 
 from official.transformer.model import model_utils
+from official.utils.accelerator import tpu as tpu_utils
 
 
 class EmbeddingSharedWeights(tf.layers.Layer):
   """Calculates input embeddings and pre-softmax linear with shared weights."""
 
-  def __init__(self, vocab_size, hidden_size):
+  def __init__(self, vocab_size, hidden_size, method="gather"):
+    """Specify characteristic parameters of embedding layer.
+
+    Args:
+      vocab_size: Number of tokens in the embedding. (Typically ~32,000)
+      hidden_size: Dimensionality of the embedding. (Typically 512 or 1024)
+      method: Strategy for performing embedding lookup. "gather" uses tf.gather
+        which performs well on CPUs and GPUs, but very poorly on TPUs. "matmul"
+        one-hot encodes the indicies and formulates the embedding as a sparse
+        matrix multiplication. The matmul formulation is wasteful as it does
+        extra work, however matrix multiplication is very fast on TPUs which
+        makes "matmul" considerably faster than "gather" on TPUs.
+    """
     super(EmbeddingSharedWeights, self).__init__()
     self.vocab_size = vocab_size
     self.hidden_size = hidden_size
+    if method not in ("gather", "matmul"):
+      raise ValueError("method {} must be 'gather' or 'matmul'".format(method))
+    self.method = method
 
   def build(self, _):
     with tf.variable_scope("embedding_and_softmax", reuse=tf.AUTO_REUSE):
@@ -53,18 +69,24 @@ class EmbeddingSharedWeights(tf.layers.Layer):
         locations of the padding tokens in x.
     """
     with tf.name_scope("embedding"):
-      embeddings = tf.gather(self.shared_weights, x)
+      # Create binary mask of size [batch_size, length]
+      mask = tf.to_float(tf.not_equal(x, 0))
+
+      if self.method == "gather":
+        embeddings = tf.gather(self.shared_weights, x)
+      else:  # matmul
+        embeddings = tpu_utils.embedding_matmul(
+            embedding_table=self.shared_weights,
+            values=tf.cast(x, dtype=tf.int32),
+            mask=mask
+        )
+      embeddings *= tf.expand_dims(mask, -1)
 
       # Scale embedding by the sqrt of the hidden size
       embeddings *= self.hidden_size ** 0.5
 
-      # Create binary array of size [batch_size, length]
-      # where 1 = padding, 0 = not padding
-      padding = model_utils.get_padding(x)
-
-      # Set all padding embedding values to 0
-      embeddings *= tf.expand_dims(1 - padding, -1)
       return embeddings
+
 
   def linear(self, x):
     """Computes logits by running x through a linear layer.
--- a/official/transformer/model/model_params.py
+++ b/official/transformer/model/model_params.py
@@ -15,45 +15,64 @@
 """Defines Transformer model parameters."""
 
 
-class TransformerBaseParams(object):
-  """Parameters for the base Transformer model."""
-  # Input params
-  batch_size = 2048  # Maximum number of tokens per batch of examples.
-  max_length = 256  # Maximum number of tokens per example.
+BASE_PARAMS = dict(
+    # Input params
+    default_batch_size=2048,  # Maximum number of tokens per batch of examples.
+    default_batch_size_tpu=32768,
+    max_length=256,  # Maximum number of tokens per example.
 
-  # Model params
-  initializer_gain = 1.0  # Used in trainable variable initialization.
-  vocab_size = 33708  # Number of tokens defined in the vocabulary file.
-  hidden_size = 512  # Model dimension in the hidden layers.
-  num_hidden_layers = 6  # Number of layers in the encoder and decoder stacks.
-  num_heads = 8  # Number of heads to use in multi-headed attention.
-  filter_size = 2048  # Inner layer dimensionality in the feedforward network.
+    # Model params
+    initializer_gain=1.0,  # Used in trainable variable initialization.
+    vocab_size=33708,  # Number of tokens defined in the vocabulary file.
+    hidden_size=512,  # Model dimension in the hidden layers.
+    num_hidden_layers=6,  # Number of layers in the encoder and decoder stacks.
+    num_heads=8,  # Number of heads to use in multi-headed attention.
+    filter_size=2048,  # Inner layer dimension in the feedforward network.
 
-  # Dropout values (only used when training)
-  layer_postprocess_dropout = 0.1
-  attention_dropout = 0.1
-  relu_dropout = 0.1
+    # Dropout values (only used when training)
+    layer_postprocess_dropout=0.1,
+    attention_dropout=0.1,
+    relu_dropout=0.1,
 
-  # Training params
-  label_smoothing = 0.1
-  learning_rate = 2.0
-  learning_rate_decay_rate = 1.0
-  learning_rate_warmup_steps = 16000
+    # Training params
+    label_smoothing=0.1,
+    learning_rate=2.0,
+    learning_rate_decay_rate=1.0,
+    learning_rate_warmup_steps=16000,
 
-  # Optimizer params
-  optimizer_adam_beta1 = 0.9
-  optimizer_adam_beta2 = 0.997
-  optimizer_adam_epsilon = 1e-09
+    # Optimizer params
+    optimizer_adam_beta1=0.9,
+    optimizer_adam_beta2=0.997,
+    optimizer_adam_epsilon=1e-09,
 
-  # Default prediction params
-  extra_decode_length = 50
-  beam_size = 4
-  alpha = 0.6  # used to calculate length normalization in beam search
+    # Default prediction params
+    extra_decode_length=50,
+    beam_size=4,
+    alpha=0.6,  # used to calculate length normalization in beam search
 
+    # TPU specific parameters
+    use_tpu=False,
+    static_batch=False,
+    allow_ffn_pad=True,
+)
 
-class TransformerBigParams(TransformerBaseParams):
-  """Parameters for the big Transformer model."""
-  batch_size = 4096
-  hidden_size = 1024
-  filter_size = 4096
-  num_heads = 16
+BIG_PARAMS = dict(BASE_PARAMS)
+BIG_PARAMS.update(dict(
+    default_batch_size=4096,
+
+    # default batch size is smaller than for BASE_PARAMS due to memory limits.
+    default_batch_size_tpu=16384,
+
+    hidden_size=1024,
+    filter_size=4096,
+    num_heads=16,
+))
+
+TINY_PARAMS = dict(BASE_PARAMS)
+TINY_PARAMS.update(dict(
+    default_batch_size=1024,
+    default_batch_size_tpu=1024,
+    hidden_size=32,
+    num_heads=4,
+    filter_size=256,
+))
--- a/official/transformer/transformer_main.py
+++ b/official/transformer/transformer_main.py
@@ -22,6 +22,7 @@ from __future__ import absolute_import
 from __future__ import division
 from __future__ import print_function
 
+import functools
 import os
 import tempfile
 
@@ -39,7 +40,9 @@ from official.transformer.model import model_params
 from official.transformer.model import transformer
 from official.transformer.utils import dataset
 from official.transformer.utils import metrics
+from official.transformer.utils import schedule
 from official.transformer.utils import tokenizer
+from official.utils.accelerator import tpu as tpu_util
 from official.utils.flags import core as flags_core
 from official.utils.logs import hooks_helper
 from official.utils.logs import logger
@@ -47,8 +50,9 @@ from official.utils.misc import model_helpers
 
 
 PARAMS_MAP = {
-    "base": model_params.TransformerBaseParams,
-    "big": model_params.TransformerBigParams,
+    "tiny": model_params.TINY_PARAMS,
+    "base": model_params.BASE_PARAMS,
+    "big": model_params.BIG_PARAMS,
 }
 DEFAULT_TRAIN_EPOCHS = 10
 BLEU_DIR = "bleu"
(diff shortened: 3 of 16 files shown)


official/transformer/model/model_params.py
```diff
@@ -50,10 +53,17 @@ class TransformerBaseParams(object):
   beam_size = 4
   alpha = 0.6  # used to calculate length normalization in beam search
 
+  # TPU specific parameters
+  use_tpu = False
+  static_batch = False
+  no_ffn_pad = False
```
karmel: The double negative is confusing here. Can we just do ffn_pad = True?

official/transformer/transformer_main.py
```diff
@@ -74,32 +77,107 @@ def model_fn(features, labels, mode, params):
     # When in prediction mode, the labels/targets is None. The model output
     # is the prediction
     if mode == tf.estimator.ModeKeys.PREDICT:
+      if params["use_tpu"]:
+        raise NotImplementedError("prediction is not yet supported.")
```
karmel: nit: Capitalize appropriately, and fully specify "on TPUs"

official/transformer/transformer_main.py
```diff
@@ -74,32 +77,107 @@ def model_fn(features, labels, mode, params):
...
+        logits, targets, params["label_smoothing"], params["vocab_size"])
     loss = tf.reduce_sum(xentropy) / tf.reduce_sum(weights)
 
     # Save loss as named tensor that will be logged with the logging hook.
     tf.identity(loss, "cross_entropy")
 
     if mode == tf.estimator.ModeKeys.EVAL:
+      if params["use_tpu"]:
+        metric_fn = functools.partial(metrics.get_eval_metrics, params=params)
+        eval_metrics = (metric_fn, [logits, labels])
```
karmel: Comment what is happening here.

official/transformer/transformer_main.py
```diff
@@ -74,32 +77,107 @@ def model_fn(features, labels, mode, params):
     # When in prediction mode, the labels/targets is None. The model output
     # is the prediction
     if mode == tf.estimator.ModeKeys.PREDICT:
+      if params["use_tpu"]:
+        raise NotImplementedError("prediction is not yet supported.")
       return tf.estimator.EstimatorSpec(
           tf.estimator.ModeKeys.PREDICT,
           predictions=output)
 
     logits = output
```
karmel: nit: pre-existing, but why not just have the var name of output be logits above, and avoid this rename?

official/transformer/transformer_main.py
```diff
@@ -74,32 +77,107 @@ def model_fn(features, labels, mode, params):
...
+            mode=mode, loss=loss, predictions={"predictions": logits},
+            eval_metrics=eval_metrics)
       return tf.estimator.EstimatorSpec(
           mode=mode, loss=loss, predictions={"predictions": logits},
           eval_metric_ops=metrics.get_eval_metrics(logits, labels, params))
     else:
-      train_op = get_train_op(loss, params)
+      train_op, metric_dict = get_train_op(loss, params)
+
+      # Epochs can be quite long. This give some intermediate information
```
karmel: nit: gives

official/transformer/transformer_main.py
```diff
@@ -74,32 +77,107 @@ def model_fn(features, labels, mode, params):
...
+        return tf.contrib.tpu.TPUEstimatorSpec(
+            mode=mode, loss=loss, train_op=train_op,
+            host_call=construct_host_call(metric_dict, params)
+        )
+      record_scalars(metric_dict)
       return tf.estimator.EstimatorSpec(mode=mode, loss=loss, train_op=train_op)
 
 
+def construct_host_call(metric_dict, params):
+  def host_call_fn(gs, ls, lr, gn):
```
karmel: Move to utils?

official/transformer/transformer_main.py
```diff
@@ -30,6 +31,7 @@
 from absl import app as absl_app
 from absl import flags
 import tensorflow as tf
+from tensorflow.contrib import summary as contrib_summary
```
karmel: This is exposed directly.

```
>>> tf.contrib.summary
<module 'tensorflow.contrib.summary.summary' from '/usr/local/google/home/karmel/.local/lib/python2.7/site-packages/tensorflow/contrib/summary/summary.pyc'>
```

official/transformer/transformer_main.py
```diff
@@ -74,32 +77,107 @@ def model_fn(features, labels, mode, params):
...
+      gs: `Tensor with shape `[batch]` for the global_step
+      lr: `Tensor` with shape `[batch]` for the learning_rate.
+      gn: `Tensor` with shape `[batch]` for the gradient_norm.
+
+    Returns:
+      List of summary ops to run on the CPU host.
+    """
+    gs = gs[0]
+    with contrib_summary.create_file_writer(params["model_dir"]).as_default():
+      with contrib_summary.always_record_summaries():
```
karmel:
```
>>> tf.contrib.summary.always_record_summaries
<function always_record_summaries at 0x7f6866e12758>
```

official/transformer/transformer_main.py
```diff
@@ -74,32 +77,107 @@ def model_fn(features, labels, mode, params):
...
+  ls_t = tf.reshape(metric_dict["minibatch_loss"], [1])
+  lr_t = tf.reshape(metric_dict["learning_rate"], [1])
+  gn_t = tf.reshape(metric_dict["global_norm/gradient_norm"], [1])
+
+  return host_call_fn, [gs_t, ls_t, lr_t, gn_t]
+
+
+def record_scalars(metric_dict):
+  for key, value in metric_dict.items():
+    tf.summary.scalar(name=key, tensor=value)
```
karmel: Incidentally, contrib summary is the future. Can the main loop share the same contrib summary code?

official/transformer/transformer_main.py
```diff
@@ -74,32 +77,107 @@ def model_fn(features, labels, mode, params):
...
+    any Tensors in the rest of the `model_fn`. To pass Tensors from the
+    model to the `metric_fn`, provide as part of the `host_call`. See
+    https://www.tensorflow.org/api_docs/python/tf/contrib/tpu/TPUEstimatorSpec
+    for more information.
+
+    Arguments should match the list of `Tensor` objects passed as the second
+    element in the tuple passed to `host_call`.
+
+    Args:
+      gs: `Tensor with shape `[batch]` for the global_step
```
karmel: nit: may as well spell these out?

official/transformer/transformer_main.py
```diff
@@ -116,26 +194,28 @@ def get_learning_rate(learning_rate, hidden_size, learning_rate_warmup_steps):
...
     tf.identity(learning_rate, "learning_rate")
-    # Save learning rate value to TensorBoard summary.
-    tf.summary.scalar("learning_rate", learning_rate)
 
     return learning_rate
 
 
 def get_train_op(loss, params):
-  """Generate training operation that updates variables based on loss."""
+  """Generate training op and summary dict."""
```
karmel: Why make this comment less explanatory? Also, update func name to reflect new duties.

official/transformer/utils/schedule.py
```diff
@@ -0,0 +1,140 @@
+# Copyright 2018 The TensorFlow Authors. All Rights Reserved.
```
karmel: Much of this logic seems very specific to Transformer. Why in utils? The TPU logic should be separated out, to make things more clear.

official/transformer/transformer_main.py
```diff
@@ -269,18 +328,26 @@ def train_schedule(
...
     # Train the model for single_iteration_train_steps or until the input fn
     # runs out of examples (if single_iteration_train_steps is None).
     estimator.train(
-        dataset.train_input_fn, steps=single_iteration_train_steps,
+        dataset.train_input_fn,
+        steps=schedule_manager.single_iteration_train_steps,
         hooks=train_hooks)
 
-    eval_results = estimator.evaluate(dataset.eval_input_fn)
+    schedule_manager.post_train()
```
karmel: This is mysterious. I would prefer this belong to a TPU util that handle this operation, rather than a generic magic-box.

robieta:
@frankchn @saeta BUMP

official/transformer/transformer_main.py
```diff
@@ -269,18 +294,28 @@ def train_schedule(
       train_eval_iterations = INF
```
k-w-w: This line should modify `schedule_manager.train_eval_iterations`

official/transformer/model/embedding_layer.py
```diff
@@ -53,19 +56,25 @@ def call(self, x):
...
 
       # Scale embedding by the sqrt of the hidden size
       embeddings *= self.hidden_size ** 0.5
 
-      # Create binary array of size [batch_size, length]
-      # where 1 = padding, 0 = not padding
-      padding = model_utils.get_padding(x)
-
-      # Set all padding embedding values to 0
-      embeddings *= tf.expand_dims(1 - padding, -1)
```
k-w-w: Is this line added back in somewhere?

official/transformer/transformer_main.py
```diff
@@ -69,37 +73,71 @@ def model_fn(features, labels, mode, params):
...
     model = transformer.Transformer(params, mode == tf.estimator.ModeKeys.TRAIN)
 
-    output = model(inputs, targets)
+    logits = model(inputs, targets)
 
     # When in prediction mode, the labels/targets is None. The model output
     # is the prediction
     if mode == tf.estimator.ModeKeys.PREDICT:
+      if params["use_tpu"]:
+        raise NotImplementedError("Prediction is not yet supported on TPUs.")
```
k-w-w: Should this `NotImplementedError` be raised before the line `logits = model(inputs, targets)`?

official/utils/accelerator/tpu.py
```diff
@@ -0,0 +1,108 @@
...
+
+import tensorflow as tf
+from tensorflow.python.framework import dtypes
+from tensorflow.python.framework import ops
+from tensorflow.python.ops import array_ops
+from tensorflow.python.ops import embedding_ops
+from tensorflow.python.ops import math_ops
+
+
+DRY_RUN = "dry_run"
```
frankchn: I wonder if the `dry_run` name is actually necessary. I think passing in `--tpu=local` and `--use_tpu=False` has the same effect without the extra code needed to check.

robieta: Yeah, I'm not sure that how I put this in is the cleanest. (Truth be told it was for my own debugging as much as anything else.) The difference between this and say, mnist_tpu.py is that use_tpu=False is not just a debug flag in Transformer since the TPU and non-TPU loops are the same. So I want use_tpu to be True even during a dry run. I like `--tpu=local` better than `--tpu=dry_run` though.

frankchn: Yeah, part of the reason of using `--tpu=local` is that I think you can remove the `if tpu == "dry_run": don't use cluster_resolver` stuff, since cluster resolver accepts an empty string or "local" and won't try to resolve those to a TPU name.

frankchn: Otherwise I think it looks OK after resolving karmel and k-w-w's comments.

official/transformer/utils/schedule.py
robieta: It's in `official/transformer/utils`, not `official/utils`. That should be fine, no?

official/transformer/transformer_main.py
robieta: Happily this is all moot, as infra has fixed the underlying issue. Yay!

official/transformer/model/embedding_layer.py
robieta: `embeddings *= tf.expand_dims(mask, -1)` is equivalent, right?

official/transformer/transformer_main.py
robieta: It should be fine here. Just defining the ops isn't an issue, and I like to keep it close to the non-TPU PREDICT case.

robieta:
@karmel @k-w-w PTAL

official/transformer/model/embedding_layer.py
```diff
@@ -53,19 +68,25 @@ def call(self, x):
         locations of the padding tokens in x.
     """
     with tf.name_scope("embedding"):
-      embeddings = tf.gather(self.shared_weights, x)
+      # Create binary mask of size [batch_size, length]
+      mask = tf.to_float(tf.not_equal(x, 0))
+
+      if self.method == "gather" and False:
```
k-w-w: Might want to remove the `and False` here

official/transformer/model/embedding_layer.py
k-w-w: just realized it was in the code for method == "gather". This should be applied to the matmul embeddings as well.

official/transformer/model/embedding_layer.py
robieta: ...oops.

robieta:
Embeddings should be fixed. Thanks for catching that.

official/transformer/translate.py
```diff
@@ -167,11 +167,11 @@ def main(unused_argv):
       os.path.join(FLAGS.data_dir, FLAGS.vocab_file))
 
   # Set up estimator and params
-  params = transformer_main.PARAMS_MAP[FLAGS.param_set]
-  params.beam_size = _BEAM_SIZE
-  params.alpha = _ALPHA
-  params.extra_decode_length = _EXTRA_DECODE_LENGTH
-  params.batch_size = _DECODE_BATCH_SIZE
+  params = transformer_main.PARAMS_MAP[FLAGS.param_set]().dict
```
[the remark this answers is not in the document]
k-w-w: This may have been overlooked

official/transformer/utils/dataset.py
```diff
@@ -67,6 +67,16 @@
 _BOUNDARY_SCALE = 1.1
 
 
+NUM_EXAMPLES = {
+    tf.estimator.ModeKeys.TRAIN: 4572160,
+    # TODO(robieta): Find out why the actual number is slightly lower.
```
karmel: nit: Todos.

Also, @k-w-w , do you know?

official/utils/accelerator/tpu.py
```diff
@@ -0,0 +1,124 @@
...
+
+
+# "local" is a magic word in the TPU cluster resolver; it informs the resolver
+# to use the local CPU as the compute device. This is useful for testing and
+# debugging; the code flow is ostensibly identical, but without the need to
+# actually have a TPU on the other end.
+LOCAL = "local"
+
+
+# TODO(robieta): See if some version of this can be rolled into TPUEstimator.
```
karmel: Ditto. Perhaps create a bug instead.

official/transformer/transformer_main.py
```diff
@@ -403,54 +447,105 @@ def _check_bleu_files(flags_dict):
         tf.gfile.Exists(flags_dict["bleu_ref"]),
         tf.gfile.Exists(vocab_file_path)])
 
+  flags_core.require_cloud_storage(["data_dir", "model_dir"])
```
karmel: I must be confused about how this is working, but where does this actually check if TPU=true? In the func definition below, it seems that all directories would be required to be on cloud, regardless of TPU or not?

official/utils/logs/hooks_helper.py
```diff
@@ -97,6 +104,9 @@ def get_profiler_hook(save_steps=1000, **kwargs):  # pylint: disable=unused-argu
   Returns:
     Returns a ProfilerHook that writes out timelines that can be loaded into
     profiling tools like chrome://tracing.
+
+  Raises:
+    NotImplemetedError if use_tpu is True
```
karmel: Erroneous comment?

official/transformer/utils/dataset.py
k-w-w: Yes, I believe the real number appears smaller because some examples are filtered out by the dataset (where the sequences have lengths > max_length)

official/transformer/transformer_main.py
robieta: Nope, you're absolutely correct. Thanks for catching this.

PR #4298, opened 2018-05-17
Yet another ResNet

robieta:


research/resnet/resnet_model.py
```diff
@@ -35,6 +35,10 @@
                      'relu_leakiness, optimizer')
 
 
+tf.logging.warning("models/research/resnet is deprecated. "
+                   "Please use models/official/resnet instead.")
+
```
karmel: Have you tried to run this, and it still works?

robieta:
Yes, the logging works. Although it occurs to me that *we're* actually the "another resnet".

PR #4308, opened 2018-05-18
Update the benchmark run schema with the new columns.

qlzh727:
This is based on the discussion with Toby and existing tf_cnn_benchmark.

### DIFF
--- a/official/benchmark/datastore/schema/benchmark_run.json
+++ b/official/benchmark/datastore/schema/benchmark_run.json
@@ -18,6 +18,12 @@
     "type": "TIMESTAMP"
   },
   {
+    "description": "The unique name for a test by the combination of key parameters, eg batch size, num of GPU, etc. It is hardware independent.",
+    "mode": "NULLABLE",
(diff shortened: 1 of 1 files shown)


official/benchmark/datastore/schema/benchmark_run.json
```diff
@@ -17,6 +17,12 @@
     "name": "run_date",
     "type": "TIMESTAMP"
   },
+  {
+    "description": "The unique name for a test by the combination of key parameters, eg batch size, num of GPU, etc. It is hardware independent.",
+    "mode": "NULLABLE",
```
karmel: Is this auto-generated or human-set? Presumably a single test_id can have multiple run_ids?

qlzh727: Actually both. We can set the unique name from commandline if we want. If the value is not set, then our lib can populate the value based on few important hyper param. Will add more comment in the code when we define this flag and lib.

PR #4333, opened 2018-05-21
Adding a google cloud container builder spec to build cpu and gpu based images

vishh:
To setup continuous builds of container images that will includes all the models in this repo via [google cloud container builder](https://cloud.google.com/container-builder/), this PR introduces a container builder configuration file that explicitly specified the images that will be built for every commit to this branch and more importantly the need for a larger size VM to run these builds in the background.

To submit builds manually to container builder run the following command:
```shell
$ cd official 
$ gcloud container builds submit --config cloudbuild.yaml . --project <your-gcp-project> --substitutions=COMMIT_SHA=$(git rev-parse HEAD)
```

Builds will be automated via container builder and you can list the images available periodically as follows:
```shell
$ gcloud container images list-tags gcr.io/tensorflow/models
$ gcloud container images list-tags gcr.io/tensorflow/models-gpu
```

### DIFF
--- a/official/cloudbuild.yaml
+++ b/official/cloudbuild.yaml
@@ -1,0 +1,29 @@
+steps:
+- name: 'gcr.io/cloud-builders/docker'
+  args: ['build',
+  '-t', 'gcr.io/$PROJECT_ID/models-gpu:$COMMIT_SHA',
+  '-t', 'gcr.io/$PROJECT_ID/models-gpu:latest',
+  '-f', 'Dockerfile.gpu',
+  '.'
+  ]
+  
+- name: 'gcr.io/cloud-builders/docker'
+  args: ['build',
+  '-t', 'gcr.io/$PROJECT_ID/models:$COMMIT_SHA',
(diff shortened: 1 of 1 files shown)


vishh:
@karmel This is needed to run gpu builds automatically.

robieta:
Can you put a blurb at the top about what this is?

karmel:
This is great, @vishh , thanks. It is hard for me to review the actual config, but other than the request for docs from @robieta above, LGTM. Is there a link/demo/doc we can reference as well that shows this in action?

vishh:
@karmel @robieta I have updated the PR description. PTAL.

tfboyd:
Cleaning up all PRs targeting official.  This is 6+ months old and never merged.  We can bring this back and if so we may want to add to a README.md as a user would find this .yaml file at root with no description and wonder what it is.  

I would love to have a continuous builder for containers for different GPU configs via docker without wasting my GPU test servers cycles.  Have not had time to dig into the tool.

PR #4336, opened 2018-05-22
resolve 'float' to 'int' conversion, StringIO import, conacat str + byte

j143:
Hi, when running the training script I've encountered.
1. `StringIO` import error.
2. `float' to 'int' conversion error.
3. concatenating bytes to str.

I tried a fix and successfully trained the model.

Thanks.


j143:
Thank you. :)

PR #4340, opened 2018-05-22
Add minor performance improvements to resnet input pipeline

guptapriya:
Added the following changes to improve input pipeline.

1. Remove usage of one_hot labels (~2% improvement)
2. Add drop_remainder to batching. This helps in input shape being determined ahead of time which helps improve performance. (~2% improvement)
3. Use parallel_interleave for imagenet dataset reading. (~3% improvement)

The improvements were observed when testing with 8 V100 GPUs on DGX-1 using mirrored strategy (fp16, resnet v2). Also saw improvements in GCE V100. No improvement observed on K80s but that is expected since they are not bottlenecked on input.


guptapriya:
This is just a re-do of https://github.com/tensorflow/models/pull/4247 which I messed up.

PR #4341, opened 2018-05-22
Script for quantizing ResNet50

parvizp:
I find the example scripts very useful for demonstrating end-to-end flows. Are you interested in contributions like this for quantizing models and converting to quantized TF-Lite models via TOCO?


parvizp:
If anyone is interested, you can find the quantized TF-Lite model here: https://drive.google.com/open?id=1b2tmDFzyU5NwRzxQmJaO7v-TrjipyGXT

PR #4343, opened 2018-05-22
Add unit test, official flags, and benchmark logs for recommendation model

yhliang2018:
Hi All,

I have add official flags and benchmark logs for the recommendation model. A simple unit test for dataset.py is also added based on previous comments. As dataset.py needs to read csv files, I add a folder "unittest_data" for the csv files.


official/recommendation/ncf_main.py
```diff
@@ -91,14 +101,18 @@ def _get_ndcg(ranklist, true_item):
       return math.log(2) / math.log(ranklist.index(true_item) + 2)
     return 0
 
+  # Return estimator's last checkpoint as global_step for estimator
+  def _get_global_step(estimator):
+    return int(estimator.latest_checkpoint().split("-")[-1])
```
karmel: Hmm. This is fragile and unreliable. Can you try estimator.get_variable_value(tf.GraphKeys.GLOBAL_STEP)? (Sorry, that's from memory, so may not be exactly the right syntax.) Or actually reading in the checkpoint and grabbing the global step?

official/recommendation/ncf_main.py
```diff
@@ -201,100 +210,135 @@ def main(_):
...
+      name="num_factors", short_name="nf", default=8,
+      help=flags_core.help_wrap("The Embedding size of MF model."))
+
+  flags.DEFINE_list(
+      name="layers", short_name="ly", default=[64, 32, 16, 8],
+      help=flags_core.help_wrap("The size of hidden layers for MLP."))
+
+  flags.DEFINE_float(
+      name="mf_regularization", short_name="mr", default=0.0,
+      help=flags_core.help_wrap("The Regularization for MF embeddings."))
```
karmel: nit: Regularization what? Factor? Feels like it's missing a noun. Also, maybe some advice on reasonable values, defaults, etc?

official/recommendation/ncf_main.py
```diff
@@ -201,100 +210,135 @@ def main(_):
...
+      batch_size=256,
+      hooks="ProfilerHook")
+
+  # Add ncf-specific flags
+  flags.DEFINE_enum(
+      name="dataset", short_name="ds", default="ml-1m",
+      enum_values=["ml-1m", "ml-20m"], case_sensitive=False,
+      help=flags_core.help_wrap(
+          "Dataset to be trained and evaluated. Two datasets are "
+          "available for now: ml-1m and ml-20m."))
```
karmel: There should be a way with flags to display the choices and defaults for each of these without hard-coding them.

official/recommendation/ncf_main.py
```diff
@@ -201,100 +210,135 @@ def main(_):
...
+
+  flags.DEFINE_integer(
+      name="num_factors", short_name="nf", default=8,
```
karmel: See above discussion re:short names

official/recommendation/ncf_main.py
```diff
@@ -201,100 +210,135 @@ def main(_):
...
+      help=flags_core.help_wrap("The Embedding size of MF model."))
+
+  flags.DEFINE_list(
+      name="layers", short_name="ly", default=[64, 32, 16, 8],
+      help=flags_core.help_wrap("The size of hidden layers for MLP."))
```
karmel: This should be a list? How does one enter the list? Comma-separated ints? Please define expectations for the user, display default.

official/recommendation/ncf_main.py
```diff
@@ -201,100 +210,135 @@ def main(_):
...
+      name="learning_rate", short_name="lr", default=0.001,
+      help=flags_core.help_wrap("The learning rate."))
+
+  flags.DEFINE_float(
+      name="hr_threshold", short_name="ht", default=None,
+      help=flags_core.help_wrap(
+          "If passed, training will stop when the evaluation metric HR is "
+          "greater than or equal to hr_threshold. For dataset ml-1m, the "
+          "desired hr_threshold is 0.68; For dataset ml-20m, the threshold can "
+          "be set as 0.95."))
```
karmel: Are these set auto-magically? They should be, and that should be clear from the help text as well.

official/recommendation/data_download.py
```diff
@@ -335,14 +339,23 @@ def _progress(count, block_size, total_size):
     parse_file_to_csv(FLAGS.data_dir, FLAGS.dataset)
 
 
+def define_data_download_flags():
+  """Add flags specifying data download arguments."""
+  flags.DEFINE_string(
+      name="data_dir", short_name="dd", default="/tmp/movielens-data/",
```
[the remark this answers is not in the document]
yhliang2018: I will remove the short_name for now. We can always add it back if necessary. :)

official/recommendation/ncf_main.py
robieta: absl will automatically display allowed values for enums, so it can be removed altogether.

yhliang2018:
Thank you for the comments, Karmel, especially the regularization one, which helps me identify a model layer bug. Thanks a lot!

PR #4351, opened 2018-05-23
Use SyntheticDataset for creating synthetic input

guptapriya:
`SyntheticDataset` allows to easily build a performant synthetic input pipeline. Use that in resnet examples.


guptapriya:
Do not review yet, this is just example PR to showcase usage of the new API. Will send for review once the API is in master.

PR #4357, opened 2018-05-24
Update object_detection_tutorial.ipynb

nithishdivakar:
When number of objects are more than uint8's range, the earlier code crates problems. This is a fix.


nithishdivakar:
I signed it!

PR #4361, opened 2018-05-24
Intel-MKL: add inter/intra-op and performance tuning parameters specific to MKL

jinghuangintel:
In this PR, We added following parameters to improve the training and evaluation out-of-box experiences on CPU and make it easier for users to control the parallelism on CPU for object detection API.
1) inter_op and intra_op to control the parallelism.
2) performance tuning parameters specific to MKL (kmp_block_time and kmp_affinity) if MKL is enabled.

### DIFF
--- a/research/object_detection/eval.py
+++ b/research/object_detection/eval.py
@@ -78,6 +78,27 @@ flags.DEFINE_string('model_config_path', '',
 flags.DEFINE_boolean('run_once', False, 'Option to only run a single pass of '
                      'evaluation. Overrides the `max_evals` parameter in the '
                      'provided config.')
+
+# Performance tuning specific to MKL
+flags.DEFINE_boolean('mkl', False, 'If true, set MKL environment variables.')
+flags.DEFINE_integer('kmp_blocktime', 1,
+                     'The time, in milliseconds, that a thread should wait, '
+                     'after completing the execution of a parallel region, '
+                     'before sleeping')
+flags.DEFINE_string('kmp_affinity', 'granularity=fine,verbose,compact,1,0',
+                    'Restricts execution of certain threads (virtual execution '
+                    'units) to a subset of the physical processing units in a '
+                    'multiprocessor computer.')
+flags.DEFINE_integer('kmp_settings', 1,
+                     'If set to 1, MKL settings will be printed.')
+
+flags.DEFINE_integer('num_intra_threads', 0,
+                     'Number of threads to use for intra-op parallelism. If '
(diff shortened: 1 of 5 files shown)


research/object_detection/eval.py
```diff
@@ -78,6 +78,27 @@
 flags.DEFINE_boolean('run_once', False, 'Option to only run a single pass of '
                      'evaluation. Overrides the `max_evals` parameter in the '
                      'provided config.')
+
+# Performance tuning specific to MKL
```
derekjchow: Strictly speaking, are these flags needed? It seems to me that they just set environment variables. Can the user set these variables before running the eval script?

research/object_detection/eval_util.py
```diff
@@ -206,6 +206,8 @@ def _run_checkpoint_once(tensor_dict,
                          master='',
                          save_graph=False,
                          save_graph_dir='',
+                         inter_op=0,
```
derekjchow: Please update doc string with these variables.

research/object_detection/train.py
```diff
@@ -82,6 +82,26 @@
 flags.DEFINE_string('model_config_path', '',
                     'Path to a model_pb2.DetectionModel config file.')
 
+# Performance tuning specific to MKL
```
derekjchow: Same comment about eval job. Could these be set outside of the python script?

research/object_detection/trainer.py
```diff
@@ -214,7 +214,9 @@ def train(create_tensor_dict_fn,
           worker_job_name,
           is_chief,
           train_dir,
-          graph_hook_fn=None):
+          graph_hook_fn=None,
+          inter_op=0,
```
derekjchow: Please update doc string.

research/object_detection/eval.py
jinghuangintel: @derekjchow thanks for your useful comments! yeah, those are environment variables and parallelism control parameters, very similar to https://github.com/tensorflow/benchmarks/blob/master/scripts/tf_cnn_benchmarks/benchmark_cnn.py (line 178, 323),  we are trying to improve user's OOB experiences while they use MKL.

research/object_detection/eval_util.py
jinghuangintel: thanks for the comments, i have added the inter_op/intra_op in the doc string already.

jinghuangintel:
@derekjchow thanks for your comments. i have also squashed multiple PR into one single PR already, let me know if you have any further suggestions. thanks.

PR #4376, opened 2018-05-25
WIP: Strawman for dependency managed tensors in hooks.

robieta:
Not to be merged.

This is an example of how if hooks could access the train op, they could be written in a way that ensures deterministic, correct behavior.

### DIFF
--- a/official/utils/logs/hooks.py
+++ b/official/utils/logs/hooks.py
@@ -73,13 +73,29 @@ class ExamplesPerSecondHook(tf.train.SessionRunHook):
     self._total_steps = 0
     self._batch_size = batch_size
     self._warm_steps = warm_steps
+    self._train_op = None
+
+  def specify_train_op(self, train_op):
+    self._train_op = train_op
 
   def begin(self):
     """Called once before using the session to check global step."""
-    self._global_step_tensor = tf.train.get_global_step()
-    if self._global_step_tensor is None:
+    if self._train_op is None:
+      raise ValueError("train op not passed to hook.")
+
+    global_step_tensor = tf.train.get_global_step()
+
+    if global_step_tensor is None:
       raise RuntimeError(
           "Global step should be created to use StepCounterHook.")
+
+    # One nice thing about this pattern is that it isn't limited to global
+    # step tensor; and tensor can easily be forcibly evaluated AFTER the train
+    # step has occured.
+    with tf.control_dependencies([self._train_op]):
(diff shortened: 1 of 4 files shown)


official/utils/logs/hooks.py
```diff
@@ -73,14 +73,30 @@ def __init__(self,
     self._total_steps = 0
     self._batch_size = batch_size
     self._warm_steps = warm_steps
+    self._train_op = None
+
+  def specify_train_op(self, train_op):
```
qlzh727: I would prefer to define it as a property so that it gets both getter and setter method.

official/utils/misc/model_helpers.py
```diff
@@ -53,3 +53,41 @@ def past_stop_threshold(stop_threshold, eval_metric):
...
+  the estimator provides the train_op to hooks, which allows them to structure
+  the dependencies of their calls.
+  """
+  def _train_with_estimator_spec(self, estimator_spec, worker_hooks, hooks,
+                                 global_step_tensor, saving_listeners):
+
+    for h in hooks:
+      # This check is for backward compatibility, just to demonstrate that it
+      # wouldn't be an issue rolling it back into estimator.
+      if (hasattr(h, "specify_train_op") and callable(h.specify_train_op) and
```
qlzh727: This can be simplified as hasattr(h, "train_op")

official/utils/logs/hooks.py
```diff
@@ -73,14 +73,30 @@ def __init__(self,
...
+    self._train_op = train_op
 
   def begin(self):
     """Called once before using the session to check global step."""
-    self._global_step_tensor = tf.train.get_global_step()
-    if self._global_step_tensor is None:
+    if self._train_op is None:
```
qlzh727: Since you have a if check for estimator.train_op is None below, it seems that it is possible for estimator to not have a train_op. Should we handle it more gracefully, eg, backoff to the original behavior if the train_op is not specified.

official/utils/logs/hooks.py
```diff
@@ -73,14 +73,30 @@ def __init__(self,
     self._total_steps = 0
     self._batch_size = batch_size
     self._warm_steps = warm_steps
+    self._train_op = None
```
qlzh727: I think this might be applicable for all the hooks that requires accurate value for global step tensor. So probably worthwhile to push it to run_session_hook itself.

robieta:
We have determined that this is unnecessary.

PR #4378, opened 2018-05-26
Update running_pets.md

fajarnuha:
tensorboard logdir should be in train directory


fajarnuha:
I agree

pkulzc:
This is not necessary since tensorboard will automatically pick up everything (train/eval) in that folder.

PR #4389, opened 2018-05-28
Catch errors during mkdirs, prevent duplicated download

sbrodehl:
This PR catches errors during the directory creating, which previously resulted in executing the whole script again.
Now, if the script was previously interrupted the script will continue if the needed directories already exist.

Furthermore, the `ILSVRC2012_bbox_train_v2.tar.gz` is now only downloaded once, in the case where no `BASE_URL_CHANGE` was neccessary.


vincentvanhoucke:
I'll let @shlens approve this one.

PR #4394, opened 2018-05-29
Update README.md

wangershi:
In '*_test.py', no function set the GPU device, default is 0. So if the GPU 0 is used by others and the memory is full, there will an exception like
'''failed to allocate 6.04G (6487007744 bytes) from device: CUDA_ERROR_OUT_OF_MEMORY'''
so there is a necessity to set the GPU device(and it's optional).


wangershi:
I signed it!

PR #4402, opened 2018-05-29
Record the status for a benchmark run.

qlzh727:
The status is recorded into a separate bigquery table, due to the limit of streaming uploading of run that block the record from being updated. The new status table is only inserted/updated via standard SQL query.

### DIFF
--- a/official/benchmark/benchmark_uploader.py
+++ b/official/benchmark/benchmark_uploader.py
@@ -27,6 +27,7 @@ from __future__ import print_function
 import json
 
 from google.cloud import bigquery
+from google.cloud import exceptions
 
 import tensorflow as tf
 
@@ -139,7 +140,10 @@ class BigQueryUploader(object):
              "(run_id, status) "
              "VALUES('{rid}', '{status}')").format(
                  ds=dataset_name, tb=table_name, rid=run_id, status=run_status)
-    self._bq_client.query(query=query).result()
+    try:
+      self._bq_client.query(query=query).result()
+    except exceptions.GoogleCloudError as e:
+      tf.logging.error("Failed to insert run status: %s", e)
 
   def update_run_status(self, dataset_name, table_name, run_id, run_status):
     """Update the run status in in Bigquery run status table."""
@@ -147,4 +151,7 @@ class BigQueryUploader(object):
              "SET status = '{status}' "
              "WHERE run_id = '{rid}'").format(
                  ds=dataset_name, tb=table_name, status=run_status, rid=run_id)
-    self._bq_client.query(query=query).result()
+    try:
+      self._bq_client.query(query=query).result()
+    except exceptions.GoogleCloudError as e:
+      tf.logging.error("Failed to update run status: %s", e)
--- a/official/benchmark/benchmark_uploader.py
+++ b/official/benchmark/benchmark_uploader.py
@@ -133,10 +133,18 @@ class BigQueryUploader(object):
       tf.logging.error(
           "Failed to upload benchmark info to bigquery: {}".format(errors))
 
+  def insert_run_status(self, dataset_name, table_name, run_id, run_status):
+    """Insert the run status in to Bigquery run status table."""
+    query = ("INSERT {ds}.{tb} "
+             "(run_id, status) "
+             "VALUES('{rid}', '{status}')").format(
+                 ds=dataset_name, tb=table_name, rid=run_id, status=run_status)
+    self._bq_client.query(query=query).result()
+
   def update_run_status(self, dataset_name, table_name, run_id, run_status):
-    """Update the run status in to Bigquery run record."""
+    """Update the run status in in Bigquery run status table."""
     query = ("UPDATE {ds}.{tb} "
              "SET status = '{status}' "
              "WHERE run_id = '{rid}'").format(
                  ds=dataset_name, tb=table_name, status=run_status, rid=run_id)
-    self._bq_client.query(query=query)
+    self._bq_client.query(query=query).result()
--- a/official/benchmark/benchmark_uploader.py
+++ b/official/benchmark/benchmark_uploader.py
@@ -132,3 +132,11 @@ class BigQueryUploader(object):
     if errors:
       tf.logging.error(
           "Failed to upload benchmark info to bigquery: {}".format(errors))
+
+  def update_run_status(self, dataset_name, table_name, run_id, run_status):
+    """Update the run status in to Bigquery run record."""
+    query = ("UPDATE {ds}.{tb} "
+             "SET status = '{status}' "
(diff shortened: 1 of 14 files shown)


official/benchmark/benchmark_uploader.py
```diff
@@ -132,3 +132,19 @@ def _upload_json(self, dataset_name, table_name, json_list):
...
       tf.logging.error(
           "Failed to upload benchmark info to bigquery: {}".format(errors))
+
+  def insert_run_status(self, dataset_name, table_name, run_id, run_status):
+    """Insert the run status in to Bigquery run status table."""
+    query = ("INSERT {ds}.{tb} "
+             "(run_id, status) "
+             "VALUES('{rid}', '{status}')").format(
+                 ds=dataset_name, tb=table_name, rid=run_id, status=run_status)
+    self._bq_client.query(query=query).result()
```
karmel: Does the result need to be validated at all? Or it will raise an error on failure? In the case of an error, presumably the  whole run will fail. Is that the desired behavior?

official/benchmark/benchmark_uploader.py
```diff
@@ -132,3 +132,19 @@ def _upload_json(self, dataset_name, table_name, json_list):
...
+
+  def update_run_status(self, dataset_name, table_name, run_id, run_status):
+    """Update the run status in in Bigquery run status table."""
+    query = ("UPDATE {ds}.{tb} "
+             "SET status = '{status}' "
+             "WHERE run_id = '{rid}'").format(
+                 ds=dataset_name, tb=table_name, status=run_status, rid=run_id)
+    self._bq_client.query(query=query).result()
```
karmel: Same question as above-- how do we want to handle failure scenarios?

official/benchmark/benchmark_uploader_main.py
```diff
@@ -54,6 +54,10 @@ def main(_):
   uploader.upload_metric_file(
       flags.FLAGS.bigquery_data_set, flags.FLAGS.bigquery_metric_table, run_id,
       metric_json_file)
+  # Assume the run finished successfully before user invoke the upload script.
+  uploader.insert_run_status(
+      flags.FLAGS.bigquery_data_set, flags.FLAGS.bigquery_run_status_table,
+      run_id, "success")
```
karmel: The statuses should be constants, probably in the uploader module.

official/benchmark/benchmark_uploader_main.py
```diff
@@ -54,6 +54,10 @@ def main(_):
   uploader.upload_metric_file(
```
karmel: It would be nice if we did a three part status-- Running, Success, Failure. Before running, mark running; update to success on success, or failure on failure.

official/benchmark/benchmark_uploader_main.py
```diff
@@ -54,6 +54,10 @@ def main(_):
...
       flags.FLAGS.bigquery_data_set, flags.FLAGS.bigquery_metric_table, run_id,
       metric_json_file)
+  # Assume the run finished successfully before user invoke the upload script.
+  uploader.insert_run_status(
+      flags.FLAGS.bigquery_data_set, flags.FLAGS.bigquery_run_status_table,
+      run_id, "success")
```
karmel: In the event of run failure, the status should be updated. Can we `try, catch: update_run_status(failure), raise`? In the case of catastrophic failure (ie, machine down or sigterm), the run will be left in limbo, but I don't think we need to handle that currently, since the status is for tracking rather than transaction locking or anything particularly sensitive. We can always clear after the fact.

official/benchmark/benchmark_uploader_test.py
```diff
@@ -102,6 +102,22 @@ def test_upload_metric_file(self):
...
+                      'VALUES(\'run_id\', \'status\')')
+    self.mock_client.query.assert_called_once_with(query=expected_query)
+
+  def test_update_run_status(self):
+    self.benchmark_uploader.update_run_status(
+        'dataset', 'table', 'run_id', 'status')
+    expected_query = ('UPDATE dataset.table '
+                      'SET status = \'status\' '
+                      'WHERE run_id = \'run_id\'')
+    self.mock_client.query.assert_called_once_with(query=expected_query)
```
karmel: nit:  I think glint allows you to use double quotes when single quotes require escaping.

official/utils/flags/_benchmark.py
```diff
@@ -66,6 +66,12 @@ def define_benchmark(benchmark_log_dir=True, bigquery_uploader=True):
         help=help_wrap("The Bigquery table name where the benchmark run "
                        "information will be uploaded."))
 
+    flags.DEFINE_string(
+        name="bigquery_run_status_table", short_name="brst",
+        default="benchmark_run_status",
+        help=help_wrap("The Bigquery table name where the benchmark run "
+                       "status information will be uploaded."))
```
karmel: Rather than requiring this as an arg, can we expect the name to be some derivative of the main table name? Or take a prefix and use it for all the tables? Ends up being a lot of args in at the command line otherwise.

official/utils/logs/logger.py
```diff
@@ -246,6 +255,20 @@ def log_run_info(self, model_name, dataset_name, run_params):
...
+         self._run_id,
+         "running"))
+
+  def on_finish(self):
+    thread.start_new_thread(
+        self._bigquery_uploader.update_run_status,
+        (self._bigquery_data_set,
+         self._bigquery_run_status_table,
+         self._run_id,
+         "success"))
```
karmel: When does this actually get called? Will it have to get called explicitly in each run loop? Any way we can avoid that?

qlzh727: Adding the benchmark_context which manage the status update.

qlzh727:
Ping

official/wide_deep/wide_deep.py
```diff
@@ -254,25 +254,26 @@ def eval_input_fn():
...
-
-    for key in sorted(results):
-      tf.logging.info('%s: %s' % (key, results[key]))
-
-    benchmark_logger.log_evaluation_result(results)
-
-    if model_helpers.past_stop_threshold(
-        flags_obj.stop_threshold, results['accuracy']):
-      break
+  with logger.benchmark_context(benchmark_logger):
```
karmel: I'm struggling with this. On the one hand, it makes sense; on the other, we are adding more and more boilerplate and infrastructure into what is supposed to be TF model code just for the sake of getting streaming benchmarks out. Is there a way that is less intrusive? An official.main() that wraps some of the benchmarking code? I am open to discussion here. I just fear we are crossing the line into making ourselves less usable for our primary users.

qlzh727:
PTAL. Update the rest of the models to use the context within main.

PR #4410, opened 2018-05-30
[Diff Privacy] Update README and fix code exec wrt Python 3.6

LaRiffle:
Hi,
The README was unclear about how to get the data from MNIST and the use of Bazel.
I've tried to make it more simple, or fixed a pb of zip() linked to the python version which caused a NoVariable Exception in TF.


LaRiffle:
Thanks!

PR #4414, opened 2018-05-30
Fixing typo in Figure 3 for 1.8.1 branch

dandouthit:
"algorithms" was mis-spelled as "algorthims"


karmel:
Unless this is merged into master, the change will not propagate for future releases. It would be better to not rewrite the past, change master now, and, since non-critical, just wait until it gets into the released docs. Trying to change protected branches will lead to all sorts of discrepancies that are hard to untangle. @dandouthit - can you create a new PR against master?

MarkDaoust:
Thanks karmel. 

It's fixed in master. I just wasn't sure about your policy on patching branches. That makes it totally clear.

Thanks anyway for the fix Dan.

PR #4426, opened 2018-06-01
Add new test ID and test env info to the benchmark run.

qlzh727:
The ID can be used to differentiate between different runs (eg resnet convergence vs performance).

### DIFF
--- a/official/utils/flags/_benchmark.py
+++ b/official/utils/flags/_benchmark.py
@@ -44,7 +44,7 @@ def define_benchmark(benchmark_log_dir=True, bigquery_uploader=True):
                      "BaseBenchmarkLogger which logs to STDOUT. Different "
                      "loggers will require other flags to be able to work."))
   flags.DEFINE_string(
-      name="benchmark_test_id", short_name="bti",
+      name="benchmark_test_id", short_name="bti", default=None,
       help=help_wrap("The unique test ID of the benchmark run. It could be the "
                      "combination of key parameters. It is hardware "
                      "independent and could be used compare the performance "
--- a/official/utils/flags/_benchmark.py
+++ b/official/utils/flags/_benchmark.py
@@ -43,6 +43,12 @@ def define_benchmark(benchmark_log_dir=True, bigquery_uploader=True):
       help=help_wrap("The type of benchmark logger to use. Defaults to using "
                      "BaseBenchmarkLogger which logs to STDOUT. Different "
                      "loggers will require other flags to be able to work."))
+  flags.DEFINE_string(
+      name="benchmark_test_id", short_name="bti",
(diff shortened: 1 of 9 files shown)


official/utils/flags/_benchmark.py
```diff
@@ -43,6 +43,12 @@ def define_benchmark(benchmark_log_dir=True, bigquery_uploader=True):
       help=help_wrap("The type of benchmark logger to use. Defaults to using "
                      "BaseBenchmarkLogger which logs to STDOUT. Different "
                      "loggers will require other flags to be able to work."))
+  flags.DEFINE_string(
+      name="benchmark_test_id", short_name="bti", default=None,
+      help=help_wrap("The unique test ID of the benchmark run. It could be the "
```
karmel: Maybe a note that it is designed for human consumption? That is, nothing depends on this within the system.

official/utils/logs/logger.py
```diff
@@ -30,6 +30,7 @@
 import threading
 import uuid
 
+import requests
```
karmel: Isn't this a builtin? Why not with the builtins above?

official/utils/logs/logger.py
```diff
@@ -39,6 +40,10 @@
 BENCHMARK_RUN_LOG_FILE_NAME = "benchmark_run.log"
 _DATE_TIME_FORMAT_PATTERN = "%Y-%m-%dT%H:%M:%S.%fZ"
 
+GCP_TEST_ENV = "GCP"
+GCP_METADATA_URL = "http://metadata/computeMetadata/v1/instance/hostname"
+GCP_METADATA_HEADER = {"Metadata-Flavor": "Google"}
```
karmel: Might be wise to start separating out some of the system-specific logic. For example, having a gcp_utils in this dir that has a function that is is_gcp() or something like that. Otherwise, AWS/etc. will feel "tacked on" and it will get messy.

official/utils/flags/_benchmark.py
qlzh727: Done.

official/utils/logs/logger.py
qlzh727: Its actually not a builtin. I was bite by lint for this.

official/utils/logs/logger.py
qlzh727: Done. extract to another lib.

official/utils/logs/logger.py
```diff
@@ -36,13 +36,17 @@
 import tensorflow as tf
 from tensorflow.python.client import device_lib
 
+from official.utils.logs import cloud_lib
```
karmel: Note to selves: BUILD rules required for this and the test above.

qlzh727: ACK. Will do

PR #4428, opened 2018-06-01
Restore ResNet checkpoints

robieta:
We haven't had pretrained checkpoints since the recent topology changes. This adds them back.

### DIFF
--- a/official/resnet/README.md
+++ b/official/resnet/README.md
@@ -60,3 +60,18 @@ num_gpus:
 + 0:  Use OneDeviceStrategy and train on CPU.
 + 1:  Use OneDeviceStrategy and train on GPU.
 + 2+: Use MirroredStrategy (data parallelism) to distribute a batch between devices.
+
+### Pre-trained model
+You can download 190 MB pre-trained versions of ResNet-50. Reported accuracies are top-1 single-crop accuracy for the ImageNet validation set. Simply download and uncompress the file, and point the model to the extracted directory using the `--model_dir` flag.
+
+ResNet-50 v2 (Accuracy 76.05%):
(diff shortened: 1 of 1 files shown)


official/resnet/README.md
```diff
@@ -60,3 +60,18 @@ num_gpus:
...
+
+### Pre-trained model
+You can download 190 MB pre-trained versions of ResNet-50. Reported accuracies are top-1 single-crop accuracy for the ImageNet validation set. Simply download and uncompress the file, and point the model to the extracted directory using the `--model_dir` flag.
+
+ResNet-50 v2 (Accuracy 76.05%):
+* [Checkpoint](http://download.tensorflow.org/models/official/20180601_resnet_v2_imagenet_checkpoint.tar.gz)
+* [SavedModel](http://download.tensorflow.org/models/official/20180601_resnet_v2_imagenet_savedmodel.tar.gz)
+
+ResNet-50 v2 (fp16, Accuracy 75.56%):
+* [Checkpoint](http://download.tensorflow.org/models/official/20180601_resnet_v2_fp16_imagenet_checkpoint.tar.gz)
```
qlzh727: Is there any reason to only provide fp16 in v2?

robieta: It's what I had handy from other fp16 testing.

qlzh727: It will be nice to make them consistent, have fp16 for both or neither.

robieta: Alright, I'll fire off a v1 fp16 run at some point.

robieta: Looks like v1 fp16 still has numerical issues, so no checkpoint for that configuration.

PR #4436, opened 2018-06-03
Fix a buffer overflow error when reading Higgs csv in python 3.

robieta:


karmel:
I thought pandas gzip was stripped out of the code for py2 compatibility? Cc @yhliang2018

yhliang2018:
I tested with both py2 and py3, and the code works well. What error did you get? @robieta

robieta:
I was running from an out of date repo. My apologies.

PR #4453, opened 2018-06-05
Fixing wrong indices in _fixed_padding

jiahaowork:
The indices of kernel_size were wrong when computing the kernel_size_effective.


jiahaowork:
CLA signed.

Issue #3913, opened 2018-04-09
Cannot train the mask-rcnn models


sjwhhhi:
I want to train a mask-rcnn models by my personal dataset. I use create_pascal_tf_record.py to make it in tf-format. However, I cannot train it with this error.

> 2018-04-09 13:52:34.408287: W tensorflow/core/common_runtime/bfc_allocator.cc:277] ****************************************************************************************************
> 2018-04-09 13:52:34.408300: W tensorflow/core/framework/op_kernel.cc:1198] Resource exhausted: OOM when allocating tensor with shape[4,160,56,67] and type float on /job:localhost/replica:0/task:0/device:GPU:0 by allocator GPU_0_bfc
> Traceback (most recent call last):
>   File "train.py", line 167, in <module>
>     tf.app.run()
>   File "/usr/local/lib/python2.7/dist-packages/tensorflow/python/platform/app.py", line 124, in run
>     _sys.exit(main(argv))
>   File "train.py", line 163, in main
>     worker_job_name, is_chief, FLAGS.train_dir)
>   File "/home/sjw/models/object_detection/trainer.py", line 370, in train
>     saver=saver)
>   File "/usr/local/lib/python2.7/dist-packages/tensorflow/contrib/slim/python/slim/learning.py", line 782, in train
>     ignore_live_threads=ignore_live_threads)
>   File "/usr/lib/python2.7/contextlib.py", line 35, in __exit__
>     self.gen.throw(type, value, traceback)
>   File "/usr/local/lib/python2.7/dist-packages/tensorflow/python/training/supervisor.py", line 998, in managed_session
>     self.stop(close_summary_writer=close_summary_writer)
>   File "/usr/local/lib/python2.7/dist-packages/tensorflow/python/training/supervisor.py", line 826, in stop
>     ignore_live_threads=ignore_live_threads)
>   File "/usr/local/lib/python2.7/dist-packages/tensorflow/python/training/coordinator.py", line 387, in join
>     six.reraise(*self._exc_info_to_raise)
>   File "/usr/local/lib/python2.7/dist-packages/tensorflow/python/training/coordinator.py", line 295, in stop_on_exception
>     yield
>   File "/usr/local/lib/python2.7/dist-packages/tensorflow/python/training/coordinator.py", line 492, in run
>     self.run_loop()
>   File "/usr/local/lib/python2.7/dist-packages/tensorflow/python/training/supervisor.py", line 1028, in run_loop
>     self._sv.global_step])
[...]

PR #4462, opened 2018-06-06
Pets' mask type should be default to `PNG_MASKS`

quxiaofeng:
The pets data set provides PNG masks. The default mask type should be set to `PNG_MASKS` according to the input reader source code https://github.com/tensorflow/models/blob/master/research/object_detection/protos/input_reader.proto#L22. Currently, the setting is null, which means the default is `NUMERICAL_MASKS` according to https://github.com/tensorflow/models/blob/master/research/object_detection/protos/input_reader.proto#L20, which causes errors.

This issue hindered fresh users to train pets data set for instance segmentation. A friendly config file could save a lot of people's time.


quxiaofeng:
I signed it!

quxiaofeng:
This PR fixed #3913.

FreedoomFighter:
@googlebot quxiaofeng  is right

PR #4466, opened 2018-06-06
Update __init__.py

bob329:
there should use absolute path


bob329:
I signed it!

bob329:
I signed it!

PR #4473, opened 2018-06-06
Fix/resnet take

robieta:
This trims the ResNet pipeline to guarantee that the number of batches during training is divisible by the number of devices. (It is not necessary during eval because eval currently only uses one device, even in MirrorStrategy) Once all_reduce is cancel-able this can be reverted.

### DIFF
--- a/official/resnet/resnet_run_loop.py
+++ b/official/resnet/resnet_run_loop.py
@@ -75,7 +75,7 @@ def process_record_dataset(dataset, is_training, batch_size, shuffle_buffer,
   # dataset for the appropriate number of epochs.
   dataset = dataset.repeat(num_epochs)
 
-  if is_training and num_gpus is not None and num_gpus > 0:
+  if is_training and num_gpus and examples_per_epoch:
     total_examples = num_epochs * examples_per_epoch
     # Force the number of batches to be divisible by the number of devices.
     # This prevents some devices from receiving batches while others do not,
--- a/official/resnet/resnet_run_loop.py
+++ b/official/resnet/resnet_run_loop.py
@@ -42,7 +42,8 @@ from official.utils.misc import model_helpers
 # Functions for input processing.
 ################################################################################
 def process_record_dataset(dataset, is_training, batch_size, shuffle_buffer,
-                           parse_record_fn, num_epochs=1):
+                           parse_record_fn, num_epochs=1, num_gpus=None,
+                           examples_per_epoch=None):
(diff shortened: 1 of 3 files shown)


official/resnet/resnet_run_loop.py
```diff
@@ -72,6 +75,16 @@ def process_record_dataset(dataset, is_training, batch_size, shuffle_buffer,
   # dataset for the appropriate number of epochs.
   dataset = dataset.repeat(num_epochs)
 
+  if is_training and num_gpus is not None and num_gpus > 0:
```
karmel: nit: `if is_training and num_gpus` should be sufficient.

official/resnet/resnet_run_loop.py
```diff
@@ -72,6 +75,16 @@ def process_record_dataset(dataset, is_training, batch_size, shuffle_buffer,
...
+    total_examples = num_epochs * examples_per_epoch
+    # Force the number of batches to be divisible by the number of devices.
+    # This prevents some devices from receiving batches while others do not,
+    # which can lead to a lockup. This case will soon be handled directly by
+    # distribution strategies, at which point this .take() operation will no
+    # longer be needed.
+    total_batches = total_examples // batch_size // num_gpus * num_gpus
+    dataset.take(total_batches * batch_size)
```
karmel: To consider: if we calculate this number in the fn that calls this one, we don't have to pass num_gpus and total_examples so far in, and can just pass an ... examples_to_use or better-named equivalent through to here.

robieta: I considered this, but I like keeping the pipeline logic together and passing run info one layer deeper was an acceptable tradeoff.

robieta:
I'm testing it now.

official/resnet/resnet_run_loop.py
```diff
@@ -72,6 +75,16 @@ def process_record_dataset(dataset, is_training, batch_size, shuffle_buffer,
   # dataset for the appropriate number of epochs.
   dataset = dataset.repeat(num_epochs)
 
+  if is_training and num_gpus is not None and num_gpus > 0:
+    total_examples = num_epochs * examples_per_epoch
```
guptapriya: do you need to also check that examples_per_epoch is not None?

robieta: I don't believe so, because it will only be None during evaluation. But I suppose it doesn't hurt to add it.

PR #4474, opened 2018-06-06
Improve the BenchmarkFileLogger performance.

qlzh727:
Open and preserve the file handler of the metric log file during
init, which reduce the overhead of open/close the file for each
log_metric call.

### DIFF
--- a/official/utils/logs/logger.py
+++ b/official/utils/logs/logger.py
@@ -162,6 +162,8 @@ class BenchmarkFileLogger(BaseBenchmarkLogger):
     self._logging_dir = logging_dir
     if not tf.gfile.IsDirectory(self._logging_dir):
       tf.gfile.MakeDirs(self._logging_dir)
+    self._metric_file_fh = tf.gfile.GFile(
+        os.path.join(self._logging_dir, METRIC_LOG_FILE_NAME), "a")
 
   def log_metric(self, name, value, unit=None, global_step=None, extras=None):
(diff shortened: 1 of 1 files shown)


official/utils/logs/logger.py
```diff
@@ -162,6 +162,8 @@ def __init__(self, logging_dir):
     self._logging_dir = logging_dir
     if not tf.gfile.IsDirectory(self._logging_dir):
       tf.gfile.MakeDirs(self._logging_dir)
+    self._metric_file_fh = tf.gfile.GFile(
```
karmel: nit: "fh" is going to be hard to interpret unless you know what is happening. _metric_file_handle, maybe?

official/utils/logs/logger.py
```diff
@@ -179,14 +181,13 @@ def log_metric(self, name, value, unit=None, global_step=None, extras=None):
...
-        try:
-          json.dump(metric, f)
-          f.write("\n")
-        except (TypeError, ValueError) as e:
-          tf.logging.warning("Failed to dump metric to log file: "
-                             "name %s, value %s, error %s", name, value, e)
+      try:
+        json.dump(metric, self._metric_file_fh)
+        self._metric_file_fh.write("\n")
+        self._metric_file_fh.flush()
```
karmel: Flush should probably be in a finally block? Realistically, it's unlikely to matter, I guess. Your call.

PR #4475, opened 2018-06-06
remove unused imports and lint

robieta:
@karmel Apologies, I didn't address your initial cleanup requests.

### DIFF
--- a/official/utils/accelerator/tpu.py
+++ b/official/utils/accelerator/tpu.py
@@ -14,13 +14,9 @@
 # ==============================================================================
 """Functions specific to running TensorFlow on TPUs."""
 
-import time
-
 import tensorflow as tf
-from tensorflow.python.framework import dtypes
 from tensorflow.python.framework import ops
 from tensorflow.python.ops import array_ops
-from tensorflow.python.ops import embedding_ops
 from tensorflow.python.ops import math_ops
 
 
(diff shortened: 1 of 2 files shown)


official/utils/accelerator/tpu.py
```diff
@@ -105,7 +101,7 @@ def embedding_matmul(embedding_table, values, mask, name='embedding_matmul'):
   """
 
   with ops.name_scope(name):
-    n_embeddings, embedding_dim = embedding_table.get_shape().as_list()
+    n_embeddings, embedding_dim = embedding_table.get_shape().as_list()  # pylint: disable=unused-variable
```
qlzh727: If embedding_dim is not used, can we just replace it with _?

official/utils/accelerator/tpu.py
```diff
@@ -14,13 +14,9 @@
 # ==============================================================================
 """Functions specific to running TensorFlow on TPUs."""
 
-import time
-
 import tensorflow as tf
-from tensorflow.python.framework import dtypes
 from tensorflow.python.framework import ops
```
k-w-w: Just curious, is there a reason the dtypes and ops are directly imported? (e.g. array_ops.reshape instead of tf.reshape)

robieta: No! Looks like everything can be done with `tf.*` ops. Thanks for pointing this out.

official/utils/accelerator/tpu.py
robieta: Sure. I just had it there for readability, but I can just as easily see _ as clearer.

robieta:
@qlzh727 @k-w-w PTAL

PR #4476, opened 2018-06-07
Keras model benchmark

yhliang2018:
Hi All,

Could you help to review this PR of keras application model benchmark? It includes:
- benchmark_main.py: the main function to run benchmark pipeline
- model_callbacks: customized callbacks for benchmark.
- README 

Note that, the current benchmark runs with synthetic dataset to test the pipeline. Will test it with ImageNet dataset in next PR. Thanks!

### DIFF
--- a/official/keras_application_models/README.md
+++ b/official/keras_application_models/README.md
@@ -1,9 +1,11 @@
 # Keras Application Models Benchmark
 ## Overview
-This provides a single scaffold to benchmark the nine Keras built-in application [models](https://keras.io/applications/). All the models are for image classification application, and include:
+This provides a single scaffold to benchmark the nine Keras built-in application [models](https://keras.io/applications/). All the models are for image classification applications, and include:
+
  - Xception
  - VGG16
  - VGG19
+ - ResNet50
  - InceptionV3
  - InceptionResNetV2
  - MobileNet
@@ -11,10 +13,10 @@ This provides a single scaffold to benchmark the nine Keras built-in application
  - NASNet
 
 ## Dataset
-ImageNet data is used for the benchmark. To begin, you will need to download the ImageNet dataset and convert it to TFRecord format. Follow along with the [Inception guide](https://github.com/tensorflow/models/tree/master/research/inception#getting-started) in order to prepare the dataset.
+ImageNet dataset is used for the benchmark. To begin, you will need to download the ImageNet dataset and convert it to TFRecord format. Follow along with the [Inception guide](https://github.com/tensorflow/models/tree/master/research/inception#getting-started) in order to prepare the dataset.
 
 ## Callbacks
-Two customized callbacks are provided for model benchmark: ExamplesPerSecondCallback and LoggingMetricCallback. For each callback, `epoch_based` and `batch_based` options are available to set the benchmark level. Check [model_callbacks.py](model_callbacks.py) for more details.
+Two custom callbacks are provided for model benchmarking: ExamplesPerSecondCallback and LoggingMetricCallback. For each callback, `epoch_based` and `batch_based` options are available to set the benchmark level. Check [model_callbacks.py](model_callbacks.py) for more details.
 
 ## Running Code
 To benchmark a model, use `--model` to specify the model name, and issue the following command:
@@ -27,5 +29,3 @@ Arguments:
   * `--callbacks`: To specify a list of callbacks.
 
 Use the `--help` or `-h` flag to get a full list of possible arguments.
-
-## Benchmark Results (TODO)
--- a/official/keras_application_models/README.md
+++ b/official/keras_application_models/README.md
@@ -1,6 +1,6 @@
 # Keras Application Models Benchmark
 ## Overview
-This provides a single scaffold to benchmark the nine Keras built-in application [models](https://keras.io/applications/). All the models are for image classification applications, and include:
+This provides a single scaffold to benchmark the Keras built-in application [models](https://keras.io/applications/). All the models are for image classification applications, and include:
 
  - Xception
  - VGG16
@@ -13,7 +13,7 @@ This provides a single scaffold to benchmark the nine Keras built-in application
  - NASNet
 
 ## Dataset
-ImageNet dataset is used for the benchmark. To begin, you will need to download the ImageNet dataset and convert it to TFRecord format. Follow along with the [Inception guide](https://github.com/tensorflow/models/tree/master/research/inception#getting-started) in order to prepare the dataset.
+Synthetic dataset is used for the benchmark.
 
 ## Callbacks
 Two custom callbacks are provided for model benchmarking: ExamplesPerSecondCallback and LoggingMetricCallback. For each callback, `epoch_based` and `batch_based` options are available to set the benchmark level. Check [model_callbacks.py](model_callbacks.py) for more details.
--- a/official/keras_application_models/README.md
+++ b/official/keras_application_models/README.md
@@ -25,7 +25,6 @@ python benchmark_main.py --model=resnet
 ```
 Arguments:
   * `--model`: Which model to be benchmarked. The model name is defined as the keys of `MODELS` in [benchmark_main.py](benchmark_main.py).
-  * `--benchmark_level`: In which level to benchmark the model. Two options are provided: `epoch_based` and `batch_based`.
   * `--callbacks`: To specify a list of callbacks.
 
 Use the `--help` or `-h` flag to get a full list of possible arguments.
--- a/official/keras_application_models/README.md
+++ b/official/keras_application_models/README.md
@@ -1,0 +1,31 @@
+# Keras Application Models Benchmark
+## Overview
+This provides a single scaffold to benchmark the nine Keras built-in application [models](https://keras.io/applications/). All the models are for image classification application, and include:
+ - Xception
+ - VGG16
+ - VGG19
+ - InceptionV3
+ - InceptionResNetV2
+ - MobileNet
+ - DenseNet
+ - NASNet
+
+## Dataset
+ImageNet data is used for the benchmark. To begin, you will need to download the ImageNet dataset and convert it to TFRecord format. Follow along with the [Inception guide](https://github.com/tensorflow/models/tree/master/research/inception#getting-started) in order to prepare the dataset.
+
+## Callbacks
+Two customized callbacks are provided for model benchmark: ExamplesPerSecondCallback and LoggingMetricCallback. For each callback, `epoch_based` and `batch_based` options are available to set the benchmark level. Check [model_callbacks.py](model_callbacks.py) for more details.
+
+## Running Code
+To benchmark a model, use `--model` to specify the model name, and issue the following command:
+```
+python benchmark_main.py --model=resnet
+```
+Arguments:
+  * `--model`: Which model to be benchmarked. The model name is defined as the keys of `MODELS` in [benchmark_main.py](benchmark_main.py).
+  * `--benchmark_level`: In which level to benchmark the model. Two options are provided: `epoch_based` and `batch_based`.
+  * `--callbacks`: To specify a list of callbacks.
+
+Use the `--help` or `-h` flag to get a full list of possible arguments.
+
+## Benchmark Results (TODO)
--- a/official/keras_application_models/benchmark_main.py
+++ b/official/keras_application_models/benchmark_main.py
@@ -1,5 +1,21 @@
-# import the necessary packages
-import os
+# Copyright 2018 The TensorFlow Authors. All Rights Reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+# ==============================================================================
+"""Benchmark on the keras built-in application models."""
+from __future__ import absolute_import
+from __future__ import division
+from __future__ import print_function
 
 # pylint: disable=g-bad-import-order
 import numpy as np
@@ -8,38 +24,55 @@ from absl import flags
 import tensorflow as tf
 # pylint: enable=g-bad-import-order
 
+from official.keras_application_models import model_callbacks
+from official.resnet import resnet_run_loop
 from official.utils.flags import core as flags_core
 from official.utils.logs import logger
-
-# import official.keras_application_models.model_callbacks as model_callbacks
-import model_callbacks
-import official.resnet.imagenet_main as imagenet_main
-import official.resnet.imagenet_preprocessing as imagenet_preprocessing
-import official.resnet.resnet_run_loop as resnet_run_loop
 
 _NUM_CHANNELS = 3
 _NUM_CLASSES = 1000
 
 # Define a dictionary that maps model names to their model classes inside Keras
 MODELS = {
-  "vgg16": tf.keras.applications.VGG16,
-  "vgg19": tf.keras.applications.VGG19,
-  "inception": tf.keras.applications.InceptionV3,
-  "xception": tf.keras.applications.Xception, # TensorFlow ONLY
-  "resnet": tf.keras.applications.ResNet50,
-  "inceptionresnet": tf.keras.applications.InceptionResNetV2,
-  "mobilenet": tf.keras.applications.MobileNet,
-  "densenet121": tf.keras.applications.DenseNet121,
-  "densenet169": tf.keras.applications.DenseNet169,
-  "densenet201": tf.keras.applications.DenseNet201,
-  "nasnetlarge": tf.keras.applications.NASNetLarge,
-  "nasnetmobile": tf.keras.applications.NASNetMobile,
+    "vgg16": tf.keras.applications.VGG16,
+    "vgg19": tf.keras.applications.VGG19,
+    "inception": tf.keras.applications.InceptionV3,
+    "xception": tf.keras.applications.Xception,
+    "resnet": tf.keras.applications.ResNet50,
+    "inceptionresnet": tf.keras.applications.InceptionResNetV2,
+    "mobilenet": tf.keras.applications.MobileNet,
+    "densenet121": tf.keras.applications.DenseNet121,
+    "densenet169": tf.keras.applications.DenseNet169,
+    "densenet201": tf.keras.applications.DenseNet201,
+    "nasnetlarge": tf.keras.applications.NASNetLarge,
+    "nasnetmobile": tf.keras.applications.NASNetMobile,
 }
 
 
-def get_model_callbacks(
-    benchmark_level, callbacks_list, batch_size=None, epoch_size=None,
-    metrics=None, metric_logger=None):
+def get_model_callbacks(benchmark_level, callbacks_list, batch_size=None,
+                        epoch_size=None, metrics=None, metric_logger=None):
+  """Create a list of callbacks.
+
+  Args:
+    benchmark_level: a string to indicate if the benchmark is `batch_based` or
+      `epoch_based`. Check `model_callbacks.py` for more details.
+    callbacks_list: a list of strings to name desired callbacks. Allowed:
+      ExamplesPerSecondCallback, LoggingMetricCallback, which are defined
+      in CALLBACKS.
+    batch_size: an int of batch size.
+    epoch_size: an int of epoch size, which is the total number of training
+      samples in one epoch.
+    metrics: `dict` that maps string-valued tags to metric names, or `iterable`
+      of metric names.
+    metric_logger: instance of `BenchmarkLogger`, the benchmark logger that
+      callback should use to write the log.
+
+  Returns:
+    callbacks: a list of callbacks.
+
+  Raises:
+    ValueError: if an unrecognized callback name is passed.
+  """
   if not callbacks_list:
     return []
 
@@ -50,7 +83,7 @@ def get_model_callbacks(
   for callback in callbacks_list:
     callback_name = callback.strip().lower()
     if callback_name not in model_callbacks.CALLBACKS:
-      raise ValueError('Unrecognized callback requested: {}'.format(callback))
+      raise ValueError("Unrecognized callback requested: {}".format(callback))
 
     if callback_name == model_callbacks.CALLBACKS[0]:
       exp_per_second_callback = model_callbacks.ExamplesPerSecondCallback(
@@ -68,20 +101,21 @@ def get_model_callbacks(
 
 
 def generate_synthetic_input_dataset(image_size, batch_size):
-  # Generate synthetic dataset
-  input_shape = (batch_size, ) + image_size + (_NUM_CHANNELS, )
+  """Generate synthetic dataset."""
+  input_shape = (batch_size,) + image_size + (_NUM_CHANNELS,)
 
   images = tf.zeros(input_shape, dtype=tf.float32)
   labels = tf.zeros((batch_size, _NUM_CLASSES), dtype=tf.float32)
 
   return tf.data.Dataset.from_tensors((images, labels)).repeat()
 
+
 def get_default_image_size(model):
-  """Provide default image size for each model. """
+  """Provide default image size for each model."""
   image_size = (224, 224)
-  if model in ("inception", "xception", "inceptionresnet"):
+  if model in ["inception", "xception", "inceptionresnet"]:
     image_size = (299, 299)
-  elif model in ("nasnetlarge"):
+  elif model in ["nasnetlarge"]:
     image_size = (331, 331)
   return image_size
 
@@ -94,13 +128,13 @@ def main(_):
 
   # Load the model
   tf.logging.info("Benchmark on {} model...".format(FLAGS.model))
-  Network = MODELS[FLAGS.model]
-  model = Network(weights=None)
+  keras_model = MODELS[FLAGS.model]
+  model = keras_model(weights=None)
 
   image_size = get_default_image_size(FLAGS.model)
   if FLAGS.use_synthetic_data:
     tf.logging.info("Using synthetic dataset...")
-    batch_size=resnet_run_loop.per_device_batch_size(
+    batch_size = resnet_run_loop.per_device_batch_size(
         FLAGS.batch_size, flags_core.get_num_gpus(FLAGS))
     train_dataset = generate_synthetic_input_dataset(image_size, batch_size)
     val_dataset = generate_synthetic_input_dataset(image_size, batch_size)
@@ -113,11 +147,12 @@ def main(_):
   if num_gpus > 0:
     model = tf.keras.utils.multi_gpu_model(model, gpus=num_gpus)
 
-  # Configure the model using the optimizer sgd
+  # Configure the model
   model.compile(loss="categorical_crossentropy",
                 optimizer="sgd",
                 metrics=["accuracy"])
 
+  # Create benchmark logger for benchmark logging
   run_params = {
       "batch_size": FLAGS.batch_size,
       "synthetic_data": FLAGS.use_synthetic_data,
@@ -130,16 +165,17 @@ def main(_):
       dataset_name="ImageNet",
       run_params=run_params)
 
-  # Create callbacks
   if FLAGS.use_synthetic_data:
     train_num_images = batch_size
     val_num_images = batch_size
 
+  # Create callbacks that log metric values about the training and evaluation
   callbacks = get_model_callbacks(
       FLAGS.benchmark_level, FLAGS.callbacks,
       batch_size=FLAGS.batch_size, epoch_size=train_num_images,
       metrics=None, metric_logger=benchmark_logger)
 
+  # Train and evaluate the model
   history = model.fit(
       train_dataset,
       epochs=FLAGS.train_epochs,
@@ -152,10 +188,11 @@ def main(_):
   tf.logging.info("Logging the evaluation results...")
   for epoch in range(FLAGS.train_epochs):
     eval_results = {
-        "val_acc": history.history["val_acc"][epoch],
-        "val_loss": history.history["val_loss"][epoch],
+        "accuracy": history.history["val_acc"][epoch],
+        "loss": history.history["val_loss"][epoch],
         "epoch": epoch + 1,
-        tf.GraphKeys.GLOBAL_STEP: (epoch + 1) * np.ceil(train_num_images/batch_size)
+        tf.GraphKeys.GLOBAL_STEP: (epoch + 1) * np.ceil(
+            train_num_images/batch_size)
     }
     benchmark_logger.log_evaluation_result(eval_results)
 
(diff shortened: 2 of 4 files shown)


official/keras_application_models/benchmark_main.py
```diff
@@ -0,0 +1,242 @@
...
+  image_size = get_default_image_size(FLAGS.model)
+  if FLAGS.use_synthetic_data:
+    tf.logging.info("Using synthetic dataset...")
+    batch_size = resnet_run_loop.per_device_batch_size(
+        FLAGS.batch_size, flags_core.get_num_gpus(FLAGS))
+    train_dataset = generate_synthetic_input_dataset(image_size, batch_size)
+    val_dataset = generate_synthetic_input_dataset(image_size, batch_size)
+  else:
+    # Use the actual ImageNet dataset (TODO)
+    pass
```
qlzh727: Probably should add a tf.logging.warning here since we don't support real data.

official/keras_application_models/benchmark_main.py
```diff
@@ -0,0 +1,242 @@
...
+                metrics=["accuracy"])
+
+  # Create benchmark logger for benchmark logging
+  run_params = {
+      "batch_size": FLAGS.batch_size,
+      "synthetic_data": FLAGS.use_synthetic_data,
+      "benchmark_level": FLAGS.benchmark_level,
+      "train_epochs": FLAGS.train_epochs
+  }
+  benchmark_logger = logger.config_benchmark_logger(FLAGS)
```
qlzh727: We have changed how benchmark logger is created and closed, please find the latest example in other model main.

official/keras_application_models/benchmark_main.py
```diff
@@ -0,0 +1,242 @@
...
+  benchmark_logger.log_run_info(
+      model_name=FLAGS.model,
+      dataset_name="ImageNet",
```
qlzh727: synthetic_data is not real data, so dataset_name should have synthetic as suffix.

official/keras_application_models/model_callbacks.py
```diff
@@ -0,0 +1,159 @@
...
+}
+
+
+class ExamplesPerSecondCallback(tf.keras.callbacks.Callback):
+  """ExamplesPerSecond callback.
+
+  This callback records the average examples per second during training.
+  """
+
+  def __init__(self, batch_size=None, epoch_size=None, batch_based=None,
```
qlzh727: I think having both boolean param and also int param is bit redundant. If user specified batch_size, I think its clear that they want to measure based on batch.

official/keras_application_models/model_callbacks.py
```diff
@@ -0,0 +1,159 @@
...
+               epoch_based=None, metric_logger=None):
+    if (batch_based is None) == (epoch_based is None):
```
qlzh727: Does it have to be either of the option, but not both? Batch based provides a micro view of training speed, and epoch based provides a overview of speed. User might want to get both.

official/keras_application_models/model_callbacks.py
```diff
@@ -0,0 +1,159 @@
...
+  def on_batch_begin(self, batch, logs=None):
+    if self._batch_based:
+      self._time_start_batch_based = time.time()
+
+  def on_batch_end(self, batch, logs=None):
+    self._global_step += 1
+    if self._batch_based:
+      self._train_time_batch_based += time.time() - self._time_start_batch_based
+      examples_per_sec_batch_based = self._batch_size * (
+          self._global_step / self._train_time_batch_based)
```
qlzh727: I don't think this is correct. why the global_step is here? It should just be batch_size / time_per_batch.

official/keras_application_models/model_callbacks.py
```diff
@@ -0,0 +1,159 @@
...
+  def on_epoch_begin(self, epoch, logs=None):
+    if self._epoch_based:
+      self._time_start_epoch_based = time.time()
+
+  def on_epoch_end(self, epoch, logs=None):
+    if self._epoch_based:
+      self._epochs += 1
+      self._train_time_epoch_based += time.time() - self._time_start_epoch_based
+      examples_per_sec_epoch_based = self._epoch_size * (
+          self._epochs / self._train_time_epoch_based)
```
qlzh727: This is also not correct.

official/keras_application_models/benchmark_main.py
```diff
@@ -0,0 +1,242 @@
...
+    callbacks: a list of callbacks.
+
+  Raises:
+    ValueError: if an unrecognized callback name is passed.
+  """
+  if not callbacks_list:
+    return []
+
+  batch_based = True if benchmark_level == "batch_based" else None
+  epoch_based = True if benchmark_level == "epoch_based" else None
```
karmel: nit, for both of these:
```
epoch_based = (benchmark_level == "epoch_based")
```

official/keras_application_models/benchmark_main.py
```diff
@@ -0,0 +1,242 @@
...
+      exp_per_second_callback = model_callbacks.ExamplesPerSecondCallback(
+          batch_size=batch_size, epoch_size=epoch_size,
+          batch_based=batch_based, epoch_based=epoch_based,
+          metric_logger=metric_logger)
+      callbacks.append(exp_per_second_callback)
+    else:
+      logging_metric_callback = model_callbacks.LoggingMetricCallback(
+          metrics=metrics, metric_logger=metric_logger,
+          batch_based=batch_based, epoch_based=epoch_based)
+      callbacks.append(logging_metric_callback)
```
karmel: Can we make this a factory fn in model_callbacks, much like we do with hooks? That way the logic can stay there, and if we add more, we won't have to keep adding branches to this fn.

official/keras_application_models/benchmark_main.py
```diff
@@ -0,0 +1,242 @@
...
+
+  return callbacks
+
+
+def generate_synthetic_input_dataset(image_size, batch_size):
+  """Generate synthetic dataset."""
+  input_shape = (batch_size,) + image_size + (_NUM_CHANNELS,)
+
+  images = tf.zeros(input_shape, dtype=tf.float32)
```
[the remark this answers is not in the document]
karmel: Not any time soon, I suspect. This is also for the dataset itself, rather than anything inside the model.

official/keras_application_models/benchmark_main.py
```diff
@@ -0,0 +1,242 @@
...
+    else:
+      logging_metric_callback = model_callbacks.LoggingMetricCallback(
+          metrics=metrics, metric_logger=metric_logger,
+          batch_based=batch_based, epoch_based=epoch_based)
+      callbacks.append(logging_metric_callback)
+
+  return callbacks
+
+
+def generate_synthetic_input_dataset(image_size, batch_size):
```
karmel: We have used this function elsewhere-- move to a util in model_helpers that takes two lists of ints for input shapes?

official/keras_application_models/benchmark_main.py
```diff
@@ -0,0 +1,242 @@
...
+
+
+def get_default_image_size(model):
+  """Provide default image size for each model."""
+  image_size = (224, 224)
+  if model in ["inception", "xception", "inceptionresnet"]:
+    image_size = (299, 299)
+  elif model in ["nasnetlarge"]:
+    image_size = (331, 331)
+  return image_size
```
karmel: This should only be necessary for the synth data. Can we make the function private, and name it something like _get_default_image_size_for_synthetic_data? If the user ever needs to do this under normal circumstances, then we should expose those numbers from the keras models themselves.

official/keras_application_models/benchmark_main.py
```diff
@@ -0,0 +1,242 @@
...
+
+
+def main(_):
+  # Ensure a valid model name was supplied via command line argument
+  if FLAGS.model not in MODELS.keys():
+    raise AssertionError("The --model command line argument should "
+                         "be a key in the `MODELS` dictionary.")
```
karmel: Isn't there a flags enum option that checks this for us?

official/keras_application_models/benchmark_main.py
karmel: Or better yet, just raise an error with the note. Since the rest of this will fail if the appropriate vars are not defined.

official/keras_application_models/benchmark_main.py
```diff
@@ -0,0 +1,242 @@
...
+def define_keras_benchmark_flags():
+  """Add flags for keras built-in application models."""
+  flags_core.define_base()
+  flags_core.define_performance()
+  flags_core.define_image()
+  flags_core.define_benchmark()
+  flags.adopt_module_key_flags(flags_core)
+
+  flags_core.set_defaults(
+      data_format="channels_last",
```
karmel: Why does data format have a non-standard default?

official/keras_application_models/benchmark_main.py
```diff
@@ -0,0 +1,242 @@
...
+      use_synthetic_data=True,
+      hooks=False,
```
karmel: We should not allow hooks, correct? In that case, the define_base() should have hooks=False, so that it is not included in the set of avail flags.

official/keras_application_models/benchmark_main.py
```diff
@@ -0,0 +1,242 @@
...
+      train_epochs=2)
+
+  flags.DEFINE_enum(
+      name="model", default="resnet",
+      enum_values=MODELS.keys(), case_sensitive=False,
+      help=flags_core.help_wrap(
+          "Model to be benchmarked. Check MODELS for all available models."))
```
karmel: nit: the choices will be printed here, so the second sentence is unnecessary.

official/keras_application_models/model_callbacks.py
```diff
@@ -0,0 +1,159 @@
...
+class LoggingMetricCallback(tf.keras.callbacks.Callback):
+  """LoggingMetric callback.
+
+  By default, four metrics are logged: train accuracy, train loss, evaluation
+  accuracy and evaluation loss. Check _METRICS_TO_LOG for details.
+  """
+
+  def __init__(self, metrics=None, batch_based=None, epoch_based=None,
+               metric_logger=None):
+    if (batch_based is None) == (epoch_based is None):
```
karmel: This assumes these are None and not False, which seems fragile. Maybe if batch_based == epoch_based?

official/keras_application_models/benchmark_main.py
```diff
@@ -0,0 +1,242 @@
...
+
+  flags.DEFINE_enum(
+      name="benchmark_level", default="epoch_based",
```
karmel: I'm still confused by the level here. What is a level? Is that what Keras calls it? Looking at the callbacks, it seems it's more like bechmark_logging_schedule or something, but why does this have to be one or the other? To consider: if we are only interested in one, we may not want to provide the choice as an argument, and we should just always do the one we are interested in. If we are interested in both, why does it have to be one or the other?

official/keras_application_models/model_callbacks.py
```diff
@@ -0,0 +1,159 @@
...
+`every_n_steps` options in tensorflow Estimator hooks.
+"""
+from __future__ import absolute_import
+from __future__ import division
+from __future__ import print_function
+
+import time
+
+# pylint: disable=g-bad-import-order
+import tensorflow as tf
```
karmel: Just one lint warning at the end of the line here is sufficient.

official/keras_application_models/model_callbacks.py
```diff
@@ -0,0 +1,159 @@
...
+    if epoch_based and (epoch_size is None):
+      raise ValueError("epoch_size should be provided for "
+                       "epoch_based benchmark.")
+
+    self._batch_based = batch_based
+    self._epoch_based = epoch_based
+    self._batch_size = batch_size
+    self._epoch_size = epoch_size
+
+    self._logger = metric_logger or logger.BaseBenchmarkLogger()
```
karmel: Shouldn't you call super's init as well?

official/keras_application_models/model_callbacks.py
```diff
@@ -0,0 +1,159 @@
...
+
+  def on_train_begin(self, logs=None):
+    self._global_step = 0
```
karmel: Should this happen in init instead?

official/keras_application_models/model_callbacks.py
```diff
@@ -0,0 +1,159 @@
...
+    if self._epoch_based:
+      self._epochs += 1
+      self._train_time_epoch_based += time.time() - self._time_start_epoch_based
+      examples_per_sec_epoch_based = self._epoch_size * (
+          self._epochs / self._train_time_epoch_based)
+      self._logger.log_metric(
+          "examples_per_sec_epoch_based",
+          examples_per_sec_epoch_based,
+          global_step=self._global_step)
+
```
karmel: I feel like only one of the epoch/batch choices should be necessary, especially for ex per seconds. Let's just do batch, and then not have epoch_based, and that simplifies the code a lot, and matches the use-case just fine.

official/keras_application_models/model_callbacks.py
```diff
@@ -0,0 +1,159 @@
...
+      self._train_time_batch_based = 0
+    else:  # For epoch_based
+      self._epochs = 0
+      self._train_time_epoch_based = 0
+
+  def on_batch_begin(self, batch, logs=None):
+    if self._batch_based:
+      self._time_start_batch_based = time.time()
+
+  def on_batch_end(self, batch, logs=None):
```
karmel: For multi-gpu, does this happen once per batch per gpu? Or does this only get called once per global batch? Speaking of which-- I'm assuming the batch_size passed in is expected to be global for the Keras app models?

official/keras_application_models/model_callbacks.py
karmel: (Assuming batch_size is global, and this func is called once per global batch.)

official/keras_application_models/model_callbacks.py
```diff
@@ -0,0 +1,159 @@
...
+                       "and epoch_based should be provided.")
+
+    self._batch_based = batch_based
+    self._epoch_based = epoch_based
+
+    self._logger = metric_logger or logger.BaseBenchmarkLogger()
+    self._metrics = metrics or _METRICS_TO_LOG
+    for metric in self._metrics:
+      if metric.strip().lower() not in _METRICS_TO_LOG.keys():
+        raise ValueError("Unrecognized metric requested: {}".format(metric))
```
karmel: super's init?

official/keras_application_models/model_callbacks.py
```diff
@@ -0,0 +1,159 @@
...
+      raise ValueError("Exactly one of batch_based "
+                       "and epoch_based should be provided.")
+
+    self._batch_based = batch_based
+    self._epoch_based = epoch_based
+
+    self._logger = metric_logger or logger.BaseBenchmarkLogger()
+    self._metrics = metrics or _METRICS_TO_LOG
+    for metric in self._metrics:
+      if metric.strip().lower() not in _METRICS_TO_LOG.keys():
```
karmel: Why give a choice in that case? Let's just always log all _METRICS_TO_LOG?

official/keras_application_models/model_callbacks.py
```diff
@@ -0,0 +1,159 @@
...
+        raise ValueError("Unrecognized metric requested: {}".format(metric))
+
+  def on_train_begin(self, logs=None):
+    self._global_step = 0
```
karmel: ditto on the above-- consider putting this in the init?

official/keras_application_models/model_callbacks.py
```diff
@@ -0,0 +1,159 @@
...
+    self._global_step += 1
+    if self._batch_based:
+      for metric in self._metrics.keys():
+        # `val_acc` and `val_loss` can only be obtained after each epoch.
+        metric = metric.strip().lower()
+        if metric not in ["val_acc", "val_loss"]:
+          self._logger.log_metric(
+              _METRICS_TO_LOG[metric],
+              logs.get(metric),
+              global_step=self._global_step)
```
karmel: Let's split metrics_to_log into per_batch_metrics and per_epoch_metrics, and then log for both epoch and batch, without a choice for the user. That will simplify this code a lot, and I believe will match the use-case, though @yhliang2018 , @qlzh727 , let me know if I am incorrect.

official/keras_application_models/model_callbacks.py
```diff
@@ -0,0 +1,159 @@
...
+
+  def on_epoch_end(self, epoch, logs=None):
+    """Log metrics after each epoch."""
+    if self._epoch_based:
+      for metric in self._metrics.keys():
+        metric = metric.strip().lower()
```
karmel: This stripping and lowering is unnecessary once we just use the set of constants. Even if done, should be done on init, rather than every time we want to log.

official/keras_application_models/README.md
```diff
@@ -0,0 +1,31 @@
...
+Two customized callbacks are provided for model benchmark: ExamplesPerSecondCallback and LoggingMetricCallback. For each callback, `epoch_based` and `batch_based` options are available to set the benchmark level. Check [model_callbacks.py](model_callbacks.py) for more details.
+
+## Running Code
+To benchmark a model, use `--model` to specify the model name, and issue the following command:
+```
+python benchmark_main.py --model=resnet
+```
+Arguments:
+  * `--model`: Which model to be benchmarked. The model name is defined as the keys of `MODELS` in [benchmark_main.py](benchmark_main.py).
+  * `--benchmark_level`: In which level to benchmark the model. Two options are provided: `epoch_based` and `batch_based`.
```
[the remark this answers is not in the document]
robieta: `--benchmark_basis`?

official/keras_application_models/benchmark_main.py
```diff
@@ -0,0 +1,242 @@
...
+      batch_size=FLAGS.batch_size, epoch_size=train_num_images,
+      metrics=None, metric_logger=benchmark_logger)
+
+  # Train and evaluate the model
+  history = model.fit(
+      train_dataset,
+      epochs=FLAGS.train_epochs,
+      callbacks=callbacks,
+      validation_data=val_dataset,
+      steps_per_epoch=int(np.ceil(train_num_images / float(batch_size))),
```
robieta: nit: Is `float()` necessary with `from __future__ import division`?

official/keras_application_models/model_callbacks.py
robieta: Once per global batch.

official/keras_application_models/README.md
```diff
@@ -0,0 +1,31 @@
...
+ - VGG16
+ - VGG19
+ - InceptionV3
+ - InceptionResNetV2
+ - MobileNet
+ - DenseNet
+ - NASNet
+
+## Dataset
+ImageNet data is used for the benchmark. To begin, you will need to download the ImageNet dataset and convert it to TFRecord format. Follow along with the [Inception guide](https://github.com/tensorflow/models/tree/master/research/inception#getting-started) in order to prepare the dataset.
```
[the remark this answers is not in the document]
yhliang2018: I think the ImageNet preparation should be the same. Will update it accordingly when we are done with the actual ImageNet dataset.

official/keras_application_models/benchmark_main.py
```diff
@@ -0,0 +1,242 @@
...
+# limitations under the License.
+# ==============================================================================
+"""Benchmark on the keras built-in application models."""
+from __future__ import absolute_import
+from __future__ import division
+from __future__ import print_function
+
+# pylint: disable=g-bad-import-order
+import numpy as np
+from absl import app as absl_app
```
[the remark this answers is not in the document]
yhliang2018: somehow Kokoro test will give me a lint checking error if absl comes first, I remember. Will try to test it again.

official/keras_application_models/benchmark_main.py
```diff
@@ -0,0 +1,242 @@
...
+from absl import flags
+import tensorflow as tf
+# pylint: enable=g-bad-import-order
+
+from official.keras_application_models import model_callbacks
+from official.resnet import resnet_run_loop
```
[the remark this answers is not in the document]
yhliang2018: Will remove it after sync dataset is in utils.

official/keras_application_models/benchmark_main.py
yhliang2018: These image sizes are the default values when feeding dataset to models, and are required if no other image sizes are provided. Of course users can also specify the image size. Should we give this option to users?

official/keras_application_models/benchmark_main.py
yhliang2018: The default is None in utils, and the models require channels_last data format.

official/keras_application_models/benchmark_main.py
yhliang2018: Yes, we use callbacks for keras. By default, hooks is set as True in utils. So we set it as false here.

official/keras_application_models/model_callbacks.py
yhliang2018: You mean when users specify batch_size in command line, we should use batch_based benchmark? as we have default batch_size for the model. When users use callbacks, they just specify names. Maybe we should accept more args like hooks?

official/keras_application_models/model_callbacks.py
yhliang2018: We can support both, if users need it. (you will be the first user :) )

official/keras_application_models/model_callbacks.py
yhliang2018: Thanks for the explanation, Karmel. Right, this is the global average one, not the exp_sec for the current batch. We can add the current_exp_sec if necessary.

official/keras_application_models/model_callbacks.py
yhliang2018: Yes, it will definitely simply the code here. As for our earlier discussions with Scott , the epoch/batch choices can provide different granularity of benchmark metrics. If the batch size is small and dataset is large, batch_based option will provide lots of values which may be not desired for benchmark. @qlzh727 any idea here?

official/keras_application_models/benchmark_main.py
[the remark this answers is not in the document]
karmel: Ohh, right, because numpy is third-party and absl is Google. I keep forgetting that. Better to leave as-is, thanks.

yhliang2018:
Hi All,

Thanks a lot for the helpful comments and offline discussions. The code is updated and ready for review. Let me know if you have any comments.

The actual ImageNet dataset and unit tests will come in next PR. Thanks!

official/keras_application_models/benchmark_main.py
```diff
@@ -0,0 +1,166 @@
...
+      steps_per_epoch=int(np.ceil(train_num_images / batch_size)),
+      validation_steps=int(np.ceil(val_num_images / batch_size))
+  )
+
+  tf.logging.info("Logging the evaluation results...")
+  for epoch in range(FLAGS.train_epochs):
+    eval_results = {
+        "accuracy": history.history["val_acc"][epoch],
+        "loss": history.history["val_loss"][epoch],
+        "epoch": epoch + 1,
```
qlzh727: Usually epoch itself does not count as a metric.

official/keras_application_models/benchmark_main.py
```diff
@@ -0,0 +1,166 @@
...
+      batch_size=FLAGS.batch_size,
+      metric_logger=benchmark_logger)
+
+  # Train and evaluate the model
+  history = model.fit(
+      train_dataset,
+      epochs=FLAGS.train_epochs,
+      callbacks=callbacks,
+      validation_data=val_dataset,
+      steps_per_epoch=int(np.ceil(train_num_images / batch_size)),
```
qlzh727: why np.ceil is used here? I assume train_num_images and batch_size are just real number, rather than matrix.

official/keras_application_models/dataset.py
```diff
@@ -0,0 +1,44 @@
...
+  """Provide default image size for each model."""
+  image_size = (224, 224)
+  if model in ["inception", "xception", "inceptionresnet"]:
+    image_size = (299, 299)
+  elif model in ["nasnetlarge"]:
+    image_size = (331, 331)
+  return image_size
+
+
+def generate_synthetic_input_dataset(model, num_imgs):
```
qlzh727: Does this function need to care about the data format? eg channel first vs last?

official/keras_application_models/model_callbacks.py
```diff
@@ -0,0 +1,169 @@
...
+
+  This callback records the average_examples_per_sec and
+  current_examples_per_sec during training.
+  """
+
+  def __init__(self, batch_size, every_n_steps=1, metric_logger=None):
+    self._batch_size = batch_size
+    self._every_n_steps = every_n_steps
+    self._logger = metric_logger or logger.BaseBenchmarkLogger()
+    self._global_step = 0
```
qlzh727: I think this should be set in on_train_begin, in case user reuse the callback instance for several training cycle.

official/keras_application_models/model_callbacks.py
```diff
@@ -0,0 +1,169 @@
...
+    super(ExamplesPerSecondCallback, self).__init__()
+
+  def on_train_begin(self, logs=None):
+    self._train_time = 0
+    self._batch_times = []
```
qlzh727: There is a side effect of saving all the historical timestamp/duration since it will consume quite some memory, and you actually don't need them all when you log the metric.

In my vision, the optimal implementation should be:

```
def on_train_begin(self):
  self._train_start_timestamp = time.time()
  self._previous_recorded_timestamp = time.time()

def on_batch_end(self):
  self._global_step += 1
  current_time = time.time()

  if self._global_step % self._every_n_steps == 0:
    average_examples_per_sec = self._batch_size * self._global_step / (current_time - self._train_start_timestamp)

    current_examples_per_sec = self._batch_size * self._every_n_steps / (current_time - self._previous_recorded_timestamp)
    self._previous_recorded_timestamp = current_time
```

official/keras_application_models/model_callbacks.py
```diff
@@ -0,0 +1,169 @@
...
+class LoggingMetricCallback(tf.keras.callbacks.Callback):
+  """LoggingMetric callback.
+
+  Log the predefined _PER_BATCH_METRICS after each batch, and log the predefined
+  _PER_EPOCH_METRICS after each epoch.
+  """
+
+  def __init__(self, metric_logger=None):
+    self._logger = metric_logger or logger.BaseBenchmarkLogger()
+    self._per_batch_metrics = _PER_BATCH_METRICS
```
qlzh727: Should we expose them as the constructor params?

official/keras_application_models/model_callbacks.py
```diff
@@ -0,0 +1,169 @@
...
+    self._per_epoch_metrics = _PER_EPOCH_METRICS
+    self._global_step = 0
```
qlzh727: put it in on_train_start()

official/keras_application_models/benchmark_main.py
karmel: But it's an argument to the constructor, right? so, define_base(hooks=False) or something like that should work?

yhliang2018: Right. I will move it to the define_base function. Thanks!

official/keras_application_models/dataset.py
yhliang2018: By default, all the built-in models use channel_last format. So it's not an option here.

official/keras_application_models/model_callbacks.py
yhliang2018: Yeah, here is something actually I am not quite sure. In the last commit, I put global_step into the on_train_begin, but Karmel suggests to move it to the constructor. I thought as callback is invoked during the training phase, maybe it makes no difference here? What do you mean by "reusing the callback instance in several training cycle"?  @qlzh727 @karmel

official/keras_application_models/model_callbacks.py
yhliang2018: Smart! Thanks!!

official/keras_application_models/model_callbacks.py
karmel: on_train_begin gets called each time someone starts a train loop? Which would be for each epoch for example? Shouldn't global step be maintained for the life of the callback/model, rather than reset to 0 with each train loop? Or does on_train_begin only get called once?

official/keras_application_models/model_callbacks.py
yhliang2018: Here we decide to use the pre-defined metrics as Karmel suggested to make the code simple.

official/keras_application_models/model_callbacks.py
qlzh727: I would expect on_train_begin only gets called once when a training starts, eg a fit() call. For each epoch, it will use on_epoch_begin/end method.

karmel: From the code, it looks like on_train_begin gets called in the fit_loop, so it would be reset with each call to fit(). Since a model can be incrementally trained with fit(), we want the global step to persist across those training sessions. So, let's keep it in init for now.

yhliang2018:
Hi Karmel and Scott, 

I update this PR with latest API and also add a TODO for a WIP bug. Could you help to review it? Will merge it if everything looks fine. Thanks!

official/keras_application_models/README.md
```diff
@@ -0,0 +1,31 @@
...
+This provides a single scaffold to benchmark the nine Keras built-in application [models](https://keras.io/applications/). All the models are for image classification application, and include:
+ - Xception
+ - VGG16
+ - VGG19
+ - InceptionV3
+ - InceptionResNetV2
+ - MobileNet
+ - DenseNet
+ - NASNet
+
```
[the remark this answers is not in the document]
qlzh727: I guess its ResNet50 only

official/keras_application_models/model_callbacks.py
```diff
@@ -0,0 +1,168 @@
...
+      self._logger.log_metric(
+          _PER_BATCH_METRICS[metric],
+          logs.get(metric),
+          global_step=self._global_step)
+
+  def on_epoch_end(self, epoch, logs=None):
+    """Log metrics after each epoch."""
+    for metric in _PER_EPOCH_METRICS:
+      self._logger.log_metric(
+          _PER_EPOCH_METRICS[metric],
```
qlzh727: For the epoch based metric, there are two of them same as the ones in batch based. Should we dedup them since they will be logged by batch based anyway?

official/keras_application_models/model_callbacks.py
```diff
@@ -0,0 +1,168 @@
...
+
+
+def get_logging_metric_callback(metric_logger=None, **kwargs):  # pylint: disable=unused-argument
+  """Function to get LoggingMetricCallback."""
+  return LoggingMetricCallback(
+      metric_logger=metric_logger or logger.get_benchmark_logger())
+
+
+# A dictionary to map the callback name and its corresponding function
+CALLBACKS = {
```
qlzh727: constants should be on the top

official/keras_application_models/model_callbacks.py
```diff
@@ -0,0 +1,168 @@
...
+    self._logger = metric_logger or logger.BaseBenchmarkLogger()
+    self._global_step = 0  # Initialize it in __init__
+    super(ExamplesPerSecondCallback, self).__init__()
+
+  def on_train_begin(self, logs=None):
+    self._train_start_time = time.time()
+    self._last_recorded_time = time.time()
+
+  def on_batch_begin(self, batch, logs=None):
+    self._time_start = time.time()
```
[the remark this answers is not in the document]
yhliang2018: Good catch!! It's useless given current implementation. Have removed it.

official/keras_application_models/benchmark_main.py
```diff
@@ -0,0 +1,163 @@
...
+    tf.logging.info("Using synthetic dataset...")
+    dataset_name += "_Synthetic"
+    train_num_images = FLAGS.batch_size
+    val_num_images = FLAGS.batch_size
+    train_dataset = dataset.generate_synthetic_input_dataset(
+        FLAGS.model, train_num_images)
+    val_dataset = dataset.generate_synthetic_input_dataset(
+        FLAGS.model, val_num_images)
+  else:
+    # Use the actual ImageNet dataset (TODO)
```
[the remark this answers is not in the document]
yhliang2018: I remove it, as we will only support synthetic dataset for now.

official/keras_application_models/benchmark_main.py
```diff
@@ -0,0 +1,163 @@
...
+  flags.adopt_module_key_flags(flags_core)
+
+  flags_core.set_defaults(
+      data_format="channels_last",
+      use_synthetic_data=True,
+      batch_size=32,
+      train_epochs=2)
+
+  flags.DEFINE_enum(
+      name="model", default="resnet50",
```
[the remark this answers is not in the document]
yhliang2018: Not really. Do you think it's better if we don't set default value here?

official/keras_application_models/model_callbacks.py
yhliang2018: Yes for common constants. But here its values are two functions defined at the bottom. I think it's more clear to put it here for future extension, just like what we do for hooks.

official/keras_application_models/model_callbacks.py
yhliang2018: I think it's okay to keep it. As the number of epochs are not large compared to the number of steps, there won't be too much redundant info.

yhliang2018:
Hi Scott,
Thank you for the comments! They are all addressed. Would you help to take a look at this new commit? Thanks.

official/keras_application_models/benchmark_main.py
```diff
@@ -41,13 +41,14 @@
...
     "densenet201": tf.keras.applications.DenseNet201,
-    # TODO (b/80431378)
+    # TODO(b/80431378)
     # "nasnetlarge": tf.keras.applications.NASNetLarge,
     # "nasnetmobile": tf.keras.applications.NASNetMobile,
 }
 
 
-def main(_):
+def run_keras_model_benchmark(_):
```
qlzh727: You can choose to use the input param here which is the flag object, and that will enable the value injection during test. This can be done in a later PR.

Issue #4455, opened 2018-06-05
ValueError while Training With manual_stepping


dori-reichmann:
When training object detection with learning rate = manual_steeping, there is a ValueErrror:
(Python3.6)

ValueError: Tried to convert 't' to a tensor and failed. Error: Argument must be a dense tensor: range(0, 3) - got shape [3], but wanted [].

The error comes from this lines
```
# models/research/object_detection/utils/learning_schedules.py

warmup_steps = range(boundaries[0])

rate_index = tf.reduce_max(tf.where(tf.greater_equal(global_step, boundaries),
                                      range(num_boundaries),
                                      [0] * num_boundaries))
```

Very easy to fix, by making sure the range() returns a list:
```
# models/research/object_detection/utils/learning_schedules.py

warmup_steps = list(range(boundaries[0]))

rate_index = tf.reduce_max(tf.where(tf.greater_equal(global_step, boundaries),
                                      list(range(num_boundaries)),
                                      [0] * num_boundaries))
```

---  Edit: adding info:
What is the top-level directory of the model you are using: models/research/object-detection
OS Platform and Distribution = Ubuntu 16
TensorFlow installed from = conda tensorflow-gpu
TensorFlow version = 1.8
Bazel version = NA
CUDA/cuDNN version = 9.1 7.1.2
GPU model and memory = NA
Exact command to reproduce = you can just call:
`
import object_detection.utils.learning_schedules as  learning_schedules
learning_schedules.manual_stepping(3,[50],[0.2,0.3])
`

models/research/object-detection/utils/learning_schedules.manual_stepping

tensorflowbutler:
Thank you for your post. We noticed you have not filled out the following field in the issue template. Could you update them if they are relevant in your case, or leave them as N/A? Thanks.
What is the top-level directory of the model you are using
Have I written custom code
OS Platform and Distribution
TensorFlow installed from
TensorFlow version
Bazel version
CUDA/cuDNN version
GPU model and memory
Exact command to reproduce

dori-reichmann:
updates

karmel:
@jch1 -- looks like a py2/3 compatibility issue, and a proposed fix. @dori-reichmann , care to submit a PR with your fix?

dori-reichmann:
Done (I think)
https://github.com/tensorflow/models/pull/4477

karmel:
Thanks!

tensorflowbutler:
Hi There,
[...]

PR #4477, opened 2018-06-07
BugFix: python3 compatability

dori-reichmann:
Wrapped range() with a list, for python3 compatability.

See: https://github.com/tensorflow/models/issues/4455


dori-reichmann:
I signed it!

PR #4491, opened 2018-06-08
Update optimizer

yhliang2018:
Update optimizer from SGD to Adam for recommendation model, as the name scope issue is fixed in dist strategy.


robieta:
Do you need to specify the extra Adam parameters?

yhliang2018:
I don't think so, based on the mlperf implementation.

karmel:
MLPerf uses a different library, which may have different defaults-- have you checked that they are the same?

yhliang2018:
Yes, they use the same default values.

PR #4503, opened 2018-06-11
fixed evalution between traing

Heumi:
Fixed evaluation during training

ex) train_epochs:250, epochs_between_evals=10

before : total cycle:25, 1evalutaion / 1cycle
after : total cycle:250, 1evalution / 25cycle

### DIFF
--- a/official/resnet/resnet_run_loop.py
+++ b/official/resnet/resnet_run_loop.py
@@ -434,31 +434,31 @@ def resnet_main(
             flags_obj.batch_size, flags_core.get_num_gpus(flags_obj)),
         num_epochs=1)
 
-  total_training_cycle = (flags_obj.train_epochs //
-                          flags_obj.epochs_between_evals)
-  for cycle_index in range(total_training_cycle):
+  for cycle_index in range(flags_obj.train_epochs):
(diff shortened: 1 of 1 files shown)


Heumi:
I signed it!

official/resnet/resnet_run_loop.py
```diff
@@ -434,31 +434,31 @@ def input_fn_eval():
...
+  for cycle_index in range(flags_obj.train_epochs):
     tf.logging.info('Starting a training cycle: %d/%d',
-                    cycle_index, total_training_cycle)
+                    cycle_index, flags_obj.train_epochs)
 
     classifier.train(input_fn=input_fn_train, hooks=train_hooks,
                      max_steps=flags_obj.max_train_steps)
 
-    tf.logging.info('Starting to evaluate.')
+    if cycle_index == (flags_obj.train_epochs //
```
nio1814: This line should be 
```python
if (cycle_index % flags_obj.epochs_between_evals) == 0:
```
otherwise the condition will only be satisfied once for the case `cycle_index == flags_obj.train_epochs // flags_obj.epochs_between_evals`.

PR #4506, opened 2018-06-11
Wide Deep refactor and deep movies

robieta:
In order to prep wide-deep to run on more substantial hardware, this PR adds the Kaggle movie dataset as an option, and refactors the wide-deep code to allow for multiple datasets.


official/wide_deep/wide_deep_run_loop.py
```diff
@@ -0,0 +1,124 @@
...
+  example_input_fn = (
+      tf.estimator.export.build_parsing_serving_input_receiver_fn(feature_spec))
+  model.export_savedmodel(export_dir, example_input_fn)
+
+
+def run_loop(name, train_input_fn, eval_input_fn, model_column_fn,
+             build_estimator_fn, flags_obj, tensors_to_log, early_stop=False):
+  """Define training loop."""
+  # Clean up the model directory if present
+  shutil.rmtree(flags_obj.model_dir, ignore_errors=True)
```
k-w-w: Given that none of our other models remove the model directory, can we remove this line?

official/wide_deep/wide_deep_run_loop.py
```diff
@@ -0,0 +1,124 @@
...
+          "Download data to data_dir if it is not already present."))
+
+
+def export_model(model, model_type, export_dir, model_column_fn):
+  """Export to SavedModel format.
+
+  Args:
+    model: Estimator object
+    model_type: string indicating model type. "wide", "deep" or "wide_deep"
+    export_dir: directory to export the model.
```
k-w-w: Don't forget to add model_column_fn to args

official/wide_deep/movie_dataset.py
```diff
@@ -0,0 +1,303 @@
...
+import tempfile
+import zipfile
+
+# pylint: disable=wrong-import-order
+from absl import app as absl_app
+from absl import flags
+import numpy as np
+import pandas as pd
+import six
+from six.moves import urllib  # pylint: disable=redefined-builtin
```
k-w-w: Not used?

official/wide_deep/movie_dataset.py
```diff
@@ -0,0 +1,303 @@
...
+# pylint: enable=wrong-import-order
+
+from official.utils.flags import core as flags_core
+
+
+_RATINGS = "ratings.csv"
+_RATINGS_SMALL = "ratings_small.csv"
+_METADATA = "movies_metadata.csv"
+_FILES = ["credits.csv", "keywords.csv", "links.csv", "links_small.csv",
+          _METADATA, "ratings.csv", "ratings_small.csv"]
```
k-w-w: change "ratings.csv" and "ratings_small.csv" to _RATINGS and _RATINGS_SMALL

official/wide_deep/movie_dataset.py
```diff
@@ -0,0 +1,303 @@
...
+    "movieId": tf.FixedLenFeature([1], dtype=tf.int64),
+    "year": tf.FixedLenFeature([1], dtype=tf.int64),
+    "original_language": tf.FixedLenFeature([], dtype=tf.string),
+    "genres_0": tf.FixedLenFeature([], dtype=tf.string),
+    "genres_1": tf.FixedLenFeature([], dtype=tf.string),
+    "budget": tf.FixedLenFeature([1], dtype=tf.float32),
+    "rating": tf.FixedLenFeature([1], dtype=tf.float32)
+}
+
+_BUFFER_SIZE = {
```
k-w-w: What is this buffer size/subdir used for?

karmel:
Haven't read through all of this yet, but it appears to be the MovieLens dataset, not a kaggle dataset. Is that correct? If so, @yhliang2018 used the same for the recommendation model, IIRC. If the same, let's have one download script. If different... can it be the same?

official/wide_deep/movie_dataset.py
```diff
@@ -0,0 +1,303 @@
...
+
+  if not skip:
+    try:
+      subprocess.check_output(args=["kaggle", "-h"])
+    except:
+      tf.logging.error("Please ensure that you have installed kaggle.")
+      raise
+
+    if tf.gfile.Exists(data_dir):
+      tf.gfile.DeleteRecursively(data_dir)
```
yhliang2018: Could we reuse the downloaded data, rather than deleting it?

yhliang2018: Could we reuse the downloaded data, rather than deleting it?

official/wide_deep/movie_dataset.py
```diff
@@ -0,0 +1,303 @@
...
+import pandas as pd
+import six
+from six.moves import urllib  # pylint: disable=redefined-builtin
+import tensorflow as tf
+# pylint: enable=wrong-import-order
+
+from official.utils.flags import core as flags_core
+
+
+_RATINGS = "ratings.csv"
```
yhliang2018: Looks like all other py files (like cecus_dataset.py) use single quote for string. Should we keep it consistent?

official/wide_deep/movie_dataset.py
```diff
@@ -0,0 +1,303 @@
...
+        dataframe=dataframe, buffer_path=buffer_path, columns=all_columns)
+
+    def input_fn():
+      dataset = tf.data.TFRecordDataset(buffer_path)
+      dataset = dataset.repeat(repeat)
+      dataset = dataset.map(_deserialize)
+      dataset = dataset.batch(batch_size)
+      return dataset.prefetch(1)
+
+    return input_fn
```
yhliang2018: Is this _write_to_buffer the magic here? For the recommendation model with numpy array data, should I replace the dataframe with numpy array and perform desalinization on the serialized binary file?

yhliang2018:
@karmel Yes, recommendation model uses MovieLens dataset. But I see the two dataests have different fields and preprocessing pipelines. Eventually, we only use three columns in Movielens dataset, which is simpler compared to the Kaggle dataset.

official/wide_deep/movie_dataset.py
robieta: It is used to validate the binary representation of the data. So if the file construction is interrupted midway through, it won't have the same number of bytes and the program will know to reconstruct it.

official/wide_deep/wide_deep_run_loop.py
robieta: I added a `--clean` flag to base flags. That way it's possible to get this behavior, but by default we don't blow away user's progress.

robieta:
Well, this PR has rather sprawled.

Wide Deep and Recommendation now both use the same movielens base dataset, with application specific additional code in their respective folders. They also both use `utils.data.file_io` to run their datasets from binary files rather than from_tensor_slices().

@karmel @yhliang2018 @qlzh727 PTAL

official/datasets/movielens.py
```diff
@@ -0,0 +1,308 @@
...
+import pandas as pd
+import six
+from six.moves import urllib  # pylint: disable=redefined-builtin
+from absl import app as absl_app
+from absl import flags
+import tensorflow as tf
+# pylint: enable=g-bad-import-order
+
+from official.utils.flags import core as flags_core
+
```
qlzh727: Should have double lines between import and consts.

official/datasets/movielens.py
```diff
@@ -0,0 +1,308 @@
...
+    ML_1M: 6040,
+    ML_20M: 138493,
+}
+
+# Note: Users are indexed [1, k], not [0, k-1]
+# Both the 1m and 20m datasets use the same movie set.
+NUM_ITEM_IDS = 3952
+
+MAX_RATING = 5
+
```
qlzh727: No need of double lines between consts.

official/datasets/movielens.py
```diff
@@ -0,0 +1,308 @@
...
+def _download_and_clean(dataset, data_dir):
+  """Download MovieLens dataset in a standard format.
+
+  This function downloads the specified MovieLens format and coerces it into a
+  standard format. The only difference between the ml-1m and ml-20m datasets
+  after this point (other than size, of course) is that the 1m dataset uses
+  whole number ratings while the 20m dataset allows half integer ratings.
+  """
+  if dataset not in DATASETS:
+    raise ValueError("dataset must be one of: {}".format(",".join(DATASETS)))
```
qlzh727: Can you also log the input param value "dataset" as well? This might be helpful for troubleshooting the typo.

official/datasets/movielens.py
```diff
@@ -0,0 +1,308 @@
...
+
+    for fname in tf.gfile.ListDirectory(temp_dir):
+      tf.gfile.Copy(os.path.join(temp_dir, fname),
+                    os.path.join(data_subdir, fname))
+
+  finally:
+    tf.gfile.DeleteRecursively(temp_dir)
+
+
+def _write_csv(input_path, output_path, names, skip_first, separator=","):
```
qlzh727: From the comment, it seems that the function mainly does transformation, eg encoding. Should we just call it _transform_csv? or _cleanup_csv?

official/datasets/movielens.py
```diff
@@ -0,0 +1,308 @@
...
+
+    # Write column names to the csv.
+    f_out.write(",".join(names).encode("utf-8"))
+    f_out.write(b"\n")
+    for i, line in enumerate(f_in):
+      if i == 0 and skip_first:
+        continue  # ignore existing labels in the csv
+
+      line = line.decode("utf-8", errors="ignore")
+      fields = line.split(separator)
```
qlzh727: Just want to make sure that we process the data with ',' as separator correctly. Usually data with ',' in content will quote the comma, and in this case, we didn't handle it correctly.

official/utils/flags/_base.py
```diff
@@ -133,3 +139,9 @@ def get_num_gpus(flags_obj):
   from tensorflow.python.client import device_lib  # pylint: disable=g-import-not-at-top
   local_device_protos = device_lib.list_local_devices()
   return sum([1 for d in local_device_protos if d.device_type == "GPU"])
+
+def apply_clean(flags_obj):
```
qlzh727: I think it is not ideal to put this function in the flag module since it has nothing to do with flag itself.

official/utils/data/file_io.py
```diff
@@ -0,0 +1,199 @@
...
+      tf.logging.error("Failed to cleanup buffer files: {}".format(e))
+
+
+_GARBAGE_COLLECTOR = _GarbageCollector()
+atexit.register(_GARBAGE_COLLECTOR.purge)
+
+_ROWS_PER_CORE = 50000
+
+
+def write_to_temp_buffer(dataframe, buffer_folder, columns):
```
qlzh727: Didn't see a usage of this function in this PR. Is it intended to use it in future PR?

official/utils/data/file_io.py
```diff
@@ -0,0 +1,199 @@
...
+  Certain tasks (such as NCF Recommendation) require writing buffers to
+  temporary files. (Which may be local or distributed.) It is not generally safe
+  to delete these files during operation, but they should be cleaned up. This
+  class keeps track of temporary files created, and deletes them at exit.
+  """
+  def __init__(self):
+    self.temp_buffers = []
+
+  def register(self, filepath):
+    self.temp_buffers.append(filepath)
```
qlzh727: Probably use a set to avoid the duplication.

official/utils/data/file_io.py
```diff
@@ -0,0 +1,199 @@
...
+  if tf.gfile.Exists(buffer_path) and tf.gfile.Stat(buffer_path).length > 0:
+    actual_size = tf.gfile.Stat(buffer_path).length
+    if expected_size == actual_size:
+      return buffer_path
+    tf.logging.warning(
+        "Existing buffer {} has size {}. Expected size {}. Deleting and "
+        "rebuilding buffer.".format(buffer_path, actual_size, expected_size))
+    tf.gfile.Remove(buffer_path)
+
+  if dataframe is None:
```
qlzh727: Probably this check should happen first, to avoid the unnecessary remove of existing file.

official/utils/data/file_io.py
```diff
@@ -0,0 +1,199 @@
...
+
+
+def write_to_temp_buffer(dataframe, buffer_folder, columns):
+  if buffer_folder is None:
+    _, buffer_path = tempfile.mkstemp()
+  else:
+    tf.gfile.MakeDirs(buffer_folder)
+    buffer_path = os.path.join(buffer_folder, str(uuid.uuid4()))
+  _GARBAGE_COLLECTOR.register(buffer_path)
+  write_to_buffer(dataframe, buffer_path, columns)
```
qlzh727: There is a return value from this method, shouldn't we just return the value here?

official/utils/data/file_io.py
```diff
@@ -0,0 +1,199 @@
...
+  return buffer_path
+
+
+def iter_shard_dataframe(df, rows_per_core=1000):
```
qlzh727: I think this class and methods need some unit test

official/utils/data/file_io.py
```diff
@@ -0,0 +1,199 @@
...
+    df_shard = df[min_index:max_index]
+    n_shard = len(df_shard)
+    boundaries = np.linspace(0, n_shard, num_cores + 1, dtype=np.int64)
+    yield [df_shard[boundaries[j]:boundaries[j+1]] for j in range(num_cores)]
+
+
+def _shard_dict_to_examples(shard_dict):
+  """Converts a dict of arrays into a list of example bytes."""
+  n = [i for i in shard_dict.values()][0].shape[0]
+  feature_list = [{} for _ in range(n)]
```
qlzh727: probably can be changed to [{}] * n

official/utils/data/file_io.py
```diff
@@ -0,0 +1,199 @@
...
+    max_index = min([(i + 1) * max_batch_size, n])
+    df_shard = df[min_index:max_index]
+    n_shard = len(df_shard)
+    boundaries = np.linspace(0, n_shard, num_cores + 1, dtype=np.int64)
+    yield [df_shard[boundaries[j]:boundaries[j+1]] for j in range(num_cores)]
+
+
+def _shard_dict_to_examples(shard_dict):
+  """Converts a dict of arrays into a list of example bytes."""
+  n = [i for i in shard_dict.values()][0].shape[0]
```
qlzh727: I am bit confused about this, are u assuming all the item in shard_dict has same size and shape?

official/datasets/movielens.py
robieta: Yes. One of the two datasets uses commas as a separator, and has already quoted the fields with commas in the literal text. This is there to avoid double quoting those entries.

official/utils/data/file_io.py
robieta: The names shouldn't collide (they are generated using `tempfile.mkstemp()` or `uuid.uuid4()`), and there is a `tf.gfile.Exists()` check that would handle it anyway.

official/utils/data/file_io.py
robieta: It is used in `official/recommendation/ncf_main.py`:

`    buffer_path = file_io.write_to_temp_buffer(df, buffer_dir, _COLUMNS)`

official/utils/data/file_io.py
robieta:
```
a = [{}] * 5
a[0]["foo"] = 1
print(a)

b = [{} for _ in range(5)]
b[0]["foo"] = 1
print(b)
```

```
[{'foo': 1}, {'foo': 1}, {'foo': 1}, {'foo': 1}, {'foo': 1}]
[{'foo': 1}, {}, {}, {}, {}]
```

official/utils/data/file_io.py
robieta: The pandas dataframes cannot be passed to map because they cannot be pickled. So instead I convert them into a dict with the column names and the underlying arrays. Once inside the map function I infer the shape from the first one. In order to do this safely I have a `assert len(set([v.shape[0] for v in inp.values()])) == 1` check before entering the map. The reason that I've done this is that I try whenever possible to avoid throwing exceptions inside of a multiprocessing pool.

official/utils/data/file_io.py
robieta: I don't think this can be reordered. If a valid buffer does exist we want to exit early. Note the removal doesn't depend on the content of the input dataframe. And if it fails the check the removal is absolutely necessary, because otherwise we will concatenate onto an incomplete buffer fragment.

official/utils/flags/_base.py
robieta: Sure. Moved to `model_helpers`

official/utils/data/file_io.py
robieta: I think you're right.

official/utils/data/file_io.py
qlzh727: Ah, thanks for the explanation.

official/utils/data/file_io_test.py
```diff
@@ -0,0 +1,193 @@
...
+        [[0, 1, 2, 3]], [[4, 5, 6, 7]], [[8, 9, 10, 11]],
+        [[12, 13, 14, 15]], [[16, 17, 18, 19]], [[20]]
+    ]),
+
+    dict(row_count=1, cpu_count=4, expected=[
+        [[0]]
+    ]),
+
+    dict(row_count=10, cpu_count=4, expected=[
+        [[0, 1], [2, 3, 4], [5, 6], [7, 8, 9]]
```
qlzh727: Is there any reason that group 2 and 4 have 1 more item? It is random when assigning item to bucket? Should it have a deterministic order?

official/utils/data/file_io_test.py
```diff
@@ -0,0 +1,193 @@
...
+]
+
+_FEATURE_MAP = {
+    _RAW_ROW: tf.FixedLenFeature([1], dtype=tf.int64),
+    _DUMMY_COL: tf.FixedLenFeature([1], dtype=tf.int64),
+    _DUMMY_VEC_COL: tf.FixedLenFeature([_DUMMY_VEC_LEN], dtype=tf.float32)
+}
+
+
+class FixedCoreCount(object):
```
qlzh727: To make this shorter, you can use the contextmanager like below:

```
@contextmanager
def fixed_core_count(cpu_count):
   old_count_fn = multiprocessing.cpu_count
   multiprocessing.cpu_count = lambda: cpu_count
   yield
   multiprocessing.cpu_count = old_count_fn

```

official/utils/data/file_io_test.py
```diff
@@ -0,0 +1,193 @@
...
+  def test_small_rows_large_core(self):
+    self._test_sharding(**_TEST_CASES[6])
+
+  def test_large_rows_large_core(self):
+    self._test_sharding(**_TEST_CASES[7])
+
+  def _serialize_deserialize(self, num_cores=1):
+    n = 20
+
+    np.random.seed(34523)
```
qlzh727: What is this magic seed? Is it chosen randomly? or can we use a more normal one, eg 0 or 1?

official/utils/data/file_io_test.py
```diff
@@ -0,0 +1,193 @@
...
+    self._test_sharding(**_TEST_CASES[5])
+
+  def test_small_rows_large_core(self):
+    self._test_sharding(**_TEST_CASES[6])
+
+  def test_large_rows_large_core(self):
+    self._test_sharding(**_TEST_CASES[7])
+
+  def _serialize_deserialize(self, num_cores=1):
+    n = 20
```
qlzh727: What is n? number or rows?

official/utils/data/file_io_test.py
robieta: This is much better. Changed.

official/utils/data/file_io_test.py
robieta: Sorry I missed this. Sharding is done using `np.linspace`. I believe you see this pattern because it does the spacing in floats and then casts to ints. It should be deterministic.

official/wide_deep/census_dataset.py
```diff
@@ -0,0 +1,201 @@
...
+          'Other-relative'])
+
+  workclass = tf.feature_column.categorical_column_with_vocabulary_list(
+      'workclass', [
+          'Self-emp-not-inc', 'Private', 'State-gov', 'Federal-gov',
+          'Local-gov', '?', 'Self-emp-inc', 'Without-pay', 'Never-worked'])
+
+  # To show an example of hashing:
+  occupation = tf.feature_column.categorical_column_with_hash_bucket(
+      'occupation', hash_bucket_size=1000)
```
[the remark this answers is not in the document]
qlzh727: Please use _HASH_BUCKET_SIZE here as well.

PR #4510, opened 2018-06-12
Renamed base_name variable to checkpoint_name

AndriiTsok:
Fixed invalid variable name for checkpoint filename prefix


AndriiTsok:
I signed it!

PR #4511, opened 2018-06-12
Clarified .bashrc Configuration in G3Doc Installation

WillAyd:
Very minor documentation edit, but for those that are not overly familiar with `.bashrc` configuration I figured it's worth being explicit on what to do with the installation docs


WillAyd:
I signed it!

PR #4539, opened 2018-06-14
add gamma and hue adjustments to image augmentation

dheera:
Allows input data to be augmented using power law (gain, gamma) and hue adjustments.
Flags added with defaults set to not performing any adjustments.


dheera:
Hi @aquariusjay -- I have qualitatively observed improvements in cases where an entire dataset is taken with a single sensor at fixed white balance or at a certain time of day. We want the model to be learning, for example, the features characteristic to grass, rather than the exact RGB color ranges of grass, which may change at different times of day, seasons, and so on. I have found perturbations in hue / gamma to help avoid this type of overfitting.

It is less useful on open datasets, since those are already often taken using a variety of sensors and conditions, but when using any private dataset taken with a single sensor I find that these augmentations help.

dheera:
@aquariusjay I was wondering if you were at CVPR?

dheera:
any update on this? thanks!

PR #4540, opened 2018-06-14
Fix mnist Estimator input_fn batch size for multi-GPU.

yanboliang:
When training with multi-GPU, the input dataset batch size should be per device batch size.


robieta:
mnist.py still uses `tf.contrib.estimator.replicate_model_fn()`, rather than the newer distribution strategies API. As a result the per device batch size behavior in most of the other models is not appropriate for mnist.

yanboliang:
Got it, thanks.

PR #4542, opened 2018-06-15
Get started index

lamberta:
Staged: https://colab.sandbox.google.com/github/lamberta/models/blob/get-started-index/samples/core/get_started/_index.ipynb

### DIFF
--- a/samples/core/get_started/_index.ipynb
+++ b/samples/core/get_started/_index.ipynb
@@ -61,7 +61,14 @@
       "source": [
         "# Get Started with TensorFlow\n",
         "\n",
-        "First, import TensorFlow into your program:"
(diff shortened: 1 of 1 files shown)


lamberta:
Added. Though this page won't be rendered to HTML, but might be useful to jump between colab and github.

lamberta:
Not sure why it's not reflected in this branch's stage, maybe we should investigate?
Can see here: https://colab.sandbox.google.com/github/lamberta/models/blob/7bcae135909a74035fa3c811e4ca080990f7323e/samples/core/get_started/_index.ipynb

MarkDaoust:
> Not sure why it's not reflected in this branch's stage, maybe we should investigate?

It's showing now. I've had similar things happen before... like colab's change-detector need's work. Not sure.

PR #4555, opened 2018-06-15
Rename eager get started notebook to custom training walkthrough

lamberta:


lamberta:
Staged: https://colab.sandbox.google.com/github/lamberta/models/blob/rename-eager-nb/samples/core/get_started/custom_training_walkthrough.ipynb

PR #4567, opened 2018-06-18
Update gen_vocab.py to be python3 compatible

dreamgonfly:
Dictionary's .iteritem() raises an error in Python3.
Replaced it with iteritem from six library.


dreamgonfly:
I signed CLA.

PR #4577, opened 2018-06-19
Fix too much big value of default max_steps of 2560 epoch to 256 epoch

Hospitalization:
the default max_step is set to 100000 which makes epoch 2560.
It confuses users whether 256 epoch or 2560 epoch.
256 epoch is presented as previous comments and example results.
Also cifar-10 and this kind of network written here doesn't need too many epoch.


Hospitalization:
I signed it!)

PR #4581, opened 2018-06-19
correct base_name to checkpoint_name

bigsnarfdude:
reference to base_name appears to be incorrect. changing base_name to checkpoint_name resolves unpacking error and broken notebook example


bigsnarfdude:
I signed it!  Again. Previously signed Dec 2016

bigsnarfdude:
I signed it!

PR #4583, opened 2018-06-19
Save and restore

MarkDaoust:
Staging:

https://colab.sandbox.google.com/github/markdaoust/models/blob/save-and-restore/samples/core/get_started/save_and_restore_models.ipynb


MarkDaoust:
Collapsed license, deleted trailing cells.

PR #4607, opened 2018-06-22
Fix phrasing

lamberta:


lamberta:
https://colab.sandbox.google.com/github/lamberta/models/blob/master/samples/core/get_started/overfit_and_underfit.ipynb

PR #4623, opened 2018-06-25
Add environments for efficient-hrl

ofirnachum:
An initial commit for the "efficient-hrl" subdir based on https://arxiv.org/abs/1805.08296.  
The training code will be open-sourced at a later time.


ofirnachum:
Good catch - now added.

PR #4632, opened 2018-06-26
Add deep speech model and run loop

yhliang2018:
Hi All,

Could you help to review this PR of deep speech model? The PR includes:
1. deep speech model with keras implemenation
2. the training run-loop

@A30041839 will add the dataset.py for input_fn soon, and we are still working on the eval part. Will submit a new PR later. I am testing the train loop with multi-gpus on GCP instance, and will also add the test results soon.

Thanks a lot!

### DIFF
--- a/research/deep_speech/data/vocabulary.txt
+++ b/research/deep_speech/data/vocabulary.txt
@@ -1,0 +1,33 @@
+# List of alphabets (utf-8 encoded). Note that '#' starts a comment line, which
+# will be ignored by the parser.
+# begin of vocabulary
+a
+b
+c
+d
+e
+f
+g
+h
+i
+j
+k
+l
+m
+n
+o
+p
+q
+r
+s
+t
+u
+v
+w
+x
+y
+z
+'
+ 
+-
+# end of vocabulary
--- a/research/deep_speech/deep_speech.py
+++ b/research/deep_speech/deep_speech.py
@@ -12,23 +12,28 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 # ==============================================================================
-"""Main entry to train and evaluation Deep Speech model.
-"""
+"""Main entry to train and evaluate DeepSpeech model."""
 from __future__ import absolute_import
 from __future__ import division
 from __future__ import print_function
 
+import os
 # pylint: disable=g-bad-import-order
 from absl import app as absl_app
 from absl import flags
 import tensorflow as tf
+# pylint: enable=g-bad-import-order
 
-import deep_speech_model
 import data.dataset as dataset
+import deep_speech_model
 from official.utils.flags import core as flags_core
 from official.utils.logs import hooks_helper
 from official.utils.logs import logger
 from official.utils.misc import distribution_utils
+
+# Default vocabulary file
+_VOCABULARY_FILE = os.path.join(
+    os.path.dirname(__file__), "data/vocabulary.txt")
 
 
 def convert_keras_to_estimator(keras_model, num_gpus):
@@ -36,14 +41,18 @@ def convert_keras_to_estimator(keras_model, num_gpus):
 
   Args:
     keras_model: A Keras model object.
-    num_gpus: An integer, the number of gpus.
+    num_gpus: An integer, the number of GPUs.
 
   Returns:
     estimator: The converted Estimator.
   """
-  optimizer = tf.keras.optimizers.SGD(
-      lr=flags_obj.learning_rate, momentum=flags_obj.momentum,
-      decay=flags_obj.l2, nesterov=True)
+  # keras optimizer is not compatible with distribution strategy.
+  # Use tf optimizer instead
+  optimizer = tf.train.MomentumOptimizer(
+      learning_rate=flags_obj.learning_rate, momentum=flags_obj.momentum,
+      use_nesterov=True)
+
+  # ctc_loss is wrapped as a Lambda layer in the model.
   keras_model.compile(
       optimizer=optimizer, loss={"ctc_loss": lambda y_true, y_pred: y_pred})
 
@@ -71,20 +80,16 @@ def generate_dataset(data_dir):
   return speech_dataset
 
 
-def main(_):
-  with logger.benchmark_context(flags_obj):
-    run_deep_speech(flags_obj)
-
-
 def run_deep_speech(_):
   """Run deep speech training and eval loop."""
   # Data preprocessing
   # The file name of training and test dataset
   tf.logging.info("Data preprocessing...")
+
   train_speech_dataset = generate_dataset(flags_obj.train_data_dir)
   eval_speech_dataset = generate_dataset(flags_obj.eval_data_dir)
 
-  # Number of label classes. Label string is "abcdefghijklmnopqrstuvwxyz' -"
+  # Number of label classes. Label string is "[a-z]' -"
   num_classes = len(train_speech_dataset.speech_labels)
 
   # Input shape of each data example:
@@ -98,8 +103,6 @@ def run_deep_speech(_):
       input_shape, flags_obj.rnn_hidden_layers, flags_obj.rnn_type,
       flags_obj.is_bidirectional, flags_obj.rnn_hidden_size,
       flags_obj.rnn_activation, num_classes, flags_obj.use_bias)
-  # Check model structure
-  tf.logging.info(keras_model.summary(line_length=100))
 
   # Convert to estimator
   num_gpus = flags_core.get_num_gpus(flags_obj)
@@ -131,11 +134,11 @@ def run_deep_speech(_):
 
   def input_fn_train():
     return dataset.input_fn(
-        True, per_device_batch_size, train_speech_dataset)
+        per_device_batch_size, train_speech_dataset)
 
   def input_fn_eval():  # #pylint: disable=unused-variable
     return dataset.input_fn(
-        False, per_device_batch_size, eval_speech_dataset)
+        per_device_batch_size, eval_speech_dataset)
 
   total_training_cycle = (flags_obj.train_epochs //
                           flags_obj.epochs_between_evals)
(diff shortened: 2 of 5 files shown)


research/deep_speech/deep_speech.py
```diff
@@ -0,0 +1,269 @@
...
+  """Generate a speech dataset."""
+  audio_conf = dataset.AudioConfig(
+      flags_obj.sample_rate, flags_obj.frame_length, flags_obj.frame_step)
+  train_data_conf = dataset.DatasetConfig(
+      audio_conf,
+      data_dir,
+      flags_obj.vocabulary_file,
+  )
+  speech_dataset = dataset.DeepSpeechDataset(train_data_conf)
+  return speech_dataset
```
karmel: This is delightfully clean. Thanks, @yhliang2018 and @A30041839

research/deep_speech/deep_speech.py
```diff
@@ -0,0 +1,269 @@
...
+  input_shape = (None, train_speech_dataset.num_feature_bins, 1)
+
+  # Create deep speech model and convert it to Estimator
+  tf.logging.info("Creating Estimator from Keras model...")
+  keras_model = deep_speech_model.DeepSpeech(
+      input_shape, flags_obj.rnn_hidden_layers, flags_obj.rnn_type,
+      flags_obj.is_bidirectional, flags_obj.rnn_hidden_size,
+      flags_obj.rnn_activation, num_classes, flags_obj.use_bias)
+  # Check model structure
+  tf.logging.info(keras_model.summary(line_length=100))
```
karmel: What does this output? Do we want to have that in all cases, or was this part of the debugging process?

research/deep_speech/deep_speech.py
```diff
@@ -0,0 +1,269 @@
...
+
+    # eval_results = evaluate_model(
+    #     estimator, keras_model, data_set.speech_labels, [], input_fn_eval)
+
+    # benchmark_logger.log_evaluation_result(eval_results)
+
+    # if model_helpers.past_stop_threshold(
+    #   flags_obj.stop_threshold, eval_results["accuracy"]):
+    #   break
+
```
karmel: Why is this commented out? What happens?

research/deep_speech/deep_speech.py
```diff
@@ -0,0 +1,269 @@
...
+  if flags_obj.export_dir is not None:
+    # Exports a saved model for the given classifier.
+    input_receiver_fn = export.build_tensor_serving_input_receiver_fn(
+        shape, batch_size=flags_obj.batch_size)
```
karmel: I suspect this will raise an error, as shape is undefined. What is the expected input shape/format when predicting?

research/deep_speech/deep_speech_model.py
```diff
@@ -0,0 +1,174 @@
...
+    layer_id: an integer specifying the layer index.
+
+  Returns:
+    tensor output from the current layer.
+  """
+  cnn_output = tf.keras.layers.Conv2D(
+      filters=filters, kernel_size=kernel_size, strides=strides, padding="same",
+      activation="linear", name="cnn_{}".format(layer_id))(cnn_input)
+  output = tf.keras.layers.BatchNormalization(
+      momentum=0.1, epsilon=1e-05)(cnn_output)
```
karmel: Should these be constants at the top of the module?

research/deep_speech/deep_speech_model.py
```diff
@@ -0,0 +1,174 @@
...
+      on input states.
+    is_bidirectional: a boolean specifying whether the rnn layer is
+      bi-directional.
+
+  Returns:
+    tensor output for the current layer.
+  """
+  if is_batch_norm:
+    input_data = tf.keras.layers.BatchNormalization(
+        momentum=0.1, epsilon=1e-05)(input_data)
```
karmel: Ah, I can answer my own question: these should be constants at the top of the module.

research/deep_speech/deep_speech_model.py
```diff
@@ -0,0 +1,174 @@
...
+  rnn_output = rnn_layer(input_data)
+  return rnn_output
+
+
+def _ctc_lambda_func(args):
+  """Compute ctc loss."""
+  y_pred, labels, input_length, label_length = args
+  return tf.keras.backend.ctc_batch_cost(
+      labels, y_pred, input_length, label_length)
+
```
karmel: Nit: lint me. This should have two line breaks.

research/deep_speech/deep_speech_model.py
```diff
@@ -0,0 +1,174 @@
...
+def _cal_ctc_input_length(args):
```
karmel: nit: _calc_ctc... would be more clear, assuming the "cal" here stands for "calculate"

research/deep_speech/deep_speech_model.py
```diff
@@ -0,0 +1,174 @@
...
+    """
+    # Input variables
+    input_data = tf.keras.layers.Input(
+        shape=input_shape, name="features")
+
+    # Two cnn layers
+    conv_layer_1 = _conv_bn_layer(
+        input_data, filters=32, kernel_size=(41, 11), strides=(2, 2),
+        layer_id=1)
+    print("conv_layer_1", conv_layer_1)
```
karmel: nit: remove debugging prints

research/deep_speech/deep_speech_model.py
```diff
@@ -0,0 +1,174 @@
...
+    # output of conv_layer2 with the shape of
+    # [batch_size (N), times (T), features (F), channels (C)]
+
+    # RNN layers.
+    # Convert the conv output to rnn input
+    rnn_input = tf.keras.layers.TimeDistributed(tf.keras.layers.Flatten())(
+        conv_layer_2)
+
+    rnn_cell = _SUPPORTED_RNNS[rnn_type]
+    for layer in xrange(num_rnn_layers):
```
karmel: IIRC, xrange is not py3 compatible-- should be range or six.range

research/deep_speech/deep_speech_model.py
```diff
@@ -0,0 +1,174 @@
...
+      is_batch_norm = (layer != 0)  # No batch normalization on the first layer
```
karmel: nit: layer sounds like the Layer obj. Can we call this layer_counter or i or something that makes it more clearly an incrementer?

research/deep_speech/deep_speech_model.py
```diff
@@ -0,0 +1,174 @@
...
+      rnn_input = _rnn_layer(rnn_input, rnn_cell, rnn_hidden_size, layer + 1,
+                             rnn_activation, is_batch_norm, is_bidirectional)
+
+    # FC layer with batch norm
+    fc_input = tf.keras.layers.BatchNormalization(momentum=0.1, epsilon=1e-05)(
```
karmel: Module-level constants

research/deep_speech/deep_speech.py
```diff
@@ -0,0 +1,269 @@
...
+  speech_dataset = dataset.DeepSpeechDataset(train_data_conf)
+  return speech_dataset
+
+
+def main(_):
+  with logger.benchmark_context(flags_obj):
+    run_deep_speech(flags_obj)
+
+
+def run_deep_speech(_):
```
k-w-w: Since `main` is calling `run_deep_speach(flags_obj)`, `_` should be changed to `flags_obj` here.

research/deep_speech/deep_speech.py
```diff
@@ -0,0 +1,269 @@
...
+    classifier.export_savedmodel(flags_obj.export_dir, input_receiver_fn)
+
+  # Clear the session explicitly to avoid session delete error
+  tf.keras.backend.clear_session()
+
+
+def define_deep_speech_flags():
+  """Add flags for run_deep_speech."""
+  # Add common flags
+  flags_core.define_base()
```
k-w-w: The `data_dir` flag can be disabled here.

research/deep_speech/deep_speech.py
```diff
@@ -0,0 +1,269 @@
...
+
+
+def convert_keras_to_estimator(keras_model, num_gpus, flags_obj):
+  """Configure and convert keras model to Estimator.
+  Args:
+    keras_model: A Keras model object.
+    num_gpus: An integer, the number of gpus.
+    flags_obj: An object containing parsed flags. See define_resnet_flags()
+      for details.
+  Returns:
```
k-w-w: Nit: (Unsure if the linter complains about this) Might need an extra line between the end of Args and Returns.

research/deep_speech/deep_speech.py
```diff
@@ -0,0 +1,269 @@
...
+      help=flags_core.help_wrap("The hidden size of RNNs."))
+
+  flags.DEFINE_integer(
+      name="rnn_hidden_layers", default=3,
+      help=flags_core.help_wrap("The number of RNN layers."))
+
+  flags.DEFINE_bool(
+      name="use_bias", default=True,
+      help=flags_core.help_wrap(
+          "Use bias in the last fc layer"))
```
k-w-w: Spell out FC? (I assume it stands for fully-connected)

research/deep_speech/data/vocabulary.txt
```diff
@@ -0,0 +1,33 @@
...
+t
+u
+v
+w
+x
+y
+z
+'
+ 
+-
```
karmel: This is such a short vocab file; is it easier to just create a list in Python?

```
import string
vocab = list(string.ascii_lowercase) + ["'", " ", "-"]
```

or something to that effect?

A30041839: I think using a python list is fine for our purpose now. However, it seems like a common practice to put those vocabulary into a standalone file and let python code read from it to construct a dictionary. I guess it's because for some languages such as Chinese the character set is just too huge. At this point I prefer still using the txt file so that we can better separate code and data.

karmel: Leaving it as a file is fine in that case-- thanks

yhliang2018:
Thanks a lot for all the comments! Will create a new PR with parallel dataset processing and evaluation module.

PR #4674, opened 2018-07-02
np.Array to tf.data.Dataset using buffer files and decode_raw()

robieta:
This PR presents an initial implementation of an alternative to TFRecords as a format for feeding highly structured numeric data during training.

The context for this PR comes from performance tuning of NCF Recommendation (`official/recommendation`). NCF is arguably a worst case scenario for TFRecords:
1) New data must be generated each epoch, so the serialization is a per-epoch cost rather than a one-time upfront cost.
2) The amount of information per example is very small, consisting of one int32, one uint16, and one int8. So TFRecords is forced to spend 24 bytes to store 7 bytes of information.

I'm saving the refactor of NCF for a separate PR since it contains other optimizations and I don't want to clutter the discussion.

The basic approach is to write an array's binary buffer to a file, ingest the bytes using FixedLengthRecordDataset, and then use .map() and tf.decode_raw() to get the bytes into tensors. I experimented with using .from_generator() and a python generator that fed bytes and it was somewhat faster, but not much and involved doing dangerous things like hanging onto a memoryview() of a numpy array, so I abandoned that approach.

This is all implemented using the public API as I did not feel confident writing a new class in the c++ side, and time is somewhat important for the MLPerf deadline. However even as a strawman it performs quite well. I have included a script `official/utils/data/performance_comparison.py` that I don't intend to merge into master, but is useful for demonstration.

@mrry I'm curious what you think of this practice of looting the bytes from array buffers.

### DIFF
--- a/official/utils/data/buffer.py
+++ b/official/utils/data/buffer.py
@@ -137,7 +137,7 @@ class _ArrayBytesView(object):
     x_view = memoryview(source_array)
 
     if six.PY3:
-      nbytes = x_view.nbytes
+      nbytes = x_view.nbytes  # pylint: disable=no-member
     else:
       nbytes = int(np.product(x_view.shape) * x_view.itemsize)
 
@@ -213,7 +213,7 @@ class _FileBackedArrayBytesView(_ArrayBytesView):
         f.write(chunk)
 
   def to_dataset(self, decode_procs=2, decode_batch_size=16,
-      extra_map_fn=None, unbatch=True):
+                 extra_map_fn=None, unbatch=True):
     """Create a Dataset from the underlying buffer file.
 
     Args:
--- a/official/utils/data/buffer.py
+++ b/official/utils/data/buffer.py
@@ -1,0 +1,292 @@
+# Copyright 2018 The TensorFlow Authors. All Rights Reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+# ==============================================================================
+"""Create TensorFlow Datasets from NumPy binary buffers."""
+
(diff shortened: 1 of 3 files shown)


mrry:
Can you explain some more about this piece of context?

> New data must be generated each epoch

Does this comment refer to this function?

https://github.com/tensorflow/models/blob/32e7d660a813c11da61a2ad35055d85df8f14b63/official/recommendation/movielens_dataset.py#L320-L348

If so, does this data have to be written to a file at all? It looks like it could be generated on-the-fly in the `input_fn`, e.g. in a `Dataset.from_generator()`, and perhaps that would be more efficient?

robieta:
Hm...

Currently in master that function is serial and therefore painfully slow. In another branch I put it through a multiprocessing pool; however I have encountered extremely negative interactions between tensorflow and multiprocessing such that I don't think I can move that pool into .from_generator().

However, the more I think about it there may be a pipeline where the positives are written to shards, and then a .list_files(), .map(), .flat_map(), and .interleave() pipeline would allow the false negative generation to run in the tensorflow threadpool during training.

robieta:
@mrry I did some experiments with on-the-fly generation, but they didn't go well.

The first issue is the GIL. I was using tf.contrib.data.parallel_interleave() with tf.data.Dataset.from_generator() since I need multiple workers to be competitive on performance, but some experimentation on a toy pipeline showed that it really is just context switching across workers. It is possible to get around that by calling multiprocessing.Process from within the generator, but that seems exceptionally dangerous.

The second issue is that even if you manage to make the generator arbitrarily fast, yielding single points has massive overhead. You can yield batches from the generator which isn't too bad, but unbatching for shuffle (either with flat_map() and from_tensor_slices or with contrib.data.unbatch()) again tanks performance.

mrry:
Were all those performance experiments using Python generators? If so, then I'm not so surprised that the per-element yielding overhead is high. If not, it might be possible to optimize the ops being used, and the optimizations that @jsimsa and @rachellim have been trying could reduce the overhead by fusing/vectorizing individual transformations.

(I'm not opposed to the approach you're taking in this PR, as long as you're happy to support it. I strongly suspect that an approach based on on-the-fly generation should outperform one that reads and writes data each epoch, especially when the generation process is so simple and I/O has been a bottleneck.)

robieta:
Yes, this was using python generators.

I will say, I do think your trepidation is warranted; in a perfect world on-the-fly is the best of all worlds. So if the question shifts from "how can we best get data into Dataset" to "how can we best support on-the-fly" I'm totally on board. I'll follow up with them, and thanks for your input.

PR #4678, opened 2018-07-02
Mask R-CNN model added to models/research/mlperf_object_detection/Mask_RCNN 

MehdiShz:


FreedoomFighter:
Will it solve the issue like this https://github.com/tensorflow/models/issues/4689 ?
(How can I upgrade [command]?)

futurely:
What is the point of duplicating the object detection API? The diff should have been contributed to the API instead.

robieta:
This fork should have been removed by https://github.com/tensorflow/models/pull/4840, but I never got it in. You're absolutely right that this is bad. Thanks for pointing out that this is still an issue.

PR #4687, opened 2018-07-03
Fix typo

hsm207:


MarkDaoust:
Oh, wait.... this is against the r1.8.0 branch.

We prefer not to edit branches other than master, If the problem still exists on master can you re-submit there?

Thanks either way!

PR #4695, opened 2018-07-04
Update eval_util.py: force metrics to file

dketterer:
This fixes a problem when running evaluation with run_once. Examples get evaluated and logging shows the result, but the summary_writer doesn't write to disk.
Force the summary_writer to write the changes to file.


dketterer:
I signed it!

PR #4711, opened 2018-07-07
Small typo correction

Ty-Won:
Changed line 74 from "There many Tensorflow APIs"->"There are many Tensorflow APIs"


Ty-Won:
I Signed it!

Issue #4656, opened 2018-06-29
Asking users to change their PYTHONPATH in order to run samples is not necessary


martin-gorner:
### System information
- **What is the top-level directory of the model you are using**: official
- **Have I written custom code (as opposed to using a stock example script provided in TensorFlow)**: NO
- **OS Platform and Distribution (e.g., Linux Ubuntu 16.04)**: any
- **TensorFlow installed from (source or binary)**: binary
- **TensorFlow version (use command below)**: 1.8
- **Bazel version (if compiling from source)**: N/A
- **CUDA/cuDNN version**: N/A
- **GPU model and memory**: N?A
- **Exact command to reproduce**: see below

Current sample model documentation require users to change their PYTHONPATH before they can run the samples. This friction is not necessary. All samples work well if launched from the directory above "official". For example:

git clone https://github.com/tensorflow/models.git
cd models
python official/mnist/mnist.py

Just works. No need to change a PYTHONPATH.
Looking through the samples, all "local" imports have been consistently written as:
from official.some_model import some_feature

Therefore this change will work across all official samples.

martin-gorner:
Here is where we aske users to change their PYTHONPAT:
https://github.com/tensorflow/models/tree/master/official#requirements

martin-gorner:
Right now the documentation states that changing your PYTHONPATH is
required. This is definitely more friction than some users are comfortable
with.

martin-gorner:
Update: actually running

cd models
python official/mnist/mnist.py

will not work on all systems. It assumes that the current directory is on the PYTHONPATH but that is not always true. Python only adds the directory of the script (official/mnist/) to PYTHONPATH automatically but that is all. It does NOT add the curretn directory automatically. On my machine, I have PYTHONPATH set to "/usr/local/bin:" and the trailing ":" actually inserts and empty string "" to PYTHONPATH which Python interprets as the current directory.

k-w-w:
Removing the requirement for users to add the official models directory to the python path would definitely improve the usability. Although sounds like running from the models directory would still require modifications to the path. 

I wonder if we can use symbolic links to remove the need to make the official module a package... not sure. @tensorflow/tf-garden-team, what do you all think?

PR #4727, opened 2018-07-09
Update readme of adding python path

yhliang2018:
Hi Kathy,

There is a github issue about adding Python path in the official/README file: https://github.com/tensorflow/models/issues/4656

So I update the texts a little bit to make it clear. Feel free to re-word it. Thanks!

### DIFF
--- a/official/README.md
+++ b/official/README.md
@@ -14,10 +14,24 @@ If you are on a version of TensorFlow earlier than 1.4, please [update your inst
 ## Requirements
 Please follow the below steps before running models in this repo:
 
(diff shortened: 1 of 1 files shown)


tfboyd:
Over a quarter old, the adding of commits from other people broke the CLA and I doubt the issue will get resolved.  ping me if you want it reopened.  Cleaning up anything touching /official.

@tfboyd :  Do a quick test and update the read me manually if this approach works for the resnet example.  I have doubts given the absolute paths used to reference modules.  I think mnist just happens to work due to a lack of using other modules.

PR #4730, opened 2018-07-10
Update deep speech model with pure tensorflow API implementation

A30041839:
Hi,

This PR contains the final version of DeepSpeech 2 model implemented with tensorflow high level API. Changes included in this PR:
1) Add a README file and requirements.txt (yanhui & haoliang).
2) Code refactoring across the whole pipeline to improve code readability (yanhui).
3) Use tensorflow high level API to replace the previous Keras implementation (yanhui & haoliang)

The main reason we switched to TF API is that Kera's BatchNormalization layer could not be handled correctly by tf.keras.model_to_estimator. The issue we found is:
      a) The model_to_estimator function fails to add update_ops as a control dependency of the train_op. 
      b) Even if we write our own model_fn and add update_ops manually, with the current tf-nightly build, update_ops associated with Kera's BatchNormalization layer is not added to the tf.GraphKeys.UPDATE_OPs. So it's impossible for us to find a graceful solution to make Kera's BN layer work seamlessly with estimator.

For model evaluation, we obtain the best result from a 5-layer RNN with 800 hidden units. The network is trained on train-clean-100 and train-clean-360 dataset. At the 16000 steps, we reach WER of 0.21 (compared with mlperf pytorch 0.23).

Thanks a lot for reviewing this PR!

### DIFF
--- a/research/deep_speech/README.md
+++ b/research/deep_speech/README.md
@@ -1,0 +1,43 @@
+# DeepSpeech2 Model
+## Overview
+This is an implementation of the [DeepSpeech2](https://arxiv.org/pdf/1512.02595.pdf) model. Current implementation is based on the code from the authors' [DeepSpeech code](https://github.com/PaddlePaddle/DeepSpeech) and the implementation in the [MLPerf Repo](https://github.com/mlperf/reference/tree/master/speech_recognition).
+
+DeepSpeech2 is an end-to-end deep neural network for automatic speech
+recognition (ASR). It consists of 2 convolutional layers, 5 bidirectional RNN
+layers and a fully connected layer. The feature in use is linear spectrogram
+extracted from audio input. The network uses Connectionist Temporal Classification (CTC) as the loss function.
+
+## Dataset
+The [OpenSLR LibriSpeech Corpus](http://www.openslr.org/12/) are used for model training and evaluation.
+
+The training data is a combination of train-clean-100 and train-clean-360 (~130k
+examples in total). The validation set is dev-clean which has 2.7K lines.
+The download script will preprocess the data into three columns: wav_filename,
+wav_filesize, transcript. data/dataset.py will parse the csv file and build a
+tf.data.Dataset object to feed data. Within each training cycle (except for the
+first if sortagrad is enabled), the training data will be shuffled batch-wise.
+
+## Running Code
+
+### Download and preprocess dataset
+To download the dataset, issue the following command:
+```
+python data/download.py
+```
+Arguments:
+  * `--data_dir`: Directory where to download and save the preprocessed data. By default, it is `/tmp/librispeech_data`.
+
+Use the `--help` or `-h` flag to get a full list of possible arguments.
+
+### Train and evaluate model
+To train and evaluate the model, issue the following command:
+```
+python deep_speech.py
+```
+Arguments:
+  * `--model_dir`: Directory to save model training checkpoints. By default, it is `/tmp/deep_speech_model/`.
+  * `--train_data_dir`: Directory of the training dataset.
+  * `--eval_data_dir`: Directory of the evaluation dataset.
+  * `--num_gpus`: Number of GPUs to use.
+
+There are other arguments about DeepSpeech2 model and training/evaluation process. Use the `--help` or `-h` flag to get a full list of possible arguments with detailed descriptions.
--- a/research/deep_speech/README.md
+++ b/research/deep_speech/README.md
@@ -5,7 +5,7 @@ This is an implementation of the [DeepSpeech2](https://arxiv.org/pdf/1512.02595.
 DeepSpeech2 is an end-to-end deep neural network for automatic speech
 recognition (ASR). It consists of 2 convolutional layers, 5 bidirectional RNN
 layers and a fully connected layer. The feature in use is linear spectrogram
-extracted from audio input. The network uses Connectionist Temporal Classification (CTC) as the loss function.
+extracted from audio input. The network uses Connectionist Temporal Classification [CTC](https://www.cs.toronto.edu/~graves/icml_2006.pdf) as the loss function.
 
 ## Dataset
 The [OpenSLR LibriSpeech Corpus](http://www.openslr.org/12/) are used for model training and evaluation.
@@ -14,10 +14,27 @@ The training data is a combination of train-clean-100 and train-clean-360 (~130k
 examples in total). The validation set is dev-clean which has 2.7K lines.
 The download script will preprocess the data into three columns: wav_filename,
 wav_filesize, transcript. data/dataset.py will parse the csv file and build a
-tf.data.Dataset object to feed data. Within each training cycle (except for the
+tf.data.Dataset object to feed data. Within each epoch (except for the
 first if sortagrad is enabled), the training data will be shuffled batch-wise.
 
 ## Running Code
+
+### Configure Python path
+Add the top-level /models folder to the Python path with the command:
+```
+export PYTHONPATH="$PYTHONPATH:/path/to/models"
+```
+
+### Install dependencies
+
+First install shared dependencies before running the code. Issue the following command:
+```
+pip3 install -r requirements.txt
+```
+or
+```
+pip install -r requirements.txt
+```
 
 ### Download and preprocess dataset
 To download the dataset, issue the following command:
@@ -38,6 +55,7 @@ Arguments:
   * `--model_dir`: Directory to save model training checkpoints. By default, it is `/tmp/deep_speech_model/`.
   * `--train_data_dir`: Directory of the training dataset.
   * `--eval_data_dir`: Directory of the evaluation dataset.
-  * `--num_gpus`: Number of GPUs to use.
+  * `--num_gpus`: Number of GPUs to use (specify -1 if you want to use all available GPUs).
 
 There are other arguments about DeepSpeech2 model and training/evaluation process. Use the `--help` or `-h` flag to get a full list of possible arguments with detailed descriptions.
+
--- a/research/deep_speech/data/dataset.py
+++ b/research/deep_speech/data/dataset.py
@@ -128,7 +128,7 @@ def _preprocess_data(file_path):
   tf.logging.info("Loading data set {}".format(file_path))
   with tf.gfile.Open(file_path, "r") as f:
     lines = f.read().splitlines()
-  # Skip the tsv header in lines[0].
+  # Skip the csv header in lines[0].
   lines = lines[1:]
   # The metadata file is tab separated.
   lines = [line.split("\t", 2) for line in lines]
--- a/research/deep_speech/data/dataset.py
+++ b/research/deep_speech/data/dataset.py
@@ -17,13 +17,14 @@ from __future__ import absolute_import
 from __future__ import division
 from __future__ import print_function
 
-import functools
-import multiprocessing
-
+import math
+import random
+# pylint: disable=g-bad-import-order
 import numpy as np
-import scipy.io.wavfile as wavfile
 from six.moves import xrange  # pylint: disable=redefined-builtin
+import soundfile
 import tensorflow as tf
+# pylint: enable=g-bad-import-order
 
 import data.featurizer as featurizer  # pylint: disable=g-bad-import-order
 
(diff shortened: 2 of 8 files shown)


research/deep_speech/README.md
```diff
@@ -0,0 +1,43 @@
...
+
+The training data is a combination of train-clean-100 and train-clean-360 (~130k
+examples in total). The validation set is dev-clean which has 2.7K lines.
+The download script will preprocess the data into three columns: wav_filename,
+wav_filesize, transcript. data/dataset.py will parse the csv file and build a
+tf.data.Dataset object to feed data. Within each training cycle (except for the
+first if sortagrad is enabled), the training data will be shuffled batch-wise.
+
+## Running Code
+
```
yhliang2018: I see you have a requirement.txt file for the third-party library dependencies. Could you add a section here to ask users to install them first? Something like:
### Install dependencies
First install shared dependencies before running the code. Issue the following command:
`pip3 install -r requirements.txt`
or
`pip install -r requirements.txt`

research/deep_speech/deep_speech.py
```diff
@@ -17,6 +17,7 @@
 from __future__ import division
 from __future__ import print_function
 
+
```
yhliang2018: Could you add a random seed to the main loop?

research/deep_speech/deep_speech.py
```diff
@@ -41,53 +42,86 @@
...
+def ctc_loss(args):
+  """Computes the ctc loss for the current batch of predictions."""
+  label_length, ctc_input_length, labels, logits = args
+  label_length = tf.to_int32(tf.squeeze(label_length))
+  ctc_input_length = tf.to_int32(tf.squeeze(ctc_input_length))
+  sparse_labels = tf.to_int32(
+      tf.keras.backend.ctc_label_dense_to_sparse(labels, label_length))
+  y_pred = tf.log(tf.transpose(
+      logits, perm=[1, 0, 2]) + tf.keras.backend.epsilon())
```
qlzh727: What's the reason to add a small epsilon here?

research/deep_speech/deep_speech.py
```diff
@@ -103,45 +137,76 @@ def evaluate_model(
...
+    predictions = {
+        "classes": tf.argmax(logits, axis=2),
+        "probabilities": tf.nn.softmax(logits),
+        "logits": logits
+    }
+    return tf.estimator.EstimatorSpec(
+        mode=mode,
+        predictions=predictions)
+
+  # In training mode.
```
qlzh727: I guess you don't handle the EVAL model here, which probably should throw exception if the mode == EVALUATE.

research/deep_speech/deep_speech.py
```diff
@@ -41,53 +42,86 @@
...
+      given dataset.
     input_fn_eval: data input function for evaluation.
 
   Returns:
     Evaluation result containing 'wer' and 'cer' as two metrics.
   """
   # Get predictions
-  predictions = estimator.predict(
-      input_fn=input_fn_eval, yield_single_examples=False)
+  predictions = estimator.predict(input_fn=input_fn_eval)
```
qlzh727: What's the reason to use predict instead of eval method here? Is that because you can generate proper eval metric here?

research/deep_speech/deep_speech_model.py
```diff
@@ -12,183 +12,173 @@
...
-    ctc_loss = tf.keras.layers.Lambda(
-        _ctc_lambda_func, output_shape=(1,), name="ctc_loss")(
-            [y_pred, labels, ctc_input_length, label_length])
-
-    super(DeepSpeech, self).__init__(
-        inputs=[input_data, labels, input_length, label_length],
-        outputs=[ctc_input_length, ctc_loss, y_pred])
+      inputs = _rnn_layer(
+          inputs, rnn_cell, self.rnn_hidden_size, layer_counter + 1,
+          is_batch_norm, self.is_bidirectional, training=training)
```
qlzh727: nit: probably can skip "training=" since its not a named variable.

research/deep_speech/data/featurizer.py
```diff
@@ -19,42 +19,66 @@
...
+  if max_freq > sample_rate / 2:
+    raise ValueError("max_freq must not be greater than half of sample rate.")
 
-  # Perform transpose to set its shape as [time_steps, feature_num_bins]
-  spectrogram = np.transpose(np.absolute(stft), (1, 0))
-  return spectrogram
+  if stride_ms > window_ms:
+    raise ValueError("Stride size must not be greater than window size.")
+
+  stride_size = int(0.001 * sample_rate * stride_ms)
```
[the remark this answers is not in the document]
A30041839: sample_rate is particularly very large, such as 16000. Basically stride_size is number of samples per stride_ms, and should always be greater than 1. If somehow it reaches 0, I guess that's a mis-configuration and spectro-extract will probably fail.
Our default configuration for those are:
sample_rate: 16000
stride_ms: 10ms

research/deep_speech/decoder.py
```diff
@@ -18,189 +18,78 @@
...
+  def __init__(self, labels, blank_index=28):
     """Decoder initialization.
 
     Arguments:
-      labels (string): mapping from integers to characters.
-      blank_index (int, optional): index for the blank '_' character.
-        Defaults to 0.
-      space_index (int, optional): index for the space ' ' character.
+      labels: a string specifying the speech labels for the decoder to use.
+      blank_index: an integer specifying index for the blank character.
```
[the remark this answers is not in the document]
A30041839: '_' appears as the last element of our vocabulary as a pure placeholder character for blank output, so this is not really meant for outputting an underscore:) To work with CTC loss, it expects that we manually insert this placeholder into our vocabulary. In the last fully connected layer, the network will learn to output a blank symbol if there is no enough meaningful features at a time frame. For example, if I utter something like 'hello' and with the utterance may contain some disconnections, this will be captured by the network as an output form like 'h-ell-o'. The special '-' sign will be trimmed by CTC loss function when computing the actual loss.

research/deep_speech/deep_speech.py
A30041839: Adding an epsilon will prevent tf.log to occur NaN error if logits are too small (for example 0).

research/deep_speech/deep_speech.py
A30041839: Yeah, we will use the result (logits) from predict to compute metrics ourselves. Since we haven't implemented this in eval mode, so here we will just call predict.

research/deep_speech/deep_speech.py
A30041839: Discussed offline, we don't need setting random seeds as it's behavior is not fully expected.

research/deep_speech/README.md
```diff
@@ -0,0 +1,43 @@
+# DeepSpeech2 Model
+## Overview
+This is an implementation of the [DeepSpeech2](https://arxiv.org/pdf/1512.02595.pdf) model. Current implementation is based on the code from the authors' [DeepSpeech code](https://github.com/PaddlePaddle/DeepSpeech) and the implementation in the [MLPerf Repo](https://github.com/mlperf/reference/tree/master/speech_recognition).
+
+DeepSpeech2 is an end-to-end deep neural network for automatic speech
+recognition (ASR). It consists of 2 convolutional layers, 5 bidirectional RNN
+layers and a fully connected layer. The feature in use is linear spectrogram
+extracted from audio input. The network uses Connectionist Temporal Classification (CTC) as the loss function.
```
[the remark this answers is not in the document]
A30041839: Appending a link to paper.

research/deep_speech/data/dataset.py
```diff
@@ -33,40 +34,36 @@ class AudioConfig(object):
...
 
-  def __init__(self, audio_config, data_path, vocab_file_path):
+  def __init__(self, audio_config, data_path, vocab_file_path, sortagrad):
     """Initialize the configs for deep speech dataset.
 
     Args:
       audio_config: AudioConfig object specifying the audio-related configs.
       data_path: a string denoting the full path of a manifest file.
       vocab_file_path: a string specifying the vocabulary file path.
+      sortagrad: a boolean to control whether sorting the audio in the first
```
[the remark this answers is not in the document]
A30041839: Add more comments:)

research/deep_speech/data/dataset.py
```diff
@@ -178,22 +154,48 @@ def __init__(self, dataset_config):
...
+  Returns:
+    The shuffled data entries.
+  """
+  shuffled_entries = []
+  if epoch_index == 0 and sortagrad:
+    # No need to shuffle.
+    shuffled_entries = entries
+  else:
+    # Shuffle entries batch-wise.
+    max_buckets = int(math.ceil(len(entries) / batch_size))
```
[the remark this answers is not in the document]
A30041839: Sure. I choose to shuffle on full batches and just append the remaining batch into the result.

research/deep_speech/deep_speech.py
A30041839: Raise error here instead.

research/deep_speech/data/dataset.py
```diff
@@ -126,19 +122,20 @@ def _preprocess_data(file_path):
...
-    A list of tuples (audio_file_path, file_size, transcript) sorted by file
-    size.
+    A list of tuples (wav_filename, wav_filesize, transcript) sorted by
+    file_size.
   """
   tf.logging.info("Loading data set {}".format(file_path))
   with tf.gfile.Open(file_path, "r") as f:
     lines = f.read().splitlines()
-  # Skip the csv header in lines[0].
+  # Skip the tsv header in lines[0].
```
yhliang2018: what is tsv?

A30041839: tab-separated values? https://en.wikipedia.org/wiki/Tab-separated_values

research/deep_speech/data/dataset.py
```diff
@@ -178,22 +154,48 @@ def __init__(self, dataset_config):
...
+    The shuffled data entries.
+  """
+  shuffled_entries = []
+  if epoch_index == 0 and sortagrad:
+    # No need to shuffle.
+    shuffled_entries = entries
+  else:
+    # Shuffle entries batch-wise.
+    max_buckets = int(math.ceil(len(entries) / batch_size))
+    total_buckets = [i for i in xrange(max_buckets)]
```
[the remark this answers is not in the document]
yhliang2018: Sadly, I think here we cannot guarantee that the dataset length can be dividable by batch_size.

research/deep_speech/data/featurizer.py
```diff
@@ -23,10 +23,14 @@
...
-  """Compute the spectrograms for the input samples(waveforms)."""
+  """Compute the spectrograms for the input samples(waveforms).
+
+  More about spectrogram computation, please refer to:
+  https://en.wikipedia.org/wiki/Short-time_Fourier_transform.
+  """
   if max_freq is None:
-    max_freq = sample_rate / 2
-  if max_freq > sample_rate / 2:
+    max_freq = sample_rate // 2
```
yhliang2018: Do we need max_freq to be integer here?

research/deep_speech/data/featurizer.py
```diff
@@ -112,3 +116,4 @@ def __init__(self, vocab_file):
       self.index_to_token[index] = line
       self.speech_labels += line
       index += 1
+
```
yhliang2018: nit: remove the extra blank line.

qlzh727:
Thanks for the update. This overall looks good, and i think its good to checkin to meet the tight timeline here. We still have chance to address issue in future if needed.

research/deep_speech/data/featurizer.py
A30041839: I think so.. frequency should be treated as an integer.

research/deep_speech/deep_speech.py
yhliang2018: The EVAL mode will be handled by the following codes if someone calls estimator.evaluate(), and it just has no eval metrics (which is not mandatory in eval mode). I don't think raise error is a good idea.

A30041839:
Thanks Yanhui & Scott for your review! I appreciate that we can check-in code to meet timeline first.

research/deep_speech/data/dataset.py
yhliang2018: I think csv is more general and understandable. We can use csv here to keep consistent with line 122.

research/deep_speech/data/featurizer.py
yhliang2018: But if you check line 61 of variable freqs, it's a floating number. Could you confirm this part? I think we need to make it concise.

PR #4740, opened 2018-07-11
Use six and feature detection in string conversion

cclauss:
Leverage [__six.ensure_text()__](https://github.com/benjaminp/six/blob/master/six.py#L890) to deliver Unicode text in both Python 2 and Python 3.

Follow Python porting best practice [use feature detection instead of version detection](https://docs.python.org/3/howto/pyporting.html#use-feature-detection-instead-of-version-detection) in ___unicode_to_native()__.

### DIFF
--- a/official/transformer/utils/tokenizer.py
+++ b/official/transformer/utils/tokenizer.py
@@ -202,17 +202,14 @@ def _load_vocab_file(vocab_file, reserved_tokens=None):
 
 def _native_to_unicode(s):
   """Convert string to unicode (required in Python 2)."""
-  if six.PY2:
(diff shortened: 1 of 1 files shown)


karmel:
Thanks, @cclauss. @robieta , you let one by! Sacrilege!

robieta:
@karmel I've brought shame to my family!
@cclauss it looks like ensure_text() hasn't made it into the python 2 six PyPi package, so I don't think we can use that yet. So we can either remove that change or wait until it is in the package and then set a `six>=x.y.z` in requirements.txt.

PR #4748, opened 2018-07-11
update module to tensorflow 1.9.0

ricoms:
Update cifar10_estimator example to tensorflow 1.9.0


ricoms:
I filled a Google Individual Contributor License Agreement.

tfboyd:
Thanks @ricoms.  I cannot believe this was written back with 1.2. It feels so long ago but also not long ago at all.

PR #4752, opened 2018-07-12
Removing research models.

riteshkumarshah:
Pulling available TursorFlow Java samples


cclauss:
What is the reason for this change?

karmel:
We remove research models from tagged releases, but they should remain in master indefinitely. Closing this under the assumption that it was made in error.

PR #4749, opened 2018-07-11
Add notebook fixing "wide" tutorial.

MarkDaoust:
Colab staging link: 

https://colab.sandbox.google.com/github/lamberta/models/blob/mark-estimator-wide-2/samples/core/tutorials/estimators/linear.ipynb

fixes: #18929

PR #4762, opened 2018-07-13
Add shorter timeout for GCP util.

qlzh727:
This is trying to address the long pause in https://github.com/tensorflow/models/pull/4749.

### DIFF
--- a/official/utils/logs/cloud_lib.py
+++ b/official/utils/logs/cloud_lib.py
@@ -25,6 +25,8 @@ GCP_METADATA_HEADER = {"Metadata-Flavor": "Google"}
 def on_gcp():
   """Detect whether the current running environment is on GCP."""
   try:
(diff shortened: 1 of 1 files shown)


official/utils/logs/cloud_lib.py
```diff
@@ -25,7 +25,8 @@
 def on_gcp():
   """Detect whether the current running environment is on GCP."""
   try:
-    response = requests.get(GCP_METADATA_URL, headers=GCP_METADATA_HEADER)
+    response = requests.get(
+        GCP_METADATA_URL, headers=GCP_METADATA_HEADER, timeout=5)
```
karmel: What unit is 5 in? Please add comment describing why this is there, or someone will happily remove it next time they wished they had a longer timeout.

qlzh727: Done. PTAL.

PR #4765, opened 2018-07-13
Correct prediction probabilities in custom training walkthrough

erssebaggala:
Correct the prediction probabilities for Iris setosa and Iris virginica 
to match the prediction values in the image


MarkDaoust:
Thanks for the fix.

PR #4792, opened 2018-07-17
Fix small mistakes in docstring.

SimeonZhang:
correct the description of padding or clipping tensor method.


SimeonZhang:
I signed it!

PR #4802, opened 2018-07-17
Performance overhaul of NCF Recommendation

robieta:
Firstly, the branch name is misleading; this has nothing to do with file buffered datasets. This branch has gone a very different direction than I expected.

Secondly, while I know that this is a very large PR, a lot of those lines are either documentation or machine generated files.

Summary of changes:
1) The TFRecords based data pipeline has been replaced by a tf.contrib.rpc.rpc() based one.
https://github.com/tensorflow/models/tree/feat/file_buffered_datasets/official/recommendation/data_server
This is the bulk of the PR

2) Vectorized and parallelized both preprocessing and evaluation steps. Wherever possible I moved operations out of Python and into NumPy.

3) Replaced the Keras model-to-estimator with a normal Estimator (which still uses the Keras layers).

4) Updated the MLPerf run script.

I still have to do a bit more delinting and adjustments to the unit tests; however I decided to open the PR now since you should be able to assess it in the mean time.


jsimsa:
1) Yeah the proto registration is both tricky and undocumented -- you need generate and link a C++ shared-object with the proto definition to your Python binary. See the initializer [here](https://github.com/tensorflow/tensorflow/blob/master/tensorflow/contrib/proto/python/kernel_tests/proto_op_test_base.py) for an example. Stopping when an empty batch is encountered is independent of this and raising `StopIteration` is appropriate.

2) `.from_tensor_slices().batch()` will be slow indeed, but the following is more idiomatic than `from_generator` and should be comparable in performance:

```
dataset = tf.data.Dataset.from_tensors(x).map(lambda x: tf.reshape(x, [-1, 10000, 1])).apply(tf.contrib.data.unbatch())
```

karmel:
Leaving for others to review, but I approve of your ascii art.

robieta:
@jsimsa That is annoying, but it strikes me that a larger problem is what happens when I want to run this on TPUs. (Which was a lot of the motivation for going to a GRPC based pipeline.) Even if *my* tensorflow can decode the proto, my understanding is that the dataset is going to get packaged up and shipped off to the TPU where I have no ability to inject into the runtime. (Not that I can use py_func there either)

@saeta Is is currently possible to use custom protos in datasets that run with cloud TPUs?
https://github.com/tensorflow/models/blob/7ca191804a5c8fe3cbad8416219d25c4f6b28c6e/official/recommendation/data_server/pipeline.py#L150-L166 (TL;DR of the proposed pipeline.)

robieta:
Alas, this PR is not obsolete. Closing.

robieta:
*now obsolete

PR #4817, opened 2018-07-18
Fix undefined variable error

pereirfe:
The usage of the variable 'base_name' raise a NameError since it was never defined.
The value which 'base_name' is expected to have is contained in 'checkpoint_name' 
The solution was to change the variable name 'base_name' to 'checkpoint_name'


pereirfe:
I signed it!

jaeyounkim:
Fixed

PR #4823, opened 2018-07-18
Use print function for Python 3 compatibility

gar1t:


derekjchow:
I actually have a fix that uses logging.info checked into our internal version. It will be in our next Github release.

PR #4825, opened 2018-07-18
Add eager for keras benchmark

yhliang2018:
Hi All,

Could you help to review this PR? This PR is created to refine the keras model benchmark by:
1) Use models_helper for synthetic dataset generation
2) Add eager mode for keras model. 

Some notes about eager mode: to run with eager mode, the optimizer should be TF native; Also, currently eager mode and keras multi_gpu_model are not compatible. With eager mode enabled, only one GPU is used even multiple GPUs are provided.

Thanks,
Yanhui

### DIFF
--- a/official/keras_application_models/benchmark_main.py
+++ b/official/keras_application_models/benchmark_main.py
@@ -73,9 +73,9 @@ def run_keras_model_benchmark(_):
     tf.logging.info("Using synthetic dataset...")
     dataset_name += "_Synthetic"
     train_dataset = dataset.generate_synthetic_input_dataset(
-        FLAGS.model, FLAGS.num_imgs_train)
+        FLAGS.model, FLAGS.batch_size)
     val_dataset = dataset.generate_synthetic_input_dataset(
-        FLAGS.model, FLAGS.num_imgs_eval)
+        FLAGS.model, FLAGS.batch_size)
   else:
     raise ValueError("Only synthetic dataset is supported!")
 
@@ -85,15 +85,23 @@ def run_keras_model_benchmark(_):
     model = tf.keras.utils.multi_gpu_model(model, gpus=num_gpus)
 
   # Configure the model
+  if FLAGS.eager:
+    # Only TF native optimizers are supported in Eager mode
+    optimizer = tf.train.AdamOptimizer()
+  else:
+    optimizer = "Adam"
+
   model.compile(loss="categorical_crossentropy",
-                optimizer="sgd",
+                optimizer=optimizer,
                 metrics=["accuracy"])
 
   # Create benchmark logger for benchmark logging
   run_params = {
       "batch_size": FLAGS.batch_size,
       "synthetic_data": FLAGS.use_synthetic_data,
-      "train_epochs": FLAGS.train_epochs
+      "train_epochs": FLAGS.train_epochs,
+      "num_train_images": FLAGS.num_imgs_train,
+      "num_eval_images": FLAGS.num_imgs_eval,
   }
 
   benchmark_logger = logger.get_benchmark_logger()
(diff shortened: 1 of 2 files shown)


official/keras_application_models/benchmark_main.py
```diff
@@ -54,6 +58,10 @@ def run_keras_model_benchmark(_):
     raise AssertionError("The --model command line argument should "
                          "be a key in the `MODELS` dictionary.")
 
+  # Check if eager mode is enabled
```
karmel: Can we add the caveats you mentioned in the PR message to the comments somewhere too? Maybe to the eager flag help.

official/keras_application_models/benchmark_main.py
```diff
@@ -46,6 +46,10 @@
     # "nasnetmobile": tf.keras.applications.NASNetMobile,
 }
 
+# Number of synthetic images as the same number of images in ImageNet
+_IMAGENET_TRAIN_IMAGES = 1281167
+_IMAGENET_EVAL_IMAGES = 50000
```
karmel: I wouldn't bother making this complicated for synthetic data. We can re-evaluate if we add real data processing, but for now, maybe a single flag that defaults to 1000?

official/keras_application_models/benchmark_main.py
```diff
@@ -64,30 +71,31 @@ def run_keras_model_benchmark(_):
...
+        FLAGS.model, FLAGS.batch_size)
     val_dataset = dataset.generate_synthetic_input_dataset(
-        FLAGS.model, val_num_images)
+        FLAGS.model, FLAGS.batch_size)
   else:
     raise ValueError("Only synthetic dataset is supported!")
 
   # If run with multiple GPUs
+  # If eager mode is enabled, only one GPU is utilized even multiple GPUs
+  # are provided
```
anj-s: Do you want to raise a warning or error message if eager execution is enabled and we use multi_gpu_model?

official/keras_application_models/benchmark_main.py
```diff
@@ -146,6 +154,18 @@ def define_keras_benchmark_flags():
...
+      name="num_images", default=_NUM_IMAGES,
+      help=flags_core.help_wrap(
+          "The number of synthetic images for training and evaluation. The "
+          "default value is 1000."))
+
+  flags.DEFINE_boolean(
+      name="eager", default=False, help=flags_core.help_wrap(
+          "To enable eager execution. Note that if eager execution is enabled, "
+          "only one GPU is utilized even multiple GPUs are provided and "
+          "multi_gpu_model is used."))
```
anj-s: s/even/even if

official/keras_application_models/benchmark_main.py
```diff
@@ -146,6 +154,18 @@ def define_keras_benchmark_flags():
       help=flags_core.help_wrap(
           "Model to be benchmarked."))
 
+  flags.DEFINE_integer(
+      name="num_images", default=_NUM_IMAGES,
+      help=flags_core.help_wrap(
+          "The number of synthetic images for training and evaluation. The "
+          "default value is 1000."))
```
anj-s: Maybe I missed something, but is this the only place this constant is used? If so then do we just want the default to be specified here?

official/keras_application_models/benchmark_main.py
yhliang2018: Sure, I will add a warning message here. Then it keeps consistent with the help message in the --eager flag.

official/keras_application_models/benchmark_main.py
yhliang2018: Yeah, we use 1000 as the default number of images for both train and eval with synthetic dataset. Will just put a value here.

yhliang2018:
Hi Anjali and Karmel,

I have addressed all the comments. The benchmark is already set up. As long as this PR is approved, we should be able to get some benchmark numbers, like exp/sec.

Thanks,
Yanhui

official/keras_application_models/dataset.py
```diff
@@ -19,6 +19,9 @@
 
 import tensorflow as tf
 
+# pylint: disable=g-bad-import-order
```
karmel: This should be at the end of the line, or re-enabled after the relevant section

PR #4839, opened 2018-07-19
Add autograph keras layers example

MarkDaoust:
+ Use meaningful function names

Colab link to this PR:

https://colab.research.google.com/github/markdaoust/models/blob/autograph/samples/core/guide/autograph.ipynb


samples/core/guide/autograph.ipynb
```diff
@@ -609,6 +646,217 @@
...
+    {
+      "metadata": {
+        "id": "XY4UspHmZNdL",
+        "colab_type": "text"
+      },
+      "cell_type": "markdown",
+      "source": [
+        "## Interoperation with `tf.Keras`\n",
+        "\n",
+        "Now that you've seen the basics, let's build some real model components with autograph.\n",
```
lamberta: autograph => AutoGraph

samples/core/guide/autograph.ipynb
```diff
@@ -609,6 +646,217 @@
...
+        "\n",
+        "It's relatively simple to integrate `autograph` with `tf.keras`. But remember that batchng is essential for performance. So the best candidate code for conversion to autograph is code where the control flow is decided at the _batch_ level. If decisions are made at the individual _example_ level you will still need to index and batch your examples to maintain performance while appling the control flow logic. \n",
```
lamberta: Edit:
It's easy to use `autograph` with `tf.keras`, but batching is essential for performance. The best code to convert to AutoGraph is code where the control flow is decided at the *batch* level. If making decisions at the individual *example* level, you must index and batch the examples to maintain performance while applying the control flow logic.

samples/core/guide/autograph.ipynb
```diff
@@ -609,6 +646,217 @@
...
+        "\n",
+        "\n",
+        "### Stateless functions\n",
+        "\n",
+        "For stateless functions like `collatz`, below, the easiest way to include them in a keras model is to wrap them up as a layer uisng `tf.keras.layers.Lambda`."
```
lamberta: For stateless functions, like `collatz` shown below, the easiest way to include them in a Keras model is to wrap them up as a layer using `tf.keras.layers.Lambda`:

samples/core/guide/autograph.ipynb
```diff
@@ -609,6 +646,217 @@
...
+    {
+      "metadata": {
+        "id": "k9LEoa3ud9hA",
+        "colab_type": "text"
+      },
+      "cell_type": "markdown",
+      "source": [
+        "### Custom Layers and Models\n",
+        "\n",
+        "<!--TODO(markdaoust) link to full examples  or these referenced models.-->\n",
```
lamberta: TODO now or later?

samples/core/guide/autograph.ipynb
```diff
@@ -609,6 +646,217 @@
...
+        "\n",
+        "The easiest way to use autograph is keras layers and models is to `@autograph.convert()` the `call` method. See the [keras guide](https://tensorflow.org/guide/keras#build_advanced_models) for details on how to build on these classes. \n",
```
lamberta: The easiest way to use AutoGraph with Keras layers and models is to `@autograph.convert()` the `call` method. See the [TensorFlow Keras guide](https://tensorflow.org/guide/keras#build_advanced_models) for details to build these classes.

samples/core/guide/autograph.ipynb
```diff
@@ -620,9 +868,7 @@
         "\n",
         "Since writing control flow in AutoGraph is easy, running a training loop in a TensorFlow graph should also be easy.  \n",
         "\n",
-        "<!--TODO(markdaoust) link to examples showing autograph **in** keras models when ready-->\n",
-        "\n",
-        "Important: While this example wraps a `tf.keras.Model` using AutoGraph, `tf.contrib.autograph` is compatible with `tf.keras` and can be used in [Keras custom layers and models](https://tensorflow.org/guide/keras#build_advanced_models). The easiest way is to `@autograph.convert()` the `call` method.\n",
+        "Important: While this example wraps a `tf.keras.Model` using AutoGraph, `tf.contrib.autograph` is compatible with `tf.keras` and can be used in [Keras custom layers and models](https://tensorflow.org/guide/keras#build_advanced_models). \n",
```
lamberta: Use a DevSite callout, options are:
Note:, Caution:, Warning:, Key Point:, etc

samples/core/guide/autograph.ipynb
```diff
@@ -377,7 +379,7 @@
         "  result = fizzbuzz(num)\n",
         "  with tf.Session() as sess:\n",
         "    for n in range(10,16):\n",
```
mdanatg: Not related to this PR, but I feel that we should avoid examples that place control flow outside sess.run - in a way, AutoGraph was made so that one doesn't have to use that, which is a bit of an antipattern. For the point above, that converted code works like a regular op, I think we can use anything, like for instance taking the output and passing it through some other op, so we don't strictly need the fizzbuzz example for that. LMK what you think.

samples/core/guide/autograph.ipynb
```diff
@@ -415,14 +417,14 @@
       "cell_type": "code",
       "source": [
         "@autograph.convert()\n",
-        "def f(x):\n",
-        "  assert x != 0, 'Do not pass zero!'\n",
-        "  return x * x\n",
+        "def inverse(x):\n",
+        "  assert x != 0.0, 'Do not pass zero!'\n",
+        "  return 1.0/x\n",
```
mdanatg: Style nit: for expressions, the Google style guide recommends placing a space around operators, e.g. "1.0 / x"; I think it would be useful to be consistent with it in this notebook.

samples/core/guide/autograph.ipynb
```diff
@@ -415,14 +417,14 @@
...
         "\n",
         "with tf.Graph().as_default():  \n",
         "  with tf.Session():\n",
         "    try:\n",
-        "      print(f(tf.constant(0)).eval())\n",
+        "      print(inverse(tf.constant(0.0)).eval())\n",
```
mdanatg: could you use sess.run(inverse(...)) instead of .eval() here and elsewhere?

samples/core/guide/autograph.ipynb
mdanatg: I suggest keeping the observation about batching as a separate callout, e.g. in a separate paragraph with "Tip: remember that the inputs are usually batched examples, ...". The guidance is otherwise not strongly related to the keras interop.

samples/core/guide/autograph.ipynb
```diff
@@ -609,6 +646,217 @@
...
+        "    super().__init__(*args,**kwargs)\n",
+        "        \n",
+        "  def build(self,input_shape):\n",
+        "    super().build(input_shape.as_list())\n",
+        "    self.depth = len(self.layers)\n",
+        "    self.plims = np.linspace(self.pfirst, self.plast, self.depth+1)[:-1]\n",
+        "    \n",
+        "  @autograph.convert()\n",
+        "  def call(self, inputs):\n",
+        "    training = tf.cast(K.learning_phase(), dtype=bool)  \n",
```
mdanatg: FYI, we're looking into modifying the default behavior to avoid the need for these explicit casts.

samples/core/guide/autograph.ipynb
mdanatg: Another way this could be framed is that AutoGraph is compatible with Keras: AutoGraph code can be used inside Keras models and layers as well; Keras models can also be used in AutoGrahp code, as this example shows.

samples/core/guide/autograph.ipynb
MarkDaoust: Done

samples/core/guide/autograph.ipynb
MarkDaoust: > Edit:

Done. 

> not strongly related to the keras interop.

Moved up to the of examples.

samples/core/guide/autograph.ipynb
MarkDaoust: I see your point.

Here I'm trying to show that FizzBuzz is working, without getting into too-complex of an example.

This would be a step in the right direction:

```
@autograph.convert()
def fizzbuzz(num):
  if num % 3 == 0 and num % 5 == 0:
      return 'FizzBuzz'
  elif num % 3 == 0:
      return 'Fizz'
  elif num % 5 == 0:
      return 'Buzz'
  else:
      return tf.as_string(num)
    
@autograph.convert()
def fizzbuzz_loop(start, stop):
  for n in range(start, stop):
    print(fizzbuzz(n))
  return tf.constant(stop)

with tf.Graph().as_default():
  # The result works like a regular op: takes tensors in, returns tensors.
  # You can inspect the graph using tf.get_default_graph().as_graph_def()
  with tf.Session() as sess:
    sess.run(fizzbuzz_loop(10,16))
```

But `print` is currently not working...

samples/core/guide/autograph.ipynb
MarkDaoust: +1 we just don't have an integrated notebook linter (yet?).

Done.

samples/core/guide/autograph.ipynb
MarkDaoust: done

samples/core/guide/autograph.ipynb
MarkDaoust: Done

samples/core/guide/autograph.ipynb
MarkDaoust: When I can show that this performs as the paper: https://colab.research.google.com/drive/14YkY78YjBhqyKn9pQ50rC36hWkuWHFb7#scrollTo=nzst0_CH2uXV

samples/core/guide/autograph.ipynb
MarkDaoust: Done

samples/core/guide/autograph.ipynb
MarkDaoust: +1, good to know.

samples/core/guide/autograph.ipynb
MarkDaoust: !

I was sure "important" was one of the allowed ones, maybe that was in g3doc....

Fixed/reworded.

MarkDaoust: reworded, wdyt?

PR #4845, opened 2018-07-20
Fix notebook text

linbojin:
test data -> training data


linbojin:
Hi @MarkDaoust I agree, it's clear now.

PR #4852, opened 2018-07-20
A3C Blog Post

raymond-yuan:
Created PR for a3c for cartpole blog post.

### DIFF
--- a/research/a3c_blogpost/a3c_cartpole.py
+++ b/research/a3c_blogpost/a3c_cartpole.py
@@ -1,0 +1,390 @@
+import os
+os.environ["CUDA_DEVICE_ORDER"] = "PCI_BUS_ID"   # see issue #152
+os.environ["CUDA_VISIBLE_DEVICES"] = ""
+
+
+import threading
+import gym
+import multiprocessing
+import numpy as np
+from queue import Queue
+import argparse
+import matplotlib.pyplot as plt
+
+
+import tensorflow as tf
+from tensorflow.python import keras
+from tensorflow.python.keras import layers
+
+tf.enable_eager_execution()
+print("Eager execution: {}".format(tf.executing_eagerly()))
+
+def str2bool(v):
+  if v.lower() in ('yes', 'true', 't', 'y', '1'):
+    return True
+  elif v.lower() in ('no', 'false', 'f', 'n', '0'):
+    return False
+  else:
+    raise argparse.ArgumentTypeError('Boolean value expected.')
+
+
+parser = argparse.ArgumentParser(description='Run A3C algorithm on the game '
+                                             'Cartpole.')
+parser.add_argument('--algorithm', default='a3c', type=str,
+                    help='Choose between \'a3c\' and \'random\'.')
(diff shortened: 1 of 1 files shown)


research/a3c_blogpost/a3c_cartpole.py
```diff
@@ -0,0 +1,390 @@
...
+import argparse
+import matplotlib.pyplot as plt
+
+
+import tensorflow as tf
+from tensorflow.python import keras
+from tensorflow.python.keras import layers
+
+tf.enable_eager_execution()
+print("Eager execution: {}".format(tf.executing_eagerly()))
```
alextp: Redundant print?

research/a3c_blogpost/a3c_cartpole.py
```diff
@@ -0,0 +1,390 @@
...
+
+def str2bool(v):
```
alextp: I think this is nice on production code but not so nice on code that we expect people to read carefully. Just use type=bool

research/a3c_blogpost/a3c_cartpole.py
```diff
@@ -0,0 +1,390 @@
...
+  def call(self, inputs):
+    # Forward pass
+    x = self.dense1(inputs)
+    logits = self.policy_logits(x)
+    v1 = self.dense2(inputs)
+    values = self.values(v1)
+    return logits, values
+
+
+def tf_wrap(np_array, dtype=np.float32):
```
alextp: Do you need this? I think a tf.cast(np_array, tf.dtype) should do the trick if a tf.convert_to_tensor(np_array, tf_dtype) doesn't work

research/a3c_blogpost/a3c_cartpole.py
```diff
@@ -0,0 +1,390 @@
...
+
+  def train(self):
+    if args.algorithm == 'random':
+      random_agent = RandomAgent(self.game_name, args.max_eps)
+      random_agent.run()
+      return
+
+    res_queue = Queue()
+
+    # We run the algorithm on cpu only!
```
alextp: Why?

research/a3c_blogpost/a3c_cartpole.py
MarkDaoust: One option is:

```
parser.add_argument("--no_train", action="store_false", dest="train", default=True)
```

The only advantage is that it's a tiny bit less code.

research/a3c_blogpost/a3c_cartpole.py
```diff
@@ -0,0 +1,365 @@
+import os
+os.environ["CUDA_DEVICE_ORDER"] = "PCI_BUS_ID"   # see issue #152
```
alextp: Does the device order matter if we're setting visible devices to empty?

raymond-yuan: I was referring to https://github.com/keras-team/keras/issues/4613, but that line doesn't seem to affect anything. Removed.

alextp:
Will let Mark or Francois merge when they approve

PR #4871, opened 2018-07-23
Deep Contextual Bandits code for tensorflow/models

rikel:
The code corresponds to the following paper:

Deep Bayesian Bandits Showdown: An Empirical Comparison of Bayesian Deep Networks for Thompson Sampling (https://arxiv.org/abs/1802.09127), published in ICLR 2018.


rikel:
Thanks Lukasz!

PR #4874, opened 2018-07-24
Avoid hard example mining NMS when iou_threshold=1.0

irmowan:
When iou_threshold = 1.0, code will also do NMS but actually do nothing, as num_hard_examples == num_boxes. But this operation is very slow in CPU, and have a side effect on the train speed. So if iou_threshold = 1.0, we can simply skip the NMS operation.


pkulzc:
We have updated the nms op and this PR is obsolete.

PR #4880, opened 2018-07-24
Fixed error while converting between data formats.

tejaswid:
Fixed an error where data of type 'channels_last' was not being correctly transformed to 'channels_first'.

1. The if statement was incorrectly checking for the string 'channels_first' instead of checking for 'channels_last'.
2. After transposing the data the variable self.data_format has to be updated to 'channels_first' to reflect the changes made.


tejaswid:
I signed it!

tejaswid:
Please refer to the discussion in #4930

PR #4884, opened 2018-07-24
Update learning_schedules.py

VETURISRIRAM:
The following error doesn't occur with the above change in code.

Error: Argument must be a dense tensor: range(0, 3) - got shape [3], but wanted []

The range function on the vairable 'num_boundaries' should be a list! Please merge this request!


VETURISRIRAM:
Thank you for merging 😊

PR #4897, opened 2018-07-25
fix batch_size in transformer_main.py

theJiangYu:
fix batch_size in transformer_main.py which causes ResourceExhaustedError: OOM during training Transformer models using models/official/transformer

### DIFF
--- a/official/transformer/transformer_main.py
+++ b/official/transformer/transformer_main.py
@@ -555,9 +555,10 @@ def run_transformer(flags_obj):
 
   params["use_synthetic_data"] = flags_obj.use_synthetic_data
 
-  # Set batch size parameter, which depends on TPU and distribution settings.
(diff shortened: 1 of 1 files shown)


robieta:
Good catch!
Can you change it to
```
params["batch_size"] = (flags_obj.batch_size or (
    params["default_batch_size_tpu"] if params["use_tpu"] 
    else params["default_batch_size"]))
```
(so it will pass lint tests)

theJiangYu:
Format change is done. Thanks, Taylor (robieta) !

robieta:
Looks like the linter is still complaining. If you run `official/utils/testing/scripts/presubmit.sh` with no arguments it will lint the whole repo.

theJiangYu:
Removed trailing white space etc, and Linting test passed on my local machine. Please try again @robieta

PR #4902, opened 2018-07-26
Update object_detection_tutorial.ipynb

HalfLemon:
run_inference_for_single_image


pkulzc:
It's better not to add one more dependency.

PR #4918, opened 2018-07-27
Support batch inference for exported frozen models

dheera:
This PR allows exported models to be fed with batches of images for inference, e.g.
`    feed_dict = { "ImageTensor:0": [ image0, image1, image2, ... ] }`
Inference times for batches of size 1 are unaffected.

Sorry for the mess of commits -- I had to sync this to upstream/master since I had accidentally used dheera/master for another pull request in-progress and created another branch for this particular PR which is separate. This pull request contains only the code necessary for the batch inference. The only file changed should be export_model.py (see the Files changed section)


dheera:
@YknZhu @aquariusjay

aquariusjay:
Hi dheera,

Thanks for working on this!
This will be a useful feature. However, I think the current change may break some use cases. The main concern is that we still need the "input_preprocess.preprocess_image_and_label" to (1) resize the the input w.r.t. (min_resize_value, max_resize_value, resize_factor, and (2) pad the input to the desired crop size w.r.t. (crop_height, crop_width).
What you are doing currently is only resize input image to [513, 513], which is not sufficient enough.
Could you please try using the input_preprocess.preprocess_image_and_label? Maybe we also need some changes there.

Thanks,

jaeyounkim:
I am closing this PR since there is no update from dheera.

PR #4930, opened 2018-07-29
Fix/error in metrics order

tejaswid:
The order in which metrics were computed for distributed strategies was incorrect and had to interchanged.

Error with regards to data_format was also fixed. This was mentioned in a separate pull request.

### DIFF
--- a/official/resnet/resnet_model.py
+++ b/official/resnet/resnet_model.py
@@ -493,11 +493,12 @@ class Model(object):
     """
 
     with self._model_variable_scope():
-      if self.data_format == 'channels_first':
+      if self.data_format == 'channels_last':
         # Convert the inputs from channels_last (NHWC) to channels_first (NCHW).
         # This provides a large performance boost on GPU. See
         # https://www.tensorflow.org/performance/performance_guide#data_formats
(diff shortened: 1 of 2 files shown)


robieta:
I don't believe this is correct.

1) Data formats: The images come from the input function in NHWC, specifically (?, 224, 224, 3). So they need to be transposed when the convolutions expect NCHW. (So I believe the `self.data_format == 'channels_first'` condition is correct.)

2) Rewriting `self.data_format` in the conditional would cause future invocations of __call__ to behave incorrectly.

I believe you are correct regarding the ordering. (I'm surprised that it ever worked.) Metrics should now work with Distribution Strategies, so if you'd like to simply remove that conditional altogether feel free.

tejaswid:
OK. I think I understand.
The data_format is the format that the convolutional layers **expect the data** to be in.
The assumption being made in the code is that the input data is always in 'channels_last' format and so it is being transposed to 'channels_first' . Could you please confirm if this is correct?

robieta:
That is correct.

tejaswid:
I am closing this PR here without merging.
I have put up a new PR #4942  with the conditional over distribution strategies removed when computing metrics as suggested.
Thanks!

Issue #4865, opened 2018-07-23
NCF Recommendation training cannot converge when set the -layers parameter as 256,128,64


lingjiew:
Hi， 
I have tried the commit version eb4f6f5 of mlperf for a small model with layer parameter as 256,128,64, and it converged within 7 epochs.
But when I set the same -layers params in this project, the training cannot converged even in 20 epochs. 
Why the training is so different between this project and mlperf?

Here are my training params.
tensorflow/models:
python ncf_main.py -dataset ml-20m -hooks "" -hr_threshold 0.9562 -train_epochs 100 -learning_rate 0.0005 -batch_size 16384 -layers 256,128,64 -num_factors 64

mlperf:
python ncf.py ml-20m -l 0.0005 -b 16384 --layers 256 128 64 -f 64 --seed 1 --threshold 0.9562

lingjiew:
Thank you for your reply! Do you plan to improve the accuracy in short term?

robieta:
Yes, I've been working on a large scale rewrite of much of the NCF data pipeline that should be ready soon and help resolve the TF and PyTorch differences.

lingjiew:
OK! Thank you very much!

Issue #4787, opened 2018-07-16
[recommendation] protobuf::FatalException error


christ1ne:
Please go to Stack Overflow for help and support:

http://stackoverflow.com/questions/tagged/tensorflow

Also, please understand that many of the models included in this repository are experimental and research-style code. If you open a GitHub issue, here is our policy:

1. It must be a bug, a feature request, or a significant problem with documentation (for small docs fixes please send a PR instead).
2. The form below must be filled out.

**Here's why we have that policy**: TensorFlow developers respond to issues. We want to focus on work that benefits the whole community, e.g., fixing bugs and adding features. Support only helps individuals. GitHub also notifies thousands of people when issues are filed. We want them to see you communicating an interesting problem, rather than being redirected to Stack Overflow.

------------------------

### System information
- **What is the top-level directory of the model you are using**:  models/official/recommendation
- **Have I written custom code (as opposed to using a stock example script provided in TensorFlow)**: No
- **OS Platform and Distribution (e.g., Linux Ubuntu 16.04)**: Ubuntu 16.04
- **TensorFlow installed from (source or binary)**:
pip anaconda  https://storage.googleapis.com/tensorflow/linux/cpu/tensorflow-1.9.0-cp34-cp34m-linux_x86_64.whl
- **TensorFlow version (use command below)**: branch r1.9.0
- **Bazel version (if compiling from source)**: 1.9.0
- **CUDA/cuDNN version**: cuda 9.2 / cuDNN7.0.5
- **GPU model and memory**:  Titan Xp, 12192MiB
- **Exact command to reproduce**: python ncf_main.py --model_dir model_20m  --data_dir data --dataset ml-20m

### Describe the problem
Describe the problem clearly here. Be sure to convey here why it's a bug in TensorFlow or a feature request.

code stops running with error messages

### Source code / logs
Include any logs or source code that would be helpful to diagnose the problem. If including tracebacks, please include the full traceback. Large logs and files should be attached. Try to provide a reproducible test case that is the bare minimum necessary to generate the problem.

$ python
Python 2.7.15 |Anaconda custom (64-bit)| (default, May  1 2018, 23:32:55) 

$ python ncf_main.py --model_dir model_20m  --data_dir data --dataset ml-20m
[...]

PR #4935, opened 2018-07-30
NCF pipeline refactor (take 2) and initial TPU port.

robieta:
This PR moves data file generation into a side process, replaces lots of serial code with vectorized equivalents, and updates the MLPerf run scripts.

(https://github.com/tensorflow/models/issues/4865) Improved shuffling helps accuracy significantly.
(https://github.com/tensorflow/models/issues/4787) I observed a max memory use of 10.5 GB while testing this branch, so this should address the OOM errors occurring in master.

__CC__
@petermattson 
@reedwm 
@tayo
@codyaustun
@christ1ne


official/recommendation/neumf_model.py
```diff
@@ -33,102 +33,153 @@
...
+  items = tf.cast(features[movielens.ITEM_COLUMN], tf.int32)
+
+  logits = construct_model(users=users, items=items, params=params)
+
+  if mode == tf.estimator.ModeKeys.PREDICT:
+    predictions = {
+        movielens.RATING_COLUMN: logits,
+    }
+
+    if params["use_tpu"]:
```
bitfort: Generally I think we should constants instead of strings... but this isn't that objectionable compared to unblocking our perf teams.

PR #4943, opened 2018-07-30
removed optional input_shape

DecentGradient:


DecentGradient:
No problem. I had "tested" the unchanged code so i didn't realize it.
Just FYI .summary() works after .fit() even without input shape. But that doesn't help us.

PR #4949, opened 2018-07-31
update ssd_meta_arch.py to solve issue in #4940

yuezhilanyi:
to solve ssd_resnet_50_v1_fpn training problems: not available in checkpoint #4940 
update var_name

### DIFF
--- a/research/object_detection/meta_architectures/ssd_meta_arch.py
+++ b/research/object_detection/meta_architectures/ssd_meta_arch.py
@@ -867,7 +867,7 @@ class SSDMetaArch(model.DetectionModel):
         if var_name.startswith(self._extract_features_scope):
           if fine_tune_checkpoint_type == 'classification':
             var_name = (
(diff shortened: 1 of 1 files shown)


research/object_detection/meta_architectures/ssd_meta_arch.py
```diff
@@ -867,7 +867,7 @@ def restore_map(self,
         if var_name.startswith(self._extract_features_scope):
           if fine_tune_checkpoint_type == 'classification':
             var_name = (
-                re.split('^' + self._extract_features_scope + '/',
```
pkulzc: This is not the right thing to do. When loading from a classification checkpoint, this line prune the 'FeatureExtractor' substring so that your FeatureExtractor/resnet* vars can read from resnet* vars in classification checkpoint. 

My guess is that you are reading from a detection checkpoint and maybe you didn't set fine_tune_checkpoint_type to 'detection'?

PR #4954, opened 2018-07-31
Fix FSNS dataset set path construction on Windows.

alexgorban:


Abhirup-K:
The fix is isn't still overcoming the bug , that is the the dataset path isn't still accurate . 
I have come up with a minor change in your fix itself . 
Please review the changes @alexgorban .

PR #4965, opened 2018-08-01
Refactor object detection box predictors and fix some issues with model_main.

pkulzc:


research/object_detection/anchor_generators/multiscale_grid_anchor_generator.py
```diff
@@ -57,14 +57,12 @@ def __init__(self, min_level, max_level, anchor_scale, aspect_ratios,
     self._scales_per_octave = scales_per_octave
     self._normalize_coordinates = normalize_coordinates
 
+    scales = [2**(float(scale) / scales_per_octave)
+              for scale in xrange(scales_per_octave)]
```
dcyoung: py3 compatibility. Perhaps six? or are these small enough to just use a list?

PR #4972, opened 2018-08-01
Added github and colab links

raymond-yuan:
Sorry for the nit PR (again), was told it was finalized and then later asked to update.


alextp:
There's a conflict which needs to be resolved

PR #4973, opened 2018-08-01
updated the path for fsns dataset in the code  fsns.py

abhirupkamath:
Fixed the path for fsns dataset (DEFAULT_DATASET_DIR) on windows . 

The previous code would create an error : 

"C:\Users\DELL\Anaconda3\envs\python35DL\lib\site-packages\tensorflow\python\framework\errors_impl.py", line 519, in __exit__
    c_api.TF_GetCode(self.status.status))
tensorflow.python.framework.errors_impl.NotFoundError: NewRandomAccessFile failed to Create/Open .. The system cannot find the path specified .

Above changes have fixed this error.

### DIFF
--- a/research/attention_ocr/python/datasets/fsns.py
+++ b/research/attention_ocr/python/datasets/fsns.py
@@ -22,7 +22,7 @@ import tensorflow as tf
 from tensorflow.contrib import slim
 import logging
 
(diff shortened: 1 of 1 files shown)


jaeyounkim:
Fixed

Issue #4900, opened 2018-07-25
[Transformer] InvalidArgumentError at the beginning of training a transformer model on CPU 


theJiangYu:
It throws the following InvalidArgumentError at the beginning of training a Transformer model (models/official/transformer/transformer_main.py) using CPUs. 
The training steps is on: 
https://github.com/tensorflow/models/tree/master/official/transformer

Full Error Trace is as follows: 
2018-07-25 16:16:27.734555: I tensorflow/core/platform/cpu_feature_guard.cc:141] Your CPU supports instructions that this TensorFlow binary was not compiled to use: AVX2 AVX512F FMA
[...]

PR #4974, opened 2018-08-01
Change vocab_size from 33708 to 33945 in model_params.py

theJiangYu:
Change vocab_size from 33708 to 33945 in models/official/transformer/model/model_params.py
The original vocab_size causes InvalidArgumentError when training Transformer using CPU.

### DIFF
--- a/official/transformer/model/model_params.py
+++ b/official/transformer/model/model_params.py
@@ -27,7 +27,7 @@ BASE_PARAMS = defaultdict(
 
     # Model params
     initializer_gain=1.0,  # Used in trainable variable initialization.
-    vocab_size=33708,  # Number of tokens defined in the vocabulary file.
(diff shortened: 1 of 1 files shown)


k-w-w:
This might actually be a Python 2 vs Python 3 error. Can you paste the entire error stacktrace?

theJiangYu:
@k-w-w Hi, I used python 3
Please check the full error trace at: 
https://github.com/tensorflow/models/issues/4900

alexeyr:
I ran into this in `transformer/v2` too.

Zantares:
Is any update for this issue?
The training is going into an infinity loop with this PR, maybe we need another way or specify max step for the model if we want to run on CPU.

jaeyounkim:
I am closing this PR because we made many changes already.

PR #4989, opened 2018-08-02
Fix bug where data_async_generation.py would freeze.

reedwm:
The data_async_generation.py process would print to stderr, but the main process would redirect it's stderr to a pipe. The main process never read from the pipe, so when the pipe was full, data_async_generation.py would stall on a write to stderr. This change makes data_async_generation.py not write to stdout/stderr.

### DIFF
--- a/official/recommendation/data_async_generation.py
+++ b/official/recommendation/data_async_generation.py
@@ -23,7 +23,6 @@ import contextlib
 import datetime
 import gc
 import functools
-import logging
 import multiprocessing
 import json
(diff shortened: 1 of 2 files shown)


official/recommendation/data_async_generation.py
```diff
@@ -395,61 +396,54 @@ def _generation_loop(
 
 
 def main(_):
+  global _log_file
```
robieta: Yuk. Oh well.

PR #5012, opened 2018-08-06
Update config_util.py

ratandec:
Fix invalid attribute name while updating eval steps.


pkulzc:
Thanks for sending the PR! This commit actually will change the behaviour of our current code. We assume that when eval_steps is not provided, it evaluate all images(defined by num example) in the dataset.

I'm closing this request, please let me know if you have other questions. Thanks!

ratandec:
Thanks for your clarification.

I thought that eval steps corresponds to the "max evals" in eval_config and
I was convinced that it was a bug.

Could you, please give some details about this?


Best regards

Radoslaw Tandecki

pon., 6 sie 2018 o 19:54 pkulzc <notifications@github.com> napisał(a):

PR #5015, opened 2018-08-07
fix a bug

GeneralLi95:
change xrange with range to adapt python3


pkulzc:
Thanks for the fix. This has already been fixed internally and we will release it soon!

PR #5021, opened 2018-08-07
fix for constant extension length

sumsuddin:
Fixed the constant extension length which assumes extensions are always of length 3 but fails for other lengths. eg. for file types **tiff**.


YknZhu:
Thank you!

PR #5026, opened 2018-08-07
Update object detection post processing and fixes boxes padding/clipping issue.

pkulzc:
- Uses the new non_max_suppression op in tf 1.9 with extra score_threshold argument in post processing, so that data flow will not jump between CPU and GPU back and forth.
 - Fixes the tensor slicing [issue](https://github.com/tensorflow/tensorflow/issues/21320) when max_number_of_boxes is less than number of groundtruth boxes.


apacha:
With the changes from this PR, estimator-based training seems to work finally, but is training entirely on the CPU, despite allocating the entire GPU memory...

jsjs2009:
Now running object detection API must install Tensorflow1.9+?????? @pkulzc

pkulzc:
Right, the new nms ops is only available in tf 1.9 and it's very important in reducing inference time.

moyans:
the problem still exists

jsjs2009:
That is to say that the latest object detection API models only match with the tf 1.9??

pkulzc:
1.9 or newer should all work.

jsjs2009:
@pkulzc Thank you!

jsjs2009:
tf1.9 matches with python2  better or python3?

Jilliansea:
＠pkulzc I counter similar problem, and I use the new version of the api. 
When I train my data with SSD_Inception_v2，I counter problem as follow:
INFO:tensorflow:Error reported to Coordinator: <class 'tensorflow.python.framework.errors_impl.InvalidArgumentError'>, Expected size[0] in [0, 10], but got -1343651776
	 [[Node: Slice_9 = Slice[Index=DT_INT32, T=DT_FLOAT, _device="/job:localhost/replica:0/task:0/device:GPU:0"](unstack_16, Loss/Const_30, unstack_17)]]

Why is negative value？ My config about post_processing is like that:
post_processing {
      batch_non_max_suppression {
        score_threshold: 1e-8
        iou_threshold: 0.6
        max_detections_per_class: 100
        max_total_detections: 100
      }
      score_converter: SIGMOID
    }

PallawiSinghal:
> post_processing

hi were you able to solve the issue?

PR #5032, opened 2018-08-08
Fix a doc typo fashion-mnist example in #21199

HughKu:
Fix a doc typo in fashion-mnist example in [tensorflow #21199](https://github.com/tensorflow/tensorflow/issues/21199)


HughKu:
Signed it, thanks.

PR #5036, opened 2018-08-08
resnet fine tuning

zacwellmer:
warm start a resent with all but the dense layer and only update the final layer weights when fine tuning

### DIFF
--- a/official/resnet/cifar10_main.py
+++ b/official/resnet/cifar10_main.py
@@ -221,7 +221,8 @@ def cifar10_model_fn(features, labels, mode, params):
       resnet_version=params['resnet_version'],
       loss_scale=params['loss_scale'],
       loss_filter_fn=loss_filter_fn,
-      dtype=params['dtype']
+      dtype=params['dtype'],
+      fine_tune=params['fine_tune']
   )
 
 
--- a/official/resnet/imagenet_main.py
+++ b/official/resnet/imagenet_main.py
@@ -303,7 +303,8 @@ def imagenet_model_fn(features, labels, mode, params):
       resnet_version=params['resnet_version'],
       loss_scale=params['loss_scale'],
       loss_filter_fn=None,
-      dtype=params['dtype']
+      dtype=params['dtype'],
+      fine_tune=params['fine_tune']
   )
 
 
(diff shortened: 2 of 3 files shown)


zacwellmer:
I signed it!

karmel:
Thanks, @zacwellmer -- I like this as an example of how to use WarmStartSettings. That said, it's not clear to me that this is a generally useful change, and it does complicate an already complicated model. Can you clarify the motivation here? cc @robieta

zacwellmer:
As the resnet project stands there does not exist a way to:
1) load a pretrained model with a different number of output classes
2) freeze earlier layers for fine tuning
Both of the above cases are key features to users who want to implement a resnet on their own datasets. 

I'm open to approaching the above two cases in a different way if you have any other ideas. My original idea was to use the same approach that the [inception](https://github.com/tensorflow/models/tree/master/research/inception) repo uses for fine tuning. However, that repo makes heavy use of tf slim which isn't the case in the resnet official repo. In the end it seemed best to keep the same flags the same but solve the problem in a  different way.

karmel:
I think the changes you have made seem fine for your use-case, and I would suggest forking the official models and modifying the code as fits your use-case. If you can help us determine that this is a generally useful feature, rather than one that is not much requested, we can include the changes in the official version as well. How does that sound?

zacwellmer:
That sounds good to me. What would you like to see to help you determine if this is a generally useful feature? 

In terms motivation from research: The [DeCAF paper](https://arxiv.org/abs/1310.1531) was one of the earlier works on transfer learning.

In terms of motivations from a practical standpoint: If you passed the pretrained imagenet weights to the pretrained model checkpoint flag and set the fine_tune checkpoint to True. Then trained on a smaller dataset, you could expect better generalization than if you trained on the smaller dataset alone.

robieta:
I am always wary of adding more knobs to an already somewhat unwieldy model. However, this is a very modest change in terms of code complexity, and I think would be of general interest. (Both to fine tune models and as an example of how to "transfer learning-ize" a model.) So my vote would be for adding it. The one request that I would make is to add a blob in the README about what the fine tuning knob is and how to invoke it.

karmel:
That WFM-- @robieta , you can take it from here in that case.

zacwellmer:
I am closing this pull request b/c of issues git config issues and have reopened [here](https://github.com/tensorflow/models/pull/5047)

PR #5049, opened 2018-08-10
Update VariationalAutoencoder.py

### LINKED ISSUE #2243: Cost calculation in Variational Autoencoder model is wrong
### System information
- **What is the top-level directory of the model you are using**: autoencoder
- **OS Platform and Distribution (e.g., Linux Ubuntu 16.04)**: Ubuntu 14.04
- **TensorFlow installed from (source or binary)**: not relevant
- **TensorFlow version (use command below)**: not relevant
- **Bazel version (if compiling from source)**:not relevant
- **CUDA/cuDNN version**: not relevant
- **GPU model and memory**: not relevant
- **Exact command to reproduce**: not relevant

### Describe the problem
The [reconstruction error](https://github.com/tensorflow/models/blob/master/autoencoder/autoencoder_models/VariationalAutoencoder.py#L24) in variational autoencoder is calculated without specifying the "axis" argument, resulting in calculating the sum of reconstruction error across the whole batch. However, the [latent error](https://github.com/tensorflow/models/blob/master/autoencoder/autoencoder_models/VariationalAutoencoder.py#L25) is calculated with axis=1, resulting in a 1D vector representing the latent error for each sample in the batch. When we add them up and take the average to get the [cost](https://github.com/tensorflow/models/blob/master/autoencoder/autoencoder_models/VariationalAutoencoder.py#L26), the 0-d reconstruction error will get broadcast to each element of the latent error, resulting in a larger cost than it should be. The right way should be either mean + mean, or mean(vector + vector).

### Source code / logs
N/A

accraze:
Added `axis=1` for reconstruction error calculation. This will avoid taking the sum of cost so that the training objective will no longer be batch-size dependent.

Fixes #2243

### DIFF
--- a/research/autoencoder/autoencoder_models/VariationalAutoencoder.py
+++ b/research/autoencoder/autoencoder_models/VariationalAutoencoder.py
@@ -21,7 +21,7 @@ class VariationalAutoencoder(object):
         self.reconstruction = tf.add(tf.matmul(self.z, self.weights['w2']), self.weights['b2'])
 
         # cost
-        reconstr_loss = 0.5 * tf.reduce_sum(tf.pow(tf.subtract(self.reconstruction, self.x), 2.0))
+        reconstr_loss = 0.5 * tf.reduce_sum(tf.pow(tf.subtract(self.reconstruction, self.x), 2.0), 1)
         latent_loss = -0.5 * tf.reduce_sum(1 + self.z_log_sigma_sq
                                            - tf.square(self.z_mean)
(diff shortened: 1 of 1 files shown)


karmel:
CC @snurkabill -- can you take a look?

snurkabill:
hi. currently I can't do anything. Will be able to work on september

jaeyounkim:
@snurkabill Can you check this PR?

snurkabill:
done.

PR #5055, opened 2018-08-10
README for Domain Separation Network

KosukeArase:
Thank you for open sourcing your great work! However, the code have some problems.
It depends on a binary file that does not work on different architectures (ex. MacOS).
I also found problem in creating TFRecords, and  I fixed them.


jaeyounkim:
domain_adaptation is deprecated.

PR #5058, opened 2018-08-10
minor bug fix

raymond-yuan:
- fixed entropy sign 
- changed entropy implementation for numeric stability

### DIFF
--- a/research/a3c_blogpost/a3c_cartpole.py
+++ b/research/a3c_blogpost/a3c_cartpole.py
@@ -348,12 +348,12 @@ class Worker(threading.Thread):
 
     # Calculate our policy loss
     policy = tf.nn.softmax(logits)
-    entropy = -tf.nn.softmax_cross_entropy_with_logits_v2(labels=policy, logits=logits)
+    entropy = tf.nn.softmax_cross_entropy_with_logits_v2(labels=policy, logits=logits)
 
     policy_loss = tf.nn.sparse_softmax_cross_entropy_with_logits(labels=memory.actions,
                                                                  logits=logits)
     policy_loss *= tf.stop_gradient(advantage)
-    policy_loss -= 0.01 * entropy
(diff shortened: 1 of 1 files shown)


raymond-yuan:
I signed it!

research/a3c_blogpost/a3c_cartpole.py
```diff
@@ -347,15 +347,13 @@ def compute_loss(self,
...
-    entropy = tf.reduce_sum(policy * tf.log(policy + 1e-20), axis=1)
+    entropy = tf.nn.softmax_cross_entropy_with_logits_v2(labels=policy, logits=logits)
 
-    policy_loss = tf.nn.softmax_cross_entropy_with_logits_v2(labels=actions_one_hot,
-                                                             logits=logits)
+    policy_loss = tf.nn.sparse_softmax_cross_entropy_with_logits(labels=memory.actions,
+                                                                 logits=logits)
     policy_loss *= tf.stop_gradient(advantage)
-    policy_loss -= 0.01 * entropy
+    policy_loss = 0.01 * entropy
```
qlzh727: I think "policy_loss -= 0.01 * entropy" might be correct, overwriting the policy_loss does not make sense.

MarkDaoust:
@qlzh727 I think that negative is there because of the `-=` later on.
I removed the double negative.

But that didn't look right. This "entropy regularization" is supposed to _discourage_ premature convergence, so it should encourage a high entropy not a low entropy.

There seems to be a lot of confusion around this point in other tutorials on the subject, because of the way the paper is written.

I think this is right. I'm merging as is. 

@raymond-yuan LMK if you disagree.

research/a3c_blogpost/a3c_cartpole.py
MarkDaoust: You are totally correct. I fixed this before merging.

PR #5060, opened 2018-08-10
Fixes for Style transfer tutorial.

MarkDaoust:
Clear output, move buttons to the top, shrink images, keep image aspect ratio, add plots during optimization.

Staging:

https://colab.research.google.com/github/MarkDaoust/models/blob/nst_fixes/research/nst_blogpost/4_Neural_Style_Transfer_with_Eager_Execution.ipynb


MarkDaoust:
Some of the early commits include output so it's probably best to squash this on merge.

I've run this final-version end-to-end it completes with no errors.

alextp:
Mark, I don't know how to visualize this diff. Can you preview the before and after notebooks somewhere, or split this into smaller changes?

MarkDaoust:
You can now see the diff here: eaf3137

I've rewritten the PR's git history to make it more diff-able. 

This diff one of the reasons it's best to  avoid including notebook output git.

The first commit has no content changes: I loaded the file with colab, cleared the output, set the notebook to python3/GPU and saved it to github using colab.

The second commit (eaf3137) has the content changes. There there you can see the diff.

The other commits update the source images (3MB --> 30KB), and fix the aspect ratio on wave_turtle.png.

PR #5077, opened 2018-08-13
Update resnet_model.py

kangtop729:
There is a typing error.


kangtop729:
I signed it!

kangtop729:
I signed it!

PR #5084, opened 2018-08-14
Modify flag name for the checkpoint path

yesmung:
change flag name to checkpoint_dir according to the variable name
used by the checkpoint_utils within tensorflow python framework.

The important point is that when run the run_eval script, an error
occurs due to the different flag name.

Signed-off-by: MyungSung Kwak <yesmung@gmail.com>

### DIFF
--- a/research/learning_unsupervised_learning/run_eval.py
+++ b/research/learning_unsupervised_learning/run_eval.py
@@ -35,13 +35,13 @@ import sonnet as snt
 
 from tensorflow.contrib.framework.python.framework import checkpoint_utils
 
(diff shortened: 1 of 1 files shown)


yesmung:
@lukemetz, @nirum 
Hello, My name is Myungsung Kwak.

I found a different name for the variable that stores the checkpoint path.
It occurs an error while running run_eval.py in learning_unsupervised_learning.
So I changed the variable name that stores the checkpoint path according to checkpoint_utils
https://github.com/tensorflow/tensorflow/blob/master/tensorflow/contrib/framework/python/framework/checkpoint_utils.py

Please review and merge it :)

Thanks!

yesmung:
@nirum 
Is there any update?

PR #5092, opened 2018-08-14
Transformer partial fix

alope107:
Tensorflow raises an error when tf_inspect.getfullargspec is called on a functools.partial in Python 2.X. This issue would be hit during the eval stage of the Transformer TPU model. This change replaces the call to functools.partial with a lambda to work around the issue.


robieta:
Nooooo. I don't want to give up my functools.partial's. Oh well, if it must be.

PR #5094, opened 2018-08-15
Add a pointer to a Colab for the embeddings

MalcolmSlaney:
Added a pointer to a Colab that illustrates how to use the public AudioSet models to generate embeddings for user-specified sounds.


MalcolmSlaney:
I signed it

PR #5096, opened 2018-08-15
Update trainer.py - fix CUDNN_STATUS_INTERNAL_ERROR

giacomobartoli:
Some user still prefers to run concurrently legacy/train.py and legacy/eval.py instead of /model_main.py. What happens is that both files cannot be executed at the same time because the first one saturates the GPU memory. With just these 3 loc it is possible to avoid this problem.
At this [link](https://gist.github.com/giacomobartoli/694986a2d9a0d89bd9a92d58a3c1743f) you can find a gist of this reported error.


giacomobartoli:
I signed it!

PR #5098, opened 2018-08-15
Deterministic dataset order fix

JulesGM:
In order for the order of the files to be deterministic, in `tf.data.Dataset.list_files(..., shuffle)`, shuffle needs to be True, otherwise different iterator inits will yield different file orders, and, as such, different dataset orders

### DIFF
--- a/official/transformer/utils/dataset.py
+++ b/official/transformer/utils/dataset.py
@@ -58,8 +58,6 @@ import tensorflow as tf
 
 from official.utils.misc import model_helpers
 
-# Use the number of training files as the shuffle buffer.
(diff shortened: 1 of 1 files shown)


robieta:
Good catch. Presumably this can replace the `if shuffle` block below?

JulesGM:
This is correct, as `Dataset.list_files` already shuffles the files if shuffle is `True`.  Updated the patch.

JulesGM:
(accidentally closed)

robieta:
And could you delete the `_FILE_SHUFFLE_BUFFER` definition? Then it should be good.

JulesGM:
Sure, done.

PR #5102, opened 2018-08-16
DeepLab: Add Custom Dataset

jamesmcclain:
This allows one to use DeepLab on a custom dataset whose details are supplied via the environment.  The motivating use case is when the dataset(s) is (are) built by a tool rather than being a one-off with a well-known name.

### DIFF
--- a/research/deeplab/datasets/segmentation_dataset.py
+++ b/research/deeplab/datasets/segmentation_dataset.py
@@ -49,6 +49,7 @@ References:
   through ADE20K dataset", In Proc. of CVPR, 2017.
 """
 import collections
+import os
 import os.path
 import tensorflow as tf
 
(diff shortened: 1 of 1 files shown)


jamesmcclain:
This is a small change.  Is there any hope of discussing it?

jamesmcclain:
Hello,

> First of all, thanks for preparing this change.  I feel like that users should be careful about which dataset they are using (and even the dataset setting). 

Thank you for your response.

The specific use case that motivated this PR was the fact that I have some tooling that produces datasets which are broadly similar but differ in the particular values of the parameters referenced in the PR.  It is my understanding that there are a fair number of other people with similar use cases.

It seems to me that there are (at least three) possible resolutions *(a)* what I am doing is reasonable and this pull request is acceptable *(b)* What I am doing is reasonable, but this pull request is not warranted at this time, *(c)* what I am doing is incorrect and can perhaps be improved by doing the following (...).

I am okay with either of those three responses, but I would like to benefit from your experience so I am curious to know which one you would give.

Thanks

jamesmcclain:
> Thanks James for the PR! This change seems to benefit most for a particular use case (multiple similar datasets). Maybe creating a single python flag pointing to a protobuf/json file which contains dataset information? Your dataset creation tool can write the proto/json as well so that is less error-prone.

Thank you for the suggestion, I will close this PR.

jamesmcclain:
Perhaps I or someone else will implement the idea above.

PR #5105, opened 2018-08-16
Nits in basic_regression.ipynb

MaxGhenis:
* Remove trailing spaces
* Fix "dependant" spelling error
* Remove periods at the end of comments*
* Add spaces after commas*

\* For internal consistency and consistency with PEP8 examples


MaxGhenis:
I signed it!

lamberta:
Thanks! I just spent more time than I should have googling dependent vs dependant 😝 Not as clear cut as one would think.

PR #5108, opened 2018-08-16
Fix NasNet used hiddenstates to match https://arxiv.org/abs/1707.07012

guillaume-michel:
I think there is a bug in NasNet normal and reduction cell: used_hiddenstates does not match the authors article. The current code introduces additional skip connections from prev prev cell to concat for normal cell and from second block to concat for reduction cell that are not present in the paper. This claim is easily verified by observing the graph of the provided checkpoint in tensorboard.

There are 3 possibilities:
- The schemas in the paper are wrong and should be fixed. Minor changes in the text should describe the motivations for these additional connections.
- The code is not correct and the present request can be used to fix it.
- I am missing something and it would be helpful if someone could clarify for me.

In case the code is wrong, provided checkpoints should be fixed.

### DIFF
--- a/research/slim/nets/nasnet/nasnet_utils.py
+++ b/research/slim/nets/nasnet/nasnet_utils.py
@@ -450,6 +450,19 @@ class NasNetABaseCell(object):
     return net
 
 
+def get_used_hiddenstates(hiddenstate_indices):
+  assert len(hiddenstate_indices) % 2 == 0 # 2 indices per block
+
+  B = len(hiddenstate_indices) // 2
+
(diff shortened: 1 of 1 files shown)


guillaume-michel:
CLA for commit author has been submitted

gmichel-netatmo:
Confirmed

guillaume-michel:
Hi, do you have any update on the status of this request?

jaeyounkim:
@guillaume-michel I have confirmed that we can accept your code. 
@sguada Please review this pull request.

saberkun:
Slim is no longer maintained. Sorry for the delay.
There is no reviewer for this PR so we need to close. Thanks

Issue #5054, opened 2018-08-10
TensorFlow Slim: Inception V2 Confusion


dichotomies:
On https://github.com/tensorflow/models/tree/master/research/slim, `Inception V2` refers to [Batch Normalization: Accelerating Deep Network Training by Reducing Internal Covariate Shift](https://arxiv.org/abs/1502.03167).

This reference seems contrary to the definition of Inception V2 in the paper [Rethinking the Inception Architecture for Computer Vision](https://arxiv.org/pdf/1512.00567.pdf), because it is introduced there.

The paper you refer to is about BN-Inception if I'm not mistaken.

tensorflowbutler:
Thank you for your post. We noticed you have not filled out the following field in the issue template. Could you update them if they are relevant in your case, or leave them as N/A? Thanks.
What is the top-level directory of the model you are using
Have I written custom code
OS Platform and Distribution
TensorFlow installed from
TensorFlow version
Bazel version
CUDA/cuDNN version
GPU model and memory
Exact command to reproduce

karmel:
The Slim Inception model is no longer actively maintained. That said, if you submit a PR, I can find someone with expertise to review.

dichotomies:
I've submitted a PR: https://github.com/tensorflow/models/pull/5151.

tensorflowbutler:
Hi There,
 We are checking to see if you still need help on this, as this seems to be considerably old issue. Please update this issue with the latest information, code snippet to reproduce your issue and error you are seeing.
 If we don't hear from you in the next 7 days, this issue will be closed automatically. If you don't need help on this issue any more, please consider closing this.

PR #5151, opened 2018-08-21
Rename Inception V2 to BN-Inception.

dichotomies:
As discussed in https://github.com/tensorflow/models/issues/5054, I'm opening a initial PR. I've only edited the readme file, but further changes will be necessary.

The problem I'm addressing is that the link of Inception V2 in the readme file refers [Batch Normalization: Accelerating Deep Network Training by Reducing Internal Covariate Shift](https://arxiv.org/abs/1502.03167). The implemented [network architecture](https://github.com/tensorflow/models/blob/master/research/slim/nets/inception_v2.py) has the same output sizes as in the paper (p. 11), so I think that the reference to the paper is correct. It is named BN-Inception.

Inception V2 (with V3) was introduced in [Rethinking the Inception Architecture for Computer Vision](https://arxiv.org/abs/1512.00567) and it has a different network architecture compared to the implemented one (see p. 6).

Therefore I think that the name of the network architecture in the repository should be changed from Inception V2 to BN-Inception.

### DIFF
--- a/research/slim/README.md
+++ b/research/slim/README.md
@@ -248,7 +248,7 @@ crops at multiple scales.
 Model | TF-Slim File | Checkpoint | Top-1 Accuracy| Top-5 Accuracy |
 :----:|:------------:|:----------:|:-------:|:--------:|
 [Inception V1](http://arxiv.org/abs/1409.4842v1)|[Code](https://github.com/tensorflow/models/blob/master/research/slim/nets/inception_v1.py)|[inception_v1_2016_08_28.tar.gz](http://download.tensorflow.org/models/inception_v1_2016_08_28.tar.gz)|69.8|89.6|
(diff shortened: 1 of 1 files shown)


dichotomies:
I signed the CLA.

karmel:
Thanks @dichotomies -- though I think I misunderstood the intent of the original issue. I thought you were suggesting changing/adding a paper reference, not renaming the model altogether. Given the existing links to InceptionV2 as it is named, it will be hard to attempt a rename for the entirety. Maybe adding an additional link and note to the paper is more reasonable? @sguada , can you comment?

dichotomies:
@karmel While submitting the PR, I've noticed that the actual problem is not the link to the paper (links to BN-Inception), but the name of the link, i.e. the implemented model seems to be BN-Inception and not Inception V2. So you haven't misunderstood it, my initial guess was just inaccurate.

karmel:
Given historical references to the name as-is, I suspect it would be more confusing to rename now than to not. In that case, let's leave as-is. If you would like to add a note or comment referring to BN-inception, we can do that.

dichotomies:
Something like: "Note that the correct model name is BN-Inception, but it was defined as Inception V2 when implemented. Due to historical references to the name as-is, it might be more confusing to rename it now. Therefore, the name is left unchanged in the code and is only corrected in this README."?

PR #5152, opened 2018-08-21
Docs: fix escape character

staff0rd:
### Before

before

### After

image

### DIFF
--- a/samples/core/tutorials/keras/basic_regression.ipynb
+++ b/samples/core/tutorials/keras/basic_regression.ipynb
@@ -496,7 +496,7 @@
       },
       "cell_type": "markdown",
       "source": [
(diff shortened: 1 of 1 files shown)


lamberta:
Need to make sure this works when it's rendered to HTML on the site: https://www.tensorflow.org/tutorials/keras/basic_regression#train_the_model

MarkDaoust:
We're in the middle of a repository migration. 

This file has [moved](https://github.com/tensorflow/docs/blob/master/site/en/tutorials/keras/basic_regression.ipynb), if you want to move foreward with this PR, please resubmit there.

This does point out a larger problem: `$` is not handled consistently between jupyter, github, colab, tensorflow.org 

I need to find a fix or compromise that looks reasonable on all of them.

PR #5156, opened 2018-08-21
StringIO is not supported in python3

pinakinathc:
putting `import StrinIO` in try-except block with `except ImportError` ensures that the code runs in python2 and python3

Author: Pinaki Nath Chowdhury <pinakinathc@gmail.com>


jaeyounkim:
Fixed

Issue #5016, opened 2018-08-07
Create TFRecord files from the MNIST-M data set


Aksei:
Hi, 

i am trying to create the TFRecord files from the MNIST-M dataset. I did download and extract the dataset, however when I run the command : 

python download_and_convert_mnist_m.py -- --dataset_dir $DSN_DATA_DIR

I always get the following error: 

Traceback (most recent call last):
  File "download_and_convert_mnist_m.py", line 240, in <module>
    tf.app.run()
  File "/Users/myfolder/lib/python3.6/site-packages/tensorflow/python/platform/app.py", line 126, in run
    _sys.exit(main(argv))
  File "download_and_convert_mnist_m.py", line 234, in main
    assert FLAGS.dataset_dir
AssertionError

I have been working at it for quite I while, but I can't fix it. 

If I use bazel (bazel run domain_adaptation/datasets:download_and_convert_mnist_m -- --dataset_dir $DSN_DATA_DIR), I run into the same problem. 

Traceback (most recent call last):
  File "/private/var/tmp/_bazel_myname/e9545550cccabf838338cbdd5e4c6269/execroot/__main__/bazel-out/darwin-fastbuild/bin/research/domain_adaptation/datasets/download_and_convert_mnist_m.runfiles/__main__/research/domain_adaptation/datasets/download_and_convert_mnist_m.py", line 240, in <module>
    tf.app.run()
  File "/Users/myFolder/lib/python3.6/site-packages/tensorflow/python/platform/app.py", line 126, in run
    _sys.exit(main(argv))
  File "/private/var/tmp/_bazel_myname/e9545550cccabf838338cbdd5e4c6269/execroot/__main__/bazel-out/darwin-fastbuild/bin/research/domain_adaptation/datasets/download_and_convert_mnist_m.runfiles/__main__/research/domain_adaptation/datasets/download_and_convert_mnist_m.py", line 235, in main
    run(FLAGS.dataset_dir)
  File "/private/var/tmp/_bazel_myname/e9545550cccabf838338cbdd5e4c6269/execroot/__main__/bazel-out/darwin-fastbuild/bin/research/domain_adaptation/datasets/download_and_convert_mnist_m.runfiles/__main__/research/domain_adaptation/datasets/download_and_convert_mnist_m.py", line 219, in run
    train_validation_filenames_to_class_ids, dataset_dir)
[...]

PR #5158, opened 2018-08-21
Version1

Aksei:
Unicode decode error--   Issue  #5016


Aksei:
I signed it!

Aksei:
Should be solved.

research/slim/BUILD
```diff
@@ -709,7 +709,7 @@ py_library(
 py_binary(
     name = "train_image_classifier",
     srcs = ["train_image_classifier.py"],
-    paropts = ["--compress"],
+    #paropts = ["--compress"],
```
karmel: Please remove the changes in slim-- that will also reduce the complexity of this change.

PR #5166, opened 2018-08-22
Docs: Add plt.show() method

AlvaroBernalG:
Hello good people of tensorflow models, 

Change:
Adding missing method call `plt.show()` in the 'basic classification' example. 

Reason: 
It can be confusing for those who are running the examples without a Jupyter notebook.


MarkDaoust:
Hi, thanks for the PR.

I'm not convinced on this one.

There's no clear way to enforce it across notebooks, and this would add code that does nothing for typical users.

I feel a bit like: "If you're running this without a notebook you're an advanced user, and can figure this out."

If we do want to support this sort of thing I would do it with `matplotlib.ion()` just after the import.

@lamberta , what do you think?

PR #5167, opened 2018-08-22
Text classification tutorial clarification

bananabowl:
Rename "num_examples" to "num_reviews" to be consistent with the "one-hot-encode" size description: "num_words * num_reviews".


MarkDaoust:
Thanks!

Issue #5056, opened 2018-08-10
TypeError: non_max_suppression() got an unexpected keyword argument 'score_threshold'


aseelam:
### System information
- **What is the top-level directory of the model you are using**:
/models/research/
- **Have I written custom code (as opposed to using a stock example script provided in TensorFlow)**:
No
- **OS Platform and Distribution (e.g., Linux Ubuntu 16.04)**:
Linux Ubuntu 16.04
- **TensorFlow installed from (source or binary)**:
Binary
- **TensorFlow version (use command below)**:
1.10.0,1.8.0
- **Bazel version (if compiling from source)**: NA
- **CUDA/cuDNN version**:9.0,7.0
- **GPU model and memory**:Nvidia Quadro P3000
- **Exact command to reproduce**:
python object_detection/model_main.py     --pipeline_config_path=${PIPELINE_CONFIG_PATH}     --model_dir=${MODEL_DIR}     --num_train_steps=${NUM_TRAIN_STEPS}     --num_eval_steps=${NUM_EVAL_STEPS}     --alsologtostderr

### Describe the problem
Describe the problem clearly here. Be sure to convey here why it's a bug in TensorFlow or a feature request.

TypeError: non_max_suppression() got an unexpected keyword argument 'score_threshold' when running TF object detection API. Looks like the recent commit by @pkulzc 59f7e80 has some bug.

### Source code / logs
Include any logs or source code that would be helpful to diagnose the problem. If including tracebacks, please include the full traceback. Large logs and files should be attached. Try to provide a reproducible test case that is the bare minimum necessary to generate the problem.

Traceback (most recent call last):
  File "object_detection/model_main.py", line 101, in <module>
    tf.app.run()
  File "/home/anikethseelam/anaconda2/lib/python2.7/site-packages/tensorflow/python/platform/app.py", line 126, in run
    _sys.exit(main(argv))
  File "object_detection/model_main.py", line 97, in main
    tf.estimator.train_and_evaluate(estimator, train_spec, eval_specs[0])
  File "/home/anikethseelam/anaconda2/lib/python2.7/site-packages/tensorflow/python/estimator/training.py", line 439, in train_and_evaluate
    executor.run()
  File "/home/anikethseelam/anaconda2/lib/python2.7/site-packages/tensorflow/python/estimator/training.py", line 518, in run
    self.run_local()
[...]

PR #5170, opened 2018-08-23
Update the runtime version in bash script

wentaoxu415:
The bash script to submit training job for pets detection has runtime-version of 1.8. This will trigger `TypeError: non_max_suppression() got an unexpected keyword argument 'score_threshold'` on the Google Cloud since 1.8 and older does not support this keyword argument. Therefore, update this runtime version to 1.9, which is the most recent runtime version that is published on June 27, 2018. 

See 
https://github.com/tensorflow/models/issues/5056
https://cloud.google.com/ml-engine/docs/tensorflow/runtime-version-list


wentaoxu415:
Signed the Contributor License Agreement (CLA)

PR #5176, opened 2018-08-23
Add 5 epoch warmup to resnet

tfboyd:
Using this LR improve accuracy significantly for larger batch sizes and has no harm to small batch sizes, and gets a higher accuracy with less epochs than the pre-trained models presented.  Testing was focused on ResNetV1.5 which the official code lists as version=1.

A higher base lr (.128) is preferred and will be in another PR.  Results below from this change.

GPUs | dtype | batch_size | system | top_1 (90 epochs) | top_5 (90 epochs) | top_1 (85 epochs) | top_1 (81 epochs) | command | initial_lr
-- | -- | -- | -- | -- | -- | -- | -- | -- | --
8xV100 | fp16 | 2048 | gce | 0.7614 |   | 0.7612 | N/A | python -m  official.resnet.imagenet_main --data_dir=/data/imagenet --resnet_version=1 --intra_op_parallelism_threads=1 --num_gpus=8 --tf_gpu_thread_mode=gpu_shared  --model_dir=/workspace/model_ckpt/test_0_128_2 --dtype=fp16 --batch_size=2048 --epochs_between_evals=5 --train_epochs=90 | 0.1
8xV100 | fp16 | 2048 | gce | 0.76396 |   | 0.7633 | N/A | python -m  official.resnet.imagenet_main --data_dir=/data/imagenet --resnet_version=1 --intra_op_parallelism_threads=1 --num_gpus=8 --tf_gpu_thread_mode=gpu_shared  --model_dir=/workspace/model_ckpt/test_0_128_2 --dtype=fp16 --batch_size=2048 --epochs_between_evals=5 --train_epochs=90 | 0.128
8xV100 | fp32 | 1024 | gce | 0.7636 |   | 0.7633 | 0.7625 | python -m  official.resnet.imagenet_main --data_dir=/data/imagenet --resnet_version=1 --intra_op_parallelism_threads=1 --num_gpus=8 --model_dir=/workspace/model_ckpt/test3 --dtype=fp32 --batch_size=1024 | 0.1


official/resnet/resnet_run_loop.py
```diff
@@ -168,8 +168,14 @@ def learning_rate_with_decay(
   vals = [initial_learning_rate * decay for decay in decay_rates]
 
   def learning_rate_fn(global_step):
+    """Builds scaled learning rate function with 5 epoch warm up."""
     global_step = tf.cast(global_step, tf.int32)
```
karmel: I don't know if it matters, but I imagine global_step is either an int or a float at the outset, so if we cast to int here, we can avoid the second cast to float in line 176 if we just preserve the original global_step. Or vice versa. Not sure how much that additional cast would matter in this case.

official/resnet/resnet_run_loop.py
```diff
@@ -168,8 +168,14 @@ def learning_rate_with_decay(
...
-    return tf.train.piecewise_constant(global_step, boundaries, vals)
+    lr = tf.train.piecewise_constant(global_step, boundaries, vals)
+    warmup_steps = int(batches_per_epoch * 5)
```
karmel: Similarly, I think the cast to int here should be unnecessary? Unlikely to matter for performance, but makes things simpler.

tfboyd:
I have no idea if it applies to CIFAR.  I would not do a warmup with CIFAR as I have not seen that listed in any papers.  There really should be flags to control if you want a warmup or not as well as control the starting learning rate.  

Let me look at how to turn this on and off based on the model being trained.

karmel:
The Imagenet and Cifar run loops separately define their learning rate function:
```
learning_rate_fn = resnet_run_loop.learning_rate_with_decay(
```
So maybe add a flag to learning_rate_with_decay that is, warmup_steps=0, and leave as 0 for cifar, set as 5 for imagenet, and then use that in the fn you added? Does that make sense?

tfboyd:
@karmel 
I removed the casts, will test if that is an issue but it looks fine.  I went with an attribute rather than a flag as I am not aware of CIFAR-10 having a warmup and if it does then deal with that later as that was not my goal.  For now leave it as-is no worse off.  If we find a need for a flag we can always do that and this way it "just works" 

Well, I just thought of something else.  **Does the fine tuning use this learning rate schedule?**

tfboyd:
@robieta   I pushed a new version please take a look.  

FYI and maybe not worth reading as I am just trying to explain why you might see something silly and my dev workflow is sloppy.  I work from a branch with many changes and cut them down to submit so I might be missing things but I am trying to run unittests and be careful.  Just an FYI that I am often working in a "dirty" branch due to also having performance tunings.  I will make a PR with the per tweaks this week.

tfboyd:
@robieta   Let me check on the flags, I think I know what I am missing.

robieta:
@tfboyd https://github.com/tensorflow/models/blob/84a76838867643b2f613b61effd040d418c3cc76/official/resnet/resnet_run_loop.py#L505-L512

(You probably already found this.)

tfboyd:
@robieta  All fixed.  I had to get working code to run tests and then start cutting it up for PRs.  And run individual tests on each PR.  Much hoop jumping.  I did a end-to-end (a few steps V1 + FP16) and ran the imagenet / cifar10 unit tests.  The only test that fails is an OOM for imagenet due to the 200 batch-size and my GTX-1080.

robieta:
Alright! Make us fast.

tfboyd:
SWEET  thank you @robieta  and @karmel.

PR #5179, opened 2018-08-23
Add eval_only flag as optional base flag.

tfboyd:
I can can see where this flag might not fit into base.  

I only half care about this as a feature but I did it and found it useful.  If there is an easy way to make this work and fit with your philogphy that is great.  If not, moving on.

Side Note:
I really think it should eval either all the checkpoints in the dir or really better and more concise let you specify a specific checkpoint rather than the model directory.  I didn't try it so I am look like a fool but I suspect it works like tf_cnn_bench does and it uses some util class to find which checkpoint to load, what we did is if the path contains what looks like a checkpoint file then don't guess just us it.

### DIFF
--- a/official/resnet/resnet_run_loop.py
+++ b/official/resnet/resnet_run_loop.py
@@ -430,29 +430,33 @@ def resnet_main(
 
   total_training_cycle = (flags_obj.train_epochs //
                           flags_obj.epochs_between_evals)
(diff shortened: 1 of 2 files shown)


karmel:
Will let @robieta review as the owner of Garden arg parsing.

tfboyd:
better solution from someone else has been committed.

Issue #5069, opened 2018-08-12
object_detection_tutorial comparison tf.__version__ < '1.4.0' for version 1.10.0


gabrielnpol:
Minor stuff,
Context: 
**models/research/object_detection/object_detection_tutorial**
```
if tf.__version__ < '1.4.0':
  raise ImportError('Please upgrade your tensorflow installation to v1.4.* or later!')
```
Issue:
However,
if `tf.__version__` is '1.10.0'
**`'1.10.0' < '1.4.0' `=> True, which will raise ImportError even if 1.10.0 is higher.**

mutantmassXXX:
Delete the if case. Thats how i fixed mine :)

dcyoung:
the check is performed on strings, instead of ints... which is causing a failure with comparing the '4' and '10'.

I'm sure there is a  more elegant solution... but you could use:

```python
MIN_TF_VERSION_MAJOR = 1
MIN_TF_VERSION_MINOR = 9
tf_ver = [int(s) for s in tf.__version__.split('.')]

if tf_ver[0] < MIN_TF_VERSION_MAJOR or (tf_ver[0] == MIN_TF_VERSION_MAJOR and tf_ver[1] < MIN_TF_VERSION_MINOR):
  raise ImportError('Please upgrade your tensorflow installation to v{0}.{1}.* or later!'.format(MIN_TF_VERSION_MAJOR, MIN_TF_VERSION_MINOR))
```

wt-huang:
Closing as this is resolved

Issue #3033, opened 2017-12-20
Object Detection Tutorial Jupyter Notebook checks for version equality, making non versions > 1.4.0 fail


vikramg1:
### System information
- **What is the top-level directory of the model you are using**:
models/research/object_detection

- **Have I written custom code (as opposed to using a stock example script provided in TensorFlow)**:
No
- **OS Platform and Distribution (e.g., Linux Ubuntu 16.04)**:
Linux Ubuntu 14.03

- **TensorFlow installed from (source or binary)**:
binary, virtual env

- **TensorFlow version (use command below)**:
1.4.1

- **Bazel version (if compiling from source)**:
- **CUDA/cuDNN version**:
6
- **GPU model and memory**:
Titan X
- **Exact command to reproduce**:
models/research/object_detection$jupyter notebook
choose object_detection_tutorial.ipynb in the brower
press SHIFT+ENTER in the first box

### Describe the problem
#version check below:
if tf.__version__ != '1.4.0':

#fails as version is 1.4.1
---------------------------------------------------------------------------
ImportError                               Traceback (most recent call last)
<ipython-input-3-a27dab2761c6> in <module>()
     14 #if tf.__version__ != '1.4.0':
     15 if tf.__version__ != '1.4.0':
---> 16   raise ImportError('Please upgrade your tensorflow installation to v1.4.0!')

ImportError: Please upgrade your tensorflow installation to v1.4.0!

Suggested code change:
[reference: https://stackoverflow.com/questions/11887762/compare-version-strings-in-python]

from distutils.version import LooseVersion
if LooseVersion(tf.__version__) < LooseVersion('1.4.0'):

bignamehyp:
@jch1

tensorflowbutler:
Hi There,
 We are checking to see if you still need help on this, as this seems to be considerably old issue. Please update this issue with the latest information, code snippet to reproduce your issue and error you are seeing.
 If we don't hear from you in the next 7 days, this issue will be closed automatically. If you don't need help on this issue any more, please consider closing this.

PR #5182, opened 2018-08-24
 Fixed TensorFlow version check in object_detection_tutorial.ipynb

roffel:
The Object_detection_tutorial is compatible with TensorFlow 1.10 but the check says that version 1.4 is newer than version 1.10. This change uses the StrictVersion check from [PEP-0386 ](https://www.python.org/dev/peps/pep-0386/) to ensure good version checks in the future.

Solves #5069, #3033

I have also changed `tensorflow` to `TensorFlow` in the error message.

### DIFF
--- a/research/object_detection/object_detection_tutorial.ipynb
+++ b/research/object_detection/object_detection_tutorial.ipynb
@@ -54,8 +54,8 @@
         "sys.path.append(\"..\")\n",
         "from object_detection.utils import ops as utils_ops\n",
         "\n",
-        "if StrictVersion(tf.__version__) \u003c StrictVersion('1.4.0'):\n",
-        "  raise ImportError('Please upgrade your TensorFlow installation to v1.4.* or later!')\n"
+        "if StrictVersion(tf.__version__) \u003c StrictVersion('1.9.0'):\n",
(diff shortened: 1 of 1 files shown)


dcyoung:
Might want to increase the minimum version from `v1.4.0` to `v1.9.0` per the requirements.

research/object_detection/object_detection_tutorial.ipynb
```diff
@@ -53,8 +54,8 @@
         "sys.path.append(\"..\")\n",
         "from object_detection.utils import ops as utils_ops\n",
         "\n",
-        "if tf.__version__ \u003c '1.4.0':\n",
-        "  raise ImportError('Please upgrade your tensorflow installation to v1.4.* or later!')\n"
+        "if StrictVersion(tf.__version__) \u003c StrictVersion('1.4.0'):\n",
+        "  raise ImportError('Please upgrade your TensorFlow installation to v1.4.* or later!')\n"
```
pkulzc: v1.9

roffel: I'll update it in a couple of minutes!

roffel: @pkulzc done!

roffel:
@pkulzc Done!

PR #5199, opened 2018-08-29
Fix python 3 compatibility problem in ObjectDetectionEvaluator.evalua…

dnuffer:
…te()

unicode() doesn't exist in python 3, and when executed, a NameError exception is thrown.


pkulzc:
Catch an error and do nothing is not the right behaviour. We will fix this soon.

PR #5223, opened 2018-09-01
Update resnet defaults v1

tfboyd:
Changes the default ResNet to use to v1 from v2 and lowers the epochs from 100 to 90.  

FYI: I feel less great about the README update and it could use more tweaks to tell the story completely and concisely.  The topic is confusing and I do not want to gloss over it.

ResNet50v1 (which is really v1.5 in this example) trains to 76.2+% top_1  and 93%+ top_5 at 81 epochs most times if not always and "always" at 90 epochs.  90 epochs is also the general standard without using different sized images and learning rate tweaks.  I short, it is a nice test.

The purpose of this is also to have the internal CI tests change from v2 to v1 with 90 epochs, which is more useful than running V2 for 100 epochs.


robieta:
To clarify: I'm proposing adding v1 and v1.5 options in a future PR.

Issue #5206, opened 2018-08-30
Add  fused_batch_norm parameter for models/gan


mpjlu:
in models/research/gan/cifar/networks.py
dcgan generator and discriminator call slim.nets.dcgan with the following code
`dcgan.generator(noise, is_training=is_training)`
`dcgan.discriminator(noise, is_training=is_training)`

I propose to add  fused_batch_norm parameter for this function call, like this:

`dcgan.generator(noise, is_training=is_training, fused_batch_norm=True)`
`dcgan.discriminator(noise, is_training=is_training, fused_batch_norm=True)`

With fused_batch_norm parameter, there is about 20%-30% performance improvement on Intel CPU.

mpjlu:
@joelshor, any comments for this? Thanks.

joel-shor:
This would be a great addition. Feel free to send a pull request with this change!

mikaelsouza:
Hello!

After seeing this issue, I've implemented this change here: #5227

First time contributing here, so I hope everything is alright.

mikaelsouza:
I think this should be closed.

PR #5227, opened 2018-09-02
Added fused_batch_norm parameter

mikaelsouza:
As asked in #5206, I added the parameters discussed there.

There were 4 places in which this parameter had to be changed, they were:
* `generator`function;
* `conditional_generator`function;
* `discriminator`function;
* `conditional_discriminator`function.


joel-shor:
Looks great. thanks!

PR #5247, opened 2018-09-05
Update data_download.py

colinzuo:
Url format doesn't depend on OS, shouldn't use os.path.join, otherwise will generate wrong format on windows, such as 
https://archive.ics.uci.edu/ml/machine-learning-databases/00280\HIGGS.csv.gz, and cause 404 error.


colinzuo:
I signed it!

PR #5252, opened 2018-09-05
Fix spurious "did not start correctly" error.

reedwm:
The error "Generation subprocess did not start correctly" would occur if the async process started up after the main process checked for the subproc_alive file.


official/recommendation/data_preprocessing.py
```diff
@@ -494,10 +502,6 @@ def make_train_input_fn(ncf_dataset):
   # type: (NCFDataset) -> (typing.Callable, str, int)
   """Construct training input_fn for the current epoch."""
 
-  if not tf.gfile.Exists(ncf_dataset.cache_paths.subproc_alive):
```
robieta: Can you keep this in, but just change the error message. The subprocess can still die, so it should still check.

PR #5253, opened 2018-09-05
Move tf.cast to tf.float16 in input pipeline 

tfboyd:
Moves the tf.cast for tf.float16 to the input pipeline which is run on cpu and pipelined.  This seemed to have a consistent ~50 images/sec boost on 8xV100 FP16 with real imagenet data, which is a small 1% gain and logical.

FYI,  ```tf.image.per_image_standardization(image)``` Returns tf.float32 even if passed a tf.float16 tensor thus the cast to tf.float16 must be done after preprocessing.

### DIFF
--- a/official/resnet/cifar10_main.py
+++ b/official/resnet/cifar10_main.py
@@ -66,7 +66,7 @@ def get_filenames(is_training, data_dir):
     return [os.path.join(data_dir, 'test_batch.bin')]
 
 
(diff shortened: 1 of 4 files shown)


robieta:
Ah, looks like the cifar test needs to be updated.

tfboyd:
No problem,  I thought I ran all the tests but I keep jumping around.  Fixing ASAP.

PR #5255, opened 2018-09-05
cifar100 dataset builder and training for resnet

erfaneshrati:
Added the Cifar100 dataset builder and training for ResNet model in official/resnet.


erfaneshrati:
I signed the CLA!

robieta:
I'm not opposed to including cifar 100, but just forking the cifar 10 scripts is not going to be worthwhile from a maintenance perspective. Can you compactly extend the existing cifar code?

PR #5262, opened 2018-09-07
don't use os.path.join for http url

colinzuo:
don't use os.path.join for http url, add flag to not delete downloaded input file, so that it can be reused if needed.

URL format doesn't depend on OS, shouldn't use os.path.join, otherwise will generate wrong format on windows, such as
https://archive.ics.uci.edu/ml/machine-learning-databases/00280\HIGGS.csv.gz, and cause 404 error.

### DIFF
--- a/official/boosted_trees/data_download.py
+++ b/official/boosted_trees/data_download.py
@@ -45,9 +45,9 @@ INPUT_FILE = "HIGGS.csv.gz"
 NPZ_FILE = "HIGGS.csv.gz.npz"  # numpy compressed file to contain "data" array.
 
 
-def _download_higgs_data_and_save_npz(data_dir):
+def _download_higgs_data_and_save_npz(data_dir, cache_data):
   """Download higgs data and store as a numpy compressed file."""
-  input_url = os.path.join(URL_ROOT, INPUT_FILE)
+  input_url = URL_ROOT + "/%s" % INPUT_FILE
   np_filename = os.path.join(data_dir, NPZ_FILE)
   if tf.gfile.Exists(np_filename):
     raise ValueError("data_dir already has the processed data file: {}".format(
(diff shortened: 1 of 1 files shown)


robieta:
The url issue is a good catch. Could you elaborate on why one might want to hold onto the downloaded zip file?

colinzuo:
Cache is for places like china, I need to use a proxy to download the file, but modify the code to enable using proxy seems over modified, also this file is too big that I'm afraid if proxy will work well, instead download the file through a proxy server then place it as desired seems more reasonable for me, and the downloaded file can also be shared by the same team.

robieta:
Ah, I see. It strikes me that this is just generally good bandwidth saving practice to use the zip if it exists, either because something went wrong making the data dir or because you already have it somewhere and just want to get a new machine ready. (Or indeed, if you live in a jurisdiction with strong feelings on particle physics datasets.) If memory serves MNIST already follows this magic number checking pattern.

I think probably the best pattern is something like:
```
if not tf.gfile.Exists(temp_download_path) or some_file_hash_fn(temp_download_path) != _HIGGS_MAGIC_NUMBER:
  urllib.download(...)

...
```
And then the only thing the cache flag does is persist the archive file.

colinzuo:
Hi robieta,

  I feel you probably misunderstood me, it's not for bandwidth saving purpose. In China we can't access lots of foreign sites such as google, facebook and in thise case uci.edu. So we have to use proxy to access uci.edu. And as other users outside China can probably access uci.edu directly, so add proxy setting is mostly probably not appropriate. In contrast, use local file when available is similar to function get_file in tensorflow/python/keras/utils/data_utils.py, so I add similar function.

Thanks,
Colin

robieta:
The context is not lost on me. I'm simply saying that the feature that you're proposing is more broadly useful. And for that reason I think the download script should always check for the existence of the intermediate archive file to see if it can skip a download step. (Basically your PR duplicates some code between the cache and non-cache options which I think can be avoided.)

PR #5290, opened 2018-09-12
cifar100 added in a single file together with cifar10.

erfaneshrati:


robieta:
It looks like something went wrong, as large parts of CIFAR have been blown removed.

robieta:
Git seems to think you deleted those files. Maybe you used `mv` instead of `git mv`?

PR #5299, opened 2018-09-13
Pass session config to device_lib list_local_devices() call

samikama:
This PR passes session config in run_params object to device_lib.list_local_devices() call. Otherwise this call instantiates a session and may ignore any gpu configuration that user might have.

### DIFF
--- a/official/utils/logs/logger.py
+++ b/official/utils/logs/logger.py
@@ -308,9 +308,9 @@ def _gather_run_info(model_name, dataset_name, run_params, test_id):
       "run_date": datetime.datetime.utcnow().strftime(
           _DATE_TIME_FORMAT_PATTERN)}
   if "session_config" in run_params:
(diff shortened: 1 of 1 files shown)


official/utils/logs/logger.py
```diff
@@ -307,11 +307,15 @@ def _gather_run_info(model_name, dataset_name, run_params, test_id):
       "test_id": test_id,
       "run_date": datetime.datetime.utcnow().strftime(
           _DATE_TIME_FORMAT_PATTERN)}
+  if "session_config" in run_params:
```
aaroey: How about:
```
session_config = None
if xxx:
  session_config = xxx
```

samikama:
pinging @karmel and @tensorflow/tf-garden-team for merge?

karmel:
@aaroey should be able to handle-- please merge if desired.

PR #5334, opened 2018-09-19
Fix/ncf mlperf tweaks

robieta:
* Cleanup run script. Fix hr threshold and add ml_perf flag
* Make exit behavior more stable
* Add seed to reduce variation. There is still unaccounted for variation, but this at least removes some of it.


official/recommendation/data_preprocessing.py
```diff
@@ -439,20 +445,29 @@ def instantiate_pipeline(dataset, data_dir, batch_size, eval_batch_size,
...
 
   tf.logging.info(
       "Generation subprocess command: {}".format(" ".join(subproc_args)))
 
   proc = subprocess.Popen(args=subproc_args, shell=False, env=subproc_env)
 
-  atexit.register(_shutdown, proc=proc)
-  atexit.register(tf.gfile.DeleteRecursively,
-                  ncf_dataset.cache_paths.cache_root)
+  cleanup_called = {"finished": False}
```
reedwm: Where do you set this to True?

official/recommendation/ncf_main.py
```diff
@@ -390,6 +396,8 @@ def run_ncf(_):
     if model_helpers.past_stop_threshold(FLAGS.hr_threshold, hr):
       break
 
+  cleanup_fn()  # Cleanup data construction artifacts and subprocess.
```
reedwm: No need to call this, since it's registered with atexit

official/recommendation/ncf_main.py
```diff
@@ -297,11 +299,14 @@ def run_ncf(_):
...
 
+  if FLAGS.seed is not None:
+    np.random.seed(FLAGS.seed)
+
   num_gpus = flags_core.get_num_gpus(FLAGS)
   batch_size = distribution_utils.per_device_batch_size(
       int(FLAGS.batch_size), num_gpus)
   eval_batch_size = int(FLAGS.eval_batch_size or FLAGS.batch_size)
-  ncf_dataset = data_preprocessing.instantiate_pipeline(
+  ncf_dataset, cleanup_fn = data_preprocessing.instantiate_pipeline(
```
reedwm: Pass `deterministic=FLAGS.seed is not None`

official/recommendation/data_preprocessing.py
robieta: Oops. Good catch.

official/recommendation/ncf_main.py
robieta: Good catch. I wonder if this is the reason that I wasn't seeing deterministic behavior. (I'll rerun to test.)

official/recommendation/ncf_main.py
robieta: This comes from Shawn. He is running in a loop, and is accumulating files because the atexit hasn't triggered yet. I neglected that use case when I decided to rely on atexits.

PR #5340, opened 2018-09-20
Added 'rb' as a parameter for FastGFile

SimoneNigro:
If you do not add 'rb', python will think you're opening a text file, and the program would fail with an UnicodeDecodeError


SimoneNigro:
I signed the CLA

jaeyounkim:
compression is deprecated.

PR #5350, opened 2018-09-21
Update CloudML job counts in running_pets.md

feynmanliang:
As per https://github.com/tensorflow/models/blob/master/research/object_detection/samples/cloud/cloud.yml#L5

---
This change is [Reviewable](https://reviewable.io/reviews/tensorflow/models/5350)


derekjchow:
Thanks for the update!

PR #5355, opened 2018-09-22
Fix BSD compatibility of '\n' in sed create_pycocotools_package.sh

feynmanliang:
See https://stackoverflow.com/questions/46082397/insert-newline-n-using-sed

---
This change is [Reviewable](https://reviewable.io/reviews/tensorflow/models/5355)


saberkun:
this pr is stale.

PR #5366, opened 2018-09-23
adding note only SSD models are supported

woolfel:
save people the trouble of training a model and trying to convert only to realize they should have used SSD model


woolfel:
I signed the CLA

PR #5354, opened 2018-09-21
Update sim.

pkulzc:
213899768  by Sergio Guadarrama:

    Fixes #3819.

--
213493831  by Sergio Guadarrama:

    Internal change

212057654  by Sergio Guadarrama:

    Internal change

210747685  by Sergio Guadarrama:

    For FPN, when use_depthwise is set to true, use slightly modified mobilenet v1 config.

--
210128931  by Sergio Guadarrama:

    Allow user-defined current_step in NASNet.

--
209092664  by Sergio Guadarrama:

    Add quantized fine-tuning / training / eval and export to slim image classifier binaries.

--
207651347  by Sergio Guadarrama:

    Update mobilenet v1 docs to include revised tflite models.

--
207165245  by Sergio Guadarrama:

    Internal change

207095064  by Sergio Guadarrama:

    Internal change

PiperOrigin-RevId: 213899768

PR #5367, opened 2018-09-24
Cast range results to string before use as eval_spec_names

lyonzy:
This fixes an error where tensorflow training.py expects string values instead of ints:

```
    if name is not None and not isinstance(name, six.string_types):
      raise TypeError('`name` must be string, given: {}'.format(name))
```


pkulzc:
Thanks for fixing this! It's already included in another [PR](https://github.com/tensorflow/models/pull/5354).

PR #5379, opened 2018-09-25
Enabling prediction in mnist_tpu.

aman2930:
As per many users request over stack_overflow and github, I am enabling prediction in mnist_tpu.
Right now we don't have input data for prediction. So using top 10 entries of test data as input.


official/mnist/mnist_tpu.py
```diff
@@ -168,6 +186,19 @@ def main(argv):
...
+    predictions = estimator.predict(input_fn=predict_input_fn)
+
+    for pred_dict in predictions:
+      template = ('Prediction is "{}" ({:.1f}%), expected "{}"')
+
+      class_id = pred_dict['class_ids']
+      probability = pred_dict['probabilities'][class_id]
+      expected_label = pred_dict['label']
+
+      print(template.format(class_id, 100 * probability, expected_label))
```
asimshankar: I'd just inline the template string here.

official/mnist/mnist_tpu.py
```diff
@@ -168,6 +186,19 @@ def main(argv):
   if FLAGS.eval_steps:
     estimator.evaluate(input_fn=eval_input_fn, steps=FLAGS.eval_steps)
 
+  # Run prediction on the test data.
+  if FLAGS.enable_predict:
+    predictions = estimator.predict(input_fn=predict_input_fn)
+
+    for pred_dict in predictions:
+      template = ('Prediction is "{}" ({:.1f}%), expected "{}"')
```
asimshankar: In general, isn't `predict` typically used to apply the network on previously unseen data? So there won't be an "expected" answer. Since this PR is intended to provide an example, I'd leave out the "expected" business (which only happens to be available since we're using the test set). Or better still, `predict_input_fn` shouldn't even produce the "expected" tensor (as that would be the case when using `predict` in practice, no?)

(See for example: https://www.tensorflow.org/api_docs/python/tf/contrib/tpu/TPUEstimator#example_mnist_2)

official/mnist/mnist_tpu.py
```diff
@@ -82,13 +83,21 @@ def model_fn(features, labels, mode, params):
...
     image = features["image"]
 
   model = mnist.create_model("channels_last")
+
+  if mode == tf.estimator.ModeKeys.PREDICT:
+    logits = model(image, training=False)
+    predictions = {
+        'class_ids': tf.argmax(logits, axis=1),
+        'probabilities': tf.nn.softmax(logits),
+        'label': labels,
```
asimshankar: The `PREDICT` model function shouldn't produce the training label, right? (How would one know that?)

official/mnist/mnist_tpu.py
aman2930: I would like to keep it this way for better readability.

official/mnist/mnist_tpu.py
aman2930: Done. Could not remove the expected tensor from the dataset as the test() function is generating zipped dataset with features & labels. And I didn't find a suitable way to separate those out.
But as we are not using labels anywhere in the prediction call stack so it won't be a problem.

official/mnist/mnist_tpu.py
aman2930: Done.

PR #5386, opened 2018-09-26
how to get this repo isn't stated here

data-steve:
other README.md docs in tensorflow have the git clone requirement declared.

the way the documentation is set up to point PYTHON PATH to the models module path sounds like the models module was already on my machine , esp since the github repo has it as part of tensorflow master repo


jaeyounkim:
CLA not signed.

PR #5393, opened 2018-09-28
Support accepting JPEG image bytes for inference 

netfs:
Add `--image_bytes_as_serving_input` flag to export SavedModel with
serving signature that accepts JPEG image bytes instead of a fixed size
[HxWxC] image tensor.

Passing JPEG image bytes is easier for inference/serving use cases.
The model internally resizes/crops the image to required [HxWxC]
shaped tensor before passing it on for actual model inference.

This change aligns with Cloud TPU/ResNet-50 model that offers a
similar interface (jpeg bytes) for inferencing here:
https://github.com/tensorflow/tpu/tree/master/models/official/resnet

NOTE: This flag is set to `True` by default.

### DIFF
--- a/official/resnet/cifar10_main.py
+++ b/official/resnet/cifar10_main.py
@@ -238,7 +238,8 @@ def define_cifar_flags():
                           resnet_size='56',
                           train_epochs=182,
                           epochs_between_evals=10,
-                          batch_size=128)
+                          batch_size=128,
+                          image_bytes_as_serving_input=False)
 
 
 def run_cifar(flags_obj):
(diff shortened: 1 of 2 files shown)


robieta:
This all looks good. I just have two minor requests:
1) Can you make sure that this code path can't be accessed from the CIFAR side? (Just to avoid confusion.) A simple flags validator in the cifar10_main.py
2) Along the same lines, can you place appropriate comments where this function is called making it clear that this will only work for the ImageNet path?

Thanks for moving this forward! Its a tremendously important use case.

netfs:
> This all looks good. I just have two minor requests:
> 
> 1. Can you make sure that this code path can't be accessed from the CIFAR side? (Just to avoid confusion.) A simple flags validator in the cifar10_main.py
> 2. Along the same lines, can you place appropriate comments where this function is called making it clear that this will only work for the ImageNet path?
> 

Update the PR to account for your requests. Please take a look.

Thanks!

> Thanks for moving this forward! Its a tremendously important use case.

PR #5397, opened 2018-09-28
Update resnet_model.py

sahgerlad:
This should correct the if statement for self.data_format conversion from NHWC to NCHW (Line 494). If self.data_format == 'channels_first', no change should be made but if self.data_format == 'channels_last', the format should be changed to NCHW along with the variable name change.

### DIFF
--- a/official/resnet/resnet_model.py
+++ b/official/resnet/resnet_model.py
@@ -491,11 +491,12 @@ class Model(object):
     """
 
     with self._model_variable_scope():
-      if self.data_format == 'channels_first':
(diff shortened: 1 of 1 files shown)


sahgerlad:
I signed it!

robieta:
I don't believe this is correct. The data comes in as channels_last (from the data pipeline), and we need to transpose if we want channels first. Secondly, the state of the model (self.data_format) change will make subsequent calls produce incorrect results since the transpose will not be applied.

PR #5434, opened 2018-10-03
link to non-deprecated imagenet preprocessing script

tfboyd:
This imagenet script is what we are supporting.  I have used it to train the model and it trains with the output the same as it did with the previous script that we no longer support.


tfboyd:
crap I pusehd the merge button

PR #5435, opened 2018-10-04
Update train.proto

nisseb:
keep_checkpoint_every_n_hour (uint32 -> float)
Interpreted as a float in the API (saver.py).
https://github.com/tensorflow/tensorflow/blob/r1.11/tensorflow/python/training/saver.py#L699

Lets us save checkpoints more often than every hour


nisseb:
I signed it!

jaeyounkim:
Fixed

PR #5439, opened 2018-10-04
set strip_default_attrs=True for SavedModel exports

robieta:
Stripping default attributes is recommended practice for wider SavedModel portability between TensorFlow versions. (https://www.tensorflow.org/guide/saved_model#manually_build_a_savedmodel) This PR update official models to follow the convention of setting `strip_default_attrs=True` when exporting SavedModels.

### DIFF
--- a/official/resnet/resnet_run_loop.py
+++ b/official/resnet/resnet_run_loop.py
@@ -534,7 +534,8 @@ def resnet_main(
       input_receiver_fn = functools.partial(image_bytes_serving_input_fn, shape)
     else:
       input_receiver_fn = export.build_tensor_serving_input_receiver_fn(
-          shape, batch_size=flags_obj.batch_size)
+          shape, batch_size=flags_obj.batch_size,
+          dtype=flags_core.get_tf_dtype(flags_obj))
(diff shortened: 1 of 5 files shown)


official/resnet/resnet_run_loop.py
```diff
@@ -565,7 +568,7 @@ def define_resnet_flags(resnet_size_choices=None):
       help=flags_core.help_wrap('Skip training and only perform evaluation on '
                                 'the latest checkpoint.'))
   flags.DEFINE_boolean(
-      name="image_bytes_as_serving_input", default=True,
+      name="image_bytes_as_serving_input", default=False,
```
k-w-w: Any particular reason the default is set to false?

robieta: Two reasons:
1) This doesn't work properly on CIFAR, so users would have to know to go and set this to False.
2) This is a finesse in that the input is no longer just a normal tensor, and I am generally wary of putting advanced use cases as defaults.

k-w-w: I see, thanks for the response

PR #5443, opened 2018-10-05
Use AUTOTUNE, remove noop take, and comment fixes

tfboyd:
Set map_batch to use AUTOTUNE, tests on a GCE 8xV100 on GCE yielded.  Note this change needs 1.12 (cut not built yet) or nightly as of ~OCT-02:

- 5,100-5,190 images/sec.  Painful full disclosure.  My fork had dropped in performance with num_parallel_batches=1 to 4,600 from ~5K, it does not seem that the main branch (here) has that issue for some reason.  This change improved my fork from 4,600 to 5,200 and for the **"official" had change** but less code.  We need to merge in my perf args and stop using the fork for perf testing.  Still very nice to clean up the code so embarrassing and perplexing but still a win as well as a reminder as to why long-term forks are bad news.
- In my tuned version AUTOTUNE is equal or better than hand tuning num_parallel_batches. 5,400+

Other changes:

- Remove dataset.take that was not doing anything.
- Tweaked some comments to be shorter and match style guide.

Ran end-to-end test of 90 Epochs with expected accuracy achieved.

- top_1: 76.25%
- top_5: 93.077% 

{'name': 'accuracy', 'timestamp': '2018-10-05T12:30:24.004028Z', 'value': 0.7625200152397156, 'extras': [], 'unit': None, 'global_step': 56286}

I1005 12:30:24.004206 139731230779136 tf_logging.py:115] Benchmark metric: {'name': 'accuracy_top_5', 'timestamp': '2018-10-05T12:30:24.004190Z', 'value': 0.9307799935340881, 'extras': [], 'unit': None, 'global_step': 56286}

### DIFF
--- a/official/resnet/resnet_run_loop.py
+++ b/official/resnet/resnet_run_loop.py
@@ -68,36 +68,26 @@ def process_record_dataset(dataset, is_training, batch_size, shuffle_buffer,
     Dataset of (image, label) pairs ready for iteration.
   """
 
-  # We prefetch a batch at a time, This can help smooth out the time taken to
-  # load input files as we go through shuffling and processing.
+  # Sets tf.data to AUTOTUNE, e.g. num_parallel_batches in map_and_batch.
+  options = tf.data.Options()
+  options.experimental_autotune = True
+  dataset = dataset.with_options(options)
+
+  # Prefetches a batch at a time to smooth out the time taken to load input
+  # files for shuffling and processing.
   dataset = dataset.prefetch(buffer_size=batch_size)
   if is_training:
-    # Shuffle the records. Note that we shuffle before repeating to ensure
-    # that the shuffling respects epoch boundaries.
+    # Shuffles records before repeating to respect epoch boundaries.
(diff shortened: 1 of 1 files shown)


robieta:
This is awesome. Have you by any chance looked into any of the other experimental options? The two that strike me are:
1) `experimental_map_and_batch_fusion` could let us replace the somewhat ugly `dataset.apply(tf.contrib.data.map_and_batch(...))` with a more idiomatic
```
dataset = dataset.map(...)
dataset = dataset.batch(...)
```

2) Whether `experimental_hoist_random_uniform` could help with the parallelization of the map fn since it contains random transformations. (Possibly in tandem with `experimental_map_parallelization` and/or `experimental_map_vectorization`) I'm less sure on whether this one will have an effect.

tfboyd:
Doing the .map and the .batch and then setting the option to say they should be fused is still very new and ideally would be out of the box.  That would also be another PR with more testing.  Interleave also needs changed.

PR #5462, opened 2018-10-09
Allow data async generation to be run as a standalone job rather than a subprocess

shizhiw:


official/recommendation/data_preprocessing.py
```diff
@@ -416,7 +416,8 @@ def _shutdown(proc):
 
 def instantiate_pipeline(dataset, data_dir, batch_size, eval_batch_size,
                          num_data_readers=None, num_neg=4, epochs_per_cycle=1,
-                         match_mlperf=False, deterministic=False):
+                         match_mlperf=False, deterministic=False,
+                         use_subprocess=True):
```
reedwm: You never pass False to use_subprocess.

official/recommendation/data_preprocessing.py
```diff
@@ -425,44 +426,57 @@ def instantiate_pipeline(dataset, data_dir, batch_size, eval_batch_size,
...
+      "use_tf_logging": not use_subprocess,
+  }
   if ncf_dataset.deterministic:
-    subproc_args.extend(["--seed", str(int(stat_utils.random_int32()))])
-
+    flags_["seed"] = stat_utils.random_int32()
+  # We write to a temp file then atomically rename it to the final file,
+  # because writing directly to the final file can cause the data generation
+  # async process to read a partially written JSON file.
+  flagfile_temp = os.path.join(flags.FLAGS.data_dir, rconst.FLAGFILE_TEMP)
```
reedwm: You should also create the data_dir if necessary. Currently, I don't think it's created if synthetic data is used.

official/recommendation/data_async_generation.py
```diff
@@ -436,13 +440,26 @@ def remove_alive_file():
     gc.collect()
 
 
+def _parse_flagfile():
+  """Fill flags with flagfile."""
+  flagfile = os.path.join(flags.FLAGS.data_dir,
+                          rconst.FLAGFILE)
+  tf.logging.info("Waiting for flagfile to appear at {}..."
+                  .format(flagfile))
+  while not tf.gfile.Exists(flagfile):
```
reedwm: You should wait a max of rconst.TIMEOUT_SECONDS here, before giving up and exiting.

PR #5468, opened 2018-10-10
Add --use_synthetic_data option to NCF.

reedwm:


official/recommendation/data_preprocessing.py
```diff
@@ -570,9 +584,12 @@ def hash_pipeline(dataset, deterministic):
 
 
 def make_train_input_fn(ncf_dataset):
-  # type: (NCFDataset) -> (typing.Callable, str, int)
+  # type: (typing.Optional[NCFDataset]) -> (typing.Callable, str, int)
```
robieta: Ah, so that's how you do it. TIL

official/recommendation/data_preprocessing.py
```diff
@@ -644,10 +661,43 @@ def input_fn(params):
   return input_fn, record_dir, batch_count
 
 
+_SYNTHETIC_BATCHES_PER_EPOCH = 2000
```
robieta: We know that the number of train batches should just be  about `(1 or 20 million) * (1 + num_neg) / train_batch_size` and eval will just be about `num_users * (1 + num_eval_neg) / eval_batch_size`. Can you just mimic the real behavior?

official/recommendation/data_preprocessing.py
```diff
@@ -644,10 +661,43 @@ def input_fn(params):
...
+    items = tf.random_uniform([batch_size], dtype=tf.int32, minval=0,
+                              maxval=num_items)
+    labels = tf.random_uniform([batch_size], dtype=tf.int32, minval=0,
+                               maxval=2)
+
+    data = {
+        movielens.USER_COLUMN: users,
+        movielens.ITEM_COLUMN: items,
+    }, labels
+    dataset = tf.data.Dataset.from_tensors(data).repeat(
```
robieta: Didn't you want to have some interleaving here? So as to better mimic the threading behavior?

reedwm: I have no map_func, so there is nothing to interleave. Also, with synthetic data, I'm not really trying to measure the performance of the input pipeline.

official/recommendation/data_preprocessing.py
reedwm: Running for more batches is convenient to reduce noise. Added a comment stating this.

reedwm:
PTAL

robieta:
Once the tests are fixed I'll approve.

PR #5513, opened 2018-10-13
Use tf.gfile.Open to open flagfile as otherwise files on google cloud storage etc cannot be opened.

shizhiw:


official/recommendation/data_async_generation.py
```diff
@@ -455,17 +455,25 @@ def _parse_flagfile():
...
 
 def main(_):
   global _log_file
-  _parse_flagfile()
+  flagfile = _parse_flagfile()
 
   redirect_logs = flags.FLAGS.redirect_logs
   cache_paths = rconst.Paths(
       data_dir=flags.FLAGS.data_dir, cache_id=flags.FLAGS.cache_id)
+  # Rename flagfile so the next run at the same data_dir will still work.
```
robieta: This seems very brittle. Currently multiple jobs can run from the same data dir, because everything temporary is in a unique cache directory. I think we want to preserve that property. (Apologies for not catching this in the review of the last PR)

robieta: My thinking is that the cache id in ncf main would default to none. If none, generate on the fly as we do now. That way the user need not bother with it unless they need it to synchronize across workers.

shizhiw:
Commits have either been merged or addressed by #5521.

PR #5521, opened 2018-10-13
Make flagfile sharing robust to distributed filesystems and multi-worker setups.

robieta:


robieta:
Continuing discussion from: https://github.com/tensorflow/models/pull/5513

official/recommendation/data_async_generation.py
```diff
@@ -455,13 +453,23 @@ def _parse_flagfile():
       sys.exit()
     time.sleep(1)
   tf.logging.info("flagfile found.")
-  # This overrides FLAGS with flags from flagfile.
-  flags.FLAGS([__file__, "--flagfile", flagfile])
+
+  # `flags` module opens `flagfile` with `open`, which does not work on
+  # google cloud storage etc.
+  _, flagfile_temp = tempfile.mkstemp()
```
shizhiw: Are you sure this works everywhere? e.g. the green VM on cloud TPU does not seem to have local disk. Why not just pass flagfile's content as a list of string to flags.FLAGS? It seems safer and I've tested on both borg and cloud TPU.

robieta: No python code runs on the TPU host; it all runs on the user controlled instance. I believe internal implementations of tempfile work as well. Basically treating a flagfile as a flagfile seemed cleaner than string manipulating it in.

PR #5536, opened 2018-10-15
Reorder NCF data pipeline

robieta:
This PR does two things:

1) It replaces the "spillover" mechanism with a more conventional padding which is ignored by the cross entropy. @shizhiw In the last batch the padded keys all lookup at index zero. Is this hot key going to be an issue? (If so I can spread out the keys; it's just easier to read if all the paddings have the same value.)

2) It consolidates a lot of the eval generation and merges the training and eval generation. This removes some duplicate code. However, it also makes the performance properties much better:
 * Training and eval now use the same base data shards, saving a ~2GB .npz intermediate file
 * The eval set is now computed after the first training epoch, so training can begin sooner.
 * The clock start can be moved to right before the subprocess starts, because the random sampling for eval no longer happens as part of the main thread data preprocessing.

### DIFF
--- a/official/recommendation/data_async_generation.py
+++ b/official/recommendation/data_async_generation.py
@@ -92,8 +92,8 @@ def _process_shard(args):
   with tf.gfile.Open(shard_path, "rb") as f:
     shard = pickle.load(f)
 
-  users = shard[movielens.USER_COLUMN]
-  items = shard[movielens.ITEM_COLUMN]
+  users = shard[rconst.TRAIN_KEY][movielens.USER_COLUMN]
+  items = shard[rconst.TRAIN_KEY][movielens.ITEM_COLUMN]
 
   delta = users[1:] - users[:-1]
   boundaries = ([0] + (np.argwhere(delta)[:, 0] + 1).tolist() +
@@ -168,8 +168,6 @@ def _construct_training_records(
     epochs_per_cycle,     # type: int
     train_batch_size,     # type: int
     training_shards,      # type: typing.List[str]
-    spillover,            # type: bool
-    carryover=None,       # type: typing.Union[typing.List[np.ndarray], None]
     deterministic=False   # type: bool
     ):
   """Generate false negatives and write TFRecords files.
@@ -190,24 +188,19 @@ def _construct_training_records(
       to properly batch data when writing TFRecords.
     training_shards: The picked positive examples from which to generate
       negatives.
-    spillover: If the final batch is incomplete, push it to the next
-      cycle (True) or include a partial batch (False).
-    carryover: The data points to be spilled over to the next cycle.
   """
 
   st = timeit.default_timer()
   num_workers = min([num_workers, len(training_shards) * epochs_per_cycle])
-  carryover = carryover or [
-      np.zeros((0,), dtype=np.int32),
-      np.zeros((0,), dtype=np.uint16),
-      np.zeros((0,), dtype=np.int8),
-  ]
-  num_carryover = carryover[0].shape[0]
-  num_pts = num_carryover + num_train_positives * (1 + num_neg)
+
+  num_pts = num_train_positives * (1 + num_neg)
+  num_pts_with_padding = int(
+      np.ceil(num_pts / train_batch_size)) * train_batch_size
+  num_padding = num_pts_with_padding - num_pts
 
   # We choose a different random seed for each process, so that the processes
   # will not all choose the same random numbers.
-  process_seeds = [np.random.randint(2**32)
+  process_seeds = [stat_utils.random_int32()
                    for _ in training_shards * epochs_per_cycle]
   map_args = [(shard, num_items, num_neg, process_seeds[i])
               for i, shard in enumerate(training_shards * epochs_per_cycle)]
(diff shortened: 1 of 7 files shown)


robieta:
P.S. This is a prerequsite for adding logging. (Since it dictates where the tags can go.)

official/recommendation/ncf_main.py
```diff
@@ -156,8 +156,11 @@ def run_ncf(_):
         cache_id=FLAGS.cache_id)
     num_users = ncf_dataset.num_users
     num_items = ncf_dataset.num_items
-    approx_train_steps = int(ncf_dataset.num_train_positives
-                             * (1 + FLAGS.num_neg) // FLAGS.batch_size)
+    num_train_steps = int(np.ceil(
+      FLAGS.epochs_between_evals * ncf_dataset.num_train_positives *
```
shizhiw: do you need +1 here?

shizhiw: nvm, you're right.

official/recommendation/data_async_generation.py
```diff
@@ -157,75 +176,80 @@ def init_worker():
   signal.signal(signal.SIGINT, sigint_handler)
 
 
-def _construct_training_records(
+def _construct_records(
```
reedwm: Won't padding the final training batch every epoch slow down training? Why remove spillover?

official/recommendation/data_async_generation.py
```diff
@@ -387,23 +380,32 @@ def remove_alive_file():
...
-                         eval_batch_size=eval_batch_size)
+  _construct_records(
+    is_training=True, train_cycle=train_cycle, num_neg=num_neg,
+    num_positives=num_train_positives, epochs_per_cycle=epochs_per_cycle,
+    batch_size=train_batch_size, **shared_kwargs)
+
+  # Construct evaluation set.
+  shared_kwargs["num_workers"] = num_workers
+  _construct_records(
+    is_training=False, train_cycle=-1, num_neg=rconst.NUM_EVAL_NEGATIVES,
```
reedwm: Set train_cycle to None instead

official/recommendation/data_preprocessing.py
```diff
@@ -686,90 +652,39 @@ def input_fn(params):
...
 
 
-def make_train_synthetic_input_fn():
+def make_synthetic_input_fn(is_training):
   """Construct training input_fn that uses synthetic data."""
   def input_fn(params):
     """Generated input_fn for the given epoch."""
-    batch_size = params["batch_size"]
+    batch_size = (params["batch_size"] if is_training else
+                  params["eval_batch_size"])
```
reedwm: The eval batch size should default to `params["batch_size"]` if not set

official/recommendation/data_async_generation.py
```diff
@@ -104,16 +112,27 @@ def _process_shard(args):
...
 
-    negatives = stat_utils.sample_with_exclusion(
-        num_items, positive_set, n_pos * num_neg)
+    if is_training:
+      n_pos = len(positive_set)
+      negatives = stat_utils.sample_with_exclusion(
+        num_items, positive_set, n_pos * num_neg, replacement=True)
+
+    else:
+      positive_set.add(test_positive_dict[current_user])
```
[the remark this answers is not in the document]
robieta: I'll key this off of `match_mlperf`. Good catch.

official/recommendation/data_async_generation.py
robieta: Rules. Given that this thing trains at ~100 batches/sec, it is unlikely to cause too much harm.

official/recommendation/data_async_generation.py
```diff
@@ -92,8 +94,14 @@ def _process_shard(args):
...
     shard = pickle.load(f)
 
-  users = shard[movielens.USER_COLUMN]
-  items = shard[movielens.ITEM_COLUMN]
+  users = shard[rconst.TRAIN_KEY][movielens.USER_COLUMN]
+  items = shard[rconst.TRAIN_KEY][movielens.ITEM_COLUMN]
+
+  if not is_training:
+    # For eval, there is one positive which was held out from the training set.
+    test_positive_dict = {k: v for k, v in zip(
```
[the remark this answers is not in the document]
robieta: Turns out that creates a set of tuples, not a dict.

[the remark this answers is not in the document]
reedwm: I never lead you astray. You need to literal use the word `dict` instead of using the `{...}` syntax.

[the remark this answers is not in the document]
robieta: I have remedied the grievous slander to your name.

official/recommendation/ncf_main.py
```diff
@@ -206,29 +209,40 @@ def run_ncf(_):
...
-  pred_input_fn = data_preprocessing.make_pred_input_fn(ncf_dataset=ncf_dataset)
 
+  pred_input_fn = None
   total_training_cycle = FLAGS.train_epochs // FLAGS.epochs_between_evals
   for cycle_index in range(total_training_cycle):
     tf.logging.info("Starting a training cycle: {}/{}".format(
         cycle_index + 1, total_training_cycle))
 
     # Train the model
     train_input_fn, train_record_dir, batch_count = \
```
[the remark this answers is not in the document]
robieta: My inclination is that that is a non-trivial change. Lets get this in, follow up, and we can address in a subsequent PR if it turns out that that's how we want to go.

PR #5546, opened 2018-10-16
Improvements to AstroNet and add AstroWaveNet

cshallue:


cshallue:
All authors have now signed the CLA

PR #5571, opened 2018-10-19
Fixes Python 3.x error and dependency deprecation warning in research/skip_thoughts

bapfeld:
Original code attempts to apply `str.decode('utf-8')`, which throws an error in python 3.x. This now uses a try/except block to catch the error and allow backwards compatibility.

It also fixes a deprecation error on a dependency.


bapfeld:
Yes, changing from 'r' to 'rb' does work. Made the changes!

PR #5604, opened 2018-10-25
optional parameter error in calling _build_faster_rcnn_feature_extractor

haichaoyu:


haichaoyu:
I signed it!

pkulzc:
Inplace BN update is not supported in Faster RCNN. So this is set to false intentionally. See [here](https://github.com/tensorflow/models/blob/1f484095c0981e2a62403b16256cb877749dfe94/research/object_detection/builders/model_builder.py#L302-L303).

haichaoyu:
Yes, It it not supported, but it seems the parameter take the wrong place of reuse_weights.

pkulzc:
Oh that's right. Thanks for noticing this!

PR #5615, opened 2018-10-25
Split --ml_perf into two flags.

reedwm:
--ml_perf now just changes the model to make it MLPerf compliant. The compliance logging is now only done when flag --output_ml_perf_compliance_logging is specified.

Also add a test for --ml_perf.


robieta:
This will not stop the async process from logging.

reedwm:
PTAL

PR #5623, opened 2018-10-27
Add option to not use estimator.

reedwm:
To test convergence, I ran the following 48 times.
```
python ncf_main.py --model_dir=$HOME/ncf_model_dir_gpu_0 --data_dir=$HOME/ncf_data_dir_gpu_0 --dataset ml-20m --hooks= --num_gpus -1 --clean --train_epochs 14 --batch_size 65536 --eval_batch_size 65536 --layers 256,256,128,64 --num_factors 64 --hr_threshold 0.99 --ml_perf --learning_rate=0.00281936 --beta1=0.0490852 --beta2=0.965095 --epsilon=1.23757e-08 --use_xla_for_gpu --nouse_estimator
```

I hit a hit-rate of 63.5 44/48 times.

### DIFF
--- a/official/recommendation/model_runner.py
+++ b/official/recommendation/model_runner.py
@@ -64,17 +64,18 @@ class NcfModelRunner(object):
           "metric_initializer",))
 
   def __init__(self, ncf_dataset, params):
-    if params["use_xla_for_gpu"]:
-      # The XLA functions we use require resource variables.
-      tf.enable_resource_variables()
-    self._ncf_dataset = ncf_dataset
-    self._global_step = tf.train.create_global_step()
-    self._train_model_properties = self._build_model(params, is_training=True)
-    self._eval_model_properties = self._build_model(params, is_training=False)
+    with tf.Graph().as_default() as self._graph:
+      if params["use_xla_for_gpu"]:
+        # The XLA functions we use require resource variables.
+        tf.enable_resource_variables()
+      self._ncf_dataset = ncf_dataset
+      self._global_step = tf.train.create_global_step()
+      self._train_model_properties = self._build_model(params, is_training=True)
+      self._eval_model_properties = self._build_model(params, is_training=False)
 
-    initializer = tf.global_variables_initializer()
-    tf.get_default_graph().finalize()
-    self._session = tf.Session()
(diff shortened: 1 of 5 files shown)


official/recommendation/model_runner.py
```diff
@@ -0,0 +1,206 @@
...
+      initializer_feed_dict = None
+      record_dir = None
+
+    self._session.run(model_properties.iterator.initializer,
+                      initializer_feed_dict)
+    fetches = (model_properties.loss, model_properties.run_model_op)
+    mode = "Train" if is_training else "Eval"
+    start = None
+    for i in range(num_steps):
+      loss, _, = self._session.run(fetches)
```
robieta: You've tested that there is no overhead from calling this one at a time?

official/recommendation/ncf_main.py
```diff
@@ -237,37 +244,46 @@ def run_ncf(_):
...
-    if train_record_dir:
-      tf.gfile.DeleteRecursively(train_record_dir)
-
-    tf.logging.info("Beginning evaluation.")
-    if pred_input_fn is None:
-      pred_input_fn, _, eval_batch_count = data_preprocessing.make_input_fn(
-          ncf_dataset=ncf_dataset, is_training=False)
-
-      if eval_batch_count != num_eval_steps:
+    if FLAGS.use_estimator:
```
robieta: Have you confirmed that the estimator path still works? (1 run is sufficient.)

official/recommendation/model_runner.py
reedwm: I haven't measured, but this is still better than what Estimator does. If there is overhead, we can further optimize this.

official/recommendation/ncf_main.py
reedwm: I ran for two epochs on ml-1m, which worked.

official/recommendation/model_runner.py
robieta: Sounds good.

official/recommendation/ncf_main.py
robieta: good 'nuf.

PR #5631, opened 2018-10-29
Updates official NCF with TC-optimized changes

tayo:
This PR attempts to integrate the TPU-TC Estimator implementation with the main version.

Changes that should be discussed in particular:
- The eval_pred_fn creation occurs in every loop iteration.  Can this be pulled out of the loop?
- The calculation of num_train_steps was previously called 'approx_train_steps' and was slightly diffferent.  One version wraps the equation in a ceil() and the other does not.  Correctness should be confirmed.

### DIFF
--- a/official/recommendation/data_preprocessing.py
+++ b/official/recommendation/data_preprocessing.py
@@ -593,13 +593,81 @@ def hash_pipeline(dataset, deterministic):
   tf.logging.info("  [pipeline_hash] All batches hash: {}".format(overall_hash))
 
 
-def make_input_fn(ncf_dataset, is_training):
-  # type: (typing.Optional[NCFDataset], bool) -> (typing.Callable, str, int)
+def make_input_fn(
+    ncf_dataset,       # type: typing.Optional[NCFDataset]
+    is_training,       # type: bool
+    record_files=None  # type: typing.Optional[tf.Tensor]
+    ):
+  # type: (...) -> (typing.Callable, str, int)
   """Construct training input_fn for the current epoch."""
 
   if ncf_dataset is None:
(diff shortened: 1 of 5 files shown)


tayo:
Please add robieta@ as a reviewer.

robieta:
Lovely. Thanks for bringing these optimizations back to open source. To provide some context on your concerns.

1) The reason for the somewhat curious placement of the eval input function creation is to spend less time waiting. The async process first generates the first epoch of training data, then the eval data, and then subsequent epochs. And because the main thread will block on input_fn creation waiting for data, we want the eval input_fn to be created after the first epoch of training. However it will only be created once, and then reused.

2) Previously, the last partial batch of an epoch would spill over into the next epoch. This meant that every few epochs there would be an extra batch. That mechanism has been removed and training now randomly resamples the epoch for points with which to pad. This means that the number of training batches per epoch is now constant.

Hope that helps.

tayo:
That all makes sense, thanks.

robieta:
Overall this looks fine. Once https://github.com/tensorflow/models/pull/5623 lands you should rebase to that (or whatever you preferred git workflow is). The interactions between that and this PR should be pretty minor.

P.S. Because of the way that Kokoro works, any merge conflict means that you auto-fail tests without running them. (That's why they're broken.)

tayo:
Chatted with reedwm@ offline.  Seems like the diff history of this branch has been lost somehow, as the PR currently includes both commits.  I will extract the diff and re-apply it to a clean master.

tayo:
Closing this pull request. Will send a fresh one with the correct diff.

PR #5641, opened 2018-10-30
Keras-ify NCF TPU embedding lookup

robieta:
Keras throws
```
ValueError: Output tensors to a Model must be the output of a TensorFlow `Layer` (thus holding past layer metadata). Found: Tensor("rating/BiasAdd:0", shape=(12288, 1), dtype=float32)
```

when trying to run on the TPU since embedding lookups are tensors rather than Keras layers. This PR simply wraps the tensor operations in Lambda layers. There may be a more Keras-y way to do it, but this way at least works.

@tayo This doesn't seem to have a performance impact, but can you confirm that this is fine from your side?

### DIFF
--- a/official/recommendation/neumf_model.py
+++ b/official/recommendation/neumf_model.py
@@ -200,28 +200,25 @@ def construct_model(users, items, params):
 
   if params["use_tpu"]:
     with tf.variable_scope("embed_weights", reuse=tf.AUTO_REUSE):
-      cmb_embedding_user = tf.get_variable(
+      cmb_user_latent = tf.keras.layers.Lambda(lambda ids: tf.gather(
+        tf.get_variable(
           name="embeddings_mf_user",
           shape=[num_users, mf_dim + model_layers[0] // 2],
-          initializer=tf.glorot_uniform_initializer())
-      cmb_embedding_item = tf.get_variable(
+          initializer=tf.glorot_uniform_initializer()),
+        ids))(user_input)
+
+      cmb_item_latent = tf.keras.layers.Lambda(lambda ids: tf.gather(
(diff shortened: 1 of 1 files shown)


official/recommendation/neumf_model.py
```diff
@@ -200,28 +200,36 @@ def construct_model(users, items, params):
 
   if params["use_tpu"]:
     with tf.variable_scope("embed_weights", reuse=tf.AUTO_REUSE):
-      cmb_embedding_user = tf.get_variable(
+      cmb_user_latent = tf.keras.layers.Lambda(lambda ids: tf.gather(
+        tf.get_variable(
```
reedwm: If possible, move the call to `tf.get_variable` outside the lambda. It's a bit misleading to call `tf.get_variable` in a Keras layer, since if the Keras model is later called outside the variable scope, `tf.get_variable` will not be affected by the variable scope.

tayo:
I'm fine with this for now.  Feel free to merge if tests are passing.

Later on, perhaps a cleaner way might be to wrap the whole computation in one giant Lambda.

robieta:
@reedwm PTAL

official/recommendation/neumf_model.py
```diff
@@ -199,29 +199,39 @@ def construct_model(users, items, params):
   batch_size = user_input.get_shape()[0]
 
   if params["use_tpu"]:
+    cmb_embedding_user = tf.get_variable(
```
reedwm: You probably need these under the variable_scope. Otherwise, the variable_scope doesn't do anything at all

robieta: Good catch. Thanks.

PR #5646, opened 2018-10-31
Update faster_rcnn_inception_resnet_v2_feature_extractor.py

Sajjadmanal:
Shouldn't it be or instead of and?


Sajjadmanal:
I have signed it.

pkulzc:
No it shouldn't be.

PR #5648, opened 2018-10-31
Fix minor bug in images/sec calculation.

reedwm:


reedwm:
Closing, as model_runner.py has been long deleted.

PR #5652, opened 2018-11-01
Have async process end when all data is written.

reedwm:
I've noticed sometimes the async process's pool processes do not die when ncf_main.py ends and kills the async process. This commit fixes the issue.

### DIFF
--- a/official/recommendation/data_preprocessing.py
+++ b/official/recommendation/data_preprocessing.py
@@ -392,9 +392,10 @@ def _shutdown(proc):
   tf.logging.info("Shutting down train data creation subprocess.")
   try:
     try:
-      proc.send_signal(signal.SIGINT)
+      proc.terminate()
       time.sleep(5)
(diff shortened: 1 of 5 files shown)


official/recommendation/data_preprocessing.py
```diff
@@ -392,9 +392,10 @@ def _shutdown(proc):
   tf.logging.info("Shutting down train data creation subprocess.")
   try:
     try:
-      proc.send_signal(signal.SIGINT)
```
robieta: I'm reluctant to get rid of this. If a sigint is sent, the subprocess can catch the KeyboardInterrupt and exit gracefully. terminate() doesn't afford that luxury.

reedwm: Done. I thought changing this was required, but it's not.

reedwm:
PTAL

Issue #4907, opened 2018-07-26
[Object Detection] Missing oid_challenge_object_detection_metrics in eval_util.py


Akz47:
### System information
- **What is the top-level directory of the model you are using**: C:\Data\Open Images\
- **Have I written custom code (as opposed to using a stock example script provided in TensorFlow)**: No
- **OS Platform and Distribution (e.g., Linux Ubuntu 16.04)**: Windows 10
- **TensorFlow installed from (source or binary)**: Binary
- **TensorFlow version (use command below)**: Tensorflow-gpu 1.7.0
- **Bazel version (if compiling from source)**:
- **CUDA/cuDNN version**: 9.1, 7.0
- **GPU model and memory**: GTX 1070, 8Gb
- **Exact command to reproduce**: python object_detection/model_main.py

### Describe the problem
I'm having problems training & evaluating the Open Images dataset, because the "**oid_challenge_object_detection_metrics**" is not defined.

It is defined in the documentation:
[https://github.com/tensorflow/models/blob/master/research/object_detection/g3doc/evaluation_protocols.md](https://github.com/tensorflow/models/blob/master/research/object_detection/g3doc/evaluation_protocols.md)

However, the code does not exist in **eval_util.py**, which only defines the processing logic for "coco_detection_metrics" and "coco_mask_metrics".

Thus, when I run object_detection/model_main.py, it terminates with the error:
>  File "object_detection/eval_util.py", line 650, in get_eval_metric_ops_for_evaluators
>     'Found {} in the evaluation metrics'.format(metric))
> ValueError: The only evaluation metrics supported are "coco_detection_metrics" and "coco_mask_metrics". Found oid_challenge_object_detection_metrics in the evaluation metrics

It is defined in my pipeline config:
> eval_config: {
>   metrics_set: "oid_challenge_object_detection_metrics"
>   num_examples: 8000
>   max_evals: 10
> }

I found a file under **metrics/oid_od_challenge_evaluation.py**, but I am not sure how to integrate it into the eval_util.py.

How do I properly run the model training and evaluation? Thank you.

LenLen33:
As far as I can tell the oid_od_challenge_evaluation.py is not meant to be integrated in eval_util.py.

I'm currently working on a custom evaluator so treat my words with caution:

Basically what you should do is to create a file which is equivalent to the coco_evaluation.py file. You can reuse some parts, but others have to be modified.

The function you need to rewrite in that file is:
```python
def evaluate(self):
```

[...]

PR #5673, opened 2018-11-02
[Object Detection] Update evaluation protocol docs

seanpmorgan:
Updated the documentation to accurately reflect [which evaluation protocols](https://github.com/tensorflow/models/blob/master/research/object_detection/eval_util.py#L38) can be used in the current API.

Without these doc updates, users will be thrown `ValueError: Metric not found` when using a metric from the [list of eval protocols](https://github.com/tensorflow/models/blob/master/research/object_detection/legacy/evaluator.py#L33) that are now in the legacy directory.

Related issue:
https://github.com/tensorflow/models/issues/4907

### DIFF
--- a/research/object_detection/g3doc/evaluation_protocols.md
+++ b/research/object_detection/g3doc/evaluation_protocols.md
@@ -1,10 +1,66 @@
 # Supported object detection evaluation protocols
 
-The Tensorflow Object Detection API currently supports three evaluation protocols,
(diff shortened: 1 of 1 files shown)


seanpmorgan:
a1337e01db4e3a54c48d2fbd6614c772bdf0f4c5 added support for the other metrics. This PR is stale

PR #5674, opened 2018-11-02
Update data_download.py

albus7:
input_url doesn't get correct URL in Windows environment. The os.path.join() function add "\\" to the path, and it got incorrect URL for the data file.


albus7:
CLA is signed

PR #5677, opened 2018-11-04
Use explicit relative import syntax for python 3 compatibility in audioset.

dthkao:


plakal:
@dthkao did you try running the smoke test and with which version of Python?

Also, do you know why you had to do this change at all? The previous version of the code, without relative imports, worked fine with Python 3.4.3 (see README.md's Installation section for tested configurations)

dthkao:
Oh no. I ran the smoke test differently. I think what we're seeing is described here:
https://stackoverflow.com/questions/14132789/relative-imports-for-the-billionth-time

Mainly, running the script from within the audioset/ directory results in different module import names  than usage as library. In fact the smoke test isn't catching python 3 package-external usage errors; try importing prior to this PR from a `foo.py` in models/research.

FWIW, if the CWD is `models/`, the following will pass in python 3 (python 2 needs  `__init__.py` files added along the path):
`models/ $ python3 -m research.audioset.vggish_smoke_test`

A few options:
0: Revert this back for now. This also probably makes the most since while we figure out what to do.
1: Leave the imports as is, add the `__init__.py` for python 2 usage, and update instructions on smoke test execution.
2: Some form of hardcoding `__package__` inside all of the modules (not a fan of this one).
3: Something else?

dthkao:
Also, anything else executed as a script from within audioset/ (i.e., the demos) is probably hitting the same issue. :disappointed:

Issue #5678, opened 2018-11-04
Audioset SmokeCheck "Cannot Import Name" Error


BMarcelus:
Following the installation tutorial for the audioset research model results in "cannot import name" errors for all Explicit Relative Import statements
`from . import <file>`
Causes an error while
`import <file>`
does not

The change was made recently from this pull request https://github.com/tensorflow/models/pull/5677
Please let me know if I am missing an installation step to get this method of imports to work.

plakal:
Sorry for the breakage, @BMarcelus. We've just merged a rollback of the change, can you confirm that things are working again at head, and re-open if not?

PR #5692, opened 2018-11-05
Revert "Use explicit relative import syntax for python 3 compatibility in audioset."

dthkao:
As discussed in the original PR (tensorflow/models#5677), this introduces issues when running the existing code as a script.

Reverts tensorflow/models#5677


plakal:
Fixes https://github.com/tensorflow/models/issues/5678

PR #5699, opened 2018-11-06
Fix bug in dataset.py when batch_size==max_length.

oscartackstrom:
Each batch contains sequences of length in [buckets_min, buckets_max), so that the longest instance in a batch will have length buckets_max - 1. Without this fix, instances with length == batch_size will result in buckets of zero elements, causing the dataset iterator to crash.

As a bonus this adds an extra instance per batch.


oscartackstrom:
I signed it!

jaeyounkim:
I am closing this PR because we made many changes already.

PR #5716, opened 2018-11-07
Switching to more robust pysoundfile for reading wav files

SNeugber:
Scipy struggles with some WAV formats (e.g. when there is additional metadata present: https://stackoverflow.com/a/18958665/1194335). Reading the same files using the Pysoundfile works.


SNeugber:
I signed it!

SNeugber:
Sure! I'll open a new PR with that change in place (I realized that the commit was signed with my work email which I haven't signed the CLA with, so I'll need to amend that).

PR #5743, opened 2018-11-12
NCF: Enabled inter/intra-op parallelism threads by default.

bhavani-subramanian:
If these flags are not passed, a default value of 0 will be used.

### DIFF
--- a/official/recommendation/ncf_main.py
+++ b/official/recommendation/ncf_main.py
@@ -181,7 +181,10 @@ def construct_estimator(num_gpus, model_dir, params, batch_size,
     return train_estimator, eval_estimator
 
   distribution = distribution_utils.get_distribution_strategy(num_gpus=num_gpus)
-  run_config = tf.estimator.RunConfig(train_distribute=distribution)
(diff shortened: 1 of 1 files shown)


robieta:
A value of zero simply means that TensorFlow will choose default values. Why is it necessary for these flags to be enabled in NCF? (We tend not to include performance flags unless they are needed.)

bhavani-subramanian:
@robieta Setting these parameters boosts CPU performance (https://software.intel.com/en-us/articles/tensorflow-optimizations-on-modern-intel-architecture), and AFAIK it has no effect on GPU performance.

If these flags are not passed, it can easily lead to over-subscription of threads as the default behavior is to utilize all available physical cores. This may not always be desirable when the wants to explicitly run on specific cores in the system (say on a single socket).

Enabling these flags provides a way for the user to override the default settings.

tfboyd:
This was a good idea and will get added.  Sorry we did not end up taking this PR.  I do not want you to have to deal with the merge conflicts.  I do want to say thank you and we are likely adding these flags to NCF for the reasons you stated.

PR #5744, opened 2018-11-12
fixed scaling of channels

qbilius:
When loading weights for channel-scaled versions of mobilenet, such as `mobilenet_v2_0.75_224`, I get an error:

`Assign requires shapes of both tensors to match. lhs shape= [1,1,240,960] rhs shape= [1,1,240,1280]`.

This seems to stem from the fact that the layer just prior to avgpool is supposed to always have no less than 1280 channels -- yet it is incorrectly scaled down to 960 instead when the mutiplier is .75.

This PR fixes the issue that mobilenets appear to work as expected.


qbilius:
I signed it!

marksandler2:
Note you can just use finegrain_classification_mode=True parameter  in mobilenet_v2

PR #5750, opened 2018-11-13
Fix setup steps in mobilenet_example.ipynb

seanbae:
Fix the setup steps in `mobilenet_example.ipynb`.

1. The existing Jupyter notebook interchangeably uses `base_name` and `checkpoint_name`, but `base_name` is never defined.

This leads to the vanilla setup to fail with `NameError`:
screenshot 2018-11-13 13 07 12

This PR replaces all `base_name` references to `checkpoint_name`.

2. Fix an incorrect `slim` path.


jaeyounkim:
Fixed

PR #5752, opened 2018-11-14
num_replicas -> num_replicas_in_sync

josh11b:
`num_replicas` is being replaced by `num_replicas_in_sync` and something that returns the number of steps running concurrently.


robieta:
Can you also look into the `AttributeError: 'OneDeviceStrategy' object has no attribute 'call_for_each_tower'` error?

josh11b:
> Can you also look into the `AttributeError: 'OneDeviceStrategy' object has no attribute 'call_for_each_tower'` error?

That is from using an outdated version of tensorflow_estimator/python/estimator/estimator.py and is should be unrelated to this PR (I suspect that is failing without this PR).  @case540 may be able to help.

PR #5755, opened 2018-11-14
Avoid using tf.contrib.data as it's not tf2-safe

alextp:


robieta:
The tests are failing for an unrelated reason. I'll rerun when that issue is resolved.

PR #5777, opened 2018-11-19
Update running_pets.md

Vyas9296:
Line 219 runtime version updated to 1.9, current version of 1.8 will trigger 'TypeError: non_max_suppression() got an unexpected keyword argument 'score_threshold'' on the Google Cloud since 1.8 and older does not support this keyword argument.

### DIFF
--- a/research/object_detection/g3doc/running_pets.md
+++ b/research/object_detection/g3doc/running_pets.md
@@ -216,7 +216,7 @@ To start training and evaluation, execute the following command from the
 ```bash
 # From tensorflow/models/research/
 gcloud ml-engine jobs submit training `whoami`_object_detection_pets_`date +%m_%d_%Y_%H_%M_%S` \
-    --runtime-version 1.8 \
+    --runtime-version 1.9 \
     --job-dir=gs://${YOUR_GCS_BUCKET}/model_dir \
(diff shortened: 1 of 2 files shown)


Vyas9296:
I signed it!

Vyas9296:
Updated doc line 211

arinto:
should we update the [object_detection/samples/cloud/cloud.yaml](https://github.com/tensorflow/models/blob/master/research/object_detection/samples/cloud/cloud.yml#L2) as well?

Vyas9296:
Thanks @arinto, have updated the runtime version as well. Also re-emphasised in the doc that Tensorboard and Tensorflow need matching minor versions (1.x).

arinto:
I meant runtime file in `object_detection/samples/cloud/cloud.yaml` file. Is it working without updating the runtime version in that file? 
Or the runtime version from that config file is overwritten by CLI flag from README.

Vyas9296:
Hi @arinto, thanks for pointing that out. I think it is overwritten, because I ran training successfully with both 1.8 and 1.9 in the cloud.yml file. But we should update to 1.9 for consistency. Do I need to create a new pull request for that? I don't think I can add it to this one

PR #5781, opened 2018-11-19
Improved performance of deeplab inference

adrianboguszewski:
Casting predictions to float32 allows to schedule slice and resize operations on GPU, which reduces much memory copying between device and host.

In our case (xception65, 3328x3328 input size and GeForce RTX 2080Ti.) it decreased inference time by 32%.

Below screens from profiling.
Before:
before
After:
after


adrianboguszewski:
I signed it!

YknZhu:
Thanks!

PR #5786, opened 2018-11-20
WIP: Replace pipeline in NCF

robieta:
Note: this PR still needs a fair bit of polish before it is ready to merge.

Major changes:
1) Bring data generation back into the main process. (Though on a side thread.) `data_async_generation.py` has been removed, and `data_pipeline.py` is its spiritual successor.
2) Materialize all negative examples at the start, and replace the per batch set exclusion Monte Carlo with random lookups. This approach does not scale as it requires a (num_users x num_items) matrix. I have a plan for larger datasets but it will not be in this PR.
3) The last partial batch is now padded and masked instead of resampled.
4) The pandas work at the start is now cached and invalidated on a daily basis. This is just a quality of life improvement, as it means that two minutes of waiting at the start goes away.
5) I've updated the parameters in run.sh to use a larger batch size. 2k just isn't enough to keep an accelerator busy.

Performance:
8-30 second startup cost to materialize the full (user x item matrix). (depending on machine)
4-5 seconds to generate an epoch.

Unfortunately a lot of these operations haven't threaded nearly as well as I expected that they should, so on large machines most of the cores are unused. Naive multiprocessing doesn't help, because you spend more time serializing and deserializing than you gain from the extra workers. The solution is likely going to be using `multiprocessing.sharedctypes.RawArray` and `np.frombuffer` to get around the multiprocessing pickling, but I haven't gotten a chance to do it yet. Not altogether terrible, but definitely lots of room for finesse.

Breakages:
1) Model runner. I'm considering removing it altogether.
2) TPUs. I'm now using Dataset.from_generator, which I can't ship to a TPU. My plan is to write local files and us `StreamingFilesDataset`; putting TFRecords in GCS is probably not optimal going forward.
3) Multi-GPU doesn't work with XLA for unknown reasons. (I haven't tested if this is also broken on TPUs)

My plan is to just live with the sub-optimal pipeline performance for this round, and focus on getting a baseline in. That should hopefully unblock Shining for his Keras work.


robieta:
Update: This PR is ready for review. While there is still room for improvement, performance is reasonable on both GPU and TPU. The only things that are currently not implemented is deterministic behavior and pipeline hashing, as those are more difficult with the new way that things are ordered. However as those are minor features I don't intend to block on their completion.

@reedwm @seemuch @shizhiw PTAL

official/recommendation/ncf_main.py
```diff
@@ -140,105 +196,35 @@ def run_ncf(_):
...
-        use_subprocess=FLAGS.use_subprocess,
-        cache_id=FLAGS.cache_id)
-    num_users = ncf_dataset.num_users
-    num_items = ncf_dataset.num_items
-    num_train_steps = int(np.ceil(
-        FLAGS.epochs_between_evals * ncf_dataset.num_train_positives *
-        (1 + FLAGS.num_neg) / FLAGS.batch_size))
-    num_eval_steps = int(np.ceil((1 + rconst.NUM_EVAL_NEGATIVES) *
-                                 ncf_dataset.num_users / eval_batch_size))
+        deterministic=FLAGS.seed is not None, params=params)
```
shizhiw: If this is not support yet, just remove this argument for now? We can add it back when deterministic behavior is supported.

robieta: Removed.

official/recommendation/data_pipeline.py
```diff
@@ -0,0 +1,727 @@
...
+        False, stream_files, self.eval_batches_per_epoch, self._shard_root)
+
+    # Threading details
+    self._current_epoch_order_lock = threading.RLock()
+    super(BaseDataConstructor, self).__init__()
+    self.daemon = True
+    self._stop_loop = False
+    self._fatal_exception = None
+
+  def __repr__(self):
```
reedwm: This should probably be \_\_str\_\_, not \_\_repr\_\_

official/recommendation/data_pipeline.py
```diff
@@ -0,0 +1,727 @@
...
+    map_args = [i for i in range(self.eval_batches_per_epoch)]
+    with popen_helper.get_threadpool(6) as pool:
+      pool.map(self._get_eval_batch, map_args)
+    self._eval_dataset.end_construction()
+
+    tf.logging.info("Eval construction complete. Time: {:.1f} seconds".format(
+        timeit.default_timer() - start_time))
+
+  def make_input_fn(self, is_training):
+    if self._fatal_exception is not None:
```
reedwm: A fatal exception can occur after this check, in which case you won't catch it and the program will likely hang. This is hard to fix, but maybe comment that this is a bit racy.

official/recommendation/data_pipeline.py
```diff
@@ -0,0 +1,727 @@
...
+    return input_fn
+
+
+class MaterializedDataConstructor(BaseDataConstructor):
+  """Materialize a table of negative examples for fast negative generation.
+
+  This class creates a table (num_users x num_items) containing all of the
+  negative examples for each user. This table is conceptually ragged; that is to
+  say the items dimension will have elements at the end which are not used equal
+  to the number of positive elements for a given user. For instance:
```
reedwm: > that is to say the items dimension will have elements at the end which are not used equal to the number of positive elements for a given user

I don't understand that clause. Removing the word "used" makes it grammatically correct

official/recommendation/data_pipeline.py
```diff
@@ -0,0 +1,727 @@
...
+    [0         2         4         int32max  int32max],
+    [1         2         3         4         int32max],
+    [0         int32max  int32max  int32max  int32max],
+  ]
+
+  and a vector of per-user negative counts, which in this case would be:
+    [3, 4, 1]
+
+  When sampling negatives, integers are (nearly) uniformly selected from the
+  range [0, per_user_neg_count[user]) which gives a column_index, at which
```
reedwm: Change ) to ]

official/recommendation/data_pipeline.py
```diff
@@ -0,0 +1,727 @@
...
+    for i in range(self._num_users):
+      positives = self._train_pos_items[index_bounds[i]:index_bounds[i+1]]
+      negatives = np.delete(full_set, positives)
+      self._per_user_neg_count[i] = self._num_items - positives.shape[0]
+      self._negative_table[i, :self._per_user_neg_count[i]] = negatives
+
+    tf.logging.info("Negative sample table built. Time: {:.1f} seconds".format(
+        timeit.default_timer() - start_time))
+
+  def lookup_negative_items(self, negative_users, **kwargs):
```
reedwm: Why does this take **kwargs?

official/recommendation/stat_utils.py
```diff
@@ -18,71 +18,40 @@
...
-    is the number of false negatives still needed. An additional factor of
-    safety of 1.2 is used during the generation to minimize the chance of having
-    to perform another generation cycle.
 
-    While this approach generates more values than needed and then discards some
-    of them, vectorized generation is inexpensive and turns out to be much
-    faster than generating points one at a time. (And it defers quite a bit
-    of work to NumPy which has much better multi-core utilization than native
-    Python.)
+def permutation(args):
```
reedwm: Add a docstring to this function. It's unclear what "args" is without reading the code.

robieta: Done.

robieta:
@reedwm PTAL

robieta:
@reedwm I added a bisection based constructor as well. (This is not my promised super elegant one approach; that one will have to wait.) The epoch construction time is longer than the materialized constructor (expected) with a shorter startup time for precompute that makes up for it as well as much more scalability for when we try to expand the dataset. Also having it in there makes me fell slightly less dirty for having made the materialized version.

@shizhiw Setting a seed now guarantees deterministic behavior in the producer.

official/recommendation/data_preprocessing.py
```diff
@@ -22,13 +22,13 @@
...
 import hashlib
-import multiprocessing
 import json
 import os
 import pickle
 import signal
 import socket
 import subprocess
+import threading
 import time
```
reedwm: Reminder to respond to my comments in this file. You responded to all comments in every file except this one.

official/recommendation/data_pipeline.py
```diff
@@ -0,0 +1,727 @@
...
+    # We pad the users and items here so that the duplicate mask calculation
+    # will include the padding. The metric function relies on every element
+    # except the positive being marked as duplicate to mask out padded points.
+    if users.shape[0] < users_per_batch:
+      pad_rows = users_per_batch - users.shape[0]
+      padding = np.zeros(shape=(pad_rows, users.shape[1]), dtype=np.int32)
+      users = np.concatenate([users, padding.astype(users.dtype)], axis=0)
+      items = np.concatenate([items, padding.astype(items.dtype)], axis=0)
+
+    duplicate_mask = stat_utils.mask_duplicates(items, axis=1).astype(np.bool)
```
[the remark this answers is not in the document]
reedwm: I think it does need to. You use the duplicated mask [here](https://github.com/tensorflow/models/blob/064c10e80543815a7425c890d32c83d082dfa8b8/official/recommendation/neumf_model.py#L379). If you do not mark it as a duplicate, it will be used in the HR and NDCG calculations.

PR #5815, opened 2018-11-26
Fix dependency issues

dreamdragon:


dreamdragon:
1. Fix dependency issues
2. Add two other authors to the code owners

Issue #5791, opened 2018-11-21
ssd_mobilenet_v1_quantized_300x300_coco14_sync unavailable fine_tune_checkpoint


baihualinxin:
------------------------

### System information
- **What is the top-level directory of the model you are using**:models-master\research\object_detection
- **Have I written custom code (as opposed to using a stock example script provided in TensorFlow)**:NO
- **OS Platform and Distribution (e.g., Linux Ubuntu 16.04)**:ubuntu 16.04 windwos 10
- **TensorFlow installed from (source or binary)**:pip
- **TensorFlow version (use command below)**:1.10.0
- **Bazel version (if compiling from source)**:0.17.01
- **CUDA/cuDNN version**:cuda 9 cudnn 8
- **GPU model and memory**:GTX 1080Ti
- **Exact command to reproduce**:

You can collect some of this information using our environment capture script:
b'v1.10.0-rc1-19-g656e7a2b34' 1.10.0

### Describe the problem
```
# SSD with Mobilenet v1 with quantized training.
# Trained on COCO, initialized from Imagenet classification checkpoint

# Achieves 18.2 mAP on coco14 minival dataset.

# This config is TPU compatible

model {
  ssd {
    inplace_batchnorm_update: true
    freeze_batchnorm: false
    num_classes: 1
    box_coder {
      faster_rcnn_box_coder {
        y_scale: 10.0
        x_scale: 10.0
        height_scale: 5.0
        width_scale: 5.0
      }
    }
    matcher {
      argmax_matcher {
        matched_threshold: 0.5
        unmatched_threshold: 0.5
        ignore_thresholds: false
        negatives_lower_than_unmatched: true
        force_match_for_each_row: true
        use_matmul_gather: true
      }
    }
    similarity_calculator {
      iou_similarity {
      }
    }
    encode_background_as_zeros: true
    anchor_generator {
      ssd_anchor_generator {
        num_layers: 6
        min_scale: 0.2
        max_scale: 0.95
        aspect_ratios: 1.0
        aspect_ratios: 2.0
        aspect_ratios: 0.5
        aspect_ratios: 3.0
        aspect_ratios: 0.3333
      }
    }
    image_resizer {
      fixed_shape_resizer {
        height: 300
        width: 300
      }
    }
    box_predictor {
      convolutional_box_predictor {
        min_depth: 0
        max_depth: 0
        num_layers_before_predictor: 0
        use_dropout: false
        dropout_keep_probability: 0.8
[...]

PR #5816, opened 2018-11-27
This function cannot be used

baihualinxin:
This function cannot be used


pkulzc:
If you look at other [configs](https://github.com/tensorflow/models/blob/7025590841588e82f371ef0fef7dd77a8a71efb8/research/object_detection/samples/configs/rfcn_resnet101_pets.config#L2-L5), you can understand what's this for and how to config it.

baihualinxin:
https://github.com/tensorflow/models/issues/5791

If you look at other configs，There is no problem

baihualinxin:
ssd_mobilenet_v1_coco.config  There's nothing wrong with the model
quantized There's a problem

PR #5830, opened 2018-11-28
Move references to `tf_record_iterator`.

MarkDaoust:
For tf2 this will only be available in `compat.v1`.


MarkDaoust:
from cl/222414946

PR #5850, opened 2018-12-02
fix error when gradient is None

blairhan:
When I trained ResNet in FP16, it gave an error since some gradients' values are None and cannot be divided by the loss_scale.

### DIFF
--- a/official/resnet/resnet_run_loop.py
+++ b/official/resnet/resnet_run_loop.py
@@ -400,7 +400,7 @@ def resnet_model_fn(features, labels, mode, model_class,
 
       # Once the gradient computation is complete we can scale the gradients
       # back to the correct scale before passing them to the optimizer.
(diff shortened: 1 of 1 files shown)


blairhan:
I checked it and I am the only author. @googlebot

blairhan:
My git config user.email had a type-o and i fixed it. @googlebot

robieta:
What gradients are None?

blairhan:
@robieta sorry for the late reply. 

I ran the model using this script https://github.com/mlperf/training/blob/master/image_classification/tensorflow/run.sh with fp16 setting on and it shows the error. There are some issues on this like https://github.com/mlperf/training/issues/230. As far as I know, tf.gradients() returns None instead of zero and that may cause the error for loss_scale division.

jaeyounkim:
Deprecated

PR #5892, opened 2018-12-11
Add CIFAR-10 support for keras application models.

lintian06:
Add CIFAR-10 for keras application models.

- Add CIFAR-10 dataset to dataset.py;
- Add custom keras model adapted with CIFAR-10 input_shape and num_classes;
- Enable the test on CIFAR-10 when FLAGS.use_synthetic_dat is False.


official/keras_application_models/dataset.py
```diff
@@ -46,3 +47,31 @@ def generate_synthetic_input_dataset(model, batch_size):
...
+  Args:
+    batch_size: int, the number of batch size.
+
+  Returns:
+    A tuple (train_dataset, test_dataset, image_shape, num_classes) where
+      train_dataset & test_dataset have shape ((batch, 32, 32, 3), (batch, 10)),
+      input_shape is the input image shape,
+      num_classes is number of output classes.
+  """
+  input_shape, num_classes = (32, 32, 3), 10
```
karmel: These should be constants, outside of this function

official/keras_application_models/dataset.py
```diff
@@ -46,3 +47,31 @@ def generate_synthetic_input_dataset(model, batch_size):
...
+  x_train, x_test = x_train / 255.0, x_test / 255.0
+  y_train, y_test = y_train.astype(np.int64), y_test.astype(np.int64)
+  y_train = tf.keras.utils.to_categorical(y_train, num_classes)
+  y_test = tf.keras.utils.to_categorical(y_test, num_classes)
+  train_dataset = tf.data.Dataset.from_tensor_slices((x_train, y_train))
+  test_dataset = tf.data.Dataset.from_tensor_slices((x_test, y_test))
+
+  train_dataset = train_dataset.shuffle(2000).batch(batch_size)
+  test_dataset = test_dataset.shuffle(2000).batch(batch_size)
+  return train_dataset, test_dataset, input_shape, num_classes
```
karmel: If the latter two are constants, you can just return the two datasets here, which is cleaner.

official/keras_application_models/dataset.py
lintian06: These two constants are tied to CIFAR-10 only, and has nothing to do with synthesis dataset aforementioned. I change all of these to instance variables. How does that look to you?

official/keras_application_models/dataset.py
lintian06: Change from returning variables to instance variables.

lintian06:
Hi Karmel,

Change the dataset from a function to a self-contained class. PTAL. 


Thanks,
Tian

lintian06:
Manually tested with --use_synthetic_data=0: 

python official/keras_application_models/benchmark_main.py   \
	--model=resnet50   \
	--benchmark_log_dir=/tmp/metric_dir   \
	--num_gpus=-1   \
	--train_epochs=1   \
	--batch_size=64   \
	--num_train_images=50000   \
	--num_eval_images=250   \
	--use_synthetic_data=0   \
	--alsologtostderr

I can add kokoro config later.

PR #5906, opened 2018-12-13
TF cifar10 cnn tutorial: use tensorflow-datasets to load the data.

pierrot0:


pierrot0:
Thanks for looking into it. And sorry for lag.
Here are the numbers I had collected (no performances changes):
  | Before | After
-- | -- | --
Loss at 2k iterations | 1.74 | 1.70
Loss at 4k iterations | 1.08 | 1.08
Loss at 10K iterations | 0.89 (lowest around: 0.68) | 0.80 (0.70)
time at step 2k | 6’40’’ | 6’36’’
time at step 4k | 13’07’’ | 13’14’’
time at step 10k | 32’53’’ | 32’47’’

PR #5921, opened 2018-12-17
qusetions about Dataset

duwangzhe:
I have downloaded GD-Xray Casting dataset, the tota number of the dataset is 2727 that is the same as the paper. But I have found that just 842 images are of bounding box value, it is different from what the paper has mentioned. Do I loss some critical information?
I am looking forward to your answer, Thanks!


robieta:
This looks like it should be an issue rather than a PR.

PR #5928, opened 2018-12-18
Cifar keras refactor

seemuch:
Extract common code from keras_cifar_main.py and keras_imagenet_main.py into keras_common.py.


official/resnet/keras/keras_cifar_main.py
```diff
@@ -0,0 +1,220 @@
+# Copyright 2017 The TensorFlow Authors. All Rights Reserved.
```
guptapriya: 2018?

seemuch: Done.

official/resnet/keras/keras_common.py
```diff
@@ -0,0 +1,172 @@
...
+from official.resnet import resnet_run_loop
+from official.resnet.keras import keras_resnet_model
+from official.resnet.keras import resnet_model_tpu
+from official.utils.flags import core as flags_core
+from official.utils.logs import logger
+from official.utils.misc import distribution_utils
+from tensorflow.python.keras.optimizer_v2 import gradient_descent as gradient_descent_v2
+
+
+FLAGS = flags.FLAGS
```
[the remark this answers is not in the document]
seemuch: flags_obj is flags.FLAGS. They are the same thing.

official/resnet/keras/keras_cifar_main.py
```diff
@@ -0,0 +1,220 @@
...
+                                      classes=cifar_main._NUM_CLASSES,
+                                      weights=None)
+  model.compile(loss=loss,
+                optimizer=opt,
+                metrics=[accuracy],
+                distribute=strategy)
+  time_callback, tensorboard_callback, lr_callback = keras_common.get_fit_callbacks(
+      learning_rate_schedule)
+
+  steps_per_epoch = cifar_main._NUM_IMAGES['train'] // flags_obj.batch_size
```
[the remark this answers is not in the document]
seemuch: Do we want to copy the numbers over, or do we want to make it public?

official/resnet/keras/keras_common.py
```diff
@@ -0,0 +1,172 @@
...
+    #  raise ValueError('Optimizer must have a "learning_rate" attribute.')
+    self.epochs += 1
+
+  def on_batch_begin(self, batch, logs=None):
+    lr = self.schedule(self.epochs, batch, self.batches_per_epoch, self.batch_size)
+    if not isinstance(lr, (float, np.float32, np.float64)):
+      raise ValueError('The output of the "schedule" function should be float.')
+    if lr != self.prev_lr:
+      self.model.optimizer.learning_rate = lr  # lr should be a float here
+      # tf.keras.backend.set_value(self.model.optimizer.learning_rate, lr)
```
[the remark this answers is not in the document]
seemuch: Yes.

official/resnet/keras/keras_cifar_main.py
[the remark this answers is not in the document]
seemuch: I changed the variables to be public.

official/resnet/keras/keras_imagenet_main.py
```diff
@@ -0,0 +1,215 @@
...
+    return initial_learning_rate * warmup_lr_multiplier * epoch / warmup_end_epoch
+  for mult, start_epoch in LR_SCHEDULE:
+    if epoch >= start_epoch:
+      learning_rate = initial_learning_rate * mult
+    else:
+      break
+  return learning_rate
+
+
+def parse_record_keras(raw_record, is_training, dtype):
```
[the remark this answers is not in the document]
seemuch: Are you sure about this code? 
I tried your code, and it will make the loss blow up. I switch back to the old code, and it seems fine.

official/resnet/keras/resnet56.py
```diff
@@ -0,0 +1,302 @@
+# Copyright 2018 The TensorFlow Authors. All Rights Reserved.
```
tfboyd: Please rename file to something like:  resnet_cifar_model.

That is what it is and it will not always be just be Resnet56 as resnet should be configurable to be a variety of layers very easily.

official/resnet/keras/resnet50.py
```diff
@@ -0,0 +1,244 @@
...
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+# ==============================================================================
+"""ResNet50 model for Keras.
+
+Adapted from tf.keras.applications.resnet50.ResNet50().
```
tfboyd: Rename file to resnet_model.py as this is standard ResNet V1.5 and there is not need to call it say resnet_imagenet_model.py as it is implied this is that version.

official/resnet/keras/resnet50.py
```diff
@@ -0,0 +1,244 @@
...
+
+Related papers/blogs:
```
tfboyd: Indicate that this is ResNet V1.5 and not ResNet V1. This is important and frustrating to people who want to know.

official/resnet/resnet_run_loop.py
```diff
@@ -628,6 +628,12 @@ def define_resnet_flags(resnet_size_choices=None):
           'the expense of image resize/cropping being done as part of model '
           'inference. Note, this flag only applies to ImageNet and cannot '
           'be used for CIFAR.'))
+  flags.DEFINE_boolean(
+      name='use_one_device_strategy', default=True,
```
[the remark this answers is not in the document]
tfboyd: @guptapriya 
I would consider a flag that is "turn_off_distribution_strategy" default False.  I find flags that default True are a pain and often fail when you try to set them to False, but maybe I use them wrong.  The best practice should be for users to always use  `tf.contrib.distribute.MirroredStrategy(num_gpus=num_gpus)` or rather `tf.contrib.distribute.MirroredStrategy()` so they can quickly go from 1 GPU to 8 GPUs without any flags in their code. 
 I assume if you pass num_gpus=1 it ends up as onedevicestrategy.

official/utils/misc/distribution_utils.py
```diff
@@ -30,15 +31,25 @@ def get_distribution_strategy(num_gpus, all_reduce_alg=None):
...
   Returns:
     tf.contrib.distribute.DistibutionStrategy object.
   """
-  if num_gpus == 0:
+  if num_gpus == 0 and use_one_device_strategy:
     return tf.contrib.distribute.OneDeviceStrategy("device:CPU:0")
-  elif num_gpus == 1:
+  elif num_gpus == 0:
+    return None
+  elif num_gpus == 1 and use_one_device_strategy:
```
tfboyd: I feel this should just call `tf.contrib.distribute.MirroredStrategy(num_gpus=num_gpus)` so it shows users you don't have to think about this it will just work.  Then the flag is `turn_off_distr_strat` and defaults False.

official/resnet/keras/keras_common.py
```diff
@@ -0,0 +1,191 @@
...
+  def __init__(self, batch_size):
+    """Callback for logging performance (# image/second).
+
+    Args:
+      batch_size: Total batch size.
+
+    """
+    self._batch_size = batch_size
+    super(TimeHistory, self).__init__()
+    self.log_batch_size = 100
```
tfboyd: Rename this to steps or log_steps as this determines how often to log and log_batch_size is confusing.  It should be an arg but that is a FR.

official/resnet/keras/keras_common.py
```diff
@@ -0,0 +1,191 @@
...
+  def on_batch_begin(self, batch, logs=None):
+    if self.record_batch:
+      self.batch_time_start = time.time()
+      self.record_batch = False
+
+  def on_batch_end(self, batch, logs=None):
+    if batch % self.log_batch_size == 0:
+      last_n_batches = time.time() - self.batch_time_start
+      examples_per_second = (self._batch_size * self.log_batch_size) / last_n_batches
+      self.batch_times_secs.append(last_n_batches)
```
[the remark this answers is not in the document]
tfboyd: I believe this would allow us to calculate averages, but is not used.  I would remove for now and in the future we can add more features like an average and ability for the stats function to access the values with averages.

official/resnet/keras/keras_common.py
```diff
@@ -0,0 +1,191 @@
...
+    self.record_batch = True
+
+  def on_batch_begin(self, batch, logs=None):
+    if self.record_batch:
+      self.batch_time_start = time.time()
+      self.record_batch = False
+
+  def on_batch_end(self, batch, logs=None):
+    if batch % self.log_batch_size == 0:
+      last_n_batches = time.time() - self.batch_time_start
```
tfboyd: Maybe a nit but could we rename last_n_batches to something like elapsed_time as that is what it is.  I would also not call it self.batch_time_start as we are not often measuring a single batch but rather a series of steps.  I suggest `start_time` and `elapsed_time` as that is very clear.

official/resnet/keras/keras_common.py
```diff
@@ -0,0 +1,191 @@
...
+  return time_callback, tensorboard_callback, lr_callback
+
+def analyze_fit_and_eval_result(history, eval_output):
+  stats = {}
+  stats['accuracy_top_1'] = eval_output[1]
+  stats['eval_loss'] = eval_output[0]
+  stats['training_loss'] = history.history['loss'][-1]
+  stats['training_accuracy_top_1'] = history.history['categorical_accuracy'][-1]
+
+  print('Test loss:{}'.format(stats['eval_loss']))
```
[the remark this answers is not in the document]
tfboyd: You also could just remove this if you wanted.  This may have been a print statement I left in the code.  Keras already prints these numbers.  If I printed something I would format the entire (or mostly) stats dict nicely and print that out with user friendly names.

official/utils/misc/distribution_utils.py
guptapriya: great point. when this code was written, we couldn't use MirroredStrategy for 1 GPU. But I think now we can. So we can simplify this.

My question is: by default, if the user specifies only 0 or 1 GPU, do we want to use mirroredstrategy? or just not use any strategy?

official/resnet/keras/keras_common.py
```diff
@@ -0,0 +1,191 @@
...
+  print('top_1_training_accuracy:{}'.format(stats['training_accuracy_top_1']))
+
+  return stats
+
+def define_keras_flags():
+  flags.DEFINE_boolean(name='enable_eager', default=False, help='Enable eager?')
+  flags.DEFINE_boolean(name='skip_eval', default=False, help='Skip evaluation?')
+  flags.DEFINE_integer(
+      name="train_steps", default=None,
+      help="The number of steps to run for training")
```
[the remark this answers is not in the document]
tfboyd: This looks like train_steps is a nice item for benchmarks so you can a test less than 1 epoch.  If so, then say that in the help and refer the user to train_epochs as the normal arg to use.  Maybe also let them know train_steps has to be less than a full epoch, which is weird but it is for testing so "weird" is passable.

official/resnet/keras/keras_imagenet_main.py
```diff
@@ -0,0 +1,185 @@
...
+
+  history = model.fit(train_input_dataset,
+                      epochs=train_epochs,
+                      steps_per_epoch=train_steps,
+                      callbacks=[
+                        time_callback,
+                        lr_callback,
+                        tensorboard_callback
+                      ],
+                      validation_steps=num_eval_steps,
```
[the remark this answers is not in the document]
tfboyd: I would agree if you say skip_eval you mean skip it period, so I could say run 90 epochs with zero eval.  We will certainly have a feature soon to control how often we eval during training.  Ideally I could run 90 epochs and just eval at the end but that is a different use case.

official/resnet/keras/keras_common.py
```diff
@@ -0,0 +1,191 @@
...
+    if not isinstance(lr, (float, np.float32, np.float64)):
+      raise ValueError('The output of the "schedule" function should be float.')
+    if lr != self.prev_lr:
+      self.model.optimizer.learning_rate = lr  # lr should be a float here
+      # tf.keras.backend.set_value(self.model.optimizer.learning_rate, lr)
+      self.prev_lr = lr
+      tf.logging.debug('Epoch %05d Batch %05d: LearningRateBatchScheduler change '
+                   'learning rate to %s.', self.epochs, batch, lr)
+
+def get_optimizer():
```
tfboyd: I am not sure it makes any sense to have the tf.optimizer as it doesn't work.  Or if this is just for short-term perf testing put a comment and print/log a big warning stressing it will not train to convergence along with a TODO to remove it once we do not need it for performance testing.  We already don't use the the V1 keras optimizer.

official/utils/misc/distribution_utils.py
robieta: I would suggest always using a strategy (except for the `turn_off_distr_strat` case) for two reasons:
1) It does sensible optimizations (like prefetch_to_device), and I think the trend of using dist strat to always include safe best practices is a good one.
2) It can be confusing to debug issues otherwise. If a strategy is used with 2 GPUs but not 1, debugging issues is really confusing unless you know the code is doing that. And it would also make scalability numbers weird because the reference (1 GPU) is fundamentally different from the rest.

official/resnet/keras/keras_common.py
```diff
@@ -0,0 +1,191 @@
...
+  if FLAGS.use_tf_momentum_optimizer:
+    learning_rate = BASE_LEARNING_RATE * FLAGS.batch_size / 256
+    optimizer = tf.train.MomentumOptimizer(learning_rate=learning_rate, momentum=0.9)
+  else:
+    # optimizer = tf.keras.optimizers.SGD(learning_rate=0.1, momentum=0.9)
+    optimizer = gradient_descent_v2.SGD(learning_rate=0.1, momentum=0.9)
```
tfboyd: Document that the learning_rate gets overridden by the callback and that is where the learning rate scaling for batch-size happens.  I find that weird as it doesn't matter what you set here.  I would almost just set it to 0 to make it clear what is set here is not actually used.  Seeing .1 makes me think we are not doing scaling but it is just a placeholder.

official/resnet/keras/keras_common.py
```diff
@@ -0,0 +1,172 @@
...
+                   'learning rate to %s.', self.epochs, batch, lr)
+
+def get_optimizer_loss_and_metrics():
+  # Use Keras ResNet50 applications model and native keras APIs
+  # initialize RMSprop optimizer
+  # TODO(anjalisridhar): Move to using MomentumOptimizer.
+  # opt = tf.train.GradientDescentOptimizer(learning_rate=0.0001)
+  # I am setting an initial LR of 0.001 since this will be reset
+  # at the beginning of the training loop.
+  opt = gradient_descent_v2.SGD(learning_rate=0.1, momentum=0.9)
```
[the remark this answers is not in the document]
seemuch: Talked to Zhenyu and Tomer, the exposed symbol is not ready for TF 1.x.

official/resnet/keras/keras_cifar_main.py
```diff
@@ -32,27 +32,18 @@
...
 def learning_rate_schedule(current_epoch, current_batch, batches_per_epoch, batch_size):
   """Handles linear scaling rule, gradual warmup, and LR decay.
 
-  The learning rate starts at 0, then it increases linearly per step.
-  After 5 epochs we reach the base learning rate (scaled to account
-    for batch size).
-  After 30, 60 and 80 epochs the learning rate is divided by 10.
-  After 90 epochs training stops and the LR is set to 0. This ensures
-    that we train for exactly 90 epochs for reproducibility.
+  The learning rate starts at base learning_rate, then after 91, 136 and
```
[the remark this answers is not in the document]
seemuch: I do not really know why we give it such a schedule. I was under the impression that this is parameter that @tfboyd tuned for performance reasons?
@tfboyd wanna comment here?

official/resnet/keras/keras_imagenet_main.py
```diff
@@ -0,0 +1,185 @@
...
+
+  steps_per_epoch = imagenet_main.NUM_IMAGES['train'] // flags_obj.batch_size
+  num_eval_steps = (imagenet_main.NUM_IMAGES['validation'] //
+                  flags_obj.batch_size)
+
+  train_steps = imagenet_main.NUM_IMAGES['train'] // flags_obj.batch_size
+  train_epochs = flags_obj.train_epochs
+
+  if flags_obj.train_steps:
+    train_steps = min(flags_obj.train_steps, train_steps)
```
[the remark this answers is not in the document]
seemuch: I added some more documentation for the train_steps flag.
This is useful for the benchmark/regression test we want to do later. E.g. we do not want to run all the examples.

official/resnet/keras/resnet56.py
```diff
@@ -0,0 +1,302 @@
...
+      axis=bn_axis, name=bn_name_base + '1',
+      momentum=BATCH_NORM_DECAY, epsilon=BATCH_NORM_EPSILON)(
+      shortcut, training=True)
+
+  x = tf.keras.layers.add([x, shortcut])
+  x = tf.keras.layers.Activation('relu')(x)
+  return x
+
+
+def ResNet56(input_shape=None, classes=1000):
```
tfboyd: This name violate lint.  Maybe change to `resnet56` or something that passes a pylint.

official/resnet/keras/resnet50.py
```diff
@@ -0,0 +1,244 @@
...
+                                       momentum=BATCH_NORM_DECAY,
+                                       epsilon=BATCH_NORM_EPSILON,
+                                       name=bn_name_base + '1')(shortcut)
+
+  x = layers.add([x, shortcut])
+  x = layers.Activation('relu')(x)
+  return x
+
+
+def ResNet50(num_classes):
```
tfboyd: This name violate lint.  Maybe change to `resnet50` or something that passes a pylint.

official/resnet/keras/keras_cifar_main.py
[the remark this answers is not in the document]
guptapriya: sorry i didn't mean you should explain the schedule. I just meant that instead of specifying the exact epochs here just say something generic like
"Scale learning rate at epoch boundaries provided in LR_SCHEDULE by the provided scaling factor"

official/utils/misc/distribution_utils.py
guptapriya: sgtm. i think we can just use `MirroredStrategy` in all cases then, unless `turn_off_distr_strat` is set to true.

official/resnet/keras/keras_common.py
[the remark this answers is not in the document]
seemuch: You get me, Toby. I added this flag for the purpose of benchmarks.
I updated the helper string.

official/resnet/keras/keras_common.py
```diff
@@ -0,0 +1,191 @@
...
+      self.prev_lr = lr
+      tf.logging.debug('Epoch %05d Batch %05d: LearningRateBatchScheduler change '
+                   'learning rate to %s.', self.epochs, batch, lr)
+
+def get_optimizer():
+  if FLAGS.use_tf_momentum_optimizer:
+    learning_rate = BASE_LEARNING_RATE * FLAGS.batch_size / 256
+    optimizer = tf.train.MomentumOptimizer(learning_rate=learning_rate, momentum=0.9)
+  else:
+    # optimizer = tf.keras.optimizers.SGD(learning_rate=0.1, momentum=0.9)
```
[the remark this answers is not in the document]
seemuch: No, it is not ready to be used for TF 1.x.

official/resnet/keras/keras_common.py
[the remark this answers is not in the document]
seemuch: Removed all.

official/resnet/keras/keras_common.py
seemuch: Removed.

official/resnet/keras/keras_common.py
```diff
@@ -0,0 +1,173 @@
...
+  return time_callback, tensorboard_callback, lr_callback
+
+
+def define_keras_flags():
+  flags.DEFINE_boolean(name='enable_eager', default=False, help='Enable eager?')
+  flags.DEFINE_boolean(name='skip_eval', default=False, help='Skip evaluation?')
+  flags.DEFINE_integer(
+      name="train_steps", default=None,
+      help="The number of steps to run for training. If it is larger than "
+      "# batches per epoch, then use # bathes per epoch. When this flag is "
```
guptapriya: bathes - > batches

official/resnet/keras/resnet_model.py
```diff
@@ -0,0 +1,245 @@
...
+                                       momentum=BATCH_NORM_DECAY,
+                                       epsilon=BATCH_NORM_EPSILON,
+                                       name=bn_name_base + '1')(shortcut)
+
+  x = layers.add([x, shortcut])
+  x = layers.Activation('relu')(x)
+  return x
+
+
+def resnet50(num_classes):
```
guptapriya: this also needs the same `training=None` here if we want to run this with estimator. but maybe Toby is adding that later?

official/utils/misc/distribution_utils.py
```diff
@@ -30,15 +31,30 @@ def get_distribution_strategy(num_gpus, all_reduce_alg=None):
...
   Returns:
     tf.contrib.distribute.DistibutionStrategy object.
+  Raises:
+    ValueError: if turn_off_distribution_strategy is True and num_gpus is
+    larger than 1
   """
   if num_gpus == 0:
-    return tf.contrib.distribute.OneDeviceStrategy("device:CPU:0")
+      if turn_off_distribution_strategy:
+          return None
```
guptapriya: indent seems off.

official/resnet/keras/keras_cifar_main.py
```diff
@@ -0,0 +1,185 @@
...
+        time_callback,
+        lr_callback,
+        tensorboard_callback
+        ],
+      validation_steps=num_eval_steps,
+      validation_data=validation_data,
+      verbose=1)
+
+  if not flags_obj.skip_eval:
+    eval_output = model.evaluate(eval_input_dataset,
```
guptapriya: what happened to the analyze results helper? we don't need that anymore, or to return these result from here? i thought the benchmark code depends on that?

tfboyd:
I added code to handle sparse vs non sparse categorical_accuracy with unit tests.  I think it is weird (maybe a reason) it would be named differently and that is outside the scope.
https://github.com/tensorflow/models/pull/5928/commits/61ec290757e495104f1150e00449187b765805c2

official/resnet/keras/keras_cifar_main.py
tfboyd: I fixed it handled sparse non sparse with unit tests.  I did not do imagenet but that can be done in another PR as I do not have a benchmark yet.
https://github.com/tensorflow/models/pull/5928/commits/61ec290757e495104f1150e00449187b765805c2

official/resnet/keras/resnet_cifar_model.py
```diff
@@ -0,0 +1,325 @@
...
+      default_shape = (input_shape[0], default_size, default_size)
+    else:
+      if input_shape[-1] not in {1, 3}:
+        warnings.warn(
+            'This model usually expects 1 or 3 input channels. '
+            'However, it was passed an input_shape with ' +
+            str(input_shape[-1]) + ' input channels.')
+      default_shape = (default_size, default_size, input_shape[-1])
+
+  return input_shape
```
tfboyd: Should this return default_shape?  It does not seem like it does anything.  Apologies if I missed you resolving this comment last time.  Github code review is confusing.  We likely can remove this code and just assume CIFAR-10 3 channels with the only option being channels first and last.

tfboyd:
All code here is from googles the CLA check is confused.  I looked at each commit myself up to today's date.  Shinning does have come commits under a different email but they are confirmed to be him.

official/resnet/keras/resnet_model.py
tfboyd: I have not updated the imagenet version yet.  TODO for me.  Leaving message here to do it before break is over.

tfboyd:
Making noise but we have LINT and all tests green.  I really hate the google import order.  I think the pain is it sees internal projects as internal where to the outside world they are third-party.  Either way celebrating victory.

official/resnet/keras/resnet_cifar_model.py
seemuch: I delete this method and do it as resnet50 does.

official/resnet/keras/resnet_model.py
seemuch: Added a todo comment for Toby.

tfboyd:
as of last [commit](https://github.com/tensorflow/models/pull/5928/commits/7522e4dc3bb7678ee800d13ce615f863d8a3fcbd) the main test (KerasCifar10BenchmarkTests.keras_resnet56_1_gpu) had these three results (5 is the "standard" and if the finish I will update:

-  0.9324919581413269
-  0.93038862943649292 [2018-12-26 21:13:51.485667 UTC]
- 0.92988783121109009

This aligns with previous results and with the layers + estimator version as well as the paper.  Performance is 2,600 images/sec on 1xV100 and should be closer to 3,400 which I was getting on the other branch but I suspect a possible hardware issue or other small problem.  I would not block the merge as the goal was a working model/refactor not performance which is still a work in progress.

Issue #2031, opened 2017-07-25
No module named 'object_detection'


DanMossa:
### System information
- **What is the top-level directory of the model you are using**: Nothing yet
- **Have I written custom code (as opposed to using a stock example script provided in TensorFlow)**: Nope
- **OS Platform and Distribution (e.g., Linux Ubuntu 16.04)**: Windows 10 64bit
- **TensorFlow installed from (source or binary)**: pip install tensorflow-gpu
- **TensorFlow version (use command below)**:

>  b'unknown' 1.2.0

- **Bazel version (if compiling from source)**:
- **CUDA/cuDNN version**:
8.0 and 5
- **GPU model and memory**:
Geforce GTX 1050ti 4gb ddr5
- **Exact command to reproduce**:
`python object_detection/builders/model_builder_test.py`

### Describe the problem
I followed the README.MD up until the point of having to execute the above command. And when I do I an error saying that there is no module named `object_detection` I believe this has something to do with the `export PYTHONPATH=$PYTHONPATH:`pwd`:`pwd`/slim` command. Because when I look at my system environment variables, I don't see anything titled PYTHONPATH.

I'm using Anaconda with a Virtual Environment

reedwm:
@jch1 can you look at this, or should it be forwarded to someone who works on Windows support?

MaxBareiss:
Try running `pwd` on its own; `pwd` doesn't come with Windows by default, so you may not have it. If you don't have pwd, I recommend defining the environment variable manually. This is what I do; you'll have to adapt the paths to your computer:
```
set PYTHONPATH=<PATH_TO_MODELS>;<PATH_TO_MODELS>\slim
```

Also, this should be posted to Stack Overflow.

DanMossa:
Well @MaxBareiss I was able to setup `pwd` correctly using Git Bash on windows
I originally typed `source activate py35cv3`

I don't have a `PYTHONPATH` but I do have a `PYTHON_HOME` So what I did is
```
echo $PYTHON_HOME
C:\Users\DanMo\Anaconda3
```
```
export PYTHON_HOME=$PYTHON_HOME:`pwd`:`pwd`/slim
```
```
echo $PYTHON_HOME
C:\Users\DanMo\Anaconda3:/c/Users/DanMo/Anaconda3/envs/py35cv3/lib/site-packages/tensorflow/models:/c/Users/DanMo/Anaconda3/envs/py35cv3/lib/site-packages/tensorflow/models/slim
```

```
[...]

Issue #2253, opened 2017-08-18
Where is the 'tensorflow/models/' directory? 


shuuchen:
Hi All,

I am trying to install object detection API following https://github.com/tensorflow/models/blob/master/object_detection/g3doc/installation.md

In the Protobuf Compilation part, I should run 
```python
# From tensorflow/models/
protoc object_detection/protos/*.proto --python_out=.
```
in the tensorflow/models/ directory. So where is the directory?

Does it already exist or should I create one by myself?

I am using ubuntu

sguada:
You need to create it yourself, it should be where you clone this repo.

cipri-tom:
and should we put this in the tensorflow installation path ? (in site-packages)

dineshkumaramara1:
can you please help me out should i keep it in the ../site_packages/tensorflow/ directory?

cristinabustos16:
It doesn't come just with 'pip install tensorflow'

Dont forget to do 'git clone https://github.com/tensorflow/models.git' and then you configure it the path as said in the instructions

shanmugavel28:
@cristinabustos16 thank you , your idea worked for me. a long search ended with your answer. thank you

PR #5929, opened 2018-12-18
Update installation.md

rogerthat94:
I did not realize that the tensorflow/models repository had to be cloned and setup. It looks like other people also did not realize this: https://github.com/tensorflow/models/issues/2031#issuecomment-317999079

The readme files in models/ and models/research do not mention anything about this either. Therefore, I updated this repository to mention that these steps must be done.

### DIFF
--- a/research/object_detection/g3doc/installation.md
+++ b/research/object_detection/g3doc/installation.md
@@ -48,6 +48,17 @@ pip install --user jupyter
 pip install --user matplotlib
 ```
 
+## Research Models Installation
+
+Clone this repository and set up the research models, if you have not done so already.
+
+```bash
+git clone https://github.com/tensorflow/models
+cd models/research
+python setup.py build
+python setup.py install
+```
+
 <!-- common_typos_disable -->
 **Note**: sometimes "sudo apt-get install protobuf-compiler" will install
 Protobuf 3+ versions for you and some users have issues when using 3.5.


rogerthat94:
Are you sure that object detection doesn't require running python setup? I got the error "No module named 'object_detection'" when I tried running "python object_detection/builders/model_builder_test.py." This was with the latest public version of the code that I ran today. 

There are 24 upvotes on [this answer](https://github.com/tensorflow/models/issues/2031#issuecomment-317801984) so I feel like I'm not the only person who ran into this issue and was able to fix with these commands. 

I don't think it's obvious that you have to cone the repo. I thought that the models directory was created as part of the tensorflow install from pip. There are [11 upvotes](https://github.com/tensorflow/models/issues/2253) on this post, so it's clear I am not the only one who was confused about this. 

Tensorflow claims to be a "machine learning framework for everyone." Therefore, I don't think you should assume these things will be so obvious.

pkulzc:
We have carefully tested that python setup is not required. If you look at the code in setup.py, they are just about packaging object detection (for training on cloud ML engine). It has nothing to do with running locally. If you really need this for local run, something else must be wrong (and this behavior accidentally solved your problem in a weird way). 

I'm OK with adding instructions for git clone if it is that hard for people to understand. You can include it in #5930 . For minor fixes like this, there's no need to separate.

rogerthat94:
Sorry about the confusion with that. 

I updated #5930 with a note that saying that the models directory is not part of any pip TensorFlow installation that this repository must be cloned.

rogerthat94:
I actually think you are incorrect about setup.py being "just about packaging object detection (for training on cloud ML engine)." It looks like this script installs the object_detection module on the user's computer. 

Could you possibly try cloning this repository on a fresh installation of Ubuntu? If you try to run `from object_detection.utils import label_map_util`, you will get the error "ImportError: No module named 'object_detection'."

But after running python `python setup.py build` and then `python setup.py install` this will fix the problem.

rogerthat94:
This is the [setup.py](https://github.com/tensorflow/models/blob/master/research/setup.py) my readme change addresses. Based on [this documentation](https://docs.python.org/3.5/distutils/introduction.html#a-simple-example) it appears that `python setup.py install` installs the module locally on the user's computer, which is required to use this code.

rogerthat94:
My issue was that I wasn't adding models/research to my PYTHONPATH. That's why it didn't work until I ran setup.py. Sorry about my confusion with this.

PR #5953, opened 2018-12-24
move flags validator before tf.app.run()

wenliwyan:
absl.flags validation happens at tf.app.run(). See https://github.com/abseil/abseil-py/blob/master/absl/flags/_validators.py#L329

The flags validator was not working in the previous code, and the error message would be "TypeError: Expected binary or unicode string, got None" when required flags were not correctly provided in arguments, which is not very helpful for debugging.
After the modification, the error message will be like "absl.flags._exceptions.IllegalFlagValueError: flag --pipeline_config_path=None: Flag --pipeline_config_path must be specified".

### DIFF
--- a/research/object_detection/model_main.py
+++ b/research/object_detection/model_main.py
@@ -53,6 +53,8 @@ flags.DEFINE_boolean(
     'run_once', False, 'If running in eval-only mode, whether to run just '
     'one round of eval vs running continuously (default).'
 )
+flags.mark_flag_as_required('model_dir')
+flags.mark_flag_as_required('pipeline_config_path')
 FLAGS = flags.FLAGS
 
 
@@ -104,6 +106,4 @@ def main(unused_argv):
 
 
 if __name__ == '__main__':
-  flags.mark_flag_as_required('model_dir')
-  flags.mark_flag_as_required('pipeline_config_path')
   tf.app.run()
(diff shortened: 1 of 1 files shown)


research/object_detection/model_main.py
```diff
@@ -106,4 +104,6 @@ def main(unused_argv):
 
 
 if __name__ == '__main__':
+  flags.mark_flag_as_required('model_dir')
+  flags.mark_flag_as_required('pipeline_config_path')
```
derekjchow: Maybe move these before `FLAGS = flags.FLAGS` (line 55). I think this is more consistent with other examples.

chineseD:
Hi  ! I wonder if you could help me with this. I got the same error.. did you fix this problem?

chineseD:
Hi  ! I wonder if you could help me with this. I got the same error.. did you fix this problem?

chineseD:
Hi  ! I wonder if you could help me with this. I got the same error.. did you fix this problem?

wenliwyan:
@SHADOWMOOON My pull request was a fix to provide clear error message as the authors intended to, when required flags were not correctly provided in arguments. You would get the same error when you did not provide `model_dir` or `pipeline_config_path`.

jaeyounkim:
@derekjchow Please review this pull request.

PR #5955, opened 2018-12-25
fix slim exported lenet graph can not be frozen

zhangxg:
the exported Lenet graph cannot be frozen due to below error: 

```
InvalidArgumentError (see above for traceback): Restoring from checkpoint failed. This is most likely due to a mismatch between the current graph and the graph from the checkpoint. Please ensure that you have not altered the graph expected based on the checkpoint. Original error:

Assign requires shapes of both tensors to match. lhs shape= [5,5,3,32] rhs shape= [5,5,1,32]
```
The used scripts as following: 
**export graph**
```
python export_inference_graph.py \
  --alsologtostderr \
  --model_name=lenet \
  --output_file=/tmp/lenet-model/lenet.pb \
  --dataset_name=mnist
```
**frozen graph** 
```
python -m tensorflow.python.tools.freeze_graph \
  --input_graph=/tmp/lenet-model/lenet.pb \
  --input_checkpoint=/tmp/lenet-model/model.ckpt-600 \
  --input_binary=true \
  --output_graph=/tmp/lenet-model/lenet_freeze.pb \
  --output_node_names=Predictions/Reshape_1
```


zhangxg:
signed cla

Issue #4996, opened 2018-08-03
[object_detection] Python 3 incompatibilities and other errors


apacha:
### System information
- **What is the top-level directory of the model you are using**: research/object_detection
- **Have I written custom code (as opposed to using a stock example script provided in TensorFlow)**: No
- **OS Platform and Distribution (e.g., Linux Ubuntu 16.04)**: Windows
- **TensorFlow installed from (source or binary)**: binary 
- **TensorFlow version (use command below)**: 1.9.0
- **Bazel version (if compiling from source)**: -
- **CUDA/cuDNN version**: 9
- **GPU model and memory**: GeForce GTX 1080 Ti
- **Exact command to reproduce**: `model_main.py --pipeline_config_path="samples/configs/ssd_mobilenet_v1_0.75_depth_quantized_300x300_pets_sync.config" --model_dir="data/pet-train" --logtostderr`

> Note that for running the script on Windows, you need the adapted [pycocotools](https://github.com/philferriere/cocoapi).

### Describe the problem
**Summary:** I've tried to run the tutorial for training pets detection locally, but couldn't get it to run, due to numerous bugs.

The latest version of Tensorflow Object Detection API is incompatible with Python 3 and even following the [official tutorial](https://medium.com/tensorflow/training-and-serving-a-realtime-mobile-object-detector-in-30-minutes-with-cloud-tpus-b78971cf1193) does not work appropriately, when trying to train locally.

I've encountered the following issues, when trying to clone the repo and run from master-branch (commit [f0e10716](https://github.com/tensorflow/models/commit/f0e10716160cd048618ccdd4b6e18336223a172f)):
- Python 2 constructs, that are incompatible with Python 3. I've fixed [two issues in this branch of my fork](https://github.com/apacha/TensorflowObjectDetection/tree/Python3-Compatibility).
- After fixing the Python 3 incompatibilities, training started, but both the `--logtostderror` (how it used to be) and the [`--alsologtostderror`](https://github.com/tensorflow/models/blob/master/research/object_detection/g3doc/running_locally.md#running-the-training-job) flags were simply ignored by the [`model_main.py`](https://github.com/tensorflow/models/blob/master/research/object_detection/model_main.py) script and no output was produced on the std-output (information to the tensorboard was written, though).
[...]

Issue #4780, opened 2018-07-16
[object detection]  TypeError: can't pickle dict_values objects


jj19940807:
creating index...
index created!
2018-07-16 22:12:38.626883: W T:\src\github\tensorflow\tensorflow\core\framework
\op_kernel.cc:1306] Invalid argument: TypeError: can't pickle dict_values objects
Traceback (most recent call last):

  File "d:\Program Files\Anaconda3\lib\site-packages\tensorflow\python\ops\scrip
t_ops.py", line 158, in __call__
    ret = func(*args)

  File "D:\Program Files\Anaconda3\Lib\site-packages\tensorflow\models\research\
object_detection\metrics\coco_evaluation.py", line 339, in first_value_func
    self._metrics = self.evaluate()

  File "D:\Program Files\Anaconda3\Lib\site-packages\tensorflow\models\research\
object_detection\metrics\coco_evaluation.py", line 193, in evaluate
    self._detection_boxes_list)

  File "D:\Program Files\Anaconda3\Lib\site-packages\tensorflow\models\research\
object_detection\metrics\coco_tools.py", line 118, in LoadAnnotations
    results.dataset['categories'] = copy.deepcopy(self.dataset['categories'])

  File "d:\Program Files\Anaconda3\lib\copy.py", line 174, in deepcopy
    rv = reductor(4)

TypeError: can't pickle dict_values objects

Traceback (most recent call last):
  File "d:\Program Files\Anaconda3\lib\site-packages\tensorflow\python\client\se
ssion.py", line 1322, in _do_call
    return fn(*args)
  File "d:\Program Files\Anaconda3\lib\site-packages\tensorflow\python\client\se
ssion.py", line 1307, in _run_fn
    options, feed_dict, fetch_list, target_list, run_metadata)
  File "d:\Program Files\Anaconda3\lib\site-packages\tensorflow\python\client\se
ssion.py", line 1409, in _call_tf_sessionrun
    run_metadata)
tensorflow.python.framework.errors_impl.InvalidArgumentError: TypeError: can't p
ickle dict_values objects
Traceback (most recent call last):

  File "d:\Program Files\Anaconda3\lib\site-packages\tensorflow\python\ops\scrip
t_ops.py", line 158, in __call__
    ret = func(*args)

  File "D:\Program Files\Anaconda3\Lib\site-packages\tensorflow\models\research\
object_detection\metrics\coco_evaluation.py", line 339, in first_value_func
    self._metrics = self.evaluate()

[...]

PR #5978, opened 2019-01-02
py3 compatibility

DefineFC:
Fix a Python3 incompatibility error in model_lib.py.

Thanks to @gfkeith https://github.com/tensorflow/models/issues/4780#issuecomment-405441448 and @MaeThird https://github.com/tensorflow/models/issues/4996#issuecomment-411970644.


derekjchow:
Thanks for the patch!

PR #5987, opened 2019-01-03
Use the new distribution strategy API with keras (creating the model inside strategy.scope() instead of passing it to model.compile)

seemuch:


official/resnet/keras/keras_cifar_main.py
```diff
@@ -132,16 +132,16 @@ def run(flags_obj):
       num_epochs=flags_obj.train_epochs,
       parse_record_fn=parse_record_keras)
 
-  optimizer = keras_common.get_optimizer()
   strategy = distribution_utils.get_distribution_strategy(
```
guptapriya: what happens when `strategy` is `None`? Which is possible right?

seemuch: Good catch. Fixed.

PR #5992, opened 2019-01-04
remove code that was migrated to tf/privacy

npapernot:
A lot of the code that used to be in models/research/differential_privacy was improved and migrated to a new repository at tensorflow/privacy. This PR removes the old code to avoid any confusions.


ilyamironov:
LGTM.

PR #6001, opened 2019-01-07
Add repeat to datasets for Keras application set.

lintian06:
This change is to add repeat() to datasets to avoid exhausting data in training and validation.


lintian06:
Thank you for the review! :)

PR #6009, opened 2019-01-08
Create a unit test for Keras application models in TF 2.0

lintian06:
Create a unit test for Keras application models in TF 2.0. (Named models_v2_test.py)

Minor: move the shared model definition in models.py.

Tested in tf-nightly-2.0-preview. 
$ python official/keras_application_models/models_v2_test.py

### DIFF
--- a/official/keras_application_models/benchmark_main.py
+++ b/official/keras_application_models/benchmark_main.py
@@ -26,25 +26,12 @@ import tensorflow as tf
 
 from official.keras_application_models import dataset
 from official.keras_application_models import model_callbacks
+from official.keras_application_models import models
 from official.utils.flags import core as flags_core
 from official.utils.logs import logger
 from official.utils.misc import distribution_utils
 
-# Define a dictionary that maps model names to their model classes inside Keras
-MODELS = {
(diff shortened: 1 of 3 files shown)


tfboyd:
The unit tests run under TF 1.0 but I do not think they should fail. Looking at the log it looks like each of the tests fails for some reason.  I do not doubt they were successful locally but they are failing on the kokoro runs.  

Two other thoughts:  

1. The parametrized TestCase is fun and compact but I think it might be nicer to list out the tests the verbose way one at a time.
2. I did not dig in deeply but you may be able to further reduce the test time but not running models that exercise the same code, e.g. possibly VGG16 and VGG19 and various densenet sizes.  In some cases it is just another layer and that isn't really testing much.

I will try to get a clear email out soon.  We have a structure to bring on end-to-end tests and then would like to refactor your perf tests.  I also sat down with martin to identify which models should be tested end-to-end and which we can simply just have a basic perf or does it "blow up" test like you have here.

tfboyd:
Closing as we moved on from upgrading these to TF 2.0.

PR #6013, opened 2019-01-08
Handle data format and graph compile properly for better GPU performance

haoyuz:
1. When `data_format` is `channels_first`, transpose the input tensors to `NCHW` for better performance on GPU;
2. Compile only the training graph when `skip_eval` flag is `True`. This allows better memory efficiency so that a larger (per-GPU) batch size can be used for training.

### DIFF
--- a/official/resnet/keras/keras_cifar_main.py
+++ b/official/resnet/keras/keras_cifar_main.py
@@ -108,6 +108,8 @@ def run(flags_obj):
   per_device_batch_size = distribution_utils.per_device_batch_size(
       flags_obj.batch_size, flags_core.get_num_gpus(flags_obj))
 
+  tf.keras.backend.set_image_data_format(flags_obj.data_format)
+
   if flags_obj.use_synthetic_data:
     input_fn = keras_common.get_synth_input_fn(
         height=cifar_main.HEIGHT,
@@ -160,6 +162,7 @@ def run(flags_obj):
 
   validation_data = eval_input_dataset
   if flags_obj.skip_eval:
+    tf.keras.backend.set_learning_phase(1)
     num_eval_steps = None
     validation_data = None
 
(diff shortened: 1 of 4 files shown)


official/resnet/keras/keras_imagenet_main.py
```diff
@@ -28,6 +28,7 @@
 from official.utils.flags import core as flags_core
 from official.utils.logs import logger
 from official.utils.misc import distribution_utils
+from tensorflow.python.keras import backend
```
guptapriya: is this needed? we could just do `tf.keras.backend.set_...` ?

official/resnet/keras/keras_imagenet_main.py
```diff
@@ -95,6 +96,8 @@ def run(flags_obj):
     raise ValueError('dtype fp16 is not supported in Keras. Use the default '
                      'value(fp32).')
 
+  backend.set_image_data_format(flags_obj.data_format)
```
guptapriya: Can we make all these changes to the corresponding Cifar version too? at least to be consistent. I don't know how much performance benefit there will be in that case.

official/resnet/keras/keras_imagenet_main.py
haoyuz: Done.

official/resnet/keras/keras_imagenet_main.py
haoyuz: Done. Added same logic to Cifar model as well.

official/resnet/keras/keras_cifar_main.py
```diff
@@ -160,6 +162,7 @@ def run(flags_obj):
 
   validation_data = eval_input_dataset
   if flags_obj.skip_eval:
+    tf.keras.backend.set_learning_phase(1)
```
guptapriya: can you also add a comment here explaining why we do this.

haoyuz:
Thanks Priya for the comments! Regarding performance improvement, changing data format to channels_first improves throughput by 6.4%. Just finished convergence test, and the model converges to 76.5% at 90th epoch. With that I will merge this PR.

PR #6016, opened 2019-01-09
Removed INITIAL_DIR parameter from download_imagenet

ema987:
INITIAL_DIR variable is unuseful if the synsets file is not specified using a parameter because it is in the same directory as the running script, anyway probably the synsets parameter is always specified with its path and the INITIAL_DIR variable causes the script to point to a wrong non existent path.


vincentvanhoucke:
I am not eager to make changes to an old script that many people may be relying on and isn't unit-tested unless it actively breaks people. @shlens may have better context than me.

shlens:
I agree with Vincent and I would avoid changing the script unless we have some guarantees that this is a backward-compatible change.

PR #6023, opened 2019-01-10
Add end-to-end tests for Estimator.

tfboyd:
Adds end-to-end accuracy tests for cifar-10 estimator model.

### DIFF
--- a/official/resnet/estimator_cifar_benchmark.py
+++ b/official/resnet/estimator_cifar_benchmark.py
@@ -1,0 +1,98 @@
+"""Executes Estimator benchmarks and accuracy tests."""
+from __future__ import print_function
+
+import os
+
+from absl import flags
+from absl.testing import flagsaver
+import tensorflow as tf  # pylint: disable=g-bad-import-order
+
+from official.resnet import cifar10_main as cifar_main
+
+DATA_DIR = '/data/cifar10_data/'
+
+
+class EstimatorCifar10BenchmarkTests(object):
+  """Benchmarks and accuracy tests for KerasCifar10."""
+
+  local_flags = None
+
+  def __init__(self, output_dir=None):
+    self.oss_report_object = None
+    self.output_dir = output_dir
+
+  def resnet56_1_gpu(self):
+    """Test layers model with Estimator and distribution strategies."""
+    self._setup()
+    flags.FLAGS.num_gpus = 1
+    flags.FLAGS.data_dir = DATA_DIR
+    flags.FLAGS.batch_size = 128
+    flags.FLAGS.train_epochs = 182
+    flags.FLAGS.model_dir = self._get_model_dir('resnet56_1_gpu')
+    flags.FLAGS.resnet_size = 56
+    flags.FLAGS.dtype = 'fp32'
+    stats = cifar_main.run_cifar(flags.FLAGS)
+    self._fill_report_object(stats)
+
+  def resnet56_fp16_1_gpu(self):
+    """Test layers FP16 model with Estimator and distribution strategies."""
+    self._setup()
+    flags.FLAGS.num_gpus = 1
+    flags.FLAGS.data_dir = DATA_DIR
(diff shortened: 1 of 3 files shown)


official/resnet/estimator_cifar_benchmark.py
```diff
@@ -0,0 +1,98 @@
...
+from absl.testing import flagsaver
+import tensorflow as tf  # pylint: disable=g-bad-import-order
+
+from official.resnet import cifar10_main as cifar_main
+
+DATA_DIR = '/data/cifar10_data/'
+
+
+class EstimatorCifar10BenchmarkTests(object):
+  """Benchmarks and accuracy tests for KerasCifar10."""
```
haoyuz: For EstimatorCifar10?

tfboyd: fixed

official/resnet/estimator_cifar_benchmark.py
```diff
@@ -0,0 +1,96 @@
...
+      self.oss_report_object.top_5 = stats['accuracy_top_5'].item()
+    else:
+      raise ValueError('oss_report_object has not been set.')
+
+  def _get_model_dir(self, folder_name):
+    return os.path.join(self.output_dir, folder_name)
+
+  def _setup(self):
+    tf.logging.set_verbosity(tf.logging.DEBUG)
+    if EstimatorCifar10BenchmarkTests.local_flags is None:
```
guptapriya: Can you explain why all this flag magic is needed?

official/resnet/estimator_cifar_benchmark.py
```diff
@@ -0,0 +1,96 @@
...
+    flags.FLAGS.model_dir = self._get_model_dir('resnet56_fp16_2_gpu')
+    flags.FLAGS.resnet_size = 56
+    flags.FLAGS.dtype = 'fp16'
+    stats = cifar_main.run_cifar(flags.FLAGS)
+    self._fill_report_object(stats)
+
+  def _fill_report_object(self, stats):
+    # Also "available global_step"
+    if self.oss_report_object:
+      self.oss_report_object.top_1 = stats['accuracy'].item()
```
guptapriya: the actual verification of this accuracy happens elsewhere?

official/resnet/resnet_run_loop.py
```diff
@@ -588,7 +591,7 @@ def input_fn_eval():
           shape, batch_size=flags_obj.batch_size, dtype=export_dtype)
     classifier.export_savedmodel(flags_obj.export_dir, input_receiver_fn,
                                  strip_default_attrs=True)
-
+  return eval_results
```
guptapriya: btw, what is the benchmark_logger thing from before used for ?

official/resnet/estimator_cifar_benchmark.py
tfboyd: Correct, right now we are just storing the value.  I will add validation next as there are many values to validate and I need to think about how a test with multiple pass/fails would work.  But I have an idea.

official/resnet/estimator_cifar_benchmark.py
tfboyd: We all use absl flags and the keras/models flags are not loaded at the time the unit test is called.  What is happening is I am reloading all the flags after I add all of the keras/models flags, I then cache that so it can be reloaded each time as if you were to call define_keras_flags a second time it would error out as it would say the flag already exists.  I think in the next refactor I need to create a base class for the tests to avoid duplicate code.

official/resnet/resnet_run_loop.py
tfboyd: I plan on removing it as it was used as part of the previous benchmark system.  I did not want to remove it until we were back up and running.  The old system some what counted on you using the logger and my approach is `return stats` or some other method where the test gets access to the results without having to check a log file.

PR #6024, opened 2019-01-10
Add timestamp history for each batch in training

seemuch:


official/resnet/keras/keras_common.py
```diff
@@ -47,15 +55,21 @@ def __init__(self, batch_size):
...
+  def on_train_end(self, logs=None):
+    self.train_finish_time = time.time()
+
   def on_batch_begin(self, batch, logs=None):
     if self.record_batch:
       self.start_time = time.time()
       self.record_batch = False
 
+    self.train_timestamp_history.append(BatchTimestamp(batch))
+
```
tfboyd: I would do the following if I understand this correctly
- Move this under `if self.record_batch` so that index=0 will be the start time
- Also pass self.start_time rather than having a new time created, they will be almost the same but passing it will mean they are exactly the same.

official/resnet/keras/keras_common.py
```diff
@@ -34,6 +34,14 @@
 TRAIN_TOP_1 = 'training_accuracy_top_1'
 
 
+class BatchTimestamp(object):
+  """A structure to store batch time stamp."""
+
+  def __init__(self, batch_index):
+    self.batch_index = batch_index
+    self.timestamp = time.time()
```
tfboyd: Pass in the time so that the time is exactly equal to the start or end time and not off even by a nano second.

official/resnet/keras/keras_common.py
```diff
@@ -47,15 +55,21 @@ def __init__(self, batch_size):
...
   def on_batch_begin(self, batch, logs=None):
     if self.record_batch:
       self.start_time = time.time()
       self.record_batch = False
 
+    self.train_timestamp_history.append(BatchTimestamp(batch))
+
   def on_batch_end(self, batch, logs=None):
     if batch % self.log_steps == 0:
       elapsed_time = time.time() - self.start_time
```
tfboyd: add self.train_timestamp_history.append(BatchTimestamp(batch)) here and pass in the time.time() used for elapsed_time so they are exactly the same.  This will ensure each entry in history will match logging.  Meaning if I calculate elapsed time with timestamp_history it will **exactly** match the log for those 100 steps.

official/resnet/keras/keras_common.py
```diff
@@ -47,15 +55,21 @@ def __init__(self, batch_size):
     self._batch_size = batch_size
     super(TimeHistory, self).__init__()
     self.log_steps = 100
+    self.train_timestamp_history = []
 
   def on_train_begin(self, logs=None):
     self.record_batch = True
 
+  def on_train_end(self, logs=None):
+    self.train_finish_time = time.time()
```
tfboyd: Comment (not action needed):  This is likely not totally accurate as I suspect this hook is not called until the graph is compiled and grappler and even XLA run the first time.  I could be wrong.  Not a big deal and it is nice to have.

official/resnet/keras/keras_common.py
```diff
@@ -34,6 +34,14 @@
...
+class BatchTimestamp(object):
+  """A structure to store batch time stamp."""
+
+  def __init__(self, batch_index):
+    self.batch_index = batch_index
+    self.timestamp = time.time()
+
+
 class TimeHistory(tf.keras.callbacks.Callback):
   """Callback for Keras models."""
```
tfboyd: Can you update the init to accept log_steps that defaults to 100?  We use this in tf_cnn_benchmarks and it is very useful.  We often log every 10.

official/resnet/keras/keras_common.py
```diff
@@ -34,6 +34,14 @@
...
 
+class BatchTimestamp(object):
+  """A structure to store batch time stamp."""
+
+  def __init__(self, batch_index):
+    self.batch_index = batch_index
+    self.timestamp = time.time()
+
+
 class TimeHistory(tf.keras.callbacks.Callback):
```
tfboyd: It would be nice to have a unit test given we have a decent amount of logic.

official/resnet/keras/keras_common.py
seemuch: I agree on the second point: we should use the exact same timestamp. 
As to the first point, I don't think this line should be moved into the `if self.record_batch`, because if so, then we cannot warm-up (skip) for whatever number of steps we want. I think we should record the start time for *all* the batches, and end time for only a selective subset of batches (based on log_steps).

PR #6025, opened 2019-01-10
Add imagenet end-to-end accuracy test.

tfboyd:
- using batch-size 64 because nightly binaries might be before your fix?
- Will look at issues related to the unit tests that you commented on in the other thread before I finish this one.

### DIFF
--- a/official/resnet/keras/keras_imagenet_benchmark.py
+++ b/official/resnet/keras/keras_imagenet_benchmark.py
@@ -38,20 +38,6 @@ class KerasImagenetBenchmarkTests(object):
     self.oss_report_object = None
     self.output_dir = output_dir
 
-  def keras_resnet50_short_1_gpu(self):
-    """Test Keras model with Keras fit/dist_strat quick test."""
-    self._setup()
-    flags.FLAGS.num_gpus = 1
-    flags.FLAGS.data_dir = DATA_DIR
-    flags.FLAGS.batch_size = 32
-    flags.FLAGS.train_epochs = 1
-    flags.FLAGS.model_dir = self._get_model_dir('keras_resnet50_short_1_gpu')
-    flags.FLAGS.resnet_size = 50
-    flags.FLAGS.dtype = 'fp32'
-    flags.FLAGS.train_steps = 100
-    stats = keras_imagenet_main.run(flags.FLAGS)
-    self._fill_report_object(stats)
-
   def keras_resnet50_8_gpu(self):
     """Test Keras model with Keras fit/dist_strat and 8 GPUs."""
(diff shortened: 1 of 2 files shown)


official/resnet/keras/keras_imagenet_benchmark.py
```diff
@@ -0,0 +1,90 @@
...
+  def __init__(self, output_dir=None):
+    self.oss_report_object = None
+    self.output_dir = output_dir
+
+  def keras_resnet50_short_1_gpu(self):
+    """Test Keras model with Keras fit/dist_strat quick test."""
+    self._setup()
+    flags.FLAGS.num_gpus = 1
+    flags.FLAGS.data_dir = DATA_DIR
+    flags.FLAGS.batch_size = 32
```
haoyuz: For the short benchmark, do we want to run at 128 batch size (must also set `skip_eval` to `True` for now to avoid OOM)?

guptapriya: Hm, we cannot set skip_eval to True for accuracy tests (which is what these are, as of now). Perhaps we will need separate performance and convergence tests so we can use different batch size there?

official/resnet/keras/keras_imagenet_benchmark.py
```diff
@@ -0,0 +1,90 @@
...
+    flags.FLAGS.train_epochs = 1
```
guptapriya: maybe 2 epochs since sometimes we have bugs at epoch boundaries

official/resnet/keras/keras_imagenet_benchmark.py
```diff
@@ -0,0 +1,90 @@
...
+    flags.FLAGS.model_dir = self._get_model_dir('keras_resnet50_short_1_gpu')
+    flags.FLAGS.resnet_size = 50
```
guptapriya: this flag is not used, right?

official/resnet/keras/keras_imagenet_benchmark.py
tfboyd: This test is for "testing" and will not be run for real.  I also want it to run on a GTX-1080 so I need a smaller batch-size.  i will look at renaming it in the next round.

official/resnet/keras/keras_imagenet_benchmark.py
tfboyd: fixed

official/resnet/keras/keras_imagenet_benchmark.py
tfboyd: removed test

PR #6031, opened 2019-01-11
Remove duplicate elif case in _check_and_convert_legacy_input_config_key

gonzojive:


research/object_detection/utils/config_util.py
```diff
@@ -334,9 +334,6 @@ def _check_and_convert_legacy_input_config_key(key):
   elif field_name == "eval_input_path":
     key_name = "eval_input_configs"
     field_name = "input_path"
-  elif field_name == "train_input_path":
-    key_name = "train_input_config"
-    field_name = "input_path"
   elif field_name == "eval_input_path":
```
derekjchow: Are these lines duplicated too? Would you mind removing these as well?

derekjchow:
LGTM. Thanks for the patch!

PR #6032, opened 2019-01-12
Single timestamp list and return average_exp_per_second.

tfboyd:
I think this does what we need and I added returning the average although in most tests we would want to skip the first timestamp as a warmup. 

Take your time looking this over and thinking about if it works for you.  I think this works and it is as numerically correct as we can get.


seemuch:
LGTM. Approve.

PR #6044, opened 2019-01-14
fix 'dict_keys' object does not support indexing

bryanlimy:
fix `'dict_keys' object does not support indexing` at `feature_map_generators.py`


bryanlimy:
I signed it!

PR #6066, opened 2019-01-17
Change to gfile.Exists and deeper link to CIFAR-10

tfboyd:


official/resnet/cifar10_main.py
```diff
@@ -23,6 +23,7 @@
 from absl import app as absl_app
 from absl import flags
 import tensorflow as tf  # pylint: disable=g-bad-import-order
+from tensorflow.python.platform import gfile
```
robieta: Drive by comment: this is exported with tf. (so you can do `tf.gfile.Exists(...)`)

tfboyd: ahhh sweet I will swap that out.

tfboyd: HIGH FIVE drive-by

tfboyd: fixed

PR #6072, opened 2019-01-21
Added options to set validation frequency during training for Keras models

haoyuz:
* Adding `--validation_freq` and `--validarion_freq_list` flags, which take an integer and a list of comma separated integers, to specify at which epochs to invoke validation during training.
* Reduce the validation frequency of eager benchmarks tests (for both imagenet and cifar10). This is a temporary fix due to b/122886254.

### DIFF
--- a/official/resnet/keras/keras_cifar_main.py
+++ b/official/resnet/keras/keras_cifar_main.py
@@ -165,6 +165,17 @@ def run(flags_obj):
                     flags_obj.batch_size)
 
   validation_data = eval_input_dataset
+  if flags_obj.validation_freq and flags_obj.validation_freq_list:
+    raise ValueError('At most one of --validation_freq and '
+                     '--validation_freq_list must be specified.')
+  validation_freq = flags_obj.validation_freq or 1
+  if flags_obj.validation_freq_list:
+    try:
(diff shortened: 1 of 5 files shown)


official/resnet/keras/keras_cifar_main.py
```diff
@@ -165,6 +165,17 @@ def run(flags_obj):
                     flags_obj.batch_size)
 
   validation_data = eval_input_dataset
+  if flags_obj.validation_freq and flags_obj.validation_freq_list:
```
guptapriya: Could this validation go in the common file?

official/resnet/keras/keras_cifar_main.py
```diff
@@ -180,6 +191,7 @@ def run(flags_obj):
                       ],
                       validation_steps=num_eval_steps,
                       validation_data=validation_data,
+                      validation_freq=validation_freq,
```
guptapriya: what happens if skip_eval is true but validation_freq is set here?

tfboyd:
Another PR has already done this while haoyuz was on vacation.  It does not allow a list but if we need that feature we can add it back. For now it copies what the estimator version does to make testing easier.

PR #6090, opened 2019-01-23
Fix typos in morph_net/framework/README.md

stephenjfox:
In reading the documentation, I found a reference to the [original name](https://arxiv.org/abs/1711.06798v1) of the paper and a typo regarding rank of a Tensor.

These are minor changes, but will prevent any confusion of future readers


jaeyounkim:
morph_net is deprecated.

PR #6097, opened 2019-01-25
Remove Python compiled files from deep_contextual_bandits

PiotrDabrowskey:
Remove byte-compiled files from the repo.

By the way: `__pycache__/` is already added to `.gitignore`, so those `*.pyc` files must have been force added by mistake.


PiotrDabrowskey:
I signed it!

PR #6101, opened 2019-01-27
Add resnet56 short tests.

tfboyd:
- created base benchmark module
- renamed accuracy test class to contain the word Accuracy
which will result in a need to update all the jobs
and a loss of history but is worth it.
- short tests are mostly copied from shining with oss refactor

### DIFF
--- a/official/resnet/keras/keras_benchmark.py
+++ b/official/resnet/keras/keras_benchmark.py
@@ -31,12 +31,12 @@ class KerasBenchmark(object):
   """Base benchmark class with methods to simplify testing."""
   local_flags = None
 
-  def __init__(self, output_dir=None, default_flags=None):
+  def __init__(self, output_dir=None, default_flags=None, flag_methods=None):
     self.oss_report_object = None
     self.output_dir = output_dir
     self.default_flags = {}
-    if default_flags:
-      self.default_flags = default_flags
+    self.default_flags = default_flags or {}
+    self.flag_methods = flag_methods or {}
 
   def _get_model_dir(self, folder_name):
     return os.path.join(self.output_dir, folder_name)
@@ -50,19 +50,17 @@ class KerasBenchmark(object):
     """
     tf.logging.set_verbosity(tf.logging.DEBUG)
     if KerasBenchmark.local_flags is None:
-      # TODO(tobyboyd): Move to init and initialize to empty object if None.
-      if flag_methods:
-        for flag_method in flag_methods:
-          flag_method()
+      for flag_method in flag_methods:
+        flag_method()
       # Loads flags to get defaults to then override. List cannot be empty.
       flags.FLAGS(['foo'])
+      # Overrides flag values with defaults for the class of tests.
+      for k, v in self.default_flags.items():
+        setattr(FLAGS, k, v)
       saved_flag_values = flagsaver.save_flag_values()
       KerasBenchmark.local_flags = saved_flag_values
     else:
       flagsaver.restore_flag_values(KerasBenchmark.local_flags)
-    # Overrides flag values with defaults for the class of tests.
-    for k, v in self.default_flags.items():
-      setattr(FLAGS, k, v)
 
   def fill_report_object(self, stats, top_1_max=None, top_1_min=None,
                          log_steps=None, total_batch_size=None, warmup=1):
--- a/official/resnet/keras/keras_benchmark.py
+++ b/official/resnet/keras/keras_benchmark.py
@@ -34,23 +34,17 @@ class KerasBenchmark(object):
   def __init__(self, output_dir=None, default_flags=None, flag_methods=None):
     self.oss_report_object = None
     self.output_dir = output_dir
-    self.default_flags = {}
     self.default_flags = default_flags or {}
     self.flag_methods = flag_methods or {}
 
   def _get_model_dir(self, folder_name):
     return os.path.join(self.output_dir, folder_name)
 
-  def _setup(self, flag_methods=None):
-    """Sets up and resets flags before each test.
-
-    Args:
-      flag_methods: list of methods to execute that add global flags.
-
-    """
+  def _setup(self):
+    """Sets up and resets flags before each test."""
     tf.logging.set_verbosity(tf.logging.DEBUG)
     if KerasBenchmark.local_flags is None:
-      for flag_method in flag_methods:
+      for flag_method in self.flag_methods:
         flag_method()
       # Loads flags to get defaults to then override. List cannot be empty.
       flags.FLAGS(['foo'])
--- a/official/resnet/keras/keras_benchmark.py
+++ b/official/resnet/keras/keras_benchmark.py
@@ -1,0 +1,109 @@
+# Copyright 2018 The TensorFlow Authors. All Rights Reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+# ==============================================================================
+"""Executes Keras benchmarks and accuracy tests."""
+
+from __future__ import absolute_import
+from __future__ import division
+from __future__ import print_function
+
+import os
+
+from absl import flags
+from absl.testing import flagsaver
+import tensorflow as tf  # pylint: disable=g-bad-import-order
+
+
+DATA_DIR = '/data/cifar10_data/cifar-10-batches-bin'
+MIN_TOP_1_ACCURACY = 0.925
+MAX_TOP_1_ACCURACY = 0.938
+
+FLAGS = flags.FLAGS
+
+
+class KerasBenchmark(object):
+  """Base benchmark class with methods to simplify testing."""
+  local_flags = None
+
+  def __init__(self, output_dir=None, default_flags=None):
+    self.oss_report_object = None
+    self.output_dir = output_dir
+    self.default_flags = {}
+    if default_flags:
+      self.default_flags = default_flags
+
+  def _get_model_dir(self, folder_name):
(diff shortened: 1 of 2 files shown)


official/resnet/keras/keras_cifar_benchmark.py
```diff
@@ -18,118 +18,201 @@
...
+class Resnet56KerasBenchmarkBase(keras_benchmark.KerasBenchmark):
+  """Short performance tests for ResNet56 via Keras and CIFAR-10."""
 
-  def _get_model_dir(self, folder_name):
-    return os.path.join(self.output_dir, folder_name)
+  def _run_benchmark(self):
+    stats = keras_cifar_main.run(FLAGS)
+    self.fill_report_object(stats)
+
+  def benchmark_1_gpu_no_dist_strat(self):
```
guptapriya: don't you need resnet56 in the names here?

official/resnet/keras/keras_cifar_benchmark.py
```diff
@@ -18,118 +18,201 @@
...
+class Resnet56KerasBenchmarkReal(Resnet56KerasBenchmarkBase):
+  """Real data benchmarks for ResNet56 and Keras."""
+
+  def __init__(self, output_dir=None):
+    def_flags = {}
+    def_flags['skip_eval'] = True
+    def_flags['data_dir'] = DATA_DIR
+    def_flags['train_steps'] = 110
+    def_flags['log_steps'] = 10
+    FLAGS.skip_eval = True
```
guptapriya: hm, why both put it in def_flags and set it here?

official/resnet/keras/keras_benchmark.py
```diff
@@ -0,0 +1,109 @@
...
+    self.output_dir = output_dir
+    self.default_flags = {}
+    if default_flags:
+      self.default_flags = default_flags
+
+  def _get_model_dir(self, folder_name):
+    return os.path.join(self.output_dir, folder_name)
+
+  def _setup(self, flag_methods=None):
+    """Sets ups up and resets flags before each test.
```
guptapriya: typo?

lindong28:
@tfboyd I have two other PRs (https://github.com/tensorflow/benchmarks/pull/292 and https://github.com/tensorflow/models/pull/6103) that changed the interface of PerfZero and added support for benchmarks defined in leading_indicators_test.py. In particularly, these PRs would replaced `oss_report_object` with a dict and this would affect the implementation of `keras_benchmark.py` added in this patch.

If we merge this PR first, some the new code added in this PR will be replaced soon and tested again. I am wondering if we could discuss and merge the other PR first so that the new code will use the new API from the beginning.

official/resnet/keras/keras_cifar_benchmark.py
tfboyd: ResNet56 is in the classname which is part of the full test name.  Having it on the method just makes it longer and makes it longer in the UI.  I have been working to reduce the length of full test names.  At a later date I need to reconcile this style with Shining's in Google3 but I copied 95% of his structure.

tfboyd:
@lindong28   I am going to go with this one first so I can get tests up and running.  Your change impacts a bunch of tests and they will all need changed.  Which is a good move forward and we have to plan it out.  We cannot block tests for refactoring.  One option you might consider in the future is a FLAG for the execution path, where old tests use the old path and new test new path. Not needed for quick changes but your PR is a refactor touching much more than just execution and will take time to sort out.  With this change impacting old tests a flag is still likely good to avoid disruption and is fun as that is how it is often done across Google.  Then phase the flag out ASAP.

lindong28:
@tfboyd Sounds good. I was not sure about when we need the tests defined in this PR. There is tradeoff between timeliness for this PR and the amount of code change. Yes let's go with this PR first since we want to run the test soon.

Regarding the use of flags for switching between old tests and new tests, the flag reduces the coupling between repos versions at the cost of increased amount of code change (the extra logic/code for flags will eventually not be useful). Given that we currently the only user of PerfZero and oss_report_object defined in in this PR, we should be able to commit change for both repositories at around the same time without having to run different versions of these repo together. So I thought the benefits of flags may not worth the extra code/effort.

Does this make sense? I am happy to update the PRs to use the flag if the current solution (i.e. commit change for PerfZero and the benchmark classes which are run by PerZero within the same hour) has issues.

official/resnet/keras/keras_benchmark.py
```diff
@@ -0,0 +1,109 @@
...
+        for flag_method in flag_methods:
+          flag_method()
+      # Loads flags to get defaults to then override. List cannot be empty.
+      flags.FLAGS(['foo'])
+      saved_flag_values = flagsaver.save_flag_values()
+      KerasBenchmark.local_flags = saved_flag_values
+    else:
+      flagsaver.restore_flag_values(KerasBenchmark.local_flags)
+    # Overrides flag values with defaults for the class of tests.
+    for k, v in self.default_flags.items():
```
lindong28: Would it be more efficient to also setattr for `default_flags` before we initialize `KerasBenchmark.local_flags`?

official/resnet/keras/keras_cifar_benchmark.py
```diff
@@ -18,118 +18,201 @@
...
+    flag_methods = [keras_common.define_keras_flags,
+                    cifar_main.define_cifar_flags]
+    super(Resnet56KerasBenchmarkBase, self)._setup(flag_methods=flag_methods)
+
+
+class Resnet56KerasBenchmarkSynth(Resnet56KerasBenchmarkBase):
+  """Synthetic benchmarks for ResNet56 and Keras."""
+
+  def __init__(self, output_dir=None):
+    def_flags = {}
```
lindong28: The current patch specifies the values for flags in three different ways:

1) Per-class flags. Defined as `def_flags` in e.g. `Resnet56KerasBenchmarkSynth.__init__()`. This is newly added in this patch.

2) Per-class flags. Defined as e.g. `keras_common.define_keras_flags` in `Resnet56KerasBenchmarkBase._setup()`.

3) Per-method flags. Defined directly in e.g. `Resnet56KerasBenchmarkBase.benchmark_graph_1_gpu()`.

It would be nice if we can define flags in only two ways for per-class and per-method flags respectively. This may be just a matter of code taste. I can also try it in the future.

official/resnet/keras/keras_cifar_benchmark.py
```diff
@@ -18,118 +18,201 @@
...
+    FLAGS.batch_size = 128
+    FLAGS.train_epochs = 182
+    FLAGS.model_dir = self._get_model_dir('keras_resnet56_eager_1_gpu')
+    FLAGS.dtype = 'fp32'
+    FLAGS.enable_eager = True
     stats = keras_cifar_main.run(flags.FLAGS)
-    self._fill_report_object(stats)
+    self.fill_report_object(stats, FLAGS.batch_size)
 
   def keras_resnet56_eager_2_gpu(self):
```
lindong28: Since we are using a new class name (i.e. benchmark_graph_1_gpu), we can use new method name here. Would it be more consistent to use the same name patter as benchmark methods defined in `Resnet56KerasBenchmarkBase`, which starts with prefix `benchmark_` and does not duplicate words that already exist in the class name (i.e.  resnet and keras).

official/resnet/keras/keras_benchmark.py
```diff
@@ -0,0 +1,109 @@
...
+
+class KerasBenchmark(object):
+  """Base benchmark class with methods to simplify testing."""
+  local_flags = None
+
+  def __init__(self, output_dir=None, default_flags=None):
+    self.oss_report_object = None
+    self.output_dir = output_dir
+    self.default_flags = {}
+    if default_flags:
```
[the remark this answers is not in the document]
tfboyd: http://pylint-messages.wikidot.com/messages:w0102

official/resnet/keras/keras_benchmark.py
tfboyd: fixed

official/resnet/keras/keras_benchmark.py
```diff
@@ -0,0 +1,109 @@
...
+  def _setup(self, flag_methods=None):
+    """Sets ups up and resets flags before each test.
+
+    Args:
+      flag_methods: list of methods to execute that add global flags.
+
+    """
+    tf.logging.set_verbosity(tf.logging.DEBUG)
+    if KerasBenchmark.local_flags is None:
+      if flag_methods:
```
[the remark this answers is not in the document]
tfboyd: Added TODO.  This cannot be defaulted to {} because of
http://pylint-messages.wikidot.com/messages:w0102

official/resnet/keras/keras_benchmark.py
tfboyd: I added a TODO: I see what you mean and will add this during the next pass.

official/resnet/keras/keras_benchmark.py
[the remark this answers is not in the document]
lindong28: Looking at it again, it is probably not a problem here because default_flags is immutable in `KerasBenchmark`. And if it is mutable, in the future we can do `self.default_flags = dict(default_flags)`, which I personally find more concise to read. Please feel free to choose the style you prefer.

official/resnet/keras/keras_benchmark.py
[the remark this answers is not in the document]
lindong28: It is probably not a problem here because flag_methods is immutable in `_setup()`. Just want to clarify this. Feel free to choose the style you prefer :)

official/resnet/keras/keras_benchmark.py
lindong28: Sounds good.

official/resnet/keras/keras_cifar_benchmark.py
lindong28: Sounds good. Yeah in the long term we probably want to fix the legit stuff for both Google3 and OSS repositories.

official/resnet/keras/keras_benchmark.py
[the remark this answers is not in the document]
tfboyd: nice, I took a slightly different approach that I think you will find not too bad to read and a one liner.

official/resnet/keras/keras_benchmark.py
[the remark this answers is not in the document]
tfboyd: Understood it still triggers a lint error at least in my linter so I would have to add a pytlink disable which seems more a pain than just playing along.

PR #6105, opened 2019-01-28
Robust-ify recommendation launch scripts.

tayo:
1. Explicitly allow for script execution from any directory.  
2. Make envars visible in python script.

These issues were encountered while launching on TPU.

### DIFF
--- a/official/recommendation/run.sh
+++ b/official/recommendation/run.sh
@@ -6,6 +6,9 @@ if [ `id -u` != 0 ]; then
   sudo echo "Success"
 fi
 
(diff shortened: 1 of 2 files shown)


official/recommendation/run.sh
```diff
@@ -6,6 +6,9 @@ if [ `id -u` != 0 ]; then
   sudo echo "Success"
 fi
 
+SCRIPT_DIR=`dirname "$BASH_SOURCE"`
```
robieta: Huh, that's neat. TIL

PR #6119, opened 2019-01-31
Ran tf_upgrade_v2 script on resnet and utils folder.

goldiegadde:


goldiegadde:
> Here is my idea.
> 
> 1. we can get this working in the main branch (master)
> 2. then we can branch and make the keras model be 2.0 only with all the correct paths.
> 
> One issue we have is we cannot do a 2.0 saved model that works with 1.0 as the call does not exist in 1.0. We already have team members that need a 2.0 saved model.

I think thats good. Running the tests now, thanks for the setup help!

PR #3641, opened 2018-03-18
fix issue 'could not satisfy explicit device'

zengsn:
Some local operations require CPU instead of GPU.
And I found a lot of users raised the same issue in tensorflow/tensorflow, like 
https://github.com/tensorflow/tensorflow/issues/3475  
https://github.com/tensorflow/tensorflow/issues/14085  

And in tensorflow/models, #1823 

I think adding these two lines should help to avoid confusing.

PR #6122, opened 2019-01-31
fix issue 'could not satisfy explicit device'

zengsn:
2nd time pull for the same issue.
Is there any other fix to this issue? 
Please refer to https://github.com/tensorflow/models/pull/3641


zengsn:
@sguada https://github.com/tensorflow/models/pull/3641

PR #6125, opened 2019-01-31
fix issue - could not satisfy explicit device

zengsn:
2nd time pull, see - https://github.com/tensorflow/models/pull/3641


zengsn:
@robieta once again?

PR #6138, opened 2019-02-01
Add Combined NMS

pooyadavoodi:
CombinedNMS op was recently added to TensorFlow.
https://github.com/tensorflow/tensorflow/pull/23567

This PR adds CombinedNMS to the object detection API in TensorFlow.

The main difference between this new op and the old ones is that it does all of NMS in one op.
It's much more efficient, and it's also easier to swap out with a more efficient version of it from TF custom backends such as TensorRT and XLA.

### DIFF
--- a/research/object_detection/builders/post_processing_builder.py
+++ b/research/object_detection/builders/post_processing_builder.py
@@ -88,9 +88,8 @@ def _build_non_max_suppressor(nms_config):
                      'max_total_detections.')
   if nms_config.soft_nms_sigma < 0.0:
     raise ValueError('soft_nms_sigma should be non-negative.')
-  non_max_suppressor_fn = functools.partial(
 
-  if(nms_config.combined_nms):
+  if nms_config.combined_nms:
     non_max_suppressor_fn = functools.partial(
       post_processing.combined_non_max_suppression,
       score_thresh=nms_config.score_threshold,
@@ -98,7 +97,6 @@ def _build_non_max_suppressor(nms_config):
       max_size_per_class=nms_config.max_detections_per_class,
       max_total_size=nms_config.max_total_detections,
       use_static_shapes=nms_config.use_static_shapes)
-
   else:
     non_max_suppressor_fn = functools.partial(
       post_processing.batch_multiclass_non_max_suppression,
--- a/research/object_detection/builders/post_processing_builder.py
+++ b/research/object_detection/builders/post_processing_builder.py
@@ -89,6 +89,18 @@ def _build_non_max_suppressor(nms_config):
   if nms_config.soft_nms_sigma < 0.0:
     raise ValueError('soft_nms_sigma should be non-negative.')
   non_max_suppressor_fn = functools.partial(
+
+  if(nms_config.combined_nms):
+    non_max_suppressor_fn = functools.partial(
+      post_processing.combined_non_max_suppression,
+      score_thresh=nms_config.score_threshold,
+      iou_thresh=nms_config.iou_threshold,
+      max_size_per_class=nms_config.max_detections_per_class,
+      max_total_size=nms_config.max_total_detections,
+      use_static_shapes=nms_config.use_static_shapes)
+
+  else:
+    non_max_suppressor_fn = functools.partial(
       post_processing.batch_multiclass_non_max_suppression,
       score_thresh=nms_config.score_threshold,
       iou_thresh=nms_config.iou_threshold,
--- a/research/object_detection/builders/post_processing_builder.py
+++ b/research/object_detection/builders/post_processing_builder.py
@@ -88,28 +88,21 @@ def _build_non_max_suppressor(nms_config):
                      'max_total_detections.')
   if nms_config.soft_nms_sigma < 0.0:
     raise ValueError('soft_nms_sigma should be non-negative.')
-
-  if nms_config.combined_nms:
+  if nms_config.use_combined_nms:
     if nms_config.use_class_agnostic_nms:
-      raise ValueError('combined_nms does not support use_class_agnostic_nms')
-    non_max_suppressor_fn = functools.partial(
-        post_processing.combined_non_max_suppression,
-        score_thresh=nms_config.score_threshold,
-        iou_thresh=nms_config.iou_threshold,
-        max_size_per_class=nms_config.max_detections_per_class,
-        max_total_size=nms_config.max_total_detections,
-        use_static_shapes=nms_config.use_static_shapes)
-  else:
-    non_max_suppressor_fn = functools.partial(
-        post_processing.batch_multiclass_non_max_suppression,
-        score_thresh=nms_config.score_threshold,
-        iou_thresh=nms_config.iou_threshold,
-        max_size_per_class=nms_config.max_detections_per_class,
-        max_total_size=nms_config.max_total_detections,
-        use_static_shapes=nms_config.use_static_shapes,
-        use_class_agnostic_nms=nms_config.use_class_agnostic_nms,
-        max_classes_per_detection=nms_config.max_classes_per_detection,
-        soft_nms_sigma=nms_config.soft_nms_sigma)
+      raise ValueError('combined_nms does not support class_agnostic_nms')
+
+  non_max_suppressor_fn = functools.partial(
+      post_processing.batch_multiclass_non_max_suppression,
+      score_thresh=nms_config.score_threshold,
+      iou_thresh=nms_config.iou_threshold,
+      max_size_per_class=nms_config.max_detections_per_class,
+      max_total_size=nms_config.max_total_detections,
+      use_static_shapes=nms_config.use_static_shapes,
+      use_class_agnostic_nms=nms_config.use_class_agnostic_nms,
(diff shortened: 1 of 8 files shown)


pooyadavoodi:
+@azaks2 @aaroey

aaroey:
+@pengchongjin

dcyoung:
@pooyadavoodi do you happen to have any ballpark performance metrics comparing inference before and after the change?

pooyadavoodi:
@dcyoung In our inference benchmarks, I can see from 0.8x to 1.2x speedup when we use CombinedNMS. The smaller the batch size, the bigger the speedup. Note that <1x means regression.

Note that CombinedNMS only works on CPU. Someone might implement a GPU version of it in the future to make it much faster. 

TF-TRT will be able to optimize CombinedNMS (even the CPU version) to TensorRT NMS and with that I can see about 2x speedup across the board.

research/object_detection/core/post_processing.py
```diff
@@ -496,3 +496,73 @@ def _single_image_nms_fn(args):
...
+
+  Raises:
+    ValueError: if `q` in boxes.shape is not 1 or not equal to number of
+      classes as inferred from scores.shape.
+  """
+
+  with tf.name_scope(scope, 'CombinedNonMaxSuppression'):
+    nmsed_boxes, nmsed_scores, nmsed_classes, num_detections = tf.image.combined_non_max_suppression(
+            boxes, scores, max_size_per_class, max_total_size, iou_thresh,
+            score_thresh, use_static_shapes)
```
pkulzc: combined_non_max_suppression doesn't really have a input argument like this, right?

research/object_detection/core/post_processing.py
```diff
@@ -496,3 +496,73 @@ def _single_image_nms_fn(args):
...
+                                 change_coordinate_frame=False,
+                                 num_valid_boxes=None,
+                                 masks=None,
+                                 additional_fields=None,
+                                 use_static_shapes=False,
+                                 scope=None):
+  """Multi-class version of non maximum suppression that operates on a batch.
+
+  This op is similar to `multiclass_non_max_suppression` but operates on a batch
+  of boxes and scores. See documentation for `multiclass_non_max_suppression`
```
pkulzc: Are these all simply copy/pasted? You should add some comments about that this method really do here.

research/object_detection/core/post_processing.py
pkulzc: combined_non_max_suppression doesn't really have a input argument like this, right?

research/object_detection/core/post_processing.py
```diff
@@ -496,3 +496,73 @@ def _single_image_nms_fn(args):
...
+      classes as inferred from scores.shape.
+  """
+
+  with tf.name_scope(scope, 'CombinedNonMaxSuppression'):
+    nmsed_boxes, nmsed_scores, nmsed_classes, num_detections = tf.image.combined_non_max_suppression(
+            boxes, scores, max_size_per_class, max_total_size, iou_thresh,
+            score_thresh, use_static_shapes)
+
+    batch_nmsed_masks = None
+    batch_nmsed_additional_fields = None
```
pkulzc: Why are these not handled?

research/object_detection/core/post_processing.py
supriyar: Which particular input are you referring to? The CombinedNonMaxSuppression has 6 inputs as defined in https://github.com/tensorflow/tensorflow/blob/master/tensorflow/core/ops/image_ops.cc#L913

research/object_detection/core/post_processing.py
supriyar: Currently the CombinedNonMaxSuppression does not have support to handle masks or additional fields. We can handle this once the kernel is updated to use these fields.

research/object_detection/core/post_processing.py
pkulzc: There are 7 input arguments here while CombinedNonMaxSuppression only takes 6.

supriyar: Sorry I missed the last attribute - use_static_shapes is the bool argument pad_per_class.

research/object_detection/core/post_processing.py
pooyadavoodi: @pkulzc could you PTAL. I just added a better explanation of what the function does.

research/object_detection/core/post_processing.py
pkulzc: Could we use keyword argument for combined_non_max_suppression here?

research/object_detection/core/post_processing.py
pkulzc: Could you add a comment mentioning this?

research/object_detection/core/post_processing.py
```diff
@@ -496,3 +496,85 @@ def _single_image_nms_fn(args):
...
+    #TODO supriyar: Set static shapes here?
```
pkulzc: Remove this?

research/object_detection/protos/post_processing.proto
```diff
@@ -20,6 +20,9 @@ message BatchNonMaxSuppression {
 
   // Whether to use the implementation of NMS that guarantees static shapes.
   optional bool use_static_shapes = 6 [default = false];
+
+  // Whether to use the Combined NMS op
+  optional bool combined_nms = 7 [default = false];
```
pkulzc: use_combined_nms

research/object_detection/protos/faster_rcnn.proto
```diff
@@ -168,6 +168,9 @@ message FasterRcnn {
   // If True, uses implementation of ops with static shape guarantees when
   // running evaluation (specifically not is_training if False).
   optional bool use_static_shapes_for_eval = 37 [default = false];
+
+  // Whether to use Combined NMS for first stage.
+  optional bool first_stage_combined_nms = 38 [default=false];
```
pkulzc: use_combined_nms_in_first_stage

research/object_detection/core/post_processing_test.py
```diff
@@ -1077,6 +1077,53 @@ def test_batch_multiclass_nms_with_additional_fields_and_num_valid_boxes(
...
+    self.assertIsNone(nmsed_additional_fields)
+
+    with self.test_session() as sess:
+      (nmsed_boxes, nmsed_scores, nmsed_classes,
+       num_detections) = sess.run([nmsed_boxes, nmsed_scores, nmsed_classes,
+                                   num_detections])
+      self.assertAllClose(nmsed_boxes, exp_nms_corners)
+      self.assertAllClose(nmsed_scores, exp_nms_scores)
+      self.assertAllClose(nmsed_classes, exp_nms_classes)
+      self.assertAllClose(num_detections, [3, 3])
```
pkulzc: assertListEqual

pkulzc:
Pooya, we recently updated the post_processing files and there may have some conflicts with your PR. Could you please resolve them?

pooyadavoodi:
Thanks @pkulzc 
I will check. There is also some problem in the PR that causes accuracy regression on Faster-RCNN and Mask-RCNN (not in SSD). I am looking into that too.

pooyadavoodi:
@pkulzc I have made some restructuring in the code. Could you please review.

FasterRCNN doesn't work with combined_nms in this PR mainly because of unsupported arguments. I have put some checks in the code to avoid calling combined_nms with those arguments, but some of those arguments can't be checked during graph construction because they are TF tensors (for those I print warning messages instead). We can add TF conditionals to the graph to check those if you think that's necessary (assuming they don't hurt the performance).

research/object_detection/builders/post_processing_builder.py
```diff
@@ -88,6 +88,10 @@ def _build_non_max_suppressor(nms_config):
                      'max_total_detections.')
   if nms_config.soft_nms_sigma < 0.0:
     raise ValueError('soft_nms_sigma should be non-negative.')
+  if nms_config.use_combined_nms:
+    if nms_config.use_class_agnostic_nms:
```
pkulzc: Merge two if clauses?

research/object_detection/protos/post_processing.proto
```diff
@@ -39,6 +39,9 @@ message BatchNonMaxSuppression {
 
   // Soft NMS sigma parameter; Bodla et al, https://arxiv.org/abs/1704.04503)
   optional float soft_nms_sigma = 9 [default = 0.0];
+
+  // Whether to use tf.image.combined_non_max_suppression
```
pkulzc: Please add period to the end of your comments (same for your other changes in this PR as well).

research/object_detection/builders/post_processing_builder.py
pooyadavoodi: Done.

research/object_detection/protos/post_processing.proto
pooyadavoodi: Done.

pooyadavoodi:
@pkulzc I think I've answered all the comments. 
Please let me know what you think.

aaroey:
@pkulzc @pengchongjin could you help to merge this PR? Thanks.

PR #6147, opened 2019-02-04
Create a flag to specify distribution strategy

yuefengz:


dubey:
@tfboyd this is different.  The one I have adds a way to switch from separate train and evaluate calls to using train_and_evaluate.  That PR will conflict with the changes in this one, so I will wait for this to be merged before adding mine.

yuefengz:
@tfboyd This is part of what we wanted. We need a way to specify what strategy to use in the model.

guptapriya:
@yuefengz does the new distribute_strategy flag become required after this change? or is it optional?

yuefengz:
@guptapriya It should be optional. If you don't specify it, the model chooses one using the old logic.

official/utils/misc/distribution_utils.py
```diff
@@ -35,28 +36,35 @@ def get_distribution_strategy(num_gpus,
...
+        raise ValueError("`OneDeviceStrategy` can not be used for more than "
+                        "one device.")
       return tf.contrib.distribute.OneDeviceStrategy("device:GPU:0")
-  elif turn_off_distribution_strategy:
-    raise ValueError("When {} GPUs are specified, "
-                     "turn_off_distribution_strategy flag cannot be set to"
-                     "True.".format(num_gpus))
-  else:  # num_gpus > 1 and not turn_off_distribution_strategy
+
+  if distribute_strategy == "mirrored":
```
guptapriya: hm, from what i can tell, this only ever returns a mirrored strategy when distribute_strategy = "mirrored". or did i miss something?

official/utils/misc/distribution_utils.py
```diff
@@ -23,7 +23,8 @@
 
 def get_distribution_strategy(num_gpus,
                               all_reduce_alg=None,
-                              turn_off_distribution_strategy=False):
+                              turn_off_distribution_strategy=False,
+                              distribute_strategy=None):
```
guptapriya: since you're changing all the callers of this anyway, i wonder if we can clean up this method a bit. 
1) turn_off_distribution_strategy and distribute_strategy seem redundant, what if we make `distribute_strategy=None` behave as the off case?
2) all_reduce_alg should probably be the last arg as it's the least used and should not need to be specified most of the time. 
3) distribute_strategy should likely be the first one if we expect people to use it all the time. 
4) num_gpus prob needs to become optional if we are going to add TPUStrategy here in the future

official/mnist/mnist.py
```diff
@@ -165,7 +165,8 @@ def run_mnist(flags_obj):
       allow_soft_placement=True)
 
   distribution_strategy = distribution_utils.get_distribution_strategy(
-      flags_core.get_num_gpus(flags_obj), flags_obj.all_reduce_alg)
+      flags_core.get_num_gpus(flags_obj), flags_obj.all_reduce_alg
```
guptapriya: seems like a comma is missing here?

official/utils/misc/distribution_utils.py
yuefengz: Is this what we want? `distribute_strategy` override the strategy but if it is None, auto chooses one. See line 58.

official/utils/misc/distribution_utils.py
yuefengz: Do we still want three behavior?
1. distribution strategy is turned off
2. auto choose one
3. override distribution strategy if `distribution_strategy` is specified.

If we don't need 2), I agree and your suggestion makes it cleaner.

official/mnist/mnist.py
yuefengz: Done.

official/utils/misc/distribution_utils.py
guptapriya: hmm. i think maybe we can forego #2 since it makes everything more complicated. 
it would require changing all the callers though so start specifying strategy, but probably not too bad. 
@tfboyd what do you think?

official/utils/misc/distribution_utils.py
guptapriya: sorry i still don't see how `distribute_strategy=None` will ever return MirroredStrategy in this code. Line 58 is only about one device strategy it seems?

yuefengz: Sorry I misread you comment. Added commit to return MS by default.

official/utils/misc/distribution_utils.py
tfboyd: How about this and keep in mind I often make things confusing?
1) distribute_strategy=off  (this would turn it off)
2) distribute_strategy default setting in flags is "default" which means we pick, which means mirrored or onedevice
3) distribute_strategy set to something other than default we try to do what the user wants and error if we have no idea what they are asking for.

tfboyd: @guptapriya @yuefengz   what do you think ^^^^?

guptapriya: +1 to Toby's suggestion.

yuefengz:
Discarding the PR due to these weird CLA errors.

PR #6157, opened 2019-02-05
Add mobilenet_v1 to faster rcnn builder

lambdaflow:


linthieda:
When loading from a fine-tune checkpoint, it is possible that the fine-tune checkpoint is not compatible with the current model. To save the effort of removing the lines in the pipeline.config. I've added a check if legacy/trainer.py. When this is the case, the model will be initialized as if no fine-tune checkpoint is specified. And it will print an tf.logging.warning indicating no variables are loaded.

pkulzc:
Although FasterRCNNMobilenetV1FeatureExtractor has been implemented, but we haven't tested it yet. Have you train/eval'ed a model with this feature extractor?

linthieda:
> Although FasterRCNNMobilenetV1FeatureExtractor has been implemented, but we haven't tested it yet. Have you train/eval'ed a model with this feature extractor?

Yes. I am training a instance segmentation model and have exported a checkpoint to frozen model to evaluate its performance. Though it the model was frozen prematurely,  It works fine.

PR #6158, opened 2019-02-05
Change the use_synth_data flag to data_source_type flag

seemuch:
...and added a cached_real_data option to that flag.

### DIFF
--- a/official/resnet/cifar10_main.py
+++ b/official/resnet/cifar10_main.py
@@ -52,7 +52,7 @@ DATASET_NAME = 'CIFAR-10'
 ###############################################################################
 def get_filenames(is_training, data_dir):
   """Returns a list of filenames."""
-  assert tf.gfile.Exists(data_dir), (
+  assert tf.io.gfile.exists(data_dir), (
       'Run cifar10_download_and_extract.py first to download and extract the '
       'CIFAR-10 data.')
 
@@ -68,7 +68,7 @@ def get_filenames(is_training, data_dir):
 def parse_record(raw_record, is_training, dtype):
   """Parse CIFAR-10 image and label from a raw record."""
   # Convert bytes to a vector of uint8 that is record_bytes long.
-  record_vector = tf.decode_raw(raw_record, tf.uint8)
+  record_vector = tf.io.decode_raw(raw_record, tf.uint8)
 
   # The first byte represents the label, which we convert from uint8 to int32
   # and then to one-hot.
@@ -81,7 +81,7 @@ def parse_record(raw_record, is_training, dtype):
 
   # Convert from [depth, height, width] to [height, width, depth], and cast as
   # float32.
-  image = tf.cast(tf.transpose(depth_major, [1, 2, 0]), tf.float32)
+  image = tf.cast(tf.transpose(a=depth_major, perm=[1, 2, 0]), tf.float32)
 
   image = preprocess_image(image, is_training)
   image = tf.cast(image, dtype)
@@ -97,7 +97,7 @@ def preprocess_image(image, is_training):
         image, HEIGHT + 8, WIDTH + 8)
 
     # Randomly crop a [HEIGHT, WIDTH] section of the image.
-    image = tf.random_crop(image, [HEIGHT, WIDTH, NUM_CHANNELS])
+    image = tf.image.random_crop(image, [HEIGHT, WIDTH, NUM_CHANNELS])
 
     # Randomly flip the image horizontally.
     image = tf.image.random_flip_left_right(image)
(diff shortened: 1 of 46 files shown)


tfboyd:
@guptapriya   Can you review this for Shining?  I think we agree that it has to be backward compatible that is all that mattered to me.

tfboyd:
@guptapriya   ping.  Not trying to be pushy.  If we want this for testing we will want to get it approved.  These changes touch a few things and get conflicts pretty easily.

guptapriya:
@tfboyd yes, i am just waiting for the PR to be updated to respect the old flag as well (be backward compatible). @seemuch can you let us know when that is done?

seemuch:
I added a data_source flag to resnet only. The use_synthetic_data flag has been remained.

official/resnet/resnet_run_loop.py
```diff
@@ -636,6 +638,15 @@ def define_resnet_flags(resnet_size_choices=None):
       name='turn_off_distribution_strategy', default=False,
       help=flags_core.help_wrap('Set to True to not use distribution '
                                 'strategies.'))
+  flags.DEFINE_enum(
+      name='data_source_type',
+      default='real_data',
+      enum_values=[
+        "real_data",
+        "synthetic_data",
+        "cached_real_data"  # also known as "fake data"
```
guptapriya: please explain more on what this does?

official/resnet/resnet_run_loop.py
```diff
@@ -644,3 +655,11 @@ def define_resnet_flags(resnet_size_choices=None):
     flags.DEFINE_string(**choice_kwargs)
   else:
     flags.DEFINE_enum(enum_values=resnet_size_choices, **choice_kwargs)
+
+
+def resolve_flag_conflict():
+  if (flags.FLAGS.use_synthetic_data and
```
guptapriya: I don't understand how this change is still backward compatible.. you're still requiring users to provide `data_source_type` even if they provided `use_synthetic_data` ? i think this should throw an error only if both are present and do not match. 
And we should have synthetic behavior if `use_synthetic_data` is set, even if `data_source_type` is not set.

official/resnet/cifar10_main.py
```diff
@@ -252,13 +252,17 @@ def run_cifar(flags_obj):
...
+  resnet_run_loop.resolve_flag_conflict()
+
   if flags_obj.image_bytes_as_serving_input:
     tf.compat.v1.logging.fatal(
         '--image_bytes_as_serving_input cannot be set to True for CIFAR. '
         'This flag is only applicable to ImageNet.')
     return
 
-  input_function = (flags_obj.use_synthetic_data and
+  use_synthetic_data = (flags_obj.data_source_type == "synthetic_data")
```
guptapriya: This is not backward compatible. See my comment near the resolve conflict method

official/resnet/resnet_run_loop.py
```diff
@@ -109,6 +109,7 @@ def process_record_dataset(dataset,
   # allows DistributionStrategies to adjust how many batches to fetch based
   # on how many devices are present.
   dataset = dataset.prefetch(buffer_size=tf.data.experimental.AUTOTUNE)
+  dataset = dataset.take(1).cache().repeat()
```
guptapriya: Should this behavior not be conditional on `cached_real_data` being set? I don't see that value of the flag being used anywhere. Am i missing something?

official/resnet/keras/keras_imagenet_main.py
```diff
@@ -102,7 +102,9 @@ def run(flags_obj):
   tf.keras.backend.set_image_data_format(data_format)
 
   # pylint: disable=protected-access
-  if flags_obj.use_synthetic_data:
+  use_synthetic_data = (falgs_obj.data_source_type == "synthetic_data")
```
guptapriya: type in flags_obj. Is this code tested?

official/resnet/resnet_run_loop.py
```diff
@@ -636,6 +638,15 @@ def define_resnet_flags(resnet_size_choices=None):
       name='turn_off_distribution_strategy', default=False,
       help=flags_core.help_wrap('Set to True to not use distribution '
                                 'strategies.'))
+  flags.DEFINE_enum(
```
guptapriya: Can we please make sure to test all the code paths before merging this change? Namely: all combinations of keras/estimator, cifar/imagenet, all values of old flag/new flag.
It would be much more costly to discover bugs in this in the future, especially given this stuff has hardly any unittest coverage unfortunately.

official/resnet/resnet_run_loop.py
seemuch: Done.

official/resnet/resnet_run_loop.py
seemuch: I think the reason this happens is because I set the data_source_type default to real_data. I now set it to None because whatever I set the default value to be, there will be such problem.

official/resnet/resnet_run_loop.py
seemuch: Done.

guptapriya:
please sync with Toby and Yuefeng on the latest requirements for this change

PR #6162, opened 2019-02-06
Revert "tf_upgrade_v2 on resnet and utils folders. (#6154)"

goldiegadde:
This reverts commit d6b2b83c0eb668085ecf4045178d210e7bc69892.

### DIFF
--- a/official/resnet/cifar10_download_and_extract.py
+++ b/official/resnet/cifar10_download_and_extract.py
@@ -60,4 +60,4 @@ def main(_):
 
 if __name__ == '__main__':
   FLAGS, unparsed = parser.parse_known_args()
-  tf.compat.v1.app.run(argv=[sys.argv[0]] + unparsed)
+  tf.app.run(argv=[sys.argv[0]] + unparsed)
--- a/official/resnet/cifar10_main.py
+++ b/official/resnet/cifar10_main.py
@@ -52,7 +52,7 @@ DATASET_NAME = 'CIFAR-10'
 ###############################################################################
 def get_filenames(is_training, data_dir):
   """Returns a list of filenames."""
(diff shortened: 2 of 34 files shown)


goldiegadde:
> Thanks, Goldie-- can we then create a branch that is synced to the prior commit (right before this rollback) for v2 for now?

@tfboyd , so that  he is informed.

goldiegadde:
> > Thanks, Goldie-- can we then create a branch that is synced to the prior commit (right before this rollback) for v2 for now?
> 
> @tfboyd , so that he is informed. Yes I think we can do that.

tfboyd:
Why it was decided to stop tagging this feels crazy.

tfboyd:
@karmel   I want to test the following
- I want to main branch to be compat.v1 or pure TF 1.0.  So we can test TF 1.0 with compat along with some TF 2.0 testing.
- a separate branch where we are doing V2 that will cannot exist at the same time easily.

tfboyd:
Keep in mind the README says this.

The master branch of the models are in development, and they target the nightly binaries built from the master branch of TensorFlow. We aim to keep them backwards compatible with the latest release when possible (currently TensorFlow 1.5), but we cannot always guarantee compatibility.

karmel:
@tfboyd -- I think we resolved offline, but sounds like this is a problem with an older version, and the current Models HEAD works with TF 1.x HEAD, in which case a rollback/branch is not necessary. Sorry for the thrash on this, @goldiegadde

f23505106:
class ExamplesPerSecondHook(tf.estimator.SessionRunHook):
AttributeError: module 'tensorflow.python.estimator.api.estimator' has no attribute 'SessionRunHook'

PR #6167, opened 2019-02-07
Revert "Revert "tf_upgrade_v2 on resnet and utils folders. (#6154)" (…

goldiegadde:
…#6162)"

This reverts commit 57e075203f8fba8d85e6b74f17f63d0a07da233a.


tfboyd:
We accepted this before.  there are a bunch of lint errors and we can address those in another commit.  This will bit rot really fast.

PR #6187, opened 2019-02-12
Make 2 gpu tests num_gpu=2 (was wrong before)

tfboyd:


lindong28:
@tfboyd There is conflict. Could you rebase the patch onto master?

tfboyd:
I cheated and changed it inline (wrong) and then changed it with another commit.  The conflict is fixed, not in the most graceful way but such is life.

PR #6188, opened 2019-02-12
Do not toggle eager if tf 2.0 is used.

tfboyd:
As discussed.  Open to changes.  Things look different when done vs. verbally discussing.  This really is not a big deal as the the tests for eager all set eager=True.  What this does is ensure a user that uses the code will not end up running a weird code path with tf.enable.eager=False.  And in a month or so I will remove this If all together.  Once tf 2.0 performance matches tf 1.0 eager and we feel good about it we can kind of shutdown tf 1.0 testing of keras or make branch until tf 1.0 is done.

I have no idea how this works internally, e.g. how do know you are running TF 2.0?

### DIFF
--- a/official/resnet/keras/keras_common.py
+++ b/official/resnet/keras/keras_common.py
@@ -252,6 +252,11 @@ def get_synth_input_fn(height, width, num_channels, num_classes,
   return input_fn
 
 
+def is_v2_0():
+  """Returns true if using tf 2.0."""
+  return tf.__version__.startswith('2')
+
+
 def get_strategy_scope(strategy):
   if strategy:
     strategy_scope = strategy.scope()
--- a/official/resnet/keras/keras_imagenet_main.py
+++ b/official/resnet/keras/keras_imagenet_main.py
@@ -87,7 +87,9 @@ def run(flags_obj):
   Raises:
     ValueError: If fp16 is passed as it is not currently supported.
   """
-  if flags_obj.enable_eager:
+  # TODO(tobyboyd): Remove eager flag when tf 1.0 testing ends.
+  # Eager is default in tf 2.0 and should not be toggled
+  if flags_obj.enable_eager and not keras_common.is_v2_0():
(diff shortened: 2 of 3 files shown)


official/resnet/keras/keras_common.py
```diff
@@ -252,6 +252,11 @@ def input_fn(is_training, data_dir, batch_size, *args, **kwargs):
   return input_fn
 
 
+def is_v2_0():
+  """Returns true if using tf 2.0."""
```
guptapriya: This seems fine to me. cc @annarev in case she knows better ways to check for 2.0 or not that work in both 1.x and 2.0

official/resnet/keras/keras_imagenet_main.py
```diff
@@ -87,7 +87,9 @@ def run(flags_obj):
   Raises:
     ValueError: If fp16 is passed as it is not currently supported.
   """
-  if flags_obj.enable_eager:
+  # TODO(tobyboyd): Remove eager flag when tf 1.0 testing ends.
+  # Eager is default in tf 2.0 and should not be toggled
+  if flags_obj.enable_eager and not keras_common.is_v2_0():
```
guptapriya: can we add the same check in keras_cifar_main?

tfboyd: I did see above.  :-)  Unless I do not understand you comment.

official/resnet/keras/keras_common.py
tfboyd: Not having contrib is a cheat we used in distribution_utils.  hasattr(tf, 'contrib')  Yuefeng thought of that one.

official/resnet/keras/keras_imagenet_main.py
guptapriya: oh my bad, that's exactly what I meant. somehow didn't see that file.

PR #6193, opened 2019-02-13
Add examples per second history to Estimator hook.

tfboyd:
Keep in mind this will be "dead code" at some point soon.  I did not want to rework it to track the time like we do for the keras hook.  I just wanted an approximation of average examples_per_second minus a small warmup.  I am not too worried about a perfect test, but if I want to make it better I can share the log_steps from keras with this in the future.  

Adding this will also allow me to add short estimator tests.

### DIFF
--- a/official/resnet/estimator_cifar_benchmark.py
+++ b/official/resnet/estimator_cifar_benchmark.py
@@ -125,7 +125,8 @@ class EstimatorCifar10BenchmarkTests(tf.test.Benchmark):
         eval_results['accuracy_top_5'].item())
     if examples_per_sec_hook:
       exp_per_second_list = examples_per_sec_hook.current_examples_per_sec_list
-      exp_per_sec = sum(exp_per_second_list[1:]) / (len(exp_per_second_list)-1)
+      # ExamplesPerSecondHook skips the first 10 steps.
+      exp_per_sec = sum(exp_per_second_list) / (len(exp_per_second_list))
       extras['exp_per_second'] = self._json_description(exp_per_sec)
 
     self.report_benchmark(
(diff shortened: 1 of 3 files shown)


official/resnet/estimator_cifar_benchmark.py
```diff
@@ -97,22 +102,36 @@ def unit_test(self):
...
+    eval_results = stats['eval_results']
+    extras = {}
+    extras['accuracy_top_1'] = self._json_description(
+        eval_results['accuracy'].item(),
+        priority=0)
+    extras['accuracy_top_5'] = self._json_description(
+        eval_results['accuracy_top_5'].item())
+    if examples_per_sec_hook:
+      exp_per_second_list = examples_per_sec_hook.current_examples_per_sec_list
+      exp_per_sec = sum(exp_per_second_list[1:]) / (len(exp_per_second_list)-1)
```
yuefengz: Is this to remove the warmup steps? Maybe we can document it?

Looks like there is already some logic to skip warmup steps in the `ExamplesPerSecondHook `?

tfboyd: It skips the first 10.  **I will go ahead and tweak** this to just average them all and leave a note that the hook already skips the first 10 anyway.  It is a bit sloppy as we do not have flags to control that piped through the estimator example and I do not see the point at the moment.

tfboyd: FIXED, good catch and idea.

yuefengz: Thanks! Looks good to me.

PR #6195, opened 2019-02-13
Manually scale the loss in Resnet DS model

rxsang:


official/resnet/resnet_run_loop.py
```diff
@@ -357,11 +357,17 @@ def exclude_batch_norm(name):
...
+  # Add weight decay to the loss. We need to scale the regularization loss
+  # manually as losses other than in tf.losses and tf.keras.losses don't scale
+  # automatically.
   l2_loss = weight_decay * tf.add_n(
       # loss is computed using fp32 for numerical stability.
-      [tf.nn.l2_loss(tf.cast(v, tf.float32))
-       for v in tf.compat.v1.trainable_variables() if loss_filter_fn(v.name)])
+      [
+          tf.nn.l2_loss(tf.cast(v, tf.float32)) /
+          tf.distribute.get_strategy().num_replicas_in_sync
```
guptapriya: would it be better to divide at the end, after adding everything? (similar to what you did for TPU example)

rxsang: That's correct. It will reduce the computation. Done.

PR #6196, opened 2019-02-14
One device strat

tfboyd:
One device from contrib to core.


tfboyd:
Tested on nightly v1 and works.

PR #6204, opened 2019-02-14
Fix lint issues

tfboyd:


tfboyd:
This should fix lint issues in random files so we can "make" people fix lint issues with less pain.

PR #6206, opened 2019-02-15
Multi-worker support for Resnet.

dubey:
This change adds support for running on multiple workers using `CollectiveAllReduceStrategy`.  It adds 2 flags to `resnet_run_loop.py`: `worker_hosts` and `task_index`.  These flags provide information to each worker about peers in the multi-worker setup.

After this change, when the official resnet model runs with `tf.contrib.distribute.CollectiveAllReduceStrategy,` it uses `tf.estimator.train_and_evaluate` rather than separate `train` and `evaluate` calls.

### DIFF
--- a/official/resnet/resnet_run_loop.py
+++ b/official/resnet/resnet_run_loop.py
@@ -24,6 +24,7 @@ from __future__ import division
 from __future__ import print_function
 
 import functools
+import json
 import math
 import multiprocessing
 import os
@@ -463,6 +464,28 @@ def resnet_main(
   if flags_obj.tf_gpu_thread_mode:
     override_flags_and_set_envars_for_gpu_thread_pool(flags_obj)
 
+  # Configures cluster spec for distribution strategy.
+  tf_config = json.loads(os.environ.get('TF_CONFIG', '{}'))
+  if tf_config:
+    num_workers = len(tf_config['cluster']['worker'])
+  elif flags_obj.worker_hosts:
+    workers = flags_obj.worker_hosts.split(',')
+    num_workers = len(workers)
+    if num_workers > 1 and flags_obj.task_index < 0:
+      raise ValueError('Must specify task_index when number of workers > 1')
+    task_index = 0 if num_workers == 1 else flags_obj.task_index
+    os.environ['TF_CONFIG'] = json.dumps({
+        'cluster': {
+            'worker': workers
+        },
+        'task': {'type': 'worker', 'index': task_index}
+    })
+  else:
+    num_workers = 1
+    if flags_obj.all_reduce_alg == 'collective':
+      raise ValueError('Must specify worker_hosts with collective all-reduce')
+
+
   # Creates session config. allow_soft_placement = True, is required for
   # multi-GPU and is not harmful for other modes.
   session_config = tf.ConfigProto(
@@ -471,7 +494,7 @@ def resnet_main(
       allow_soft_placement=True)
 
   distribution_strategy = distribution_utils.get_distribution_strategy(
-      flags_core.get_num_gpus(flags_obj), flags_obj.all_reduce_alg)
+      flags_core.get_num_gpus(flags_obj), flags_obj.all_reduce_alg, num_workers)
 
   # Creates a `RunConfig` that checkpoints every 24 hours which essentially
   # results in checkpoints determined only by `epochs_between_evals`.
@@ -560,10 +583,17 @@ def resnet_main(
     tf.logging.info('Starting cycle: %d/%d', cycle_index, int(n_loops))
 
     if num_train_epochs:
-      classifier.train(input_fn=lambda: input_fn_train(num_train_epochs),
-                       hooks=train_hooks, max_steps=flags_obj.max_train_steps)
-
-    tf.logging.info('Starting to evaluate.')
+      if num_workers > 1:
+        train_spec = tf.estimator.TrainSpec(
+            input_fn=lambda: input_fn_train(num_train_epochs), hooks=train_hooks,
+            max_steps=flags_obj.max_train_steps)
+      else:
+        classifier.train(input_fn=lambda: input_fn_train(num_train_epochs),
+                         hooks=train_hooks, max_steps=flags_obj.max_train_steps)
+    elif num_workers > 1:
+      train_spec = tf.estimator.TrainSpec(
+          input_fn=lambda: None, hooks=train_hooks,
+          max_steps=flags_obj.max_train_steps)
 
     # flags_obj.max_train_steps is generally associated with testing and
     # profiling. As a result it is frequently called with synthetic data, which
@@ -571,8 +601,16 @@ def resnet_main(
     # eval (which is generally unimportant in those circumstances) to terminate.
     # Note that eval will run for max_train_steps each loop, regardless of the
     # global_step count.
-    eval_results = classifier.evaluate(input_fn=input_fn_eval,
-                                       steps=flags_obj.max_train_steps)
+    if num_workers > 1:
+      eval_spec = tf.estimator.EvalSpec(input_fn=input_fn_eval,
+                                        steps=flags_obj.max_train_steps)
+      tf.logging.info('Starting to train and evaluate.')
+      eval_results, _ = tf.estimator.train_and_evaluate(classifier, train_spec,
+                                                        eval_spec)
+    else:
+      tf.logging.info('Starting to evaluate.')
+      eval_results = classifier.evaluate(input_fn=input_fn_eval,
+                                         steps=flags_obj.max_train_steps)
 
     benchmark_logger.log_evaluation_result(eval_results)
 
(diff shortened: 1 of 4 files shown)


official/resnet/resnet_run_loop.py
```diff
@@ -563,24 +584,41 @@ def input_fn_eval():
...
     # eval (which is generally unimportant in those circumstances) to terminate.
     # Note that eval will run for max_train_steps each loop, regardless of the
     # global_step count.
-    eval_results = classifier.evaluate(input_fn=input_fn_eval,
-                                       steps=flags_obj.max_train_steps)
+    if use_train_and_evaluate:
+      eval_spec = tf.estimator.EvalSpec(input_fn=input_fn_eval,
+                                        steps=flags_obj.max_train_steps)
+      tf.compat.v1.logging.info('Starting to train and evaluate.')
+      eval_results, _ = tf.estimator.train_and_evaluate(classifier, train_spec,
```
yuefengz: I am not sure whether it is the best practice to call `train_and_evaluate` in a loop. Usually we make an input_fn for all epochs and training finishes when `train_and_evaluate` returns. Also `train_and_evaluate` would run evaluation on evaluator continuously until it reaches  `max_train_steps`.

official/resnet/resnet_run_loop.py
```diff
@@ -563,24 +584,41 @@ def input_fn_eval():
     schedule = [flags_obj.epochs_between_evals for _ in range(int(n_loops))]
     schedule[-1] = flags_obj.train_epochs - sum(schedule[:-1])  # over counting.
 
+  use_train_and_evaluate = isinstance(
```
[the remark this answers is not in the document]
dubey: Sure, I'll make that change in the next commit.

official/resnet/resnet_run_loop.py
dubey: Can you elaborate a bit more? Does this mean when using `train_and_evaluate` there is no loop? But then how do we specify number of epochs?

official/resnet/resnet_run_loop.py
```diff
@@ -468,6 +469,25 @@ def resnet_main(
   if flags_obj.tf_gpu_thread_mode:
     override_flags_and_set_envars_for_gpu_thread_pool(flags_obj)
 
+  # Configures cluster spec for distribution strategy.
+  tf_config = json.loads(os.environ.get('TF_CONFIG', '{}'))
```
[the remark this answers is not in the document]
dubey: Will do in next commit.

official/resnet/resnet_run_loop.py
```diff
@@ -645,6 +683,15 @@ def define_resnet_flags(resnet_size_choices=None):
           'the expense of image resize/cropping being done as part of model '
           'inference. Note, this flag only applies to ImageNet and cannot '
           'be used for CIFAR.'))
+  flags.DEFINE_string(
```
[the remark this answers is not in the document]
dubey: ack

official/resnet/resnet_run_loop.py
yuefengz: We specify the number of epochs in the `input_fn` or specify the number of steps in `train_spec`. @karmel can correct me if I understand it wrong.

official/utils/misc/distribution_utils.py
```diff
@@ -56,6 +58,10 @@ def get_distribution_strategy(distribution_strategy="default",
                        "cannot be set to 'off'.".format(num_gpus))
     return None
 
+  if distribution_strategy == "collective" or num_workers > 1:
```
guptapriya: is 1 remote worker a valid use case? which strategy do we want users to use in that case? 

also, will it be confusing if the user explicitly says "collective" with 1 worker, but we still fallback to mirrored strategy? At the very least we should say if this is the case in documentation.

official/resnet/resnet_run_loop.py
```diff
@@ -645,6 +683,15 @@ def define_resnet_flags(resnet_size_choices=None):
...
+      name='worker_hosts', default=None,
```
guptapriya: I guess the expectation is that the user would have started this program on all these hosts? If yes, should we clarify this?

official/resnet/resnet_run_loop.py
```diff
@@ -478,6 +498,7 @@ def resnet_main(
   distribution_strategy = distribution_utils.get_distribution_strategy(
       distribution_strategy=flags_obj.distribution_strategy,
       num_gpus=flags_core.get_num_gpus(flags_obj),
+      num_workers=num_workers,
       all_reduce_alg=flags_obj.all_reduce_alg)
 
   # Creates a `RunConfig` that checkpoints every 24 hours which essentially
```
guptapriya: Unrelated to your CL but i just realized we do not distribute evaluation in this case. Should we? Or is it a bad idea because we throw away any partial batch so eval may not be always complete based on the batch size..

official/resnet/resnet_run_loop.py
```diff
@@ -645,6 +683,15 @@ def define_resnet_flags(resnet_size_choices=None):
           'the expense of image resize/cropping being done as part of model '
           'inference. Note, this flag only applies to ImageNet and cannot '
           'be used for CIFAR.'))
+  flags.DEFINE_string(
+      name='worker_hosts', default=None,
+      help=flags_core.help_wrap(
+          'Comma-separated list of worker ip:port pairs for running '
+          'multi-worker models with DistributionStrategy.'))
+  flags.DEFINE_integer(
```
guptapriya: Are you planning to update the README in this directory to guide users how to use multi worker training? (Can be a separate PR but would be good to update)

official/utils/misc/distribution_utils.py
dubey: The current behavior is that we use `CollectiveAllReduceStrategy` either when user explicitly specifies `distribution_strategy=collective` or there is more than 1 worker.  So, if they specify `collective` with 1 worker it should use `CollectiveAllReduceStrategy` and not `MirroredStrategy`.

official/resnet/resnet_run_loop.py
dubey: Yes, if they are launching the program directly they need to issue similar invocations at all workers.

official/resnet/resnet_run_loop.py
dubey: Thanks for that suggestion, will do

official/resnet/resnet_run_loop.py
guptapriya: Should we clarify this in the docstring?

official/utils/misc/distribution_utils.py
guptapriya: oh my bad. I think I must've read that "or" as an "and". What you have is fine then.

official/utils/misc/distribution_utils.py
```diff
@@ -56,6 +58,10 @@ def get_distribution_strategy(distribution_strategy="default",
...
+    return tf.contrib.distribute.CollectiveAllReduceStrategy(
```
guptapriya: If Yuefeng submits his change to export the `CollectiveAllReduceStrategy` (and ParameterServerStrategy) in core TF soon, then perhaps this can be changed to use that and be 2.0 compatible?

official/resnet/resnet_run_loop.py
karmel: My understanding matches @yuefengz 's -- train and eval runs the loop for you.

official/resnet/resnet_run_loop.py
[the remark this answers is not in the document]
dubey: Done.

official/resnet/resnet_run_loop.py
dubey: Thanks, done.

official/resnet/resnet_run_loop.py
dubey: Done.

official/utils/misc/distribution_utils.py
dubey: Sure, I will update after Yuefeng's change.

dubey:
I think I've resolved all blockers for this PR.  @yuefengz PTAL.

official/utils/misc/distribution_utils.py
```diff
@@ -56,6 +60,10 @@ def get_distribution_strategy(distribution_strategy="default",
                        "cannot be set to 'off'.".format(num_gpus))
     return None
 
+  if distribution_strategy == "collective" or num_workers > 1:
```
guptapriya: @yuefengz @dubey Since the name of the strategy as exposed has now changed (MultiWorkerMirroredStrategy), i wonder if we should rename this flag too?

dubey: done.  I'll change the strategy object from contrib.distribute.collectiveallreducestrategy to distribute.multiworkermirroredstrategy in a follow-on PR.

PR #6211, opened 2019-02-17
Pass datasets_num_private_threads flag into Keras resnet model.

yuefengz:


official/resnet/keras/keras_imagenet_main.py
```diff
@@ -116,17 +116,20 @@ def run(flags_obj):
...
+      num_epochs=flags_obj.train_epochs,
+      parse_record_fn=parse_record_keras,
+      datasets_num_private_threads=flags_obj.datasets_num_private_threads)
+
+  eval_input_dataset = input_fn(
+      is_training=False,
+      data_dir=flags_obj.data_dir,
+      batch_size=flags_obj.batch_size,
+      num_epochs=flags_obj.train_epochs,
+      parse_record_fn=parse_record_keras)
```
guptapriya: what do you think about passing to the eval dataset also?

official/resnet/resnet_run_loop.py
```diff
@@ -77,7 +77,6 @@ def process_record_dataset(dataset,
   # Defines a specific size thread pool for tf.data operations.
   if datasets_num_private_threads:
     options = tf.data.Options()
-    options.experimental_threading = tf.data.experimental.ThreadingOptions()
```
guptapriya: removing because unnecessary?

official/resnet/keras/keras_imagenet_main.py
yuefengz: I am not very clear about the impact of doing so. Would that have any impact on training dataset? Before figuring this out, I would err on the side of not doing so.

official/resnet/resnet_run_loop.py
yuefengz: Right.

yuefengz:
Thanks for the review! I'll probably add it to benchmark tomorrow. cc @tfboyd

PR #6213, opened 2019-02-17
Allow user to specify root_data_dir in the benchmark class constructor

lindong28:
Currently the benchmark class harcodes the path to the dataset, e.g. /data/imagenet for imagenet benchmark. This may not always work since sometimes the dataset may not be under directory `/data`.

This patch adds optional parameter to the constructor of these benchmark classes so that user can specify a directory other than `/data` for dataset.

### DIFF
--- a/official/resnet/estimator_cifar_benchmark.py
+++ b/official/resnet/estimator_cifar_benchmark.py
@@ -29,22 +29,33 @@ import tensorflow as tf  # pylint: disable=g-bad-import-order
 from official.resnet import cifar10_main as cifar_main
 from official.utils.logs import hooks
 
-DATA_DIR = '/data/cifar10_data/cifar-10-batches-bin'
-
 
 class EstimatorCifar10BenchmarkTests(tf.test.Benchmark):
   """Benchmarks and accuracy tests for Estimator ResNet56."""
 
   local_flags = None
 
-  def __init__(self, output_dir=None):
+  def __init__(self, output_dir=None, root_data_dir=None):
+    """A benchmark class.
+
+    Args:
+      output_dir: directory where to output e.g. log files
+      root_data_dir: directory under which to look for dataset
+    """
+
     self.output_dir = output_dir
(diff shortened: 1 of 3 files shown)


official/resnet/estimator_cifar_benchmark.py
```diff
@@ -29,22 +29,33 @@
...
+      root_data_dir: directory under which to look for dataset
+    """
+
     self.output_dir = output_dir
 
+    if root_data_dir is None:
+      self.data_dir = '/data/cifar10_data/cifar-10-batches-bin'
+    else:
+      self.data_dir = os.path.join(root_data_dir,
+                                   'cifar10_data/cifar-10-batches-bin')
```
tfboyd: This path should not have /cifar-10-batches-bin in it that greatly restricts the structure the user would setup.  We can update the jobs that download CIFAR-10 data to put it directly in cifar10_data.  Long story, we had to tweak the underlying code to resolve this issue internally due to how we stored the data in CNS and realized we should not go more than one folder deep as far as assumptions of names unless the structure is necessary and well accepted as the "only" way to setup the data.

official/resnet/estimator_cifar_benchmark.py
```diff
@@ -29,22 +29,33 @@
...
+    """A benchmark class.
+
+    Args:
+      output_dir: directory where to output e.g. log files
+      root_data_dir: directory under which to look for dataset
+    """
+
     self.output_dir = output_dir
 
+    if root_data_dir is None:
```
tfboyd: We will have to figure out how this will work in CNS.  I think it will be fine to move data around to match the pattern if we don't this will be very hard and maybe impossible to manage.  Not action needed.

official/resnet/estimator_cifar_benchmark.py
lindong28: Very good point. It is fixed now.

PR #6218, opened 2019-02-17
Update crate_dataset.py

ShantanuShinde:
In the loop for initializing writers, Flags.output_shards is being used even though output_shards is passed as a parameter to the function "convert_data()".


ShantanuShinde:
I have signed it!

jaeyounkim:
CLA not signed

PR #6234, opened 2019-02-20
Benchmark Keras Application Models in TF2.0

xunkai55:
(WIP) This PR adds codes for training Keras App Models in TF2.0 and benchmarking them.

Not finished yet. Let's firstly discuss code structure and design principles. I tried to follow the principles below, but they're just my opinions and we should definitely discuss.

1. Try to use 2.0 codes and avoid "compat.v1".
2. Try to not introduce cross-directory dependency in the training code. I think it would be easier for external users adopting those examples, as they don't need to understand hierarchical code structures or non-core utils.
3. Try to adopt best practice for simple training -- dataset, keras and dist_strat.
4. Try to use APIs only in tf.data, tf.keras and tf.distribute.

As a result, the code looks less "engineered". I'm not very sure if the principles are correct. Opening the Draft PR for discussing before I go too far.

### DIFF
--- a/official/keras_application_models/v2/README.md
+++ b/official/keras_application_models/v2/README.md
@@ -1,0 +1,3 @@
+# Examples for using Keras Application Models in TF2.0
+
+TODO(xunkai)
--- a/official/keras_application_models/v2/benchmark/README.md
+++ b/official/keras_application_models/v2/benchmark/README.md
@@ -1,0 +1,3 @@
+# Benchmarks for Keras Application Models (in TF2.0)
+
+TODO(xunkai)
--- a/official/keras_application_models/v2/benchmark/resnet50_benchmark.py
+++ b/official/keras_application_models/v2/benchmark/resnet50_benchmark.py
@@ -1,0 +1,66 @@
+# Copyright 2019 The TensorFlow Authors. All Rights Reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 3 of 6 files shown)


tfboyd:
Good start.  I am going to sync to this and try to run it.  I am not the full expert but I think I see a few things we will want to maybe tweak with learning rate but in general this seems like a starting point.  

**Can you point me to how to get ImageNet in the format this script wants?**  I assume it will be ok with our normal tf.Record format?

xunkai55:
In this script I don't run ImageNet data -- I use CIFAR-10 to fine-tune and test, like transfer learning.

The CIFAR data is downloaded by `tf.keras.datasets.cifar10.load_data()`.

In the future, I want to unify all image classification datasets into such a 7 element tuple:

```
(train_ds: tf.data.Dataset, test_ds: tf.data.Dataset, num_train: int, num_test: int, num_classes: int, image_shape: (int, int, int), batch_size: int)
```

where `image_shape` and `batch_size` are configurable when creating the Dataset objects.

tfboyd:
OK I will get this running.  Can you get one setup to train ImageNet from scratch.  I would like either MobileNet or ResNet50 training from 0.  I care more about MobileNetV2 but both would be fine as I am curious about the performance of ResNet50 compared to our own version of ResNet50.

jaeyounkim:
I am closing this PR since we have updated the official directory with more training examples.

PR #6239, opened 2019-02-20
Fix import dataset_tools statement

chardch:
This import from google3.third_party doesn't work.


chardch:
Glad to help! Also, I am under the corporate CLA. How can I get the CLA check passed?

PR #6255, opened 2019-02-22
speedup caption generator on im2txt

rola93:
replace the usage of python's built in sort with numpy's argsort, which is faster. Also use the given axes to get the most_likely_words instead of calling enumerate and list (it's slower and more memory consuming)

### DIFF
--- a/research/im2txt/im2txt/inference_utils/caption_generator.py
+++ b/research/im2txt/im2txt/inference_utils/caption_generator.py
@@ -176,11 +176,12 @@ class CaptionGenerator(object):
         word_probabilities = softmax[i]
         state = new_states[i]
         # For this partial caption, get the beam_size most probable next words.
-        words_and_probs = list(enumerate(word_probabilities))
-        words_and_probs.sort(key=lambda x: -x[1])
-        words_and_probs = words_and_probs[0:self.beam_size]
-        # Each next word gives a new partial caption.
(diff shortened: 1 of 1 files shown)


rola93:
I signed it!

research/im2txt/im2txt/inference_utils/caption_generator.py
```diff
@@ -176,11 +176,12 @@ def beam_search(self, sess, encoded_image):
         word_probabilities = softmax[i]
         state = new_states[i]
         # For this partial caption, get the beam_size most probable next words.
-        words_and_probs = list(enumerate(word_probabilities))
-        words_and_probs.sort(key=lambda x: -x[1])
-        words_and_probs = words_and_probs[0:self.beam_size]
-        # Each next word gives a new partial caption.
-        for w, p in words_and_probs:
+        most_likely_words = np.argsort(word_probabilities)[:-self.beam_size][::-1]
+        # sort the indexes with numpy, select the last self.beam_size (3 by default) (ie, the most likely)
```
cshallue: - Please move this comment above the line of code it describes
- Keep lines within 80 chars
- Capitalize the first word and end with a period
- "let" -> "sort"

rola93: Solved! Thanks for your feedback!

PR #6265, opened 2019-02-25
Improvements to CIFAR10 tutorial

balakrishnanv:
Hi Jon, Sherry
I updated the cifar10.py's input pipeline to use the new DataSet API. Also the cifar10_train.py is updated to do both train and eval (after every epoch) instead of doing them in two different py files.
Please take a look


jaeyounkim:
CLA not signed

PR #6283, opened 2019-02-28
Add benchmarks for thread tuning.

yuefengz:
Also removed the bfc allocator one. We'll add GPU thread pool, inter op, intra op pool later.

### DIFF
--- a/official/resnet/keras/keras_imagenet_benchmark.py
+++ b/official/resnet/keras/keras_imagenet_benchmark.py
@@ -74,6 +74,7 @@ class Resnet50KerasAccuracy(keras_benchmark.KerasBenchmark):
     FLAGS.model_dir = self._get_model_dir('benchmark_8_gpu')
     FLAGS.dtype = 'fp32'
     FLAGS.enable_eager = True
+    # Add some thread tunings to improve performance.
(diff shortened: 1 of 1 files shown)


official/resnet/keras/keras_imagenet_benchmark.py
```diff
@@ -76,22 +76,17 @@ def benchmark_8_gpu(self):
     FLAGS.enable_eager = True
     self._run_and_report_benchmark()
 
-  def benchmark_8_gpu_bfc_allocator(self):
-    """Restricts CPU memory allocation."""
+  def benchmark_8_gpu_tweaked(self):
```
tfboyd: Can you  go ahead and delete this test completely and make the test above `def benchmark_8_gpu(self):`  run with the 14 threads.  Since the goal for this class of tests is to test accuracy first it is ok for the accuracy test to be "tuned".  The short test are our primary location for perf tests.  I realize this is a bit weird.  If the accuracy tests did not take 16 hours I would have a normal and then tweaked version.  Thank you and this is a great test to track our progress.

yuefengz: Done.

PR #6288, opened 2019-03-01
Adding option to sort input sentences by token count for inference

kdatta:
This commit adds an option to order the input sentences on token-count instead of wordcount to translate.py. Order of input sentences sorted on wordcount is quite different than sentences sorted on token-count. This is the reason possibly the padding overhead in wordcount is higher than token-count. We found that default sorting on wordcount results in slower inference throughput. Hence, the commit. We ensured that the ordering has no effect on BLEU score.

To run with this option, pass --sort_by_tokens=True from the command line to translate.py

We also add options to pass inter_op and intra_op from command line.


jaeyounkim:
CLA not signed

PR #6301, opened 2019-03-04
Migration of official/mnist to TF 2.0

tejank10:
Migrated `dataset.py`, `mnist_eager.py` and `mnist_eager_test.py` to TF 2.0

I am not sure about `defun` tests in `mnist_eager_test.py`, have skipped them for now. Can someone tell me about how to go about them?

### DIFF
--- a/official/mnist/dataset.py
+++ b/official/mnist/dataset.py
@@ -35,13 +35,13 @@ def read32(bytestream):
 
 def check_image_file_header(filename):
   """Validate that filename corresponds to images for the MNIST dataset."""
-  with tf.gfile.Open(filename, 'rb') as f:
+  with tf.io.gfile.GFile(filename, 'rb') as f:
     magic = read32(f)
     read32(f)  # num_images, unused
     rows = read32(f)
     cols = read32(f)
     if magic != 2051:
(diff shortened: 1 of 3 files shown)


tfboyd:
Thank you for starting this.  I need to create a unit test for this an an end-to-end to run nightly so we can verify the changes.  We are busy this week with the summit but I hope to get to it next week.  I am going to see if @yashk2810 can provide some guidance before we jump in.  Here are some questions I can think of:

1) Should we upgrade this at all or if we just have mnist in the examples and colab?
2) If we do want this do we want it to be TF 2.0 only or also work in TF 1.0?
3) Do we want to do something about the estimator version, should it be keras based?

I see you focused on the eager aspect and removed all the contrib calls that will not work in TF 2.0.

tfboyd:
I am thinking we may drop this because it is in the tutorial and this ends up bit rotting.  I will think about it for a week or so and see if anyone else comments.

mandroid6:
Also @tejank10 , over the summer the project is focused on migrating the research models under tensorflow/models/research.

Take up one of the simple one's like https://github.com/tensorflow/models/tree/master/research/autoencoder to get started.

jaeyounkim:
I am closing this PR because we already have MNIST examples in TF 2.

PR #6303, opened 2019-03-04
Update object detection tutorial for use in Colab

olgn:
Google Colab will only display image results if the `%matplotlib inline` _precedes_ `from matplotlib import pyplot as plt`. This pr moves the ENV code cell above the IMPORTS cell.


jaeyounkim:
@olgn Have you tested your changes? Please resolve conflicts.

PR #6309, opened 2019-03-06
Mixed precision support

reedwm:
Add mixed precision support to the Keras Resnet model.

This uses the experimental tf.keras mixed precision API. The API will eventually allow users to use mixed precision with a single line of code, but currently it requires several changes to model code.

I tested mixed precision converges with the command line:

```
python keras_imagenet_main.py --clean --num_gpus=8 --data_dir=/raid/imagenet-2012-tfrecord --batch_size=1024 --train_epochs=90 --model_dir ~/fp16_keras_mod --dtype=fp16 --enable_eager=true --datasets_num_private_threads=14
```

The final validation accuracy is 76.35%. I ran tensorflow at commit tensorflow/tensorflow@23ec33ae1418d56c3ce0294720782b636595a2b3.

### DIFF
--- a/official/resnet/keras/resnet_model.py
+++ b/official/resnet/keras/resnet_model.py
@@ -174,7 +174,7 @@ def conv_block(input_tensor,
   return x
 
 
-def resnet50(num_classes):
+def resnet50(num_classes, dtype='float32'):
   # TODO(tfboyd): add training argument, just lik resnet56.
   """Instantiates the ResNet50 architecture.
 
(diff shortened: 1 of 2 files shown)


official/resnet/keras/resnet_model.py
```diff
@@ -232,10 +232,12 @@ def resnet50(num_classes):
 
   x = layers.GlobalAveragePooling2D(name='avg_pool')(x)
   x = layers.Dense(
-      num_classes, activation='softmax',
+      num_classes,
       kernel_regularizer=regularizers.l2(L2_WEIGHT_DECAY),
       bias_regularizer=regularizers.l2(L2_WEIGHT_DECAY),
       name='fc1000')(x)
+  x = backend.cast(x, 'float32')
```
azaks2: add a comment and a todo to remove the code when layers have fp16 support

PR #6312, opened 2019-03-06
fix unitest failure in faster_rcnn_meta_arch_test_lib.py

danyfang:
The pull request contains two fixes.
The first one is the missing method `_get_box_classifier_features_shape` in class `FasterRCNNMetaArchTestBase`, this will cause the following error 'AttributeError: 'FasterRCNNMetaArchTestBase' object has no attribute '_get_box_classifier_features_shape'.

The second one is caused by python version. In `faster_rcnn_meta_arch_test_lib.py` line 1677 the `uninitialized_vars_list` is returned as a bytes array in python3 so the assertion in the following line will fail in  python3.

### DIFF
--- a/research/object_detection/meta_architectures/faster_rcnn_meta_arch_test_lib.py
+++ b/research/object_detection/meta_architectures/faster_rcnn_meta_arch_test_lib.py
@@ -143,6 +143,14 @@ class FasterRCNNMetaArchTestBase(test_case.TestCase, parameterized.TestCase):
         num_classes=num_classes,
         is_training=is_training)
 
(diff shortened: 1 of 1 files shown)


danyfang:
I signed it!

danyfang:
The exact function `_get_box_classifier_features_shape` can be found in `faster_cnn_meta_arch_test.py` so I think this function is merely put in the wrong place, it should be included in the base class `FasterRCNNMetaArchTestBase` in file `faster_rcnn_meta_arch_test_lib.py`

PR #6317, opened 2019-03-07
Add command line option for multi worker collective implementations.

dubey:


official/resnet/resnet_run_loop.py
```diff
@@ -486,10 +486,13 @@ def resnet_main(
 
   # Creates a `RunConfig` that checkpoints every 24 hours which essentially
   # results in checkpoints determined only by `epochs_between_evals`.
+  save_checkpoints_secs = None if (distribution_strategy.__class__.__name__ ==
```
yuefengz: I would prefer to completely disable checkpointing. If you don't do that, do you also want to check whether the name is "MultiWorkerMirroredStrategy"?

dubey: disabled checkpointing

dubey:
@guptapriya I'll update README in follow-on PR.

PR #6324, opened 2019-03-08
Quick Draw Tutorial: Added Dropout to `DropoutWrapper`

HanGuo97:
Currently, the dropout seems to be ignored in regular RNN given that it's not actually used in the `DropoutWrapper`:

```python
if params.dropout > 0.0:
      cells_fw = [tf.contrib.rnn.DropoutWrapper(cell) for cell in cells_fw]
      cells_bw = [tf.contrib.rnn.DropoutWrapper(cell) for cell in cells_bw]
```


jaeyounkim:
Deprecated

PR #6326, opened 2019-03-08
Update attribute call

shiyinw:
Currently, the `tf.distribute` is modified into `tf.contrib.distribute`.


shiyinw:
I signed it!

robieta:
This was an intentional change. `tf.contrib` is deprecated.

PR #6345, opened 2019-03-12
fix typo to use image_np_expanded

guoyejun:


guoyejun:
yes, it works with my change.

guoyejun:
any comment?  thanks.

PR #6355, opened 2019-03-13
Fix ncf test for keras

seemuch:
The failure is mostly caused by the flag values passed in. 
The Keras implementation is sensitive about the batch_size, which needs to be dividable by (NUM_EVAL_NEGATIVES+1). In the test, the NUM_EVAL_NEGATIVES is set to 2, while in an ordinary run, it is 999.

### DIFF
--- a/official/recommendation/ncf_keras_main.py
+++ b/official/recommendation/ncf_keras_main.py
@@ -151,18 +151,20 @@ def _get_keras_model(params):
 
 def run_ncf(_):
   """Run NCF training and eval with Keras."""
-  # TODO(seemuch): This model does not support eval_batch_size to be
-  # different from batch_size, because in the input layers, it explicitly
-  # the batch_size to 'batch_size'
-  FLAGS.eval_batch_size = FLAGS.batch_size
+  # TODO(seemuch): Support different train and eval batch sizes
+  if FLAGS.eval_batch_size != FLAGS.batch_size:
+    tf.logging.warning(
+        "The Keras implementation of NCF currently does not support batch_size "
+        "!= eval_batch_size ({} vs. {}). Overriding eval_batch_size to match "
+        "batch_size".format(FLAGS.eval_batch_size, FLAGS.batch_size)
+        )
+    FLAGS.eval_batch_size = FLAGS.batch_size
 
   params = ncf_common.parse_flags(FLAGS)
 
-  # TODO(seemuch): now that the eval_batch_size might have been adjusted, asign
-  # it back to batch_size. This is not needed once the batch_size issue is
-  # fixed.
-  if params['batch_size'] != params['eval_batch_size']:
-    params['batch_size'] = params['eval_batch_size']
+  # ncf_common rounds eval_batch_size (this is needed due to a reshape during
+  # eval). This carries over that rounding to batch_size as well.
+  params['batch_size'] = params['eval_batch_size']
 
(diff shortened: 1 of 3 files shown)


robieta:
Huh. Doesn't that imply that it will also error out in a real run if the divisibility is wrong? If that is the case, it seems better to use the same batch rounding logic that Estimator uses for the Keras code.

seemuch:
Re robieta: it does, that is why for keras, the batch size used for both training and eval is 160000, to be dividable by 1000. 
Where is the logic exactly?

official/recommendation/ncf_keras_main.py
```diff
@@ -153,6 +153,12 @@ def run_ncf(_):
   """Run NCF training and eval with Keras."""
```
robieta: I think we need to set `eval_batch_size` equal to `batch_size` here. (And put a `logging.warning` statement if we rewrite it.)

And then after the parse flags call, we need to set `batch_size` to `eval_batch_size`.

official/recommendation/ncf_test.py
```diff
@@ -185,34 +185,38 @@ def test_hit_rate_and_ndcg(self):
     self.assertAlmostEqual(ndcg, (1 + math.log(2) / math.log(3) +
                                   2 * math.log(2) / math.log(4)) / 4)
 
-  _BASE_END_TO_END_FLAGS = ['-batch_size', '1024', '-train_epochs', '1']
+  _BASE_END_TO_END_FLAGS_ESTIMATOR = [
```
robieta: Once we have the rounding behavior in Keras, we won't need to have separate flags for Keras and Estimator.

official/recommendation/ncf_keras_main.py
seemuch: Done.

official/recommendation/ncf_test.py
seemuch: Done.

official/recommendation/ncf_keras_main.py
```diff
@@ -151,8 +151,19 @@ def _get_keras_model(params):
 
 def run_ncf(_):
   """Run NCF training and eval with Keras."""
+  # TODO(seemuch): This model does not support eval_batch_size to be
+  # different from batch_size, because in the input layers, it explicitly
+  # the batch_size to 'batch_size'
+  FLAGS.eval_batch_size = FLAGS.batch_size
```
robieta: I think it's a little clearer if it is:
```
# TODO(seemuch): Support different train and eval batch sizes
if FLAGS.eval_batch_size != FLAGS.batch_size:
  logging.warning(
      "The Keras implementation of NCF currently does not support batch_size "
      "!= eval_batch_size ({} vs. {}). Overriding eval_batch_size to match "
      "batch_size".format(FLAGS.eval_batch_size, FLAGS.batch_size)
  )
  FLAGS.eval_batch_size = FLAGS.batch_size

params = ncf_common.parse_flags(FLAGS)

# ncf_common rounds eval_batch_size (this is needed due to a reshape during 
# eval). This carries over that rounding to batch_size as well.
params['batch_size'] = params['eval_batch_size']
```

I'll admit I didn't do a great job conveying what I was asking for.

seemuch: Done.

robieta:
Thanks!

PR #6359, opened 2019-03-14
change tf.to_int32 to tf.cast

tranvohuy:
In models/official/mnist/dataset.py: the use of ```tf.to_int32``` raises deprecated warning.

I've changed ```tf.to_int32(label)``` to ```tf.cast(label, tf.int32)```


tranvohuy:
I signed it

PR #6368, opened 2019-03-14
Add the option to run Keras resnet model on multiple workers.

sshrdp:
Add the option to run Keras resnet model on multiple workers.
Fix the logic that computes the number of workers to include chief if it is present.

### DIFF
--- a/official/utils/misc/distribution_utils.py
+++ b/official/utils/misc/distribution_utils.py
@@ -217,6 +217,8 @@ def configure_cluster(worker_hosts=None, task_index=-1):
   tf_config = json.loads(os.environ.get('TF_CONFIG', '{}'))
   if tf_config:
     num_workers = len(tf_config['cluster']['worker'])
+    if tf_config['cluster'].get('chief', None):
+      num_workers += 1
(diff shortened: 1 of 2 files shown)


tfboyd:
Adding Ayush, there has been focus on the estimator example as performance is cleaned up and testing LARs for huge batch sizes.  I do not know if Keras is ready to go, but he can comment and give you any tips to reduce frustration.

Thank you for the PR we certainly want to get this working.

official/utils/misc/distribution_utils.py
```diff
@@ -217,6 +217,8 @@ def configure_cluster(worker_hosts=None, task_index=-1):
   tf_config = json.loads(os.environ.get('TF_CONFIG', '{}'))
   if tf_config:
     num_workers = len(tf_config['cluster']['worker'])
+    if tf_config['cluster'].get('chief', None):
+      num_workers += 1
```
dubey: Thanks for this fix!

yuefengz:
Yeah, we need dataset sharding to make it converge. But we can add that later.

PR #6396, opened 2019-03-19
Added thread tuning and tweaked tests to improve Keras model performance

haoyuz:
Use GPU private threads to schedule and launch kernels, and limit the number of tf.data preprocessing threads to reduce contention.

This should bring the performance of real data tests close to synthetic ones.

### DIFF
--- a/official/resnet/keras/keras_common.py
+++ b/official/resnet/keras/keras_common.py
@@ -18,6 +18,8 @@ from __future__ import absolute_import
 from __future__ import division
 from __future__ import print_function
 
+import multiprocessing
+import os
 import time
 
 import numpy as np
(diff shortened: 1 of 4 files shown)


official/resnet/keras/keras_common.py
```diff
@@ -146,6 +148,31 @@ def get_config_proto():
...
+  os.environ['TF_GPU_THREAD_COUNT'] = str(per_gpu_thread_count)
+  tf.compat.v1.logging.info('TF_GPU_THREAD_COUNT: %s',
+                            os.environ['TF_GPU_THREAD_COUNT'])
+  tf.compat.v1.logging.info('TF_GPU_THREAD_MODE: %s',
+                            os.environ['TF_GPU_THREAD_MODE'])
+
+  # Limit data preprocessing threadpool to CPU cores minus number of total GPU
+  # private threads and memory copy threads.
+  total_gpu_thread_count = per_gpu_thread_count * flags_obj.num_gpus
+  num_mem_copy_threads = flags_obj.num_gpus
```
yuefengz: The estimator path uses more `monitoring_threads` and as a result less dataset threads. We probably can try it out later as well.

haoyuz: Sounds good; let's fine tune the threads a little further. Thanks for the review!

PR #6413, opened 2019-03-20
Added python script to compile *.proto files.

rapsealk:
- Recommended command on installation guide is not working.
- Tested with Python 3.6 and protoc installed.


rapsealk:
Acquired a CLA.

pkulzc:
Which part of installation guide is not working?
The tutorial you added if just a copy of our notebook, I don't think we should add it.

jaeyounkim:
@rapsealk I am closing this pull request since there is no response.

PR #6433, opened 2019-03-25
Possibility to pass loss weights as arguments

adrianboguszewski:
One can pass loss weight per class as call arguments now.


aquariusjay:
Hi adrianboguszew,

Thanks for the PR!
In fact, we are working on a similar change for this feature as well.
We will merge it once ready.
Thus, we will postpone your PR.
Sorry about that, and hope you don't mind.
In any case, we very appreciate your contribution!

Thanks,

adrianboguszewski:
Hi aquariusjay,

No problem :)

PR #6447, opened 2019-03-26
Add DS support for NCF keras 

seemuch:
With the fix in distributed_training_utils (which is currently pending review), the NCF model converges with distribution strategies.


official/recommendation/ncf_keras_main.py
```diff
@@ -74,8 +75,12 @@ def metric_fn(y_true, y_pred):
             params["match_mlperf"],
             params["use_xla_for_gpu"]))
 
+    is_learning_phase = tf.keras.backend.learning_phase()
```
guptapriya: this should be called `is_training` or something. learning_phase is something that can be 1 or 0. 

Why was this change needed now btw?

official/recommendation/ncf_keras_main.py
```diff
@@ -87,11 +92,19 @@ def metric_fn(y_true, y_pred):
 def _get_train_and_eval_data(producer, params):
   """Returns the datasets for training and evalutating."""
 
+  def preprocess_train_inpu(features, labels):
```
guptapriya: please add comments explaining why this is needed, here and elsewhere

official/recommendation/ncf_keras_main.py
```diff
@@ -121,25 +134,24 @@ def _get_keras_model(params):
   batch_size = params['batch_size']
 
   user_input = tf.keras.layers.Input(
-      shape=(),
-      batch_size=batch_size,
-      name=movielens.USER_COLUMN,
-      dtype=rconst.USER_DTYPE)
-
+      shape=(batch_size,), batch_size=1, name=movielens.USER_COLUMN, dtype=tf.int32)
```
guptapriya: is the batch_size argument needed?

official/utils/misc/distribution_utils.py
```diff
@@ -246,3 +246,21 @@ def configure_cluster(worker_hosts=None, task_index=-1):
   else:
     num_workers = 1
   return num_workers
+
+
+class MaybeDistributionScope(object):
```
guptapriya: Why not re-use what you added for resnet? 
https://github.com/tensorflow/models/blob/master/official/resnet/keras/keras_imagenet_main.py#L127

official/recommendation/ncf_keras_main.py
seemuch: This is not absolutely necessary. It's just that this metric is not used in the training phase, so I set it to zero.

official/recommendation/ncf_keras_main.py
seemuch: I believe so. With this argument, the input is of shape (1, 80000), which matches the data.

official/utils/misc/distribution_utils.py
seemuch: Good idea.

guptapriya:
Another thing - either in this PR or a new one would be to add the distributed case to the unit tests and benchmarks.

seemuch:
Re @guptapriya : I will do the benchmark thing in a separate PR.

guptapriya:
Code looks good. Please add some unit tests that exercise the new codepaths

official/recommendation/ncf_keras_main.py
```diff
@@ -121,25 +150,31 @@ def _get_keras_model(params):
   batch_size = params['batch_size']
 
   user_input = tf.keras.layers.Input(
-      shape=(),
-      batch_size=batch_size,
+      shape=(batch_size,),
```
robieta: I feel like this definitely needs a comment explaining.

seemuch: Done.

seemuch:
I will add benchmark and unit test in separate PRs.

PR #6456, opened 2019-03-27
[WIP] Migration of official/resnet and Cifar-10 to TF 2.0

Atom-101:
Migrated `resnet_model.py` to TF 2.0 and the training code for Cifar-10 to TF 2.0

Wrote two variants of the Model class, using keras functional API. One directly creates tf.keras.Model object while the other returns an object that can perform forward pass.

Refactored the Cifar-10 code to use keras API calls instead of estimators. Tried to retain as much functionality of the original scripts as possible. I have not implemented fp16 support. I removed tf flags and used defined constants instead.


tfboyd:
Thank you for putting this together.  We already have this under the keras folder and I run nightly perf and end-to-end tests on a variety of scenarios.  It also already has FP16 and we added XLA and perf has improved from ~5K images/sec to 9,900 images/sec for synthetic and ~7-8K real but that is being resolved.  

I am not sure what I can say that does not make this not a "downer".  The work looks useful, it has just already been done.  I will leave this open a few days as I feel bad just closing it abruptly.

Atom-101:
Thanks a lot for taking the time to review my work. The point of my fork was to remove all instances of tf.compat from the code and refactor it just enough to match the design philosophies of TF 2.0(eager execution, higher level keras API calls for training, etc.). I made the pull request to allow the devs to judge my preliminary work, since I am applying for Google Summer of Code at Tensorflow. 

I know I haven't hit the performance level of the official repository. I didnt want you to merge so I used the WIP tag.

tfboyd:
We did one with compatibility and then the Keras folder is almost pure V2.
There is some contamination.  My first start digging into TF, and I am not
very competent was rewriting examples to speed up perf and checking if I
broke accuracy. I benefit from near unlimited compute.  I will leave this
open until it looks like you are done.  It will not be merged.

Good project.  There are other models in official we are not converting if
you have interest we can work together to find one that you can submit.

Let me know

Atom-101:
@tfboyd Thanks a ton. It would be really great if you can point out the models that aren't receiving as much attention from the team. I had picked Resnet because I have had lots of experience dealing with various kinds of resnets. Thanks for being so supportive.

nelsongonzalez:
@Atom-101 i see there is an unresolved conflict at the end of the file `official/resnet/resnet_model.py`

tfboyd:
I am going to close this.  I forgot to come back.  I did update someone else that the only models left were the canned estimators ones and they are not interesting as the code in here was just going to call into the model code the exists in TensorFlow core.   I still feel bad about the work that was done, and I hope maybe something was learned and you do not feel it was a total waste of time.

PR #6462, opened 2019-03-28
Add trivial model tests for Resnet data input pipeline

haoyuz:
These tests should be removed once data input bottleneck issue is fixed.


tfboyd:
Will let tonight and merge, then add to a kokoro job.

haoyuz:
@tfboyd Thanks for the comments! Regrouped the trivial model tests into a separate class, and change it to run 600 steps each. For 8GPU running these many steps almost covers the entire dataset so we might run any of them as a dummy test first and then start the real ones. Open to any suggestions. Thank you!

tfboyd:
We can figure out the run time.  We can get this going now and then tweak them.

PR #6508, opened 2019-04-02
 Ensure static shapes when enabling XLA in Resnet Keras model

rxsang:


official/resnet/keras/keras_imagenet_main.py
```diff
@@ -146,7 +147,8 @@ def run(flags_obj):
       num_epochs=flags_obj.train_epochs,
       parse_record_fn=parse_record_keras,
       datasets_num_private_threads=flags_obj.datasets_num_private_threads,
-      dtype=dtype)
+      dtype=dtype,
+      drop_remainder=flags_obj.enable_xla)
```
haoyuz: Please define a `drop_remainder` boolean (instead of directly using `flags_obj.enable_xla`); and for that variable add a comment explaining why drop_remainder and XLA are related.

official/resnet/keras/keras_imagenet_main.py
```diff
@@ -166,11 +169,19 @@ def run(flags_obj):
       optimizer = tf.keras.mixed_precision.experimental.LossScaleOptimizer(
           optimizer, loss_scale=flags_core.get_loss_scale(flags_obj))
 
+    if flags_obj.enable_xla:
+      per_replica_batch_size = (
+          flags_obj.batch_size // strategy.num_replicas_in_sync)
```
haoyuz: Maybe use `flags_obj.num_gpus`?

official/resnet/keras/keras_imagenet_main.py
rxsang: Done.

official/resnet/keras/keras_imagenet_main.py
rxsang: strategy.num_replicas_in_sync is more general, especially this code may run on other devices in the future like TPU.

haoyuz: It will break the tests with no distribution strategy (get_distribution_strategy can return None). Also we'll have to make a lot of other changes for this to work with other devices.

rxsang: Makes sense. `num_gpus` doesn't cover all the cases as well because you may run the CPU XLA program. I made a change to support non-XLA case as well, PTAL.

rxsang:
Ensure static shapes when enabling XLA in Resnet Keras model.

PR #6522, opened 2019-04-04
TypeError: 'dict_keys' object does not support indexing

lstsq:
When I try to train the `ssd_mobilenet_v1_coco` downloaded from object_detection model zoo,
with `samples/configs/ssd_mobilenet_v1_ppn_shared_box_predictor_300x300_coco14_sync.config`.
There occured an TypeError as mentioned above, so I use list() function to wrap the dict_keys object of the following line in `models/feature_map_generators.py`:
```python
image_features = image_features[image_features.keys()[0]]           # problem solved !
```

Then I think, maybe there are some other similar problem which not have been fixed, so I use `grep` recursively search them and fix them. Hopefully, I have fixed most of that.

### DIFF
--- a/research/object_detection/exporter.py
+++ b/research/object_detection/exporter.py
@@ -402,7 +402,7 @@ def _export_inference_graph(input_type,
       f.write(str(inference_graph_def))
 
   if additional_output_tensor_names is not None:
-    output_node_names = ','.join(outputs.keys()+additional_output_tensor_names)
+    output_node_names = ','.join(list(outputs.keys())+additional_output_tensor_names)
   else:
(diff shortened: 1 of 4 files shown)


xunnv:
I signed it!

LongZhai:
I am still having the problem after the update and I am using python 3.7

image_features = image_features[image_features.keys()[0]]
TypeError: 'dict_keys' object does not support indexing

NPetsky:
@LongZhai you have to change the line to image_features = list(image_features.values())[0] then it works

aniketbote:
These changes need to be merged.

PR #6528, opened 2019-04-04
Add profiler callback for Keras models

haoyuz:


official/resnet/keras/keras_imagenet_main.py
```diff
@@ -214,15 +207,12 @@ def run(flags_obj):
...
-  if flags_obj.enable_e2e_xprof:
-    results = profiler.stop()
-    profiler.save(flags_obj.model_dir, results)
-
   eval_output = None
   if not flags_obj.skip_eval:
     eval_output = model.evaluate(eval_input_dataset,
                                  steps=num_eval_steps,
                                  verbose=2)
+  time_callback = callbacks[0]
```
tfboyd: What do you think about a helper to get a specific call back by type as part of common instead of counting on the index?  Then the method can return value not found or something.  Assuming the first callback is time_callback feels risky.  

The rest looks like an improvement to me.

haoyuz: Great suggestion! Fixed in the following commit.

official/resnet/keras/keras_common.py
```diff
@@ -143,27 +167,53 @@ def get_optimizer():
...
       num_images=num_images)
+  callbacks = [time_callback, lr_callback]
 
-  return time_callback, tensorboard_callback, lr_callback
+  if FLAGS.enable_tensorboard:
+    tensorboard_callback = tf.keras.callbacks.TensorBoard(
+        log_dir=FLAGS.model_dir)
+    callbacks.append(tensorboard_callback)
 
+  if FLAGS.profile_steps:
```
qqfish: Please make sure it does not include the step profiled by tensorboard callback.

haoyuz: Good point. Since the TensorBoard profile step is customizable and it's not exposed by the class, I added a warning message when both callbacks are used. Observed that if they overlap, only the first one can start the profiler.

qqfish: SGTM

PR #6546, opened 2019-04-08
Added main function to the two keras resent benchmarks

seemuch:
...for the purpose of using it as internal benchmarks as well.


seemuch:
This PR has a problem that tensorflow is not imported. 
I will fix it in https://github.com/tensorflow/models/pull/6447

PR #6548, opened 2019-04-09
Move metrics info from extras to metrics field in test_log.proto

lindong28:


lindong28:
This PR can be merged after cl/242589492 is built into tf-nightly and tf-nightly-gpu pip package.

PR #6558, opened 2019-04-10
Ensure static shapes when enabling XLA in Resnet Keras model in graph mode.

rxsang:
Reverts tensorflow/models#6517


rxsang:
We only enable static shapes in graph mode because eager mode will cause OOM.

Issue #5033, opened 2018-08-08
Syntax error: import matplotlibnmatplotlib.use('Agg')nimport matplotlib.pyplot as plt


ericlevine:
### System information
- **What is the top-level directory of the model you are using**: models/research
- **Have I written custom code (as opposed to using a stock example script provided in TensorFlow)**: No
- **OS Platform and Distribution (e.g., Linux Ubuntu 16.04)**: Mac OSX
- **TensorFlow installed from (source or binary)**: binary
- **TensorFlow version (use command below)**: ('v1.3.0-rc2-20-g0787eee', '1.3.0')
- **Bazel version (if compiling from source)**: n/a
- **CUDA/cuDNN version**: n/a
- **GPU model and memory**: n/a
- **Exact command to reproduce**: n/a

### Describe the problem

On Mac OSX, create_pycocotools_package.sh that's part of the object detection [running on cloud](https://github.com/tensorflow/models/blob/master/research/object_detection/g3doc/running_on_cloud.md) instructions uses `sed` to change the import statements in `coco.py`. It inserts newlines, but that functionality doesn't work for OSX because of [quirks in the way sed works on OSX](https://stackoverflow.com/questions/24275070/sed-not-giving-me-correct-substitute-operation-for-newline-with-mac-difference). I've addressed the problem locally by changing the replacement from newlines to semicolons.

The error that I get when I try to run the job looks like this:

`Syntax error: import matplotlibnmatplotlib.use('Agg')nimport matplotlib.pyplot as plt`

### Source code / logs

Here's the specific line that I changed locally to make it work:

    -sed "s/import matplotlib\.pyplot as plt/import matplotlib\nmatplotlib\.use\(\'Agg\'\)\nimport matplotlib\.pyplot as plt/g" pycocotools/coco.py > coco.py.updated
    +sed "s/import matplotlib\.pyplot as plt/import matplotlib; matplotlib\.use\(\'Agg\'\); import matplotlib\.pyplot as plt/g" pycocotools/coco.py > coco.py.updated

rgalbo:
This is a known issue with `sed` installed on [OSX](https://superuser.com/questions/307165/newlines-in-sed-on-mac-os-x) in [multiple](https://unix.stackexchange.com/questions/13711/differences-between-sed-on-mac-osx-and-other-standard-sed) forms. I would push for the hotfix of adding sem-colons to make this work for both linux and OSX for the time being.

aysark:
Just ran into this...

[...]

PR #6563, opened 2019-04-11
 Fix newline issue with sed on macOS

aysark:
Addresses #5033.

There was a PR for this previously but user didn't agree to cla it seems https://github.com/tensorflow/models/pull/5115


aysark:
@pkulzc @jch1 @derekjchow @dreamdragon can you review?

pkulzc:
Sorry for the late response as most of us are traveling for some conferences recently.

PR #6572, opened 2019-04-14
reshape only applies to numpy arrays, it was however being used again…

Koff:
`eat_reward` is a Pandas DataFrame, as such it can't be `reshape` without first being converted to a Numpy array.


jaeyounkim:
Fixed

PR #6586, opened 2019-04-16
fix bug in research/deep_speech model when is_bidirectional == False

xysmlx:
When is_bidirectional == False, the original version passes the output tuple (output tensor and state tensor) of unidirectional RNN to the rnn_outputs. However, it should pass the exact output tensor instead of the tuple to the rnn_outputs.


xysmlx:
I signed it!

jaeyounkim:
deep_speech is deprecated.

PR #6589, opened 2019-04-16
Add 2-GPU benchmarks for NCF

seemuch:


robieta:
Presumably you'll need to double batch size as well?

seemuch:
I am not sure if I want to double the batch_size. Because the batch_size specified by the flag is not the real batch size. The real batch size is the number of replicas.

PR #6591, opened 2019-04-16
Remove contrib imports, or move them inline

seemuch:
...and other stuff to make the model runs in TF 2.0


official/recommendation/data_pipeline.py
```diff
@@ -261,6 +261,7 @@ def get_dataset(self, batch_size, epochs_between_evals):
 
       file_pattern = os.path.join(
           epoch_data_dir, rconst.SHARD_TEMPLATE.format("*"))
+      from tensorflow.contrib.tpu.python.tpu.datasets import StreamingFilesDataset
```
guptapriya: Can you leave a TODO describing why this is the case? For either you or someone in the TPU team to fix later.

official/recommendation/ncf_common.py
```diff
@@ -110,10 +110,10 @@ def parse_flags(flags_obj):
 
 
 def get_optimizer(params):
-  optimizer = tf.train.AdamOptimizer(
```
guptapriya: fyi, i have heard there are slight numerical differences between the old and new adam optimizers. something to be aware of as you compare convergence etc.

official/recommendation/neumf_model.py
```diff
@@ -381,8 +381,8 @@ def compute_eval_loss_and_metrics_helper(logits,              # type: tf.Tensor
   # ignore padded examples
   example_weights *= tf.cast(expanded_metric_weights, tf.float32)
 
-  cross_entropy = tf.losses.sparse_softmax_cross_entropy(
-      logits=softmax_logits, labels=eval_labels, weights=example_weights)
+  cross_entropy = tf.keras.losses.sparse_categorical_crossentropy(
```
guptapriya: this change does not seem equivalent. the new one is not using weights. and also i think the original one returns a reduced loss and new one returns per example losses? Does this run?

official/recommendation/neumf_model.py
```diff
@@ -443,7 +443,8 @@ def compute_top_k_and_ndcg(logits,              # type: tf.Tensor
   position_vector = tf.reduce_sum(sparse_positions, axis=1)
 
   in_top_k = tf.cast(tf.less(position_vector, rconst.TOP_K), tf.float32)
-  ndcg = tf.log(2.) / tf.log(tf.cast(position_vector, tf.float32) + 2)
+  ndcg = tf.keras.backend.log(2.) / tf.keras.backend.log(
```
guptapriya: maybe https://www.tensorflow.org/versions/r2.0/api_docs/python/tf/math/log is a better replacement?

official/recommendation/neumf_model.py
seemuch: It does, and it seems not to affect the convergence number, weirdly. But the number does fluctuate, so it is hard to tell sometimes. 
Anyway, I changed it to include the example_weights.

seemuch:
We do not have unit tests for the keras model.

official/recommendation/data_pipeline.py
guptapriya: Can we add a username to the TODO, as well as add an explanation of why do it this way?

official/recommendation/ncf_keras_main.py
```diff
@@ -155,13 +157,13 @@ def _get_keras_model(params):
   # is designed to be of batch_size 1 for each replica.
   user_input = tf.keras.layers.Input(
       shape=(batch_size,),
-      batch_size=1,
+      batch_size=params["batches_per_step"],
```
guptapriya: hm, why is this change in this PR? Or am i seeing a bad merge result or something?

official/recommendation/ncf_common.py
```diff
@@ -109,13 +109,15 @@ def parse_flags(flags_obj):
   }
 
 
-def get_optimizer(params):
-  optimizer = tf.train.AdamOptimizer(
+def get_keras_optimizer(params):
```
guptapriya: do you need a helper method if this is used in only one place?

official/recommendation/neumf_model.py
```diff
@@ -109,7 +109,11 @@ def neumf_model_fn(features, labels, mode, params):
     mlperf_helper.ncf_print(key=mlperf_helper.TAGS.OPT_HP_ADAM_EPSILON,
                             value=params["epsilon"])
 
-    optimizer = ncf_common.get_optimizer(params)
+    optimizer = tf.train.AdamOptimizer(
```
guptapriya: `tf.train.AdamOptimizer` is not a valid 2.0 symbol, right? Does this code run in 2.0 like this? 
https://www.tensorflow.org/versions/r2.0/api_docs/python/tf/train

official/recommendation/ncf_test.py
```diff
@@ -215,7 +216,24 @@ def test_end_to_end_keras_mlperf(self):
          '-distribution_strategy', 'off',
          '-log_steps', '100'])
 
+  @mock.patch.object(rconst, "SYNTHETIC_BATCHES_PER_EPOCH", 100)
+  def test_end_to_end_keras_1_gpu(self):
```
guptapriya: you should probably skip the test if there are no GPUs available. otherwise it will just run with CPU i believe

official/recommendation/ncf_test.py
```diff
@@ -215,7 +216,24 @@ def test_end_to_end_keras_mlperf(self):
...
+    integration.run_synthetic(
+        ncf_keras_main.main, tmp_root=self.get_temp_dir(), max_train=None,
+        extra_flags=self._BASE_END_TO_END_FLAGS +
+        ['-num_gpus', '1', '-log_steps', '100'])
```
guptapriya: what's the log_steps used for ?

official/recommendation/neumf_model.py
```diff
@@ -443,7 +448,8 @@ def compute_top_k_and_ndcg(logits,              # type: tf.Tensor
   position_vector = tf.reduce_sum(sparse_positions, axis=1)
 
   in_top_k = tf.cast(tf.less(position_vector, rconst.TOP_K), tf.float32)
-  ndcg = tf.log(2.) / tf.log(tf.cast(position_vector, tf.float32) + 2)
+  ndcg = tf.math.log(2.) / tf.keras.backend.log(
```
guptapriya: what about the other `log` call in this expression?

official/recommendation/ncf_keras_main.py
seemuch: No, it is because of cl/243685134. Now we are supposed to have global batch_size in the input layers.

official/recommendation/neumf_model.py
seemuch: This is in neumf_model_fn, which is only supposed to be used in the estimator case.

official/recommendation/neumf_model.py
seemuch: Oops. Done.

official/recommendation/ncf_common.py
seemuch: There is still an if logic in the method, so I think we could keep the method. 
But it is used in keras, so I move it out of ncf_common to ncf_keras_main.

official/recommendation/ncf_test.py
seemuch: It is used for the callback that is useful for benchmark. 
I don't think it is useful here.
Deleted.

official/recommendation/ncf_common.py
guptapriya: the else part seems irrelevant for the keras case anyway - it is for TPUEstimator. so i dont see why the method is helpful in your codepath

official/recommendation/neumf_model.py
guptapriya: Does that mean you're saying the estimator code path will not be run in 2.0? I think we should just change this to tf.compat.v1 at least, so that it will run in both 1.x and 2.0, even with the estimator codepath.

official/recommendation/neumf_model.py
```diff
@@ -381,8 +385,9 @@ def compute_eval_loss_and_metrics_helper(logits,              # type: tf.Tensor
   # ignore padded examples
   example_weights *= tf.cast(expanded_metric_weights, tf.float32)
 
-  cross_entropy = tf.losses.sparse_softmax_cross_entropy(
-      logits=softmax_logits, labels=eval_labels, weights=example_weights)
+  cce = tf.keras.losses.SparseCategoricalCrossentropy(from_logits=True)
```
guptapriya: is this in the estimator code path or the new keras path? if former, i think maybe leave it as tf.compat.v1.losses.sparse_softmax_cross_entropy unless you plan to verify convergence with the old code path

official/recommendation/ncf_keras_main.py
```diff
@@ -189,11 +191,25 @@ def _get_keras_model(params):
...
 
 
+def _get_keras_optimizer(params):
+  """Returns the optimizer used by the keras model."""
+  optimizer = tf.keras.optimizers.Adam(
+      learning_rate=params["learning_rate"],
+      beta_1=params["beta1"],
+      beta_2=params["beta2"],
+      epsilon=params["epsilon"])
+  if params["use_tpu"]:
```
guptapriya: I dont think this else clause is useful in the keras codepath yet. when we add TPUStrategy to this codepath, even then it would not need any such wrapping. You probably need to move this else clause to the estimator path's model_fn which is the route that TPUEstimator takes

official/recommendation/data_pipeline.py
guptapriya: i dont see how this was resolved, the TODO looks the same as before to me.

official/recommendation/neumf_model.py
seemuch: Done.

official/recommendation/data_pipeline.py
seemuch: Must have missed this comment. Added my user name to the TODOs.

PR #6592, opened 2019-04-16
Adds original mobilenetv2 keras implementation

JonathanCMitchell:
Adds original MobileNetV2 keras implementation. Needs different imports.


JonathanCMitchell:
I signed it!

PR #6604, opened 2019-04-18
fix training crash with Unicode labels

idrissbellil:
The training seems to crash on both Python 2 and Python 3 with the presence of Unicode characters in the labels.


idrissbellil:
I signed it!

PR #6633, opened 2019-04-22
Ncf metric tweaks

tfboyd:
Changed to the following:

- exp_per_second
- hr_at_10


tfboyd:
Extra commit with the cifar-10 change is being merged now.  it will end up as a no op in few minutes.

PR #6635, opened 2019-04-22
Set static shapes in all XLA cases.

rxsang:


tfboyd:
PR is on hold waiting for XLA perf improvement.

tfboyd:
Closed, we can always repopen.  or submit again.

PR #6637, opened 2019-04-23
Fix some box realated files

Zhangxuri:
To edit these file, I can finally  use box model. But there still missing d2r_frcnn_20190411 .


Zhangxuri:
> I signed it!

andrefaraujo:
Hi @Zhangxuri, it seems like this has been fixed already, two days after your PR. So I believe this is not an issue any more, but please re-open in case you face issues.

PR #6664, opened 2019-04-24
Added none check for output_dir

seemuch:


official/resnet/keras/keras_benchmark.py
```diff
@@ -36,6 +36,9 @@ def __init__(self, output_dir=None, default_flags=None, flag_methods=None):
     self.default_flags = default_flags or {}
     self.flag_methods = flag_methods or {}
 
+    if not output_dir:
+      output_dir = "/tmp/"
```
tfboyd: single quotes please.

tfboyd: Go ahead and just set output_dir = /tmp instead of none in the __init__

seemuch: Done.

PR #6673, opened 2019-04-25
Remove contrib cross device ops and update all_reduce_alg options.

dubey:


tfboyd:
Talking with Ayush recording questions here so I do not forget:
So I can pass None and that is ok as the all_reduce_algo because None is in the dictionary.
and cross_device_ops=None is acceptable?

tfboyd:
None will result in automatic choice by the system.  Currently nccl unless dgx-1 then hierarchical copy.  This needs changed but is a different issue.  nccl should be the only algo used.

PR #6675, opened 2019-04-25
Add benchmarks with the --cloning flag to Resnet and NFC.

isaprykin:
This PR adds essential correctness and performance benchmarks for the new --cloning=False code path.  The feature was added https://github.com/tensorflow/tensorflow/commit/22962228c29fe454522ee39e2154f0703f981e3b and probably haven't made its way into tf-nightly yet.  Quite a bit of the verification happened in that feature PR.  This PR hopefully represents a trivial change that still needs to be tested as part of the performance framework.  I will do that tomorrow once tf-nightly is ready.

### DIFF
--- a/official/recommendation/ncf_common.py
+++ b/official/recommendation/ncf_common.py
@@ -104,7 +104,7 @@ def parse_flags(flags_obj):
       "epsilon": flags_obj.epsilon,
       "match_mlperf": flags_obj.ml_perf,
       "use_xla_for_gpu": flags_obj.use_xla_for_gpu,
-      "cloning": flags_obj.cloning,
+      "clone_model_in_keras_dist_strat": flags_obj.clone_model_in_keras_dist_strat,
       "epochs_between_evals": FLAGS.epochs_between_evals,
       "turn_off_distribution_strategy": FLAGS.turn_off_distribution_strategy,
   }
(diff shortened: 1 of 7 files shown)


official/recommendation/ncf_common.py
```diff
@@ -312,6 +313,11 @@ def eval_size_check(eval_batch_size):
   def xla_validator(flag_dict):
     return not flag_dict["use_xla_for_gpu"] or not flag_dict["tpu"]
 
+  flags.DEFINE_bool(
+      name="cloning", default=True, help=flags_core.help_wrap(
```
guptapriya: can we call it something more explicit like "clone_model_in_keras_dist_strat"?

tfboyd: Actually we should/need to move this option into the "global" flags because it would/could also be used with transformer and is already used for NCF and ResNet.  You just need to add it to the flag options under /utils and then each model indicates it is a valid flag for their situation.  I can help tomorrow, this is a late night drive-by.  

To fix this correctly is bit of work that would be appreciated.  You need to move this [flag logic](https://github.com/tensorflow/models/blob/master/official/resnet/resnet_run_loop.py#L709)  out of run_loop and into something like "resnet_common" so it is not so tightly associated with the estimator version and have it include everything that is mostly common for the resnet models (keras and estimator).  The other option is to just copy everything you need for keras into keras_common.define_flags.  I would be ok with the `common_resnet` was in the keras folder.  

I am all for a more descriptive name, I agree now that cloning is to vague.  Priya's name is also a bit long, maybe we can ignore tf 1.0 and call it dist_strat_clone_model. 

Sorry this got long, I tried to clean up my wording.  It is worth it to take 2 hours or so can clean this up before it gets more messy.

isaprykin: I renamed the flag.

We clarified offline the temporary nature of the flag and decide to leave the current flag plumbing as is.

isaprykin:
I ran all the benchmarks and tests that I added.  They all succeeded, at least as defined by their status and by my examining of the logs.  You can see the links at http://<internal-paste-service>/6576106209542144.

I don't understand why the lint error still there.  I hope that posting this comment will make the linter re-run.

PR #6676, opened 2019-04-25
 Add num_packs flag for MirroredStrategy's cross device ops.

dubey:


dubey:
Tested with `all_reduce_alg=nccl` and 2 different values of `num_packs` to 100 steps.

I'm looking into lint error.

official/utils/misc/distribution_utils.py
```diff
@@ -111,7 +110,8 @@ def get_distribution_strategy(distribution_strategy="default",
               all_reduce_alg))
     return tf.distribute.MirroredStrategy(
         devices=devices,
-        cross_device_ops=_MIRRORED_ALL_REDUCE_OPTIONS[all_reduce_alg])
+        cross_device_ops=_MIRRORED_ALL_REDUCE_OPTIONS[all_reduce_alg](
+            num_packs=num_packs if num_packs is not None else 1))
```
guptapriya: would it make sense to just make the default for the flag as 1 instead of None?

official/utils/misc/distribution_utils.py
```diff
@@ -111,7 +110,8 @@ def get_distribution_strategy(distribution_strategy="default",
               all_reduce_alg))
     return tf.distribute.MirroredStrategy(
         devices=devices,
-        cross_device_ops=_MIRRORED_ALL_REDUCE_OPTIONS[all_reduce_alg])
+        cross_device_ops=_MIRRORED_ALL_REDUCE_OPTIONS[all_reduce_alg](
```
guptapriya: wouldn't this fail is all_reduce_alg = None?

official/utils/misc/distribution_utils.py
dubey: Done.

official/utils/misc/distribution_utils.py
dubey: I think you reviewed an outdated snapshot.  I refactored some of the changes while trying to remove some lint, PTAL.

tfboyd:
Thanks for adding num_packs, it is worth having since there is no perfect value.

PR #6697, opened 2019-04-29
Change official owners to Priya and Toby for now.

tfboyd:
@guptapriya


tfboyd:
@karmel   18 days

karmel:
Option 2 that you might prefer is that you switch the garden team to be the relevant people.

tfboyd:
I am going to merge this and then figure out the groups next.  I know for the next few weeks Priya and I will ensure everything is covered and we want to be fast bottlenecks.  But long-term we need to alert a larger group or the model owner.  This was to end having to remove Karmel from each PR and then add someone to review it.

PR #6699, opened 2019-04-30
Update oid_od_challenge_evaluation_utils.py

ManWingloeng:
```
  data_boxes = data[data.ConfidenceImageLabel.isnull()]
  data_labels = data[data.XMin.isnull()]
```
If this code want to get the data_boxes which is not null, u should use this follow:
```
  data_boxes = data[~data.ConfidenceImageLabel.isnull()]
  data_labels = data[~data.XMin.isnull()]
```


ManWingloeng:
```I signed it!``` :+1:

ManWingloeng:
I find my thought is wrong.Sorry about that.

PR #6703, opened 2019-05-01
Add --fp16_implementation option.

reedwm:
This options allows the new tf.train.experimental.enable_mixed_precision_graph_rewrite() function to be used for fp16, instead of manual casts.

To test convergence, I ran
```
python imagenet_main.py --num_gpus=8 --data_dir=/data/imagenet/imagenet --batch_size=1024 --train_epochs=90 --epochs_between_evals=10 --model_dir ~/model_dir_amp --dtype=fp16 --hooks=ExamplesPerSecondHook --fp16_implementation=graph_rewrite
```

I got 76.26% accuracy.

To test performance, I ran

```
python imagenet_main.py --clean --synth --num_gpus=8 --batch_size=1024 --train_epochs=1 --model_dir ~/model_dir_test --dtype=fp16 --hooks=ExamplesPerSecondHook,ProfilerHook
```
I added --fp16_implementation=graph_rewrite to use enable_mixed_precision_graph_rewrite(), and I removed --synth and added --data_dir when using real data. Results (images/sec):

| |casting|graph_rewrite|
|--|--|--|
|Real data|5104|4005|
|Synthetic data|4682|4522|

On synthetic data, the graph rewrite is only 3.5% slower than manual casting. But on real data, it's 27% slower. And the fact that real data is faster than synthetic data with manual casting is very unusual.

### DIFF
--- a/official/utils/flags/_performance.py
+++ b/official/utils/flags/_performance.py
@@ -184,6 +184,7 @@ def define_performance(num_parallel_calls=True, inter_op=True, intra_op=True,
       @flags.multi_flags_validator(['fp16_implementation', 'dtype',
                                     'loss_scale'])
       def _check_fp16_implementation(flags_dict):
+        """Validator to check fp16_implementation flag is valid."""
         if (flags_dict['fp16_implementation'] == 'graph_rewrite' and
             flags_dict['dtype'] != 'fp16'):
           raise flags.ValidationError('--fp16_implementation should not be '
--- a/official/utils/flags/_performance.py
+++ b/official/utils/flags/_performance.py
@@ -171,14 +171,15 @@ def define_performance(num_parallel_calls=True, inter_op=True, intra_op=True,
 
     if fp16_implementation:
       # Currently, this flag is only defined for the estimator resnet model.
-      flags.DEFINE_enum(name="fp16_implementation", default='casting',
-                        enum_values=('casting', 'graph_rewrite'),
-                        help=help_wrap(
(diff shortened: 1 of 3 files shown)


tfboyd:
@guptapriya   Take a look with me please.  :-)

official/utils/flags/_performance.py
```diff
@@ -164,6 +169,30 @@ def _check_loss_scale(loss_scale):  # pylint: disable=unused-variable
 
       return loss_scale > 0
 
+    if fp16_implementation:
```
guptapriya: which code paths does this flag effect? estimator + resnet? doesn't seem to change the keras path yet right? should we clarify?

official/utils/flags/_performance.py
```diff
@@ -34,6 +34,10 @@
 
 
 def get_tf_dtype(flags_obj):
```
guptapriya: where all is this API used? Are all those codepaths also using the new API to support fp16?

reedwm: This is used in the estimator and keras models. Since --fp16_implementation is only defined for the estimator model, this change has no impact on the keras model.

official/utils/flags/_performance.py
reedwm: Only affects estimator for now. Added comment to clarify.

reedwm:
The failing tests fails even on master, so I'm merging.

PR #6710, opened 2019-05-02
Enable CuDNN BatchNorm spatial persistent by default

haoyuz:
* Enable CuDNN BatchNorm spatial persistent by default

* Consistently set `scale=False` and `fused=True` in all the batchnorm layers.

* Also, replace the zero padding layer with a padding attribute in max pooling layer. Zero padding adds both forward and backprop (Split) ops, each taking 1.5--2ms. Now the overhead is removed.

### DIFF
--- a/official/resnet/keras/keras_common.py
+++ b/official/resnet/keras/keras_common.py
@@ -291,9 +291,13 @@ def define_keras_flags():
       'help improve performance using EagerIterator and function. The codepath '
       'when enabling this feature is experimental and will be removed once the '
       'corresponding performance features are fully supported in TensorFlow.')
-  flags.DEFINE_boolean(name='clone_model_in_keras_dist_strat', default=True,
-                       help='If False, then the experimental code path is used'
-                       ' that doesn\'t clone models for distribution.')
+  flags.DEFINE_boolean(
+      name='batchnorm_spatial_persistent', default=True,
+      help='Enable the spacial persistent mode for CuDNN batch norm kernel.')
+  flags.DEFINE_boolean(
+      name='clone_model_in_keras_dist_strat', default=True,
+      help='If False, then the experimental code path is used that doesn\'t '
+           'clone models for distribution.')
 
 
 def get_synth_input_fn(height, width, num_channels, num_classes,
@@ -356,6 +360,15 @@ def data_prefetch_with_slack():
   """Use unstable code for perf tuning purposes."""
   if not FLAGS.use_synthetic_data:
     _monkey_patch_org_create_device_dataset()
+
+
+def set_cudnn_batchnorm_mode():
+  """Set CuDNN batchnorm mode for better performance. Note that the spatial
+     persistent mode may lead to accuracy losses for certain models."""
+  if FLAGS.batchnorm_spatial_persistent:
+    os.environ['TF_USE_CUDNN_BATCHNORM_SPATIAL_PERSISTENT'] = '1'
+  else:
+    os.environ.pop('TF_USE_CUDNN_BATCHNORM_SPATIAL_PERSISTENT', None)
 
 
 def _monkey_patch_org_assert_broadcastable():
--- a/official/resnet/keras/keras_common.py
+++ b/official/resnet/keras/keras_common.py
@@ -368,7 +368,7 @@ def set_cudnn_batchnorm_mode():
   if FLAGS.batchnorm_spatial_persistent:
     os.environ['TF_USE_CUDNN_BATCHNORM_SPATIAL_PERSISTENT'] = '1'
   else:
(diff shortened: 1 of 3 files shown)


official/resnet/keras/resnet_model.py
```diff
@@ -204,12 +218,13 @@ def resnet50(num_classes, dtype='float32', batch_size=None):
                     kernel_regularizer=regularizers.l2(L2_WEIGHT_DECAY),
                     name='conv1')(x)
   x = layers.BatchNormalization(axis=bn_axis,
+                                scale=False,
                                 momentum=BATCH_NORM_DECAY,
                                 epsilon=BATCH_NORM_EPSILON,
+                                fused=True,
                                 name='bn_conv1')(x)
   x = layers.Activation('relu')(x)
-  x = layers.ZeroPadding2D(padding=(1, 1), name='pool1_pad')(x)
```
yuefengz: Is there any quality drop after removing the zero padding?

official/resnet/keras/keras_common.py
```diff
@@ -358,6 +362,15 @@ def data_prefetch_with_slack():
     _monkey_patch_org_create_device_dataset()
 
 
+def set_cudnn_batchnorm_mode():
+  """Set CuDNN batchnorm mode for better performance. Note that the spatial
+     persistent mode may lead to accuracy losses for certain models."""
+  if FLAGS.batchnorm_spatial_persistent:
+    os.environ['TF_USE_CUDNN_BATCHNORM_SPATIAL_PERSISTENT'] = '1'
+  else:
+    os.environ.pop('TF_USE_CUDNN_BATCHNORM_SPATIAL_PERSISTENT', None)
```
tfboyd: My understanding is that this is the best way to remove an ENV VAR.

`del os.environ['MYVAR']`

official/resnet/keras/resnet_model.py
```diff
@@ -69,8 +69,10 @@ def identity_block(input_tensor, kernel_size, filters, stage, block):
                     kernel_regularizer=regularizers.l2(L2_WEIGHT_DECAY),
                     name=conv_name_base + '2a')(input_tensor)
   x = layers.BatchNormalization(axis=bn_axis,
+                                scale=False,
                                 momentum=BATCH_NORM_DECAY,
                                 epsilon=BATCH_NORM_EPSILON,
+                                fused=True,
```
tfboyd: fused is default True.  I see below it gets confusing. My concern with setting it is that people then think you need to worry about this when by default it is True.  I am ok with it, but I wanted to understand and a little digging made me more confused. Even if you say `True` it will still treat it as if you set it to none because it has to pass `_raise_if_fused_cannot_be_used`

The document says

> fused: if None or True, use a faster, fused implementation if possible. If False, use the system recommended implementation.

I guess why is this now different in V2. I think this comment should be be higher up to indicate it still "may" not use fused if the input rank is not 4.
```python
if self._USE_V2_BEHAVIOR:
      if fused:
        self._raise_if_fused_cannot_be_used()
      # We leave fused as None if self._fused_can_be_used()==True, since we
      # still may set it to False in self.build() if the input rank is not 4.
      elif fused is None and not self._fused_can_be_used():
        fused = False
    elif fused is None:
      fused = True
    self.supports_masking = True

```

official/resnet/keras/resnet_model.py
tfboyd: I am kicking off a run now and will link the kokoro job so we can wait and see.  easy to do.

haoyuz: Great thanks! Note that zero padding is not removed, it's just fused into the followup max pooling layer.

official/resnet/keras/resnet_model.py
haoyuz: I see. I knew that the underlying op is `FusedBatchNorm` by default but this logic here successfully confused me :) Will get the `fused=True` removed.

official/resnet/keras/keras_common.py
haoyuz: Got it will make the change.

official/resnet/keras/resnet_model.py
tfboyd: I am still confused but for V2 it looks like even if you set to True it can still flip back to false, but if you Fused=True then instead of silently it should raise an error.  I say should as looking at self.build I am not totally sure if it does not "secretly" flip it to false.  I also don't get why V2 is different unless virtual batch is a new thing.  If removing fused gets you the desired result we should drop it.

official/resnet/keras/resnet_model.py
haoyuz: Thanks for checking the accuracy! It's indeed hurting the accuracy, I will look into this.

For now I reverted the zero padding change. Finished a convergence test and got 76.09% which is probably acceptable. We will closely watch the accuracy numbers these days.

PR #6721, opened 2019-05-06
Fix ResNet model convergence problem

haoyuz:
* Remove `scale=False` in batchnorm layers for better accuracy;
* Fuse zero padding and maxpool into one layer;
* Avoid throwing errors when unsetting non-existing environment variable.

Verified that with this change the model converges to 76.27% in 90 epochs.


guptapriya:
Does the resnet TPU model need the same change?

PR #6725, opened 2019-05-07
Support fp16 in estimator NCF model

reedwm:
fp16 is added with the new `tf.train.experimental.enable_mixed_precision_graph_rewrite` function.

Note: I moved the `get_loss_scale` function from _performance.py to resnet_run_loop.py, and added a similar function to the NCF model. This is because the default loss scale should be different for each model.

To test convergence, I ran:

```
python ncf_estimator_main.py --model_dir ~/ncf_model_dir --data_dir ~/ncf_data_dir --dataset=ml-20m --hooks= --num_gpus=1 --clean --train_epochs=14 --batch_size=98340 --learning_rate=0.00382059 --beta1=0.783529 --beta2=0.909003 --epsilon=1.45439e-07 --layers=256,256,128,64 --num_factors=64 --hr_threshold=0.635 --ml_perf --dtype=fp16
```
I got an HR of 0.6371 on epoch 8.

Performance gain is modest: From about 67 steps/sec to 75 steps/sec with a batch size of 98340 on a V100.

### DIFF
--- a/official/recommendation/ncf_common.py
+++ b/official/recommendation/ncf_common.py
@@ -108,6 +108,8 @@ def parse_flags(flags_obj):
           flags_obj.clone_model_in_keras_dist_strat,
       "epochs_between_evals": FLAGS.epochs_between_evals,
       "turn_off_distribution_strategy": FLAGS.turn_off_distribution_strategy,
+      "dtype": flags_obj.dtype,
+      "loss_scale": flags_obj.loss_scale,
   }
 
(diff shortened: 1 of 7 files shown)


guptapriya:
Hey @reedwm sorry i missed this PR. Should I review it now or you plan to update it?

reedwm:
I'll update it once I get time. Closing for now until I update it.

PR #6728, opened 2019-05-08
Update mobilenet_example.ipynb

okdshin:


okdshin:
I signed it!

sguada:
Can you provide a description for this PR, could you explain why this change is necessary?

okdshin:
It occurs an error as `base_name` is not defined. `checkpoint_name` is defined, but it is not used later. I can run it by replacing `checkpoint_name` with `base_name`.

marksandler2:
I think this got fixed in the head. Thank you for your pull request.

PR #6734, opened 2019-05-08
Transformer instrumented for benchmarking

tfboyd:
Transformer instrumented for benchmarking along with:

- Fixed deprecation issue with gfile
- Added benchmark for 1 GPU accuracy for base model and 1 GPU benchmark base model.  Accuracy cannot reach SOTA due to small batch-size.

### DIFF
--- a/official/transformer/transformer_estimator_benchmark.py
+++ b/official/transformer/transformer_estimator_benchmark.py
@@ -180,8 +180,7 @@ class TransformerBaseEstimatorAccuracy(EstimatorBenchmark):
     FLAGS.batch_size = 4048 * 2
     FLAGS.train_steps = 100000
     FLAGS.steps_between_evals = 5000
-    FLAGS.model_dir = self._get_model_dir(
-        'benchmark_1_gpu')
+    FLAGS.model_dir = self._get_model_dir('benchmark_2_gpu')
     FLAGS.hooks = ['ExamplesPerSecondHook']
     self._run_and_report_benchmark()
 
@@ -198,8 +197,7 @@ class TransformerBaseEstimatorAccuracy(EstimatorBenchmark):
     FLAGS.batch_size = 4048 * 8
     FLAGS.train_steps = 100000
     FLAGS.steps_between_evals = 5000
-    FLAGS.model_dir = self._get_model_dir(
-        'benchmark_8_gpu')
+    FLAGS.model_dir = self._get_model_dir('benchmark_8_gpu')
     FLAGS.hooks = ['ExamplesPerSecondHook']
     self._run_and_report_benchmark()
 
--- a/official/transformer/transformer_estimator_benchmark.py
+++ b/official/transformer/transformer_estimator_benchmark.py
@@ -186,12 +186,7 @@ class TransformerBaseEstimatorAccuracy(EstimatorBenchmark):
     self._run_and_report_benchmark()
 
   def benchmark_8_gpu(self):
-    """Benchmark 2 gpu.
-
-      The paper uses 8 GPUs and a much larger effective batch-size,
-      this is unlikely to hit the target bleu score regardless of
-      number of steps.
-    """
+    """benchmark 8 gpus."""
     self._setup()
     FLAGS.num_gpus = 8
     FLAGS.data_dir = self.train_data_dir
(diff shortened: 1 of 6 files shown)


official/transformer/transformer_main.py
```diff
@@ -279,6 +279,11 @@ def run_loop(
     bleu_threshold: minimum BLEU score before training is stopped.
     vocab_file: Path to vocab file that will be used to subtokenize bleu_source.
 
+  Returns:
+    Dict of results of the run.  Contains the keys `eval_results`, `train_hooks`
+      ,`bleu_cased`, and `bleu_uncased`. `train_hooks` is a list the instances
```
reedwm: This is formatted slightly weirdly. Comma should be on same line as \`train_hooks\`, and each line in Returns section should be indented same amount.

official/transformer/transformer_estimator_benchmark.py
```diff
@@ -0,0 +1,299 @@
...
+
+class EstimatorBenchmark(tf.test.Benchmark):
+  """Base class to hold methods common to test classes in the module.
+
+     Code under test for the Transformer Estimator models that report mostly the
+     same data and require the same FLAG setup.
+  """
+  local_flags = None
+
+  def __init__(self, output_dir=None, default_flags=None, flag_methods=None):
```
reedwm: You can put output_dir='/tmp' directly in the function signature, instead of having it default to None.

Note you should not do this for default_flags or flags_methods, since those default to dicts, and dicts are mutable. It's only safe to do this for immutable default values, like strings.

official/transformer/transformer_estimator_benchmark.py
```diff
@@ -0,0 +1,299 @@
...
+    examples_per_sec_hook = None
+    for hook in stats['train_hooks']:
+      if isinstance(hook, hooks.ExamplesPerSecondHook):
+        examples_per_sec_hook = hook
+        break
+
+    eval_results = stats['eval_results']
+    metrics = []
+    if 'bleu_cased' in stats:
+      metrics.append({'name': 'bleu_score',
```
reedwm: Maybe call this blue_cased_score?

official/transformer/transformer_estimator_benchmark.py
```diff
@@ -0,0 +1,299 @@
...
+    """
+    self._setup()
+    FLAGS.num_gpus = 8
+    FLAGS.data_dir = self.train_data_dir
+    FLAGS.vocab_file = self.vocab_file
+    # Sets values directly to avoid validation check.
+    FLAGS['bleu_source'].value = self.bleu_source
+    FLAGS['bleu_ref'].value = self.bleu_ref
+    FLAGS.param_set = 'base'
+    FLAGS.batch_size = 2048 * 8
```
reedwm: Why is the per-GPU batch size lower?

official/transformer/transformer_estimator_benchmark.py
tfboyd: I think the answer is no, because the methods that extend this end up passing None as the value and I want to catch that.  I wanted to do output_dir='/tmp' but it was not as robust.  Open to other ideas and view points.

```python
   super(TransformerBaseEstimatorAccuracy, self).__init__(
        output_dir=output_dir, flag_methods=flag_methods)
```

official/transformer/transformer_estimator_benchmark.py
tfboyd: For the benchmark I want to keep them the same.  For the accuracy a larger batch gives a  better result up to a certain point.  The paper and supporting work suggest  2048 * 8 gpus.  I tried 4096*8, which worked but I am now trying 2048 as I think that better matches conventional testing.  Regardless, the value for the 1 and 2 GPUs you want to make the batches as large as possible to get the highest BLEU.  We could run them with smaller batches just to show scaling, but that isn't realy the test.  In the long-run I will likely remove the 1 and 2 GPU tests, but for now they are faster and will help with FP16 testing.    I am cool with whatever for 1 and 2.  I will keep playing with the 8 GPU batch until it is SOTA everytime, which I suspect is 2048 * 8 or 4096 * 8 and X total steps.

official/transformer/transformer_estimator_benchmark.py
tfboyd: fixed, compact and detailed.

tfboyd:
@reedwm PTAL

guptapriya:
Sorry didn't get a chance to review earlier but LGTM.

PR #6735, opened 2019-05-08
r/tf.random_uniform/tf.random.uniform

tfboyd:
This was causing the synthetic data tests to fail in TF 2.0.


tfboyd:
Fixed
https://sponge.corp.google.com/target?id=6233d383-3312-4c6d-987b-423178f0f314&target=tensorflow_models/perfzero/manual/v100_1_gpu/v100_1_gpu

PR #6756, opened 2019-05-10
Add FP16 to transformer with benchmark tests.

tfboyd:


official/transformer/transformer_main.py
```diff
@@ -392,7 +397,7 @@ def define_transformer_flags():
       intra_op=False,
       synthetic_data=True,
       max_train_steps=False,
-      dtype=False,
+      dtype=True,
```
reedwm: If you set dtype=True, it also defines --loss_scale, which is ignored. I would add a loss_scale argument to define_performance, so you can pass False to it.

official/transformer/transformer_main.py
```diff
@@ -182,6 +182,11 @@ def get_train_op_and_metrics(loss, params):
     if params["use_tpu"] and params["tpu"] != tpu_util.LOCAL:
       optimizer = tf.contrib.tpu.CrossShardOptimizer(optimizer)
 
+    # Uses automatic mixed precision FP16 training if on GPU.
+    if params["dtype"] == "fp16" and not params["use_tpu"]:
```
reedwm: I don't think the use_tpu part is necessary. fp16 isn't support on TPUs, so users should not pass --dtype=fp16 with --use_tpu.

official/transformer/transformer_estimator_benchmark.py
```diff
@@ -194,21 +201,47 @@ def benchmark_graph_8_gpu(self):
     FLAGS['bleu_source'].value = self.bleu_source
     FLAGS['bleu_ref'].value = self.bleu_ref
     FLAGS.param_set = 'base'
-    FLAGS.batch_size = 2048 * 8
+    FLAGS.batch_size = 3072 * 8
```
reedwm: Why 3072?

tfboyd: 2048 is too small the goal is over 20K tokens per batch.  It might work if I train longer, but I would rather try 3072 (4096 * 8 was close ~27.226  around 75K steps) and still do 100K steps.  If that works I can try other configs and do more research or move look at the other hparams and maybe try 4096 again.

official/transformer/transformer_main.py
tfboyd: Done and loss_scale is False by default and I updated resnet.

PR #6760, opened 2019-05-10
Fix trivial model to work properly with fp16

haoyuz:


official/resnet/keras/trivial_model.py
```diff
@@ -23,15 +23,17 @@
...
   input_shape = (224, 224, 3)
-  img_input = layers.Input(shape=input_shape)
+  img_input = layers.Input(shape=input_shape, dtype=dtype)
 
   x = layers.Lambda(lambda x: backend.reshape(x, [-1, 224 * 224 * 3]),
                     name='reshape')(img_input)
   x = layers.Dense(1, name='fc1')(x)
-  x = layers.Dense(num_classes, activation='softmax', name='fc1000')(x)
+  x = layers.Dense(num_classes, name='fc1000')(x)
+  x = backend.cast(x, 'float32')
```
yuefengz: Could you add a comment for why we need to do casting here?

haoyuz: Added by copying from the ResNet model. The final activation layer requires fp32 to function correctly. This manual casting is only a temporary solution and will be removed later. Thanks!

PR #6779, opened 2019-05-14
Corrects headers in movielens.py util.

tayo:


tfboyd:
Silly question, if this was wrong how was it still training correctly?  Asking as I will need to kick off a training run to verify this does not create any problems.

tayo:
On closer look, this change is not necessary.  The current headers are ignored.  

See here for more info:
https://github.com/tensorflow/models/blob/7f031e936aeffdbe4073b6f95bbf0efc91a4e825/official/datasets/movielens.py#L164

tayo:
Closing PR.

PR #6799, opened 2019-05-16
Fix DeepLab Python 3 compatibility issue #6797.

### LINKED ISSUE #6797: [DeepLab] PR #6681 breaks train.py and eval.py on Python 3
### System information
- **What is the top-level directory of the model you are using**: research/deeplab
- **OS Platform and Distribution (e.g., Linux Ubuntu 16.04)**: Python 3
- **TensorFlow installed from (source or binary)**: binary
- **TensorFlow version (use command below)**: 1.13.1
- **Bazel version (if compiling from source)**: n/a
- **CUDA/cuDNN version**: n/a
- **GPU model and memory**: n/a
- **Exact command to reproduce**: run `python3 deeplab/train.py` with any valid arguments.

### Describe the problem
PR  #6681 appears not to be tested with Python 3. In particular running `deeplab/train.py` fails 
with error:

```
  File "***/research/deeplab/datasets/data_generator.py", line 286, in _preprocess_image
    crop_height=self.crop_size[0],
TypeError: 'map' object is not subscriptable
```
since `Dataset.__init__` expects `crop_size` to be list, but `train.py` constructs `Dataset` as:

```
      dataset = data_generator.Dataset(
         # .....
          crop_size=map(int, FLAGS.train_crop_size),
         # .......
    )
```

There is similar issue with instantiation of ModelOptions:
```
  model_options = common.ModelOptions(
      outputs_to_num_classes=outputs_to_num_classes,
      crop_size=map(int, FLAGS.train_crop_size),
      atrous_rates=FLAGS.atrous_rates,
      output_stride=FLAGS.output_stride)
```

The same problem also affects `vis.py` and `eval.py`

I'm willing to contribute PR for this issue, but I would like to ask, what is the preferred way to fix it. I see the following options:

1. Partially revert  #6681 and replace `flags.DEFINE_list('eval_crop_size'` with `flags.DEFINE_multi_integer('` (as it was originally).
2. Replace `map(int, FLAGS.train_crop_size)` with `list(map(int, FLAGS.train_crop_size))` (there will be 4 such replacements around the code).
3.  Replace `map(int, FLAGS.train_crop_size)` with `[int(sz) for sz in FLAGS.train_crop_size]` (there will be 4 such replacements around the code).

### Source code / logs
n/a

merofeev:
Fixes #6797 by replacing `map`s with list comprehensions.

### DIFF
--- a/research/deeplab/eval.py
+++ b/research/deeplab/eval.py
@@ -88,7 +88,7 @@ def main(unused_argv):
       split_name=FLAGS.eval_split,
       dataset_dir=FLAGS.dataset_dir,
       batch_size=FLAGS.eval_batch_size,
-      crop_size=map(int, FLAGS.eval_crop_size),
+      crop_size=[int(sz) for sz in FLAGS.eval_crop_size],
       min_resize_value=FLAGS.min_resize_value,
       max_resize_value=FLAGS.max_resize_value,
       resize_factor=FLAGS.resize_factor,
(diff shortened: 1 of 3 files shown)


merofeev:
I signed it!

PR #6809, opened 2019-05-18
Include flags when reporting benchmark.

reedwm:
This will allow one to easily reproduce a benchmark by running with the flags.


reedwm:
Also, please do not merge until tensorflow/benchmarks#375 is merged. Otherwise, PerfZero will break.

PR #6828, opened 2019-05-20
Use new tensorrt API

tjakob:
In Tensorflow 1.14 a majority of the TensorRT python code was moved from contrib to python/compilers.
As part of this move, some of the API was changed.

Running this example without changes, using Tensorflow 1.14, and passing `--int8` will result an
`AttributeError: module 'tensorflow.contrib.tensorrt' has no attribute 'calib_graph_to_infer_graph'`


tfboyd:
@aaroey can take a look at this.  Thank you for doing the update.

Issue #5411, opened 2018-10-01
config_util.get_configs_from_multiple_files changed from eval_input_config to eval_input_configs and breaks  read_data_and_evaluate


afcruzs:
------------------------

### System information
- **What is the top-level directory of the model you are using**: models\research\object_detection
- **Have I written custom code (as opposed to using a stock example script provided in TensorFlow)**: Yes
- **OS Platform and Distribution (e.g., Linux Ubuntu 16.04)**: Windows 10
- **TensorFlow installed from (source or binary)**: Binary
- **TensorFlow version (use command below)**:  1.10.0
- **CUDA/cuDNN version**: 9.0/64_7
- **Bazel version (if compiling from source)**: none
- **GPU model and memory**: gtx 1050
- **Exact command to reproduce**:  python offline_eval_map_corloc.py --eval_dir=path/to/eval_dir \
        --eval_config_path=path/to/evaluation/configuration/file \
        --input_config_path=path/to/input/configuration/file

You can collect some of this information using our environment capture script:

### Describe the problem
When executing offline_eval_map_corloc.py it breaks when trying to retrieve `input_config = configs['eval_input_config']`. This is because [recently](https://github.com/tensorflow/models/commit/99256cf470df6af16808eb0e49a8354d2f9beae2#diff-cee03111e9efeb5582a53ee258ecc2a9) it changed from eval_input_config to eval_input_configs (which is a list of configs).

Also, the documentation of `get_configs_from_multiple_files` is outdated, it still says eval_input_config instead of eval_input_configs.

### Source code / logs
```
  _sys.exit(main(argv))
  File "offline_eval_map_corloc.py", line 171, in main
    input_config = configs['eval_input_config']
KeyError: 'eval_input_config'

```

afcruzs:
@Cpryslak I ended up replacing it with this in my local repo:  `configs['eval_input_configs'][0]`

Cpryslak:
Yes that passed over the error. Thank you for the help. I really appreciate it.

wt-huang:
Closing as this is resolved

paviddavid:
@afcruzs @Cpryslak  @wt-huang I totally agree that there is an error in the variables. However, how have you solved this issue? I added the  `configs['eval_input_configs'][0]` in `offline_eval_map_corloc.py` and a new error appears: `NameError: name unicode is not defined` in `utils/object_detection_evaluation.py`. Replacing unicode() to str() leads to other errors ... 

[...]

PR #6836, opened 2019-05-21
update offline_eval_map_corloc.py

tanguypommellet:
in config_util.get_configs_from_multiple_files, the key of the dict seems to be eval_input_configs instead of eval_input_config. (a 's' is missing)
found that while following the Inference and evaluation on the Open Images dataset tutorial.
I was getting the error :
input_config = configs['eval_input_config']
KeyError: 'eval_input_config'

### DIFF
--- a/research/object_detection/metrics/offline_eval_map_corloc.py
+++ b/research/object_detection/metrics/offline_eval_map_corloc.py
@@ -159,7 +159,7 @@ def main(argv):
       eval_config_path=FLAGS.eval_config_path)
 
   eval_config = configs['eval_config']
-  input_config = configs['eval_input_config']
+  input_config = configs['eval_input_configs']
 
   metrics = read_data_and_evaluate(input_config, eval_config)
 


tanguypommellet:
I signed it!

research/object_detection/metrics/offline_eval_map_corloc.py
```diff
@@ -159,7 +159,7 @@ def main(argv):
       eval_config_path=FLAGS.eval_config_path)
 
   eval_config = configs['eval_config']
-  input_config = configs['eval_input_config']
+  input_config = configs['eval_input_configs']
```
pkulzc: configs['eval_input_configs'] is a list so I think this should be configs['eval_input_configs'][0]

tanguypommellet: I tried the changes proposed here ; 
https://github.com/tensorflow/models/issues/5411#issuecomment-483992817
It does the job

techds:
The offline_eval_map_corloc.py is still pending the required update. Ran into KeyError: 'eval_input_config'. The key needs to be 'eval_input_configs' instead of 'eval_input_config'. On replacing 
`input_config = configs['eval_input_config']` 
with
`input_config = configs['eval_input_configs'][0]`
a new error is being raised: 
> AttributeError: 'NoneType' object has no attribute 'size'

PR #6838, opened 2019-05-21
Add Transformer Big Benchmarks + FP16 for other tests.

tfboyd:
Main feature is adding Transformer Big, which will run the same basic tests as Transformer Base.

- Added Transformer Big
- Add FP16 tests
- Updated batch-sizes to match final testing approach.


tfboyd:
LINT issues are old, I will fix those in another PR.

PR #6842, opened 2019-05-22
Add the transformer model after the oss conversion.

isaprykin:


guptapriya:
cc @lintian06 @xunkai55

tfboyd:
Close for now while we work on the other PR.  It will be nice when we are done and end the timezone churn.

isaprykin:
Sounds good!

PR #6843, opened 2019-05-22
Made a few changes to benchmarks

haoyuz:
* Add tweaked fp32 test
* Run accuracy tests with LR ops


tfboyd:
I do not count FP32 XLA.  Open to debate for sure.  No impact on this PR but you may want to do a tweaked fp32 non-XLA.  Pure FP32 (non-xla) is needed as using XLA gives you more buffer between synthetic and real and I compare it with the pure FP32 from tf_cnn_benchmarks as the baseline.

PR #6846, opened 2019-05-22
Merge Transformer V2 to Github

lintian06:
Merge changes related to Transformer V2 to master.

- Major changes are in official/transformer/model/v2/*


official/resnet/keras/resnet_model.py
```diff
@@ -230,8 +230,7 @@ def resnet50(num_classes, dtype='float32', batch_size=None):
   x = identity_block(x, 3, [512, 512, 2048], stage=5, block='b')
   x = identity_block(x, 3, [512, 512, 2048], stage=5, block='c')
 
-  rm_axes = [1, 2] if backend.image_data_format() == 'channels_last' else [2, 3]
```
guptapriya: why this change? how is this out of sync? 
cc @haoyuz

haoyuz: Thanks Priya for bring this to my attention! Please don't change the ResNet model code. This PR was just merged earlier today for performance improvement.

lintian06: OK. It is hard to revert from a squashed changes. I will manually revert to the original.

lintian06:
Thank you for the approval.

To add a note: I disable two tests in v2. Find the reason is due to tf.where behavior change in v2. Will fix the related code and enable those two tests again.

PR #6860, opened 2019-05-23
Transformer v2 benchmark

tfboyd:
This includes the refactor from my other PR.  I am not going to merge this until the other PR is merged.

https://github.com/tensorflow/models/pull/6859

### DIFF
--- a/official/transformer/v2/misc.py
+++ b/official/transformer/v2/misc.py
@@ -66,11 +66,13 @@ def define_transformer_flags():
   flags_core.define_device(tpu=True)
 
   flags.DEFINE_integer(
-      name='train_steps', default=None,
-      help='The number of steps to run for training. If it is larger than '
-      '# batches per epoch, then use # batches per epoch. When this flag is '
-      'set, only one epoch is going to run for training.')
-
+      name='train_steps', short_name='ts', default=None,
+      help=flags_core.help_wrap('The number of steps used to train.'))
+  flags.DEFINE_integer(
+      name='steps_between_evals', short_name='sbe', default=1000,
+      help=flags_core.help_wrap(
+          'The Number of training steps to run between evaluations. This is '
+          'used if --train_steps is defined.'))
   flags.DEFINE_boolean(
       name='enable_tensorboard', default=False,
       help='Whether to enable Tensorboard callback.')
@@ -110,15 +112,6 @@ def define_transformer_flags():
           'and static batching will always be used.'))
 
   # Flags for training with steps (may be used for debugging)
-  flags.DEFINE_integer(
-      name='init_epoch', short_name='is', default=0,
-      help=flags_core.help_wrap('The number of initial epoch for training.'))
-  flags.DEFINE_string(
-      name='init_weight_path', short_name='iwp', default=None,
-      help=flags_core.help_wrap('The initial model weights to load.'))
-  flags.DEFINE_string(
-      name='init_logdir_timestamp', short_name='ilt', default=None,
-      help=flags_core.help_wrap('The initial timestamp for logdir.'))
   flags.DEFINE_integer(
       name='validation_steps', short_name='vs', default=64,
       help=flags_core.help_wrap('The number of steps used in validation.'))
--- a/official/transformer/v2/misc.py
+++ b/official/transformer/v2/misc.py
@@ -18,27 +18,32 @@ from __future__ import absolute_import
 from __future__ import division
 from __future__ import print_function
 
+# pylint: disable=g-bad-import-order
 from absl import flags
+import tensorflow as tf
 
 from official.transformer.model import model_params
 from official.utils.flags import core as flags_core
+from official.utils.misc import keras_utils
+
+FLAGS = flags.FLAGS
 
 PARAMS_MAP = {
-    "tiny": model_params.TINY_PARAMS,
-    "base": model_params.BASE_PARAMS,
-    "big": model_params.BIG_PARAMS,
+    'tiny': model_params.TINY_PARAMS,
+    'base': model_params.BASE_PARAMS,
+    'big': model_params.BIG_PARAMS,
 }
 
 
 def get_model_params(param_set, num_gpus):
   """Gets predefined model params."""
   if num_gpus > 1:
-    if param_set == "big":
+    if param_set == 'big':
       return model_params.BIG_MULTI_GPU_PARAMS.copy()
-    elif param_set == "base":
+    elif param_set == 'base':
       return model_params.BASE_MULTI_GPU_PARAMS.copy()
     else:
-      raise ValueError("Not valid params: param_set={} num_gpus={}".format(
+      raise ValueError('Not valid params: param_set={} num_gpus={}'.format(
           param_set, num_gpus))
 
   return PARAMS_MAP[param_set].copy()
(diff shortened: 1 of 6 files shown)


official/transformer/v2/transformer_main.py
```diff
@@ -125,24 +125,38 @@ def train(self):
...
 
     init_epoch = flags_obj.init_epoch or 0
-    init_steps = init_epoch * flags_obj.steps_per_epoch
-    callbacks = self._create_callbacks(cur_log_dir, init_steps, params)
+    init_steps = init_epoch * flags_obj.train_steps
+    callbacks = self._create_callbacks(flags_obj.model_dir, init_steps, params)
 
     history = model.fit(
         train_ds,
-        initial_epoch=init_epoch,
```
guptapriya: erm, i am not sure if the logic here is all consistent if init_epoch and init_steps are being used.. maybe would be good for @lintian06 to confirm

official/transformer/v2/transformer_main.py
```diff
@@ -125,24 +125,38 @@ def train(self):
...
+        callbacks=callbacks,
+        verbose=2)
     tf.compat.v1.logging.info("\nTrain history: {}".format(history.history))
 
-    save_weight_path = os.path.join(cur_log_dir, "saves-model-weights.hdf5")
-    save_model_path = os.path.join(cur_log_dir, "saves-model.hdf5")
+    stats = misc.build_stats(history, callbacks)
+
+    if flags_obj.bleu_source and flags_obj.bleu_ref:
+      uncased_score, cased_score = evaluate_and_log_bleu(model,
```
guptapriya: oh, this might be why you saw that error. the model in transformer is different for training vs eval/predict.
you need to create another model here: `model = transformer.create_model(params, is_train=False)`

Although, that makes me think why validation during training works with the same model.. @lintian06 do you know?

official/transformer/v2/misc.py
```diff
@@ -60,113 +65,182 @@ def define_transformer_flags():
...
+
+  Returns:
+    Dictionary of normalized results.
+  """
+  stats = {}
+
+  if history and history.history:
+    train_hist = history.history
+    # Gets final loss from training.
+    stats['loss'] = train_hist['loss'][-1].item()
```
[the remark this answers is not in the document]
tfboyd: Yeah, I did not see anything that was of interest in history.  The validation is not super useful as we want to bleu score.

official/transformer/v2/transformer_benchmark.py
```diff
@@ -0,0 +1,196 @@
...
+
+from official.transformer.v2 import misc
+from official.transformer.v2 import transformer_main as transformer_main
+
+TRANSFORMER_EN2DE_DATA_DIR_NAME = 'wmt32k-en2de-official'
+EN2DE_2014_BLEU_DATA_DIR_NAME = 'newstest2014'
+FLAGS = flags.FLAGS
+
+
+class EstimatorBenchmark(tf.test.Benchmark):
```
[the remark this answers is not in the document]
tfboyd: Fixing this,  I need to create a base as a big part of this is common, but the reporting part is not.  Maybe I will do that today so we can fix this going forward.

official/transformer/v2/transformer_benchmark.py
```diff
@@ -0,0 +1,196 @@
...
+                              TRANSFORMER_EN2DE_DATA_DIR_NAME,
+                              'vocab.ende.32768')
+
+    def_flags = {}
+    def_flags['param_set'] = 'base'
+    def_flags['vocab_file'] = vocab_file
+    def_flags['data_dir'] = train_data_dir
+    def_flags['train_steps'] = 200
+    def_flags['train_epochs'] = 1
+    def_flags['epochs_between_evals'] = 10
```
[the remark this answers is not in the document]
tfboyd: I want to ensure we do not do any evals, I will add a comment.

official/transformer/v2/transformer_main.py
tfboyd: Validation doesn't actually work, well it works but it does not give the values we want.  I guess I don't get why this works just fine with Resnet and we don't set training=False.  I will dig in more today.

official/transformer/v2/transformer_main.py
tfboyd: I would like to see this stream lined to "just work" like the resnet example; we should not need to say what epoch we are starting at to start up training again.  It may take some iterations.

official/transformer/v2/transformer_main.py
guptapriya: if you look inside the `create_model` function, you will see that an different model object is created for training vs eval/predict. it's to do with how custom losses are handled during training - which is not needed in eval so the eval model is slightly simpler. in resnet it's the same model in both cases.

official/transformer/v2/transformer_main.py
tfboyd: I have fixed this and it does not allow starting from a checkpoint.  We can add that back and it should just work and not require you to know what epoch you were on.  The previous code only worked if you knew where you stopped.

official/transformer/v2/transformer_main.py
tfboyd: Fixed

official/transformer/v2/transformer_main.py
lintian06: Hi Toby, the default init_epoch is 0 and just works as another other models, so no need to set or remove.

official/transformer/v2/transformer_main.py
lintian06: Validation is only to give a number. As it also feeds ground-truth labels, it is not a correct way to get BLEU score by bringing in extra knowledge. is_train=False will build another model which is totally agnostic to labels.

official/transformer/v2/transformer_main.py
```diff
@@ -124,31 +123,49 @@ def train(self):
...
-    model.save(save_model_path)
+    callbacks = self._create_callbacks(flags_obj.model_dir, 0, params)
+
+    epochs = 0
+    predict_model = None
+    for i in xrange(1, flags_obj.train_epochs + 1):
+      print("Train Epoch:{}/{}".format(i, flags_obj.train_epochs))
+      history = model.fit(
+          train_ds,
+          initial_epoch=epochs,
```
guptapriya: this one could be `i-1`?

official/transformer/v2/transformer_main.py
```diff
@@ -124,31 +123,49 @@ def train(self):
...
+          epochs=i,
+          steps_per_epoch=flags_obj.train_steps,
+          callbacks=callbacks,
+          verbose=2)
+      epochs += 1
```
guptapriya: this can be just `i`?

official/transformer/v2/transformer_main.py
```diff
@@ -124,31 +123,49 @@ def train(self):
...
-        callbacks=callbacks)
-    tf.compat.v1.logging.info("\nTrain history: {}".format(history.history))
-
-    save_weight_path = os.path.join(cur_log_dir, "saves-model-weights.hdf5")
-    save_model_path = os.path.join(cur_log_dir, "saves-model.hdf5")
-    model.save_weights(save_weight_path)
-    model.save(save_model_path)
+    callbacks = self._create_callbacks(flags_obj.model_dir, 0, params)
+
+    epochs = 0
```
guptapriya: not entirely sure you need another variable. you can write it all in terms of `i` from what i could tell?

official/transformer/v2/transformer_main.py
```diff
@@ -177,23 +194,20 @@ def _create_callbacks(self, cur_log_dir, init_steps, params):
...
+    ckpt_full_path = os.path.join(cur_log_dir, "cp-{epoch:04d}.ckpt")
+    callbacks.append(tf.keras.callbacks.ModelCheckpoint(ckpt_full_path,
+                                                        save_weights_only=True))
+    return callbacks
 
   def _load_weights_if_possible(self, model, init_weight_path=None):
     """Loads model weights when it is provided."""
     if init_weight_path:
       tf.compat.v1.logging.info("Load weights: {}".format(init_weight_path))
-      model.load_weights(init_weight_path, by_name=True)
```
guptapriya: by_name is not needed anymore?

official/transformer/v2/transformer_main.py
lintian06: Yes. initial_epoch=i-1.

official/transformer/v2/transformer_main.py
```diff
@@ -124,31 +123,49 @@ def train(self):
...
+      epochs += 1
+      tf.compat.v1.logging.info("Train history: {}".format(history.history))
+
+      stats = misc.build_stats(history, callbacks)
+
+      if (flags_obj.bleu_source and
+          flags_obj.bleu_ref and
+          epochs%flags_obj.epochs_between_evals == 0):
+        if not predict_model:
+          predict_model = transformer.create_model(params, False)
```
lintian06: My concern is that the first model may have name_scope "model", second "model_1", and third "model_2". I don't know whether _load_weights_if_possible can be reloaded properly. Should check.

official/transformer/v2/transformer_main.py
tfboyd: Correct, that was with the other saving style.  I did multiple a 100K step run on 1 GPU and this works correctly.

official/transformer/v2/transformer_main.py
tfboyd: https://sponge.corp.google.com/target?id=48d603e1-252e-45c1-8385-52044f2c5be9&target=tensorflow_models%2Fperfzero%2Fmanual%2Fv100_1_gpu%2Fv100_1_gpu&searchFor=

There is only `model` and `model_1`.  I create the model once and just reload the weights each time.  This saves a little time and looks to work based on the log.

Until I started caching the model, I did some ignored Error in the log but I forget what it was.  But it works fine now.

official/transformer/v2/transformer_main.py
tfboyd: https://sponge.corp.google.com/target?id=48d603e1-252e-45c1-8385-52044f2c5be9&target=tensorflow_models%2Fperfzero%2Fmanual%2Fv100_1_gpu%2Fv100_1_gpu&searchFor=

Yup works perfectly.

official/transformer/v2/transformer_main.py
tfboyd: Sorry about that, much more streamlined now.

official/transformer/v2/transformer_main.py
tfboyd: fixed

official/transformer/v2/transformer_main.py
```diff
@@ -124,31 +123,49 @@ def train(self):
...
-
-    save_weight_path = os.path.join(cur_log_dir, "saves-model-weights.hdf5")
-    save_model_path = os.path.join(cur_log_dir, "saves-model.hdf5")
-    model.save_weights(save_weight_path)
-    model.save(save_model_path)
+    callbacks = self._create_callbacks(flags_obj.model_dir, 0, params)
+
+    epochs = 0
+    predict_model = None
+    for i in xrange(1, flags_obj.train_epochs + 1):
```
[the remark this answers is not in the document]
tfboyd: I changed it to range.  You likely know the fully story and I also got confused and thought they also had different signatures.  

In Python2 I understood most of the time we should just use xrange; even if the performance gain is minimal.  This creates issues with Python3 so then I started using six.  Bad pattern, with Python2 going away range should be used unless there is a clear performance need due to a huge list.

tfboyd:
@guptapriya  PTAL  please take one more pass.  I think this is much better.  I am sorry I mixed the quote changes and the other refactor into the transformer_main.py change.  Thank you for working through it.  I would rebase; but I know this is close and we can try V2 this afternoon.

tfboyd:
Ok, i'm pushing merge.  I have multiple end-to-end tests and we know there will be more refactors.  Better than it was.  :-)

PR #6865, opened 2019-05-23
NCF Keras: Add validation every epoch

guptapriya:
Adding validation every epoch allows us to view the progress during training instead of having to wait until the last eval. Mostly useful for manual runs.

### DIFF
--- a/official/recommendation/ncf_keras_main.py
+++ b/official/recommendation/ncf_keras_main.py
@@ -110,6 +110,8 @@ def _get_train_and_eval_data(producer, params):
   train_input_fn = producer.make_input_fn(is_training=True)
   train_input_dataset = train_input_fn(params).map(
       preprocess_train_input)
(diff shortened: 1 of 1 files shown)


official/recommendation/ncf_keras_main.py
```diff
@@ -242,10 +244,13 @@ def run_ncf(_):
         cloning=params["clone_model_in_keras_dist_strat"])
 
     history = keras_model.fit(train_input_dataset,
+                              steps_per_epoch=num_train_steps,
                               epochs=FLAGS.train_epochs,
                               callbacks=[
                                   IncrementEpochCallback(producer),
                                   time_callback],
+                              validation_data=eval_input_dataset,
```
guptapriya: Ideally, this is the only change that should be needed. I needed to add all the other things (steps etc) because of an odd bug in Keras. Can simplify once that bug is fixed.

PR #6871, opened 2019-05-23
 Fix bug in dataset reader.

MarvinTeichmann:
The ".mat" files loaded in the dataset are byte files. Python 3.7 requires them to be loaded using "rb".

@andrefaraujo


MarvinTeichmann:
Removed commit which is already under review in  #6870.

PR #6881, opened 2019-05-24
Use more warmup steps for 96 core tests

haoyuz:
Increase the number of warmup steps for 96 core platforms. Hope this can stabilize the reported performance.

### DIFF
--- a/official/resnet/keras/keras_benchmark.py
+++ b/official/resnet/keras/keras_benchmark.py
@@ -96,6 +96,12 @@ class KerasBenchmark(tf.test.Benchmark):
       num_examples = (
           total_batch_size * log_steps * (len(time_log) - warmup - 1))
       examples_per_sec = num_examples / elapsed
+      print('HZDEBUG: stats ---------------------------------------------------')
+      print('num_examples =', num_examples)
+      print('len(time_log) =', len(time_log))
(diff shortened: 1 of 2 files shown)


guptapriya:
why not just do more warmup steps in all cases?

haoyuz:
> why not just do more warmup steps in all cases?

I think that should be good as well; just a minor point, getting stable performance in a short period of time is also desirable so we might want to eventually figure out the slowness on 96. Keeping tests on 64 or DGX-1 short can also save resources a little bit 😺

guptapriya:
> > why not just do more warmup steps in all cases?
> 
> I think that should be good as well; just a minor point, getting stable performance in a short period of time is also desirable so we might want to eventually figure out the slowness on 96. Keeping tests on 64 or DGX-1 short can also save resources a little bit 😺

I am fine either way, check with @tfboyd as he is the expert on the benchmarks.

haoyuz:
> Let's make a new test that does the delayed or longer measurement. That way we have a constant comparison on any system. Just do it for 8-GPUs. tweaked / non-tweaked if desired and only fp16+xla as that is where it matters. `tweaked_delay_measure` or something and try to keep it short.

Added tests, PTAL. Thanks for the reviews!

PR #6883, opened 2019-05-25
Add distribute strategies to transformer.

isaprykin:


official/transformer/v2/metrics.py
```diff
@@ -160,9 +160,10 @@ def get_config(self):
 
   def call(self, inputs):
     logits, targets = inputs[0], inputs[1]
-    for mean, fn in self.metric_mean_fns:
-      m = mean(*fn(logits, targets))
-      self.add_metric(m)
+    if not tf.distribute.in_cross_replica_context():
```
tfboyd: Leave a comment and bug link if relevant as to why this is needed, it seems odd that we have to do something special.

official/transformer/v2/transformer.py
```diff
@@ -103,6 +103,9 @@ def call(self, x, training):
           outputs: [batch_size, decoded length]
           scores: [batch_size, float]}
     """
+    if not tf.distribute.in_cross_replica_context() and len(x) == 1:
```
tfboyd: Leave a comment and bug link if relevant as to why this is needed, it seems odd that we have to do something special.

official/transformer/v2/transformer_benchmark.py
```diff
@@ -142,6 +142,7 @@ def benchmark_1_gpu(self):
     FLAGS.batch_size = 4096
     FLAGS.train_steps = 200000
     FLAGS.steps_between_evals = 5000
+    FLAGS.static_batch = True
```
tfboyd: Make a new test called   `benchmark_1_gpu_static`  so we can track the difference.  legacy graph does not seem to have this issue or not as bad running the same keras code.  

Once they both converge or we are sure of the standard we can drop one of them.  I am not certain at this time what is "right".  I don't want to run both to convergence, but for now both seems correct.

official/transformer/v2/transformer_benchmark.py
```diff
@@ -150,6 +151,54 @@ def benchmark_1_gpu(self):
...
+    """
+    self._setup()
+    FLAGS.num_gpus = 8
+    FLAGS.data_dir = self.train_data_dir
+    FLAGS.vocab_file = self.vocab_file
+    # Sets values directly to avoid validation check.
+    FLAGS['bleu_source'].value = self.bleu_source
+    FLAGS['bleu_ref'].value = self.bleu_ref
+    FLAGS.param_set = 'base'
+    FLAGS.batch_size = 32768
```
tfboyd: Let's write this out as 8 * 4096 to make the per gpu batch-size obvious at a glance.  And change the steps to 100K.  Running 100k I reached 27 to 27.2 but never then 27.3 desired.

official/transformer/v2/transformer_benchmark.py
```diff
@@ -150,6 +151,54 @@ def benchmark_1_gpu(self):
                                    bleu_min=25.3,
                                    bleu_max=26)
 
+  def benchmark_8_gpu(self):
+    """Benchmark 8 gpu.
+
+      The paper uses 8 GPUs and a much larger effective batch size, this is will
+      not converge to the 27.3 BLEU (uncased) SOTA.
```
tfboyd: Change this to say:
"Should converge to 27.3 BLEU (uncased). This has not been confirmed yet."

official/transformer/v2/transformer_benchmark.py
```diff
@@ -150,6 +151,54 @@ def benchmark_1_gpu(self):
...
+    FLAGS.train_steps = 200000
+    FLAGS.static_batch = True
+    FLAGS.steps_between_evals = 5000
+    FLAGS.model_dir = self._get_model_dir('benchmark_8_gpu')
+    self._run_and_report_benchmark(total_batch_size=FLAGS.batch_size,
+                                   log_steps=FLAGS.log_steps,
+                                   bleu_min=25.3,
+                                   bleu_max=26)
+
+  def benchmark_big_8_gpu(self):
```
tfboyd: Make a new class: TransformerBigKerasAccuracy and then just use `benchmark_8_gpu`

I look for the word  `TransformerBig` as a way to know the model when I do reporting.

official/transformer/v2/transformer_benchmark.py
```diff
@@ -150,6 +151,54 @@ def benchmark_1_gpu(self):
...
+    """Benchmark 8 gpu.
+
+      The paper uses 8 GPUs and a much larger effective batch size, this is will
+      not converge to the 27.3 BLEU (uncased) SOTA.
```
tfboyd: Change this to say (you could add a date, but I am pretty sure we will remember to update when it converged like I did with the estimator version:

"Should converge to 28.4 BLEU (uncased). This has not be verified yet."

official/transformer/v2/transformer_benchmark.py
```diff
@@ -150,6 +151,54 @@ def benchmark_1_gpu(self):
...
+    self._setup()
+    FLAGS.num_gpus = 8
+    FLAGS.data_dir = self.train_data_dir
+    FLAGS.vocab_file = self.vocab_file
+    # Sets values directly to avoid validation check.
+    FLAGS['bleu_source'].value = self.bleu_source
+    FLAGS['bleu_ref'].value = self.bleu_ref
+    FLAGS.param_set = 'big'
+    FLAGS.batch_size = 24576
+    FLAGS.train_steps = 200000
```
tfboyd: I think 100,000 steps is enough.  But we have not validated yet.  Totally FYI not a big deal until we know.

official/transformer/v2/transformer_benchmark.py
```diff
@@ -150,6 +151,54 @@ def benchmark_1_gpu(self):
...
+    FLAGS.static_batch = True
+    FLAGS.steps_between_evals = 5000
+    FLAGS.model_dir = self._get_model_dir('benchmark_big_8_gpu')
+    self._run_and_report_benchmark(total_batch_size=FLAGS.batch_size,
```
tfboyd: This should be something like 28.0 to 29.0 for bleu scores, we will need to dial it in.  Ideally we would have the min be 28.4 matching the paper. More of an FYI we can dial this in later.

official/transformer/v2/transformer_benchmark.py
```diff
@@ -179,6 +228,7 @@ def benchmark_1_gpu(self):
     FLAGS.num_gpus = 1
     FLAGS.batch_size = self.batch_per_gpu
     FLAGS.model_dir = self._get_model_dir('benchmark_1_gpu')
+    FLAGS.static_batch = True
```
tfboyd: I would rather a new test with `benchmark_1_gpu_static`

official/transformer/v2/transformer_main.py
```diff
@@ -107,13 +113,22 @@ def train(self):
...
+    else:
+      model = transformer.create_model(params, is_train)
+      opt = self._create_optimizer()
+      model.compile(opt)
 
-    model.compile(opt, target_tensors=[])
     model.summary()
 
-    map_data_fn = data_pipeline.map_data_for_transformer_fn
+    if self.distribution_strategy:
```
tfboyd: Might be nice to have a comment as to why you do this for dist_strat.  This is kind of a learning model and nice to point people to it and they can understand the why.

official/transformer/v2/transformer.py
guptapriya: Figured out why this was needed and changed the mode such that this is no longer required.

official/transformer/v2/metrics.py
guptapriya: Found another way to write metrics that doesn't need this hack. Bug filed to track the issue with the old approach.

official/transformer/v2/transformer_main.py
guptapriya: added a comment. I tried hard to write the input pipeline in a way that works in both cases but failed so far. A proper fix may need a change in keras i think. this should not be the end state..

isaprykin:
I don't get these test failures.  I wonder if they are stale or related.
Are there any unresolved comments on this PR before we merge?

PR #6886, opened 2019-05-27
Add static batch benchmarks to estimator

guptapriya:
So we can distinguish how much static vs dynamic batch matter.


guptapriya:
Thanks Toby!

PR #6903, opened 2019-05-28
Move dataset creation into the loop

guptapriya:
Because keeping it outside and iterating over it multiple times does not shuffle the data every iteration, which is what we ideally want.


guptapriya:
Closing as I messed up my branch.

PR #6904, opened 2019-05-28
Turn dist strat off for 1 GPU benchmarks

guptapriya:


official/transformer/v2/transformer_benchmark.py
```diff
@@ -133,6 +133,7 @@ def benchmark_1_gpu(self):
     """
     self._setup()
     FLAGS.num_gpus = 1
+    FLAGS.distribution_strategy = off
```
isaprykin: off is probably a string literal.  ;)

PR #6905, opened 2019-05-28
Add flag to use custom training loop for keras NCF model.

bfontain:


official/recommendation/ncf_keras_main.py
```diff
@@ -280,7 +280,7 @@ def run_ncf(_):
         beta_2=params["beta2"],
         epsilon=params["epsilon"])
 
-  if params['tensorflow_v2']:
+  if params['keras_use_ctl']:
     loss_object = tf.losses.SparseCategoricalCrossentropy(
```
guptapriya: maybe throw an explicit error somewhere if the user tries to use CTL with TF 1.x?

bfontain: Done

PR #6928, opened 2019-05-30
Bug fix of cifar10_eval.py

youngleox:
line 120: changed from eval_data = FLAGS.eval_data == 'test'
to: eval_data = FLAGS.eval_data
comment: the original code assigns 'true' to eval_data, when the script is being used to evaluate networks on the evaluation set, it DOES NOT load the evaluation set as intended, but actually loads the trainning set (line 105 of cifar10_input.py).


tfboyd:
Thank you, I confirmed the issue and this is also cleaner.  Double defaults are confusing even if done correctly.

PR #6936, opened 2019-05-31
Work-around to prevent OOM with --static_batch=False.

isaprykin:
Re-batching currently doesn't happen when the batch size isn't constant as is the case with the dynamic batch.  There is an internal fix submitted, but until that makes its ways to nightly and gets verified this work around allows to prevent OOM.

The resulting model is still not as fast as it could be, because the underlying tf.function gets retraced for every unique shape of the dynamic batch size.  I'm looking into that now.

### DIFF
--- a/official/transformer/v2/data_pipeline.py
+++ b/official/transformer/v2/data_pipeline.py
@@ -252,7 +252,7 @@ def _read_and_batch_from_files(
   else:
     # Group and batch such that each batch has examples of similar length.
     # TODO: _batch_examples might need to do something special for num_replicas.
-    dataset = _batch_examples(dataset, batch_size, max_length)
+    dataset = _batch_examples(dataset, batch_size // num_replicas, max_length)
 
   dataset = dataset.repeat(repeat)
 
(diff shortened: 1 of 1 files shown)


tfboyd:
@guptapriya  will have to approve.  This seems ok, but I do not know enough to approve this one as ok.

official/transformer/v2/data_pipeline.py
```diff
@@ -252,7 +252,7 @@ def _read_and_batch_from_files(
   else:
     # Group and batch such that each batch has examples of similar length.
     # TODO: _batch_examples might need to do something special for num_replicas.
-    dataset = _batch_examples(dataset, batch_size, max_length)
+    dataset = _batch_examples(dataset, batch_size // num_replicas, max_length)
```
guptapriya: Let's add a comment here to explain that this is temporary until the bug is fixed.

isaprykin:
I will hold off submitting this change.
At this point, it's unclear that it's valuable because Jiri's fix will be in tomorrow's nightly.  If the issue is still not resolved, then I'm going to submit this to unblock Transformer with dynamic batch.

PR #6950, opened 2019-06-03
Multi-worker ResNet Estimator benchmarks for PerfZero.

dubey:
This change refactors Resnet50EstimatorBenchmark into a base class and subclasses for single worker and multiple worker benchmarks.

### DIFF
--- a/official/resnet/estimator_benchmark.py
+++ b/official/resnet/estimator_benchmark.py
@@ -188,8 +188,8 @@ class Resnet50EstimatorAccuracy(EstimatorBenchmark):
                            top_1_max=0.766)
 
 
-class Resnet50EstimatorBenchmark(EstimatorBenchmark):
-  """Benchmarks for ResNet50 using Estimator."""
+class Resnet50EstimatorBenchmarkBase(EstimatorBenchmark):
+  """Base class for benchmarks for ResNet50 using Estimator."""
   local_flags = None
 
   def __init__(self, output_dir=None, default_flags=None):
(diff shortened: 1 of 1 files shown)


official/resnet/estimator_benchmark.py
```diff
@@ -342,6 +350,63 @@ def __init__(self, output_dir=None, root_data_dir=None, **kwargs):
...
 
+class Resnet50MultiWorkerEstimatorBenchmark(Resnet50EstimatorBenchmarkBase):
+  """Benchmarks for ResNet50 using Estimator with multiple workers."""
+
+  def __init__(self, output_dir=None, default_flags=None):
+    super(Resnet50MultiWorkerEstimatorBenchmark, self).__init__(
+        output_dir=output_dir,
+        default_flags=default_flags)
+
+  def benchmark_graph_fp16_8_gpu_ring_tweaked(self):
```
tfboyd: How do we know how many workers and total GPUs used?

dubey: The benchmark expects `TF_CONFIG` environment variable to be set with the list of workers and their IP addresses.

tfboyd:
Cool let's keep moving forward and deal with reporting when we have the
first result.  I do not want to slow you down on something we can sort out
later and faster when it is running.  Merge when ready.

guptapriya:
Thanks for the PR! Is there a plan to add real data and accuracy benchmark later on?

dubey:
Yes, we'll expand the set of benchmarks and also include 2.0 tests.  I'd like to get the end to end setup working first.

PR #6956, opened 2019-06-05
NCF distribution strategy updates

guptapriya:
two things in this PR:
- Loss needs to be scaled by global batch size. 
- Training input was being handled differently in fit vs CTL case. made them the same. will observe the difference in convergence from this change and decide which version to keep.


guptapriya:
Thanks @pavithrasv for the quick review!

PR #6961, opened 2019-06-05
TF 2.0 Transformer: Remove metrics hack

guptapriya:
The underlying issue has been fixed so we can now create and update metrics in cross replica context. 

Also fixed existing unit tests and added more tests which test distribution strategy codepaths as well.

### DIFF
--- a/official/transformer/v2/transformer_main_test.py
+++ b/official/transformer/v2/transformer_main_test.py
@@ -42,21 +42,19 @@ class TransformerTaskTest(tf.test.TestCase):
 
   def setUp(self):
     temp_dir = self.get_temp_dir()
-    FLAGS.model_dir = temp_dir
-    FLAGS.init_logdir_timestamp = FIXED_TIMESTAMP
+    FLAGS.model_dir = os.path.join(temp_dir, FIXED_TIMESTAMP)
     FLAGS.param_set = param_set = "tiny"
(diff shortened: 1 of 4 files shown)


official/transformer/v2/transformer_main_test.py
```diff
@@ -42,49 +42,50 @@ class TransformerTaskTest(tf.test.TestCase):
...
+    self.temp_dir = temp_dir
+    self.vocab_file = os.path.join(temp_dir, "vocab")
+    self.vocab_size = misc.get_model_params(FLAGS.param_set, 0)["vocab_size"]
+    self.bleu_source = os.path.join(temp_dir, "bleu_source")
+    self.bleu_ref = os.path.join(temp_dir, "bleu_ref")
 
   def _assert_exists(self, filepath):
     self.assertTrue(os.path.exists(filepath))
 
   def test_train(self):
```
tfboyd: #1 thank you for working on these tests.  

Question, do we need this test if we are not checking anything?  My assumption is we will catch exceptions via the other tests. Not remotely against the test just asking.

guptapriya: Hm, these tests are basically verifying that we get to the end of train without running into any errors. 
I don't know of any other tests that verify this for transformer, are there any?

PR #6965, opened 2019-06-05
Modify tweaked tests for better performance in no cloning mode

haoyuz:


official/resnet/keras/keras_imagenet_benchmark.py
```diff
@@ -139,9 +139,8 @@ def benchmark_8_gpu_mlperf_like_tweaked(self):
     FLAGS.dtype = 'fp16'
     FLAGS.enable_eager = True
     FLAGS.enable_xla = True
-    # Tweaks to improve performance.
-    FLAGS.data_delay_prefetch = True
     FLAGS.use_tensor_lr = True
+    FLAGS.tf_gpu_thread_mode = 'gpu_private'
```
guptapriya: Was this one just missing by mistake before?

official/resnet/keras/keras_imagenet_benchmark.py
```diff
@@ -768,8 +761,8 @@ def benchmark_xla_8_gpu_fp16_tweaked_optional_next(self):
         'benchmark_xla_8_gpu_fp16_tweaked_optional_next')
     FLAGS.batch_size = 256 * 8  # 8 GPUs
     FLAGS.use_tensor_lr = True
-    # FLAGS.tf_gpu_thread_mode = 'gpu_private'
-    FLAGS.data_delay_prefetch = True
+    FLAGS.tf_gpu_thread_mode = 'gpu_private'
+    FLAGS.datasets_num_private_threads = 48
```
guptapriya: hm, why do some of these tests need the threads setting explicitly but others don't?

official/resnet/keras/keras_imagenet_benchmark.py
haoyuz: I think we previously missed this one; @tfboyd Could you confirm?

official/resnet/keras/keras_imagenet_benchmark.py
haoyuz: When we enable `gpu_private` thread mode, it automatically sets `datasets_num_private_threads` to (num of CPU cores - 3 x num of GPUs)---for each GPU, 2 private launching threads and 1 event listener thread. However we actually need 48 cores for data pipeline to saturate our model. Setting them to introduce thread contention on 64 vCPUs.

haoyuz: For some tests (like the MLPerf-like one and `_delay_measure` ones), they target the 96 vCPU platform and we leave enough cores for data threads there.

PR #6980, opened 2019-06-07
Add benchmarks for custom training loops + tf.distribute

anj-s:
Add benchmarks for running ResNet50 on the ImageNet dataset using custom training loops.
- Benchmarks(accuracy and examples_per_second) added for 1 GPU, 8 GPUs and no distrat
- real and synthetic data use cases
- This does not include benchmarks for the no_function case due to an existing OOM error


official/resnet/ctl/ctl_imagenet_main.py
```diff
@@ -0,0 +1,297 @@
...
+from official.resnet.keras import resnet_model
+from official.utils.flags import core as flags_core
+from official.utils.logs import logger
+from official.utils.misc import distribution_utils
+from official.utils.misc import model_helpers
+from official.resnet.ctl import ctl_common
+
+
+# Imagenet training and test data sets.
+APPROX_IMAGENET_TRAINING_IMAGES = 1280000  # Approximate number of images.
```
tfboyd: Can we move this into keras_common or some common resnet module, even the other main.py?  I do not want to put code clean up on you but I suspect we have this value and the other values below in many locations.

official/resnet/ctl/ctl_imagenet_main.py
```diff
@@ -0,0 +1,297 @@
...
+            lr_epoch / warmup_end_epoch)
+  for mult, start_epoch in _LR_SCHEDULE:
+    if lr_epoch >= start_epoch:
+      learning_rate = _BASE_LEARNING_RATE * mult
+    else:
+      break
+  return learning_rate
+
+
+def run(flags_obj):
```
tfboyd: This function seems massive with all the inline functions.  Is this how we would suggest people do it?

official/resnet/ctl/ctl_benchmark.py
```diff
@@ -0,0 +1,88 @@
+# Copyright 2019 The TensorFlow Authors. All Rights Reserved.
```
tfboyd: I know this follows a pattern I helped push.  Can we remove this file and move the base class (if needed) into the main benchmark.  I created a [base class](https://github.com/tensorflow/models/blob/master/official/utils/testing/perfzero_benchmark.py) you can extend that does most of what this does.  [Example usage](https://github.com/tensorflow/models/blob/master/official/transformer/v2/transformer_benchmark.py)

official/resnet/ctl/ctl_imagenet_main.py
```diff
@@ -0,0 +1,295 @@
...
+
+      # Store the last train loss and accuracy calculated
+      stats['train_loss'] = train_loss.numpy()
+      stats['train_acc'] = training_accuracy.result().numpy()
+      training_accuracy.reset_states()
+
+      if (not flags_obj.skip_eval and
+          epoch % flags_obj.epochs_between_eval == 0):
+        test_iterator = iter(test_ds)
+        for step in range(steps_per_eval):
```
guptapriya: Can we write this as `for xyz in test_ds` ? should be equivalent but nicer and handle last batch.

official/resnet/ctl/ctl_imagenet_main.py
```diff
@@ -0,0 +1,295 @@
...
+        is_training=False,
+        data_dir=flags_obj.data_dir,
+        batch_size=flags_obj.batch_size,
+        num_epochs=flags_obj.train_epochs,
+        parse_record_fn=parse_record_keras,
+        dtype=dtype)
+
+    if strategy:
+      test_ds = strategy.experimental_distribute_dataset(test_ds)
+      steps_per_eval = IMAGENET_VALIDATION_IMAGES // flags_obj.batch_size
```
guptapriya: Let's try to use the last partial batch if possible.

guptapriya: Also, should steps calculation, if needed, be outside this if block?

official/resnet/ctl/ctl_imagenet_main.py
```diff
@@ -0,0 +1,295 @@
...
+      datasets_num_private_threads=flags_obj.datasets_num_private_threads,
+      dtype=dtype)
+
+  test_ds = None
+  if not flags_obj.skip_eval:
+    test_ds = input_fn(
+        is_training=False,
+        data_dir=flags_obj.data_dir,
+        batch_size=flags_obj.batch_size,
+        num_epochs=flags_obj.train_epochs,
```
guptapriya: i don't think you need to repeat the dataset N times, as you are creating a new iterator every epoch already

official/resnet/ctl/ctl_imagenet_main.py
```diff
@@ -0,0 +1,295 @@
...
+        dtype=dtype,
+        drop_remainder=True)
+  else:
+    input_fn = imagenet_main.input_fn
+
+  train_ds = input_fn(
+      is_training=True,
+      data_dir=flags_obj.data_dir,
+      batch_size=flags_obj.batch_size,
+      num_epochs=flags_obj.train_epochs,
```
guptapriya: Since you are creating a new iterator every epoch, you don't need the num_epochs.

However, this does bring up the shuffle-repeat issue. Since you're going through the dataset from the beginning every epoch, it would be going through the data in the same order every epoch. Does that still converge?

official/resnet/ctl/ctl_imagenet_main.py
```diff
@@ -0,0 +1,295 @@
...
+    training_accuracy = tf.keras.metrics.SparseCategoricalAccuracy(
+        'training_accuracy', dtype=tf.float32)
+    test_loss = tf.keras.metrics.Mean('test_loss', dtype=tf.float32)
+    test_accuracy = tf.keras.metrics.SparseCategoricalAccuracy(
+        'test_accuracy', dtype=tf.float32)
+    logging.info('Finished building Keras ResNet-50 model')
+
+    def train_step(train_ds_inputs):
+      """Training StepFn."""
+      def step_fn(inputs):
```
guptapriya: i think you could write this as `def step_fn(images, labels)` (and modify the call accordingly). Would that work or does it make something else ugly?

official/resnet/ctl/ctl_imagenet_main.py
```diff
@@ -0,0 +1,295 @@
...
+        """Per-Replica StepFn."""
+        images, labels = inputs
+        with tf.GradientTape() as tape:
+          logits = model(images, training=True)
+
+          prediction_loss = tf.keras.losses.sparse_categorical_crossentropy(
+              labels, logits)
+          loss1 = tf.reduce_sum(prediction_loss) * (1.0/ flags_obj.batch_size)
```
guptapriya: optional: maybe use the new utilities for loss scaling?

official/resnet/ctl/ctl_imagenet_main.py
```diff
@@ -0,0 +1,295 @@
...
+                               axis=None)
+      else:
+        return step_fn(train_ds_inputs)
+
+    def test_step(test_ds_inputs):
+      """Evaluation StepFn."""
+      def step_fn(inputs):
+        images, labels = inputs
+        logits = model(images, training=False)
+        loss = tf.keras.losses.sparse_categorical_crossentropy(labels,
```
guptapriya: i remember there was some discussion around which loss to report (both for training and eval): the one scaled by number of replicas or not. What does keras fit do? 
cc @pavithrasv

official/resnet/ctl/ctl_imagenet_main.py
```diff
@@ -0,0 +1,295 @@
...
+      if strategy:
+        strategy.experimental_run_v2(step_fn, args=(test_ds_inputs,))
+      else:
+        step_fn(test_ds_inputs)
+
+    if flags_obj.enable_function:
+      train_step = tf.function(train_step)
+      test_step = tf.function(test_step)
+
+    epoch_exp_per_sec = []
```
guptapriya: you can try using the time callback that we use in keras fit case. ncf ctl does this: 
https://github.com/tensorflow/models/blob/master/official/recommendation/ncf_keras_main.py#L356

Then you don't need to repeat a lot of code and you get the standard format.

Then you do this too: `keras_common.build_stats(..)`

official/resnet/ctl/ctl_imagenet_main.py
```diff
@@ -0,0 +1,295 @@
...
+    stats = {}
+    for epoch in range(train_epochs):
+      logging.info('Starting to run epoch: %s', epoch)
+      train_iterator = iter(train_ds)
+
+      step = 0
+      total_loss = 0.0
+      batch_exp_per_sec = []
+      for step in range(train_steps):
```
guptapriya: any reason to not write this as the nicer version `for xyz in train_ds` ?

official/resnet/ctl/ctl_imagenet_main.py
```diff
@@ -0,0 +1,295 @@
...
+            epoch + 1 + (float(step) / steps_per_epoch))
+        optimizer.lr = learning_rate
+        start_time = time.time()
+        total_loss += train_step(next(train_iterator))
+        end_time = time.time()
+        elapsed_time = end_time - start_time
+        samples_per_sec = flags_obj.batch_size / elapsed_time
+        # We skip the first step for warmup purposes. We can
+        # add a flag for tuning this.
+        if step > 0:
```
guptapriya: there is already some logic somewhere for warmup etc, can we follow the same? ( i think it skips more than 1 step)

official/resnet/ctl/ctl_imagenet_main.py
```diff
@@ -0,0 +1,295 @@
...
+          "Training Metric: {'epoch':%d, 'examples_per_second': %f}" %
+          (epoch, epoch_exp_per_sec[epoch]))
+
+      # Store the last train loss and accuracy calculated
+      stats['train_loss'] = train_loss.numpy()
+      stats['train_acc'] = training_accuracy.result().numpy()
+      training_accuracy.reset_states()
+
+      if (not flags_obj.skip_eval and
+          epoch % flags_obj.epochs_between_eval == 0):
```
guptapriya: this would eval after the first epoch, even if epochs_between_eval = N>1. Is that what we want? I dont know what Keras fit does, but maybe we want to do only after the first N epochs.

official/resnet/ctl/ctl_imagenet_main.py
```diff
@@ -0,0 +1,295 @@
...
+        test_iterator = iter(test_ds)
+        for step in range(steps_per_eval):
+          test_step(next(test_iterator))
+        logging.info('Test loss: %s, accuracy: %s%%',
+                     test_loss.result().numpy(),
+                     test_accuracy.result().numpy())
+        stats['accuracy_top_1'] = test_accuracy.result().numpy()
```
guptapriya: This doesn't need to be done in the epoch loop, right?

official/resnet/ctl/ctl_common.py
```diff
@@ -0,0 +1,55 @@
...
+FLAGS = flags.FLAGS
+
+
+def define_ctl_flags():
+  """Define flags for CTL."""
+
+  flags.DEFINE_boolean(name='enable_function', default=True,
+      help='Enable tf.function')
+  flags.DEFINE_boolean(name='skip_eval', default=False, help='Skip evaluation?')
+  flags.DEFINE_boolean(name='report_accuracy_metrics', default=True,
```
guptapriya: is this a flag we have for other examples too?

official/resnet/ctl/ctl_common.py
```diff
@@ -0,0 +1,55 @@
...
+                       help='Report metrics during training and evaluation.')
+  flags.DEFINE_integer(
+      name='train_steps', default=None,
+      help='The number of steps to run for training. If it is larger than '
+      '# batches per epoch, then use # batches per epoch. When this flag is '
+      'set, only one epoch is going to run for training.')
+  # This is used when we create a synth dataset.
+  flags.DEFINE_boolean(
```
guptapriya: this is not yet used right, or is it?

official/resnet/ctl/ctl_imagenet_main.py
```diff
@@ -0,0 +1,295 @@
...
+  train_ds = input_fn(
+      is_training=True,
+      data_dir=flags_obj.data_dir,
+      batch_size=flags_obj.batch_size,
+      num_epochs=flags_obj.train_epochs,
+      parse_record_fn=parse_record_keras,
+      datasets_num_private_threads=flags_obj.datasets_num_private_threads,
+      dtype=dtype)
+
+  test_ds = None
```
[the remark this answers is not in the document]
anj-s: Done.

official/resnet/ctl/ctl_imagenet_main.py
anj-s: i do need to unpack the inputs somewhere. I think we can make that change when we iterate over the distributed dataset inside the tf.function.

official/resnet/ctl/ctl_imagenet_main.py
anj-s: Will add it in a follow up PR when I add more benchmarks.

official/resnet/ctl/ctl_imagenet_main.py
anj-s: There were outstanding issues when I started working on this PR so I chose this path. I intend to add this use case as well.

official/resnet/ctl/ctl_imagenet_main.py
anj-s: yes, makes sense. Done.

official/resnet/ctl/ctl_imagenet_test.py
```diff
@@ -0,0 +1,122 @@
...
+        extra_flags=extra_flags
+    )
+
+  def test_end_to_end_2_gpu(self):
+    """Test Keras model with 2 GPUs."""
+    if context.num_gpus() < 2:
+      self.skipTest(
+          "{} GPUs are not available for this test. {} GPUs are available".
+          format(2, context.num_gpus()))
+    tf.enable_v2_behavior()
```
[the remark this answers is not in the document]
anj-s: I agree. I added a condition where we only set it if we are running TF 1.0 binary.

official/resnet/ctl/ctl_imagenet_main.py
```diff
@@ -0,0 +1,297 @@
...
+    (1.0, 5), (0.1, 30), (0.01, 60), (0.001, 80)
+]
+_BASE_LEARNING_RATE = 0.4
+
+
+def compute_learning_rate(lr_epoch):
+  """Learning rate for each step."""
+  warmup_lr_multiplier, warmup_end_epoch = _LR_SCHEDULE[0]
+  if lr_epoch < warmup_end_epoch:
+    # Learning rate increases linearly per step.
```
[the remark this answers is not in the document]
anj-s: tfboyd@ Ah the function's argument is incorrectly named. We are calculating learning rate per step for the first 5 epochs. We pass the epoch + 1 + step/steps_per_epoch as an argument to this function. 
Changed to using learning_rate_schedule to be consistent with keras.

official/resnet/ctl/ctl_imagenet_main.py
```diff
@@ -0,0 +1,297 @@
...
+  # Keras model.
+  label = tf.cast(tf.cast(tf.reshape(label, shape=[1]), dtype=tf.int32) - 1,
+                  dtype=tf.float32)
+  return image, label
+
+# Learning rate schedule
+_LR_SCHEDULE = [    # (multiplier, epoch to start) tuples
+    (1.0, 5), (0.1, 30), (0.01, 60), (0.001, 80)
+]
+_BASE_LEARNING_RATE = 0.4
```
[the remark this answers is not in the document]
anj-s: Done.

official/resnet/ctl/ctl_imagenet_main.py
anj-s: I think Keras uses imagenet_main so I've done the same for these benchmarks as well. We should move this into a common repo. Right now we need a resnet_common.py which can be shared by keras, ctl and estimator.

official/resnet/ctl/ctl_imagenet_main.py
anj-s: Moved a few things around. Let me know if this helps simplify things a bit.

official/resnet/ctl/ctl_benchmark.py
anj-s: Done. We might need to update keras to do the same. I can do that in a follow up PR.

official/resnet/ctl/ctl_imagenet_main.py
```diff
@@ -0,0 +1,295 @@
...
+from __future__ import print_function
+
+import os
+import tempfile
+import datetime
+import time
+
+from absl import app as absl_app
+from absl import flags
+# from absl import logging
```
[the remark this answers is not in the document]
anj-s: Hmm...I might have had some version issues early on. Uncommented now.

official/resnet/ctl/ctl_imagenet_main.py
```diff
@@ -0,0 +1,295 @@
...
+        loss = tf.reduce_sum(loss) * (1.0/ flags_obj.batch_size)
+        test_loss.update_state(loss)
+        test_accuracy.update_state(labels, logits)
+
+      if strategy:
+        strategy.experimental_run_v2(step_fn, args=(test_ds_inputs,))
+      else:
+        step_fn(test_ds_inputs)
+
+    if flags_obj.enable_function:
```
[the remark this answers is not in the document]
anj-s: I think use_tf_function sounds better. I would prefer that over run_eagerly but if you think it makes more sense for us to be consistent with Keras I am happy to change it to run_eagerly.

official/resnet/ctl/ctl_imagenet_main.py
anj-s: There were a couple of issues when I started this PR so I went with this approach. The goal is to add benchmarks for this use case as well.

official/resnet/ctl/ctl_imagenet_main.py
anj-s: ah yes, removed.

official/resnet/ctl/ctl_imagenet_main.py
anj-s: Can we do that in a follow up PR? 
Moved the steps calculation to another function.

official/resnet/ctl/ctl_imagenet_main.py
anj-s: This did converge because there was a shuffle but no repeat. Wouldn't that ensure that the input is randomized every epoch? Or did you mean that if I repeat() with shuffle I would get the same set of random input every epoch?

official/resnet/ctl/ctl_imagenet_main.py
anj-s: Done.

official/resnet/ctl/ctl_imagenet_main.py
anj-s: Done.

official/resnet/ctl/ctl_imagenet_main.py
anj-s: Done.

official/resnet/ctl/ctl_common.py
anj-s: removed. CTL does not need this.

official/resnet/ctl/ctl_common.py
anj-s: I think in one of my versions it was needed to figure out if we need to drop_remainder but I now set it to the default value.

anj-s:
Thank you for the comments @guptapriya and @tfboyd ! PTAL.

official/resnet/ctl/ctl_imagenet_main.py
anj-s: This reports a loss value that is equal to what the user would see if they did not use distribution strategy. I followed the same process that we suggested to our users in the tutorials. 
@pavithrasv can maybe talk more about loss reporting in keras.

pavithrasv: fit reports the value scaled by number of replicas. Do we want to change that in Keras callbacks?

anj-s: Does this mean that if a user uses DistributionStrategy they will see different loss values? will that be confusing to the user? 
Is it a big change to report the prediction loss instead of the scaled one?

official/resnet/ctl/ctl_imagenet_main.py
```diff
@@ -25,9 +25,9 @@
 
 from absl import app as absl_app
 from absl import flags
-# from absl import logging
+from absl import logging
 
-import tensorflow.compat.v1.logging as logging
+# import tensorflow.compat.v1.logging as logging
```
guptapriya: remove this?

official/resnet/ctl/ctl_imagenet_main.py
```diff
@@ -60,58 +52,41 @@ def parse_record_keras(raw_record, is_training, dtype):
...
   Args:
-    flags_obj: An object containing parsed flag values.
-
-  Raises:
-    ValueError: If fp16 is passed as it is not currently supported.
-
+    loss: The final loss at training time.
+    eval_result: Output of the eval step. Assumes first value is eval_loss and
+      second value is accuracy_top_1.
+    time_callback: Time tracking callback likely used during keras.fit.
```
guptapriya: can remove the "used during keras.fit"

official/resnet/ctl/ctl_imagenet_main.py
guptapriya: i see. i thought most of the issues were with putting the iteration inside the tf.function, didn't know there were issues with it being outside too. 
Lets have multiple variations for the performance benchmarks. For the accuracy benchmark, we can just pick one variation perhaps? ideally it is one which handles last partial batch correctly so we get the eval on the entire dataset.

official/resnet/ctl/ctl_imagenet_main.py
```diff
@@ -127,50 +102,87 @@ def run(flags_obj):
...
+
+def get_num_train_iterations(flags_obj):
+  steps_per_epoch = imagenet_main.NUM_IMAGES['train'] // flags_obj.batch_size
   train_epochs = flags_obj.train_epochs
 
   if flags_obj.train_steps:
     train_steps = min(flags_obj.train_steps, steps_per_epoch)
     train_epochs = 1
 
+  steps_per_eval = imagenet_main.NUM_IMAGES['validation'] // flags_obj.batch_size
```
guptapriya: `steps_per_eval` name seems a bit confusing to me, perhaps just be `eval_steps`. 
steps_per_eval is sometimes used to describe how many training steps to do between 2 evals..

official/resnet/ctl/ctl_imagenet_main.py
```diff
@@ -127,50 +102,87 @@ def run(flags_obj):
...
+      distribution_strategy=flags_obj.distribution_strategy,
+      num_gpus=flags_obj.num_gpus,
+      num_workers=distribution_utils.configure_cluster(),
+      all_reduce_alg=flags_obj.all_reduce_alg,
+      num_packs=flags_obj.num_packs)
+
+  train_ds, test_ds = get_input_dataset(flags_obj, strategy)
+  train_steps, train_epochs, steps_per_eval = get_num_train_iterations(flags_obj)
+
+  time_callback = keras_utils.TimeHistory(flags_obj.batch_size, 1)
```
guptapriya: The second argument should take the flag value? FLAGS.log_steps

official/resnet/ctl/ctl_imagenet_main.py
```diff
@@ -215,70 +228,44 @@ def step_fn(inputs):
...
       test_step = tf.function(test_step)
 
-    epoch_exp_per_sec = []
-    stats = {}
+    time_callback.on_train_begin()
     for epoch in range(train_epochs):
-      logging.info('Starting to run epoch: %s', epoch)
       train_iterator = iter(train_ds)
 
       step = 0
```
guptapriya: i am confused why there seem to be 2 things called `step` here..

official/resnet/ctl/ctl_imagenet_main.py
```diff
@@ -215,70 +228,44 @@ def step_fn(inputs):
...
-        start_time = time.time()
-        total_loss += train_step(next(train_iterator))
-        end_time = time.time()
-        elapsed_time = end_time - start_time
-        samples_per_sec = flags_obj.batch_size / elapsed_time
-        # We skip the first step for warmup purposes. We can
-        # add a flag for tuning this.
-        if step > 0:
-          batch_exp_per_sec.append(samples_per_sec)
+        training_accuracy.reset_states()
```
guptapriya: hm, do you want to reset accuracy every step? usually it's at end of every epoch

official/resnet/keras/keras_common.py
```diff
@@ -38,6 +38,42 @@
 TRAIN_TOP_1 = 'training_accuracy_top_1'
 
 
+LR_SCHEDULE = [    # (multiplier, epoch to start) tuples
```
guptapriya: hm, i think this schedule and this warmup process is only applicable to imagenet.. maybe you can just use it from there?

official/resnet/ctl/ctl_imagenet_main.py
guptapriya: Hmm i think this is still slightly wrong. if epochs_between_eval = 3, you want to eval at epoch #2, #5 etc, assuming you start at 0. so what you need it `(epoch + 1) % epochs_between_eval == 0` (basically coz of the zero indexing of epoch index). would be good to confirm what keras does.

official/resnet/ctl/ctl_imagenet_main.py
```diff
@@ -215,70 +228,44 @@ def step_fn(inputs):
...
-            (epoch, test_accuracy.result().numpy()))
         test_loss.reset_states()
         test_accuracy.reset_states()
 
-  stats['exp_per_second'] = np.mean(epoch_exp_per_sec)
-  return stats
+        test_iterator = iter(test_ds)
+        for step in range(steps_per_eval):
+          test_step(next(test_iterator))
+
```
guptapriya: Would be good to print some stuff every epoch/every few steps or so? Is anything being printed now at all?

official/resnet/ctl/ctl_imagenet_benchmark.py
```diff
@@ -33,7 +33,50 @@
...
+      top_1_min: lowest passing level for top_1 accuracy.
+    """
+
+    metrics = []
+    if 'eval_acc' in stats:
+      metrics.append({'name': 'accuracy_top_1',
+                      'value': stats['eval_acc'],
+                      'min_value': top_1_min,
+                      'max_value': top_1_max})
+      metrics.append({'name': 'top_1_train_accuracy',
```
guptapriya: this seems wrongly setting loss value for accuracy metric?

official/resnet/ctl/ctl_imagenet_benchmark.py
```diff
@@ -33,7 +33,50 @@
...
+    """Report benchmark results by writing to local protobuf file.
+
+    Args:
+      stats: dict returned from keras models with known entries.
+      wall_time_sec: the during of the benchmark execution in seconds
+      top_1_max: highest passing level for top_1 accuracy.
+      top_1_min: lowest passing level for top_1 accuracy.
+    """
+
+    metrics = []
```
guptapriya: Does train loss also need to be set somewhere or is that common?

official/resnet/ctl/ctl_imagenet_benchmark.py
```diff
@@ -85,7 +128,7 @@ def _get_model_dir(self, folder_name):
     return os.path.join(self.output_dir, folder_name)
 
 
-class Resnet50CtlBenchmarkBase(ctl_benchmark.CtlBenchmark):
+class Resnet50CtlBenchmarkBase(CtlBenchmark):
```
guptapriya: these tests should set the warmup steps, otherwise we may see some slow perf in the beginning?

official/resnet/ctl/ctl_imagenet_test.py
```diff
@@ -119,4 +116,6 @@ def test_end_to_end_2_gpu(self):
     )
 
 if __name__ == '__main__':
+  if not keras_common.is_v2_0():
+    tf.enable_v2_behavior()
```
guptapriya: still suspicious of this.. ideally we just have one version of the test and run that with whatever TF is appropriate. i think for now this is nice becasue we are not running these tests with 2.0 yet. once we do, we can remove this from here and just skip the tests in 1.x?

official/resnet/ctl/ctl_imagenet_test.py
```diff
@@ -83,7 +82,6 @@ def test_end_to_end_1_gpu(self):
       self.skipTest(
```
guptapriya: we are probably missing this in other tests too, but instad of skipping, i think we can just run the same test with 1 CPU instead? (pass num_gpus=0). that way we get some coverage. 

also, after checking this in, remember to add internal BUILD files so that they run internally as well.

official/resnet/ctl/ctl_imagenet_main.py
guptapriya: so there is an issue with shuffle currently that without a repeat, it doesn't really shuffle every time you create an iterator on the same dataset. rohan is fixing it but i dont think it is yet done. 

i can explain in person tomorrow, it's a bit complex to write down =)

official/resnet/ctl/ctl_imagenet_main.py
guptapriya: ideally, we report the same loss in fit and ctl and with and without strategy.
for this PR, it's fine but let's try to come to some consensus on this topic as users do get confused when they say different loss being reported.

official/resnet/ctl/ctl_imagenet_main.py
anj-s: Done.

official/resnet/ctl/ctl_imagenet_main.py
anj-s: Done.

official/resnet/ctl/ctl_imagenet_main.py
anj-s: I just tried making the change you suggested and ran into issue. Will follow up and have that version as part of these benchmarks.

official/resnet/ctl/ctl_imagenet_main.py
anj-s: Keras starts from epoch + 1 in its calculation of validation frequency. I've done the same here as well.
why do you think evaluating at 3, 6 is incorrect? We do 4 epochs the first time(only because we skip 0) but other than that what is the value in doing 2, 5? maybe I am missing something here.

official/resnet/ctl/ctl_imagenet_main.py
anj-s: Done.

official/resnet/ctl/ctl_imagenet_main.py
anj-s: Done.

official/resnet/ctl/ctl_imagenet_main.py
anj-s: I think the first one was added when I was iterating over the distributed dataset. I removed that version since there were a few outstanding issues at that time.

official/resnet/ctl/ctl_imagenet_benchmark.py
anj-s: ah yes. reworked this a bit.

official/resnet/ctl/ctl_imagenet_benchmark.py
anj-s: Done.

official/resnet/ctl/ctl_imagenet_main.py
anj-s: Done.

official/resnet/ctl/ctl_imagenet_main.py
anj-s: Wrong indentation during refactor. Done.

official/resnet/ctl/ctl_imagenet_main.py
anj-s: We log the time at the end of training so I wasn't sure if we should be logging anything. However this is not used for the perf number calculation(only for reporting test run time on the dashboard) so I added logging back.

official/resnet/ctl/ctl_imagenet_benchmark.py
anj-s: Done.

official/resnet/ctl/ctl_imagenet_test.py
anj-s: Sure. we can do that.

official/resnet/ctl/ctl_imagenet_test.py
anj-s: Done.
Will follow up on this PR and add internal BUILD files so that we run this internally.

official/resnet/ctl/ctl_imagenet_main.py
```diff
@@ -0,0 +1,297 @@
...
+                     epoch)
+
+    time_callback.on_train_end()
+
+    eval_result = None
+    train_result = None
+    if not flags_obj.skip_eval:
+      eval_result = [test_loss.result().numpy(),
+                     test_accuracy.result().numpy()]
+      train_result = [train_loss.numpy(),
```
guptapriya: nit: setting train_result can be outside the if block?

official/resnet/ctl/ctl_imagenet_test.py
```diff
@@ -0,0 +1,121 @@
...
+    ]
+    extra_flags = extra_flags + self._extra_flags
+
+    integration.run_synthetic(
+        main=ctl_imagenet_main.run,
+        tmp_root=self.get_temp_dir(),
+        extra_flags=extra_flags
+    )
+
+  def test_end_to_end_1_gpu(self):
```
guptapriya: remove the 1 gpu maybe since it can run with and without?

official/resnet/ctl/ctl_imagenet_test.py
```diff
@@ -0,0 +1,121 @@
...
+        main=ctl_imagenet_main.run,
+        tmp_root=self.get_temp_dir(),
+        extra_flags=extra_flags
+    )
+
+  def test_end_to_end_2_gpu(self):
+    """Test Keras model with 2 GPUs."""
+    num_gpus = "2"
+    if context.num_gpus() < 2:
+      num_gpus = "0"
```
guptapriya: don't need this one here, since it would be exactly the same as the previous one if no gpus?

official/resnet/ctl/ctl_common.py
```diff
@@ -0,0 +1,48 @@
...
+                       help='Wrap the train and test step inside a '
+                       'tf.function.')
+  flags.DEFINE_boolean(name='skip_eval', default=False, help='Skip evaluation?')
+  flags.DEFINE_integer(
+      name='train_steps', default=None,
+      help='The number of steps to run for training. If it is larger than '
+      '# batches per epoch, then use # batches per epoch. When this flag is '
+      'set, only one epoch is going to run for training.')
+  flags.DEFINE_integer(
+      name='warmup_steps', default=1,
```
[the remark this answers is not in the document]
anj-s: Removed.

official/resnet/ctl/ctl_imagenet_main.py
anj-s: True but that would mean useless values would be populated when we don't run the Accuracy test. Hence moved it together with the eval results.

official/resnet/ctl/ctl_imagenet_test.py
anj-s: Done. I think the no DS and 2 GPU case should suffice.

official/resnet/ctl/ctl_imagenet_test.py
anj-s: I'll have to keep this since I removed the previous test.

PR #6985, opened 2019-06-09
Properly use steps_per_epoch and num_epoch in BERT classify benchmark

robieta:
The super `__init__` call was overriding steps_per_epoch and num_epochs to None, so this moves them after.


robieta:
Done. (Sorry, this got lost in my inbox)

I'm a bit behind on procedures for this repo. Is it now using the same sync process as tensorflow/tensorflow?

saberkun:
Yes. It has manual two-way sync. Only when there are conflict checked-in changes on one side will cause problem. I feel this particular PR is good.

PR #6989, opened 2019-06-10
Code cleanup.

rxsang:


rxsang:
> thanks for cleaning up! have you had a chance to verify that this does indeed work?

This actually falls back to the previous behavior, as we actually disables the logic that handles the zero batch in batch norm, so those code won't make a difference. So this change won't make any differences.

PR #6993, opened 2019-06-11

saberkun:
252534787  by hongkuny<hongkuny@google.com>:
    
    Transformer vocab fix to strip correctly in py2


saberkun:
CL has been checked in internally.

PR #6999, opened 2019-06-11
fix OneDeviceStrategy error

christk1:
fix AttributeError: module 'tensorflow._api.v1.distribute' has no attribute 'OneDeviceStrategy' when running models/official/mnist/mnist.py

### DIFF
--- a/official/utils/misc/distribution_utils.py
+++ b/official/utils/misc/distribution_utils.py
@@ -127,12 +127,12 @@ def get_distribution_strategy(distribution_strategy="default",
   if (distribution_strategy == "one_device" or
       (distribution_strategy == "default" and num_gpus <= 1)):
     if num_gpus == 0:
-      return tf.distribute.OneDeviceStrategy("device:CPU:0")
+      return tf.contrib.distribute.OneDeviceStrategy("device:CPU:0")
     else:
       if num_gpus > 1:
         raise ValueError("`OneDeviceStrategy` can not be used for more than "
                          "one device.")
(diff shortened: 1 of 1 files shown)


christk1:
I signed it!

tfboyd:
I have a feeling this is a version issue as contrib is not needed.  What version of TensorFlow are you using?  I run tests using the util on TF 1.0 and TF 2.0 on multiple scenarios nightly.  My guess is you have TF 1.13 or maybe earlier.  I forget when the change took place.

tfboyd:
There are tags by version that should work with your version of TensorFlow.  Example below is v1.13.
https://github.com/tensorflow/models/tree/v1.13.0

christk1:
i am using master version and tf.distribute.OneDeviceStrategy is imported only after r1.14rc

tfboyd:
It may take me time to test the RC and to be honest I might not get time.  I thought it was cut well after these changes occurred.   I would use the tagged version for 1.13 above as that used contrib.  We are removing contrib from the examples and adding it back will break the nightly tests as we run both TF 2.0 and TF 2.0 on the same code.  If I get a chance to test with the RC I will report back.

tfboyd:
Or you can install tf-nightly and that will fix the issue.  
https://pypi.org/project/tf-nightly-gpu/

guptapriya:
Yes, we are trying to keep models working with TF nightly, so it's possible the latest version does not work with older versions of TF. Should this PR be closed?

PR #7038, opened 2019-06-17
adding a new perf test for ncf, and changing some names

nnigania:


tfboyd:
Will try to take a look early today so we can get this setup and running before the next sync on Wed 19-JUN

official/recommendation/ncf_keras_benchmark.py
```diff
@@ -213,3 +213,50 @@ def benchmark_2_gpus(self):
     self._setup()
     FLAGS.num_gpus = 2
     self._run_and_report_benchmark()
+
+class NCFKerasPerf(NCFKerasBenchmarkBase):
```
tfboyd: r/NCFKerasPerf/NCFKerasBenchmark

NCF "breaks" our pattern in that "Benchmark" is often just a few steps.  For ResNet50 I put the "mlperf like" tests under accuracy.  You could also choose to do that and not have this Class at all.  My  guideline, for myself, is if the test is "long" it falls under accuracy; which is a false assumption I choose to ignore for now.  If the test is testing examples/sec and short (a few steps) then it is a benchmark test.  What you pick is fine with me as it is a grey area that is my fault.

official/recommendation/ncf_keras_benchmark.py
```diff
@@ -213,3 +213,50 @@ def benchmark_2_gpus(self):
...
+        default_flags=default_flags,
+        **kwargs)
+
+  def _extract_benchmark_report_extras(self, stats):
+    metrics = []
+    metrics.append({'name': 'exp_per_second',
+                    'value': stats['avg_exp_per_second']})
+    return metrics
+
+  def benchmark_1_gpu_ctl_perf(self):
```
tfboyd: r/benchmark_1_gpu_ctl_perf/benchmark_1_gpu_ctl_mlperf_like

Add a docstring:
https://github.com/tensorflow/models/blob/master/official/resnet/keras/keras_imagenet_benchmark.py#L120

official/recommendation/ncf_keras_benchmark.py
```diff
@@ -213,3 +213,50 @@ def benchmark_2_gpus(self):
...
+    self._setup()
+    FLAGS.keras_use_ctl = True
+    self._run_and_report_benchmark()
+
+  def benchmark_1_gpu_perf(self):
```
tfboyd: r/benchmark_1_gpu_perf/benchmark_1_gpu_mlperf_like

Add a comment about being mlperf like.  I did something similar in ResNet that I thought was a nice reminder to anyone that sees it as to what the test is doing and why.  
https://github.com/tensorflow/models/blob/master/official/resnet/keras/keras_imagenet_benchmark.py#L120

nnigania:
Please review the new changes I made.
To reproduce a test using the movielens data from gcp bucket, below is an example command:

python3 /google/data/ro/teams/tensorflow-performance/perfzero/perfzero.py  --benchmark_methods=official.recommendation.ncf_keras_benchmark.NCFKerasAccuracy.benchmark_1_gpu_ctl_mlperf_like --python_path=models --git_repos="https://github.com/nnigania/models;ncf_perfzero_new_test" --num_gpus=1 --dockerfile_path=docker/Dockerfile_ubuntu_1804_tf_v2 --label=manual --data_downloads=gs://tf-perfzero-data/movielens_data

PR #7048, opened 2019-06-18
Add XLA to transformer

tfboyd:
- Moved config logic to keras_utils
- Added enable_xla flag to _performance flags
- refactored flags for keras resnet
- created common set_session_config

### DIFF
--- a/official/resnet/keras/keras_cifar_main.py
+++ b/official/resnet/keras/keras_cifar_main.py
@@ -28,6 +28,7 @@ from official.resnet.keras import resnet_cifar_model
 from official.utils.flags import core as flags_core
 from official.utils.logs import logger
 from official.utils.misc import distribution_utils
+from official.utils.misc import keras_utils
 
 
 LR_SCHEDULE = [  # (multiplier, epoch to start) tuples
@@ -98,17 +99,7 @@ def run(flags_obj):
   Returns:
     Dictionary of training and eval stats.
   """
-  # TODO(tobyboyd): Remove eager flag when tf 1.0 testing ends.
-  # Eager is default in tf 2.0 and should not be toggled
-  if keras_common.is_v2_0():
-    keras_common.set_config_v2()
-  else:
-    config = keras_common.get_config_proto_v1()
-    if flags_obj.enable_eager:
-      tf.compat.v1.enable_eager_execution(config=config)
-    else:
-      sess = tf.Session(config=config)
-      tf.keras.backend.set_session(sess)
+  keras_utils.set_session_config(enable_eager=flags_obj.enable_eager)
 
   dtype = flags_core.get_tf_dtype(flags_obj)
   if dtype == 'fp16':
--- a/official/resnet/keras/keras_common.py
+++ b/official/resnet/keras/keras_common.py
@@ -13,7 +13,6 @@
 # limitations under the License.
 # ==============================================================================
 """Common util functions and classes used by both keras cifar and imagenet."""
-
 from __future__ import absolute_import
 from __future__ import division
 from __future__ import print_function
@@ -29,7 +28,6 @@ import tensorflow as tf
 
 from official.utils.misc import keras_utils
 # pylint: disable=ungrouped-imports
-from tensorflow.core.protobuf import rewriter_config_pb2
 from tensorflow.python.keras.optimizer_v2 import (gradient_descent as
                                                   gradient_descent_v2)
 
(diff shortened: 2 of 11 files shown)


official/resnet/keras/keras_common.py
```diff
@@ -419,8 +375,10 @@ def data_delay_prefetch():
 
 
 def set_cudnn_batchnorm_mode():
-  """Set CuDNN batchnorm mode for better performance. Note that the spatial
-     persistent mode may lead to accuracy losses for certain models."""
+  """Set CuDNN batchnorm mode for better performance. Note that the spatial.
```
reedwm: Don't split the sentence starting with "Note that the spatial". Instead, format like:

```
"""Set CuDNN batchnorm mode for better performance.

Note that the spatial persistent mode may lead to accuracy losses for certain
models.
"""

File: official/transformer/transformer_estimator_benchmark.py [hunk 2]
```diff
@@ -100,10 +101,12 @@ def _report_benchmark(self,
       exp_per_sec = sum(exp_per_second_list) / (len(exp_per_second_list))
       metrics.append({'name': 'exp_per_second',
                       'value': exp_per_sec})
-    self.report_benchmark(
-        iters=eval_results['global_step'],
-        wall_time=wall_time_sec,
-        metrics=metrics)
+
+    flags_str = flags_core.get_nondefault_flags_as_str()
```
reedwm: Thanks for adding FLAGs to the estimator version!

Now the only thing left (that I really want) is the Resnet keras version.

official/utils/flags/_performance.py
```diff
@@ -270,4 +271,9 @@ def _check_fp16_implementation(flags_dict):
             "Whether to enable tf.data's `experimental_slack` option.")
     )
 
+  if enable_xla:
+    flags.DEFINE_boolean(
```
reedwm: Maybe also have the Keras Resnet model use this version of the flag, so we don't have two definitions of the `enable_xla` flag.

official/transformer/v2/transformer_main.py
```diff
@@ -121,6 +122,12 @@ def __init__(self, flags_obj):
   def train(self):
     """Trains the model."""
     params, flags_obj, is_train = self.params, self.flags_obj, True
+    # Sets config options.
+    keras_utils.set_config_v2(
```
reedwm: I think the transformer model supports TF 1. A simple fix would be that if TF1 is used, and either enable_xla==False or enable_grappler_layout_optimizer==True, then raise an error saying those flags are only supported with TF 2.

official/utils/flags/_performance.py
tfboyd: I'll follow your "hack" (pattern) and add it in.  I really want to fix the flags correctly for keras.  I do not like that we call into run_loops to make the flags.

official/transformer/v2/transformer_main.py
tfboyd: I will add in the TF 1.0 config code.  I was not sure this works with TF 1.0.  I might have been thinking of the estimator version not being upgraded for TF 2.0.

official/resnet/keras/keras_common.py
tfboyd: Totally not thinking on my part.

official/utils/flags/_performance.py
reedwm: Yeah, the pattern is pretty ugly. It's why I used the word "Maybe" in this comment :).

official/resnet/keras/keras_common.py
tfboyd: fixed

official/transformer/v2/transformer_main.py
tfboyd: I refactored this entirely.  I believe I have thought it through and saved a lot of lines of code and made this easier to reuse.

official/utils/flags/_performance.py
tfboyd: Fixed, I think it is slightly better and ready for a bigger change soon.

official/resnet/keras/keras_cifar_main.py
```diff
@@ -98,17 +99,7 @@ def run(flags_obj):
...
-  if keras_common.is_v2_0():
-    keras_common.set_config_v2()
-  else:
-    config = keras_common.get_config_proto_v1()
-    if flags_obj.enable_eager:
-      tf.compat.v1.enable_eager_execution(config=config)
-    else:
-      sess = tf.Session(config=config)
-      tf.keras.backend.set_session(sess)
+  keras_utils.set_session_config(enable_eager=flags_obj.enable_eager)
```
reedwm: Why not pass `enable_xla` here?

official/utils/misc/keras_utils.py
```diff
@@ -128,3 +129,73 @@ def on_batch_end(self, batch, logs=None):
...
+        enable_xla=enable_xla,
+        enable_grappler_layout_optimizer=enable_grappler_layout_optimizer)
+    if enable_eager:
+      tf.compat.v1.enable_eager_execution(config=config)
+    else:
+      sess = tf.Session(config=config)
+      tf.keras.backend.set_session(sess)
+
+
+def get_config_proto_v1(enable_xla=False,
```
reedwm: For this function and below, don't give default values to arguments.

official/transformer/v2/misc.py
```diff
@@ -69,13 +69,26 @@ def define_transformer_flags():
...
+           'convolutions and batch normalizations, and this flag allows to '
+           'disable it.'
+  )
+
   flags_core.define_benchmark()
   flags_core.define_device(tpu=True)
 
   flags.DEFINE_integer(
-      name='train_steps', short_name='ts', default=None,
+      name='train_steps', short_name='ts', default=300000,
```
reedwm: Why change this default value?

official/transformer/v2/misc.py
```diff
@@ -69,13 +69,26 @@ def define_transformer_flags():
...
+      enable_xla=True
   )
+
+  # Additional performance flags
+  # TODO(b/76028325): Remove when generic layout optimizer is ready.
+  flags.DEFINE_boolean(
+      name='enable_grappler_layout_optimizer',
+      default=True,
+      help='Enable Grappler layout optimizer. Currently Grappler can '
+           'de-optimize fp16 graphs byt forcing NCHW layout for all '
```
reedwm:
```suggestion
           'de-optimize fp16 graphs by forcing NCHW layout for all '
```

official/transformer/v2/misc.py
tfboyd: This was already submitted.  None results in an error.

official/utils/misc/keras_utils.py
tfboyd: based on our conversation I think this is the best I can do and I don't want to go back to FLAGs.

official/resnet/keras/keras_cifar_main.py
tfboyd: That flag is not turned on for ResNet56 yet.  I hope to add FP16 and XLA in a future change.

PR #7055, opened 2019-06-19
Improve performance of Keras ResNet models when not using distribution strategy

haoyuz:
Setting the learning phase for some reason negatively affect the performance when not using distribution strategy. We observe additional FusedBachnorm Grad ops in the backward pass if the learning phase here is set to 1. The root cause is still unclear.

### DIFF
--- a/official/resnet/keras/keras_cifar_main.py
+++ b/official/resnet/keras/keras_cifar_main.py
@@ -178,7 +178,8 @@ def run(flags_obj):
   if not strategy and flags_obj.num_gpus == 1:
     # TODO(b/135607227): Add device scope automatically in Keras training loop
     # when not using distribition strategy.
-    tf.device('/device:GPU:0').__enter__()
+    no_dist_strat_device = tf.device('/device:GPU:0')
+    no_dist_strat_device.__enter__()
 
   history = model.fit(train_input_dataset,
                       epochs=train_epochs,
@@ -195,7 +196,7 @@ def run(flags_obj):
                                  verbose=2)
 
   if not strategy and flags_obj.num_gpus == 1:
-    tf.device('/device:GPU:0').__exit__()
+    no_dist_strat_device.__exit__()
 
   stats = keras_common.build_stats(history, eval_output, callbacks)
   return stats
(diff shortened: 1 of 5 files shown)


official/resnet/keras/keras_cifar_main.py
```diff
@@ -177,7 +177,10 @@ def run(flags_obj):
 
   validation_data = eval_input_dataset
   if flags_obj.skip_eval:
-    tf.keras.backend.set_learning_phase(1)
+    if strategy:
+      # TODO(haoyuzhang): Understand slowdown of setting learning phase when
+      # not using distribution strategy.
+      tf.keras.backend.set_learning_phase(1)
```
guptapriya: what happens if you just skip setting this altogether?

haoyuz: That's a great question. I ran a set of experiments with this removed altogether last night but that ended up making benchmarks with mirrored strategy slower.

sponge/ede73a6b-979c-4a01-a609-ddfbef90694c

I will get to understand how the learning phase (and control flow ops) affect our performance.

guptapriya: It is slightly unexpected that this would affect mirrored strategy, because we already hard code the learning phase internally: 
https://cs.corp.google.com/piper///depot/google3/third_party/tensorflow/python/keras/engine/training_arrays.py?l=149

official/resnet/keras/keras_cifar_main.py
```diff
@@ -177,7 +177,10 @@ def run(flags_obj):
 
   validation_data = eval_input_dataset
   if flags_obj.skip_eval:
-    tf.keras.backend.set_learning_phase(1)
+    if strategy:
+      # TODO(haoyuzhang): Understand slowdown of setting learning phase when
```
tfboyd: Please add a bug number, I know we are committed to resolving this but this makes the test somewhat invalid as OTB.  Should we set this up as a tweak rather than avoid it for `if strategy`. Maybe make a flag for `turn off learning phase` or something.  More work but keeps this honest.

PR #7067, opened 2019-06-20
NCF XLA and Eager tests with a refactor of resnet flags to make this cleaner.

tfboyd:
First: I am sorry this got big.  What happened is to move run_eagerly as a based flag Resnet flag usage needed refactored.  And I wanted to do it anyway to reduce our calls to resnet_run_loop from the TF 2.0/Keras code.

I have run ResNet50 and ResNet56 1 GPU tests  benchmark.*1_gpu and for ResNet50 benchmark.*8_gpu and then the NCF tests.

Very soon I will factor out the rest of the calls to imagenet_main.py from the v2 folder so we can more easily move the v1 code into some archive folder.  I think only the constants are left.

### DIFF
--- a/official/recommendation/ncf_keras_benchmark.py
+++ b/official/recommendation/ncf_keras_benchmark.py
@@ -132,6 +132,13 @@ class NCFKerasAccuracy(NCFKerasBenchmarkBase):
     FLAGS.early_stopping = True
     self._run_and_report_benchmark()
 
+  def benchmark_1_gpu_no_dist_strat_run_eagerly_early_stop(self):
+    self._setup()
+    FLAGS.distribution_strategy = 'off'
+    FLAGS.early_stopping = True
+    FLAGS.run_eagerly = True
+    self._run_and_report_benchmark()
+
   def benchmark_xla_1_gpu_early_stop(self):
     self._setup()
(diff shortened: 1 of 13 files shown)


official/recommendation/ncf_keras_benchmark.py
```diff
@@ -156,45 +176,69 @@ def benchmark_2_gpus_early_stop(self):
...
+    self._setup()
+    FLAGS.train_epochs = 7
+    FLAGS.distribution_strategy = 'off'
+    self._run_and_report_benchmark()
+
+  def benchmark_1_gpu_no_dist_strat_run_eagerly_mlperf_like(self):
+    self._setup()
+    FLAGS.train_epochs = 7
+    FLAGS.distribution_strategy = 'off'
+    FLAGS.early_stopping = True
```
guptapriya: i think you dont want early stopping in the mlperf like test iirc?

tfboyd: You rock, saving me some big time pain later.  All fixed.

tfboyd:
Test failures are Transformer.  I fixed all ResNet unit tests failures.  Hopefully I or we can clean up the transformer issues soon as well.

PR #7070, opened 2019-06-20
Add ResNet56 CPU benchmark and accuracy tests.

tfboyd:
Tested locally nothing blows up.


tfboyd:
LINT job failed to run.  I added docstrings everywhere based on the last report.  I am going to risk it given we know I will fix it in the next PR.  Worth the risk to try out these tests.

tfboyd:
OK Lint is good and errors are transformer only.

PR #7077, opened 2019-06-21
Transformer Keras: Use add_loss instead of loss layer

guptapriya:
model.add_loss is (1) simpler (2) works in both 1.x and 2.0 with distribution strategy.
Using the loss layer has a bug with 1.x + distribution strategy that we do not have an easy way to fix.

### DIFF
--- a/official/transformer/v2/metrics.py
+++ b/official/transformer/v2/metrics.py
@@ -181,4 +181,3 @@ def transformer_loss(logits, labels, smoothing, vocab_size):
   xentropy, weights = padded_cross_entropy_loss(logits, labels, smoothing,
                                                 vocab_size)
   return tf.reduce_sum(xentropy) / tf.reduce_sum(weights)
-
--- a/official/transformer/v2/metrics.py
+++ b/official/transformer/v2/metrics.py
@@ -182,24 +182,3 @@ def transformer_loss(logits, labels, smoothing, vocab_size):
                                                 vocab_size)
   return tf.reduce_sum(xentropy) / tf.reduce_sum(weights)
 
-
-class LossLayer(tf.keras.layers.Layer):
-  """Custom a layer of transformer loss for Transformer model."""
-
-  def __init__(self, vocab_size, label_smoothing):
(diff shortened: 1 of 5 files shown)


official/transformer/v2/transformer.py
```diff
@@ -43,10 +43,11 @@ def create_model(params, is_train):
       vocab_size = params["vocab_size"]
       label_smoothing = params["label_smoothing"]
       logits = metrics.MetricLayer(vocab_size)([logits, targets])
-      logits = metrics.LossLayer(vocab_size, label_smoothing)([logits, targets])
       logits = tf.keras.layers.Lambda(lambda x: x, name="logits")(logits)
-      return tf.keras.Model([inputs, targets], logits)
-
+      model = tf.keras.Model([inputs, targets], logits)
+      loss = metrics.transformer_loss(logits, targets, label_smoothing, vocab_size)
```
tfboyd: I believe lint is flagging this as too long.

official/transformer/v2/metrics.py
```diff
@@ -182,24 +182,3 @@ def transformer_loss(logits, labels, smoothing, vocab_size):
                                                 vocab_size)
   return tf.reduce_sum(xentropy) / tf.reduce_sum(weights)
 
-
```
tfboyd: ************* Module official.transformer.v2.metrics
C:184, 0: Trailing newlines (trailing-newlines)

No idea using the github UI but I suspect using sublime or some other editor you will see what is up.

guptapriya:
Thanks Toby, fixed the lint.

guptapriya:
Confirmed that the training quality doesn't change with this change by running BIG.

guptapriya:
Looks like all tests now pass. @tfboyd good with merging this? I will probably have to do a manual one.

guptapriya:
This PR seems a bit messed up due to conflicts and stuff. So I am making a new one with the same changes: https://github.com/tensorflow/models/pull/7257. Will close this one, and hopefully tests pass on the new one too.

Issue #5637, opened 2018-10-30
imagenet_main.py running error


sunzhaoxing2016:
I try to run imagenet_main.py in official/resnet/, but I got a error

![image](https://user-images.githubusercontent.com/22524061/47696134-506ae300-dc40-11e8-877f-cc48ce4f7ce9.png)

Could anyone help me to figure out why?

I  haven't changed the code, just git clone and ran the code by typing "python imagenet_main.py --help" in /tensorflow/models/official/resnet/. Now I notice the error may caused by "\ufeff" in line110 of /tensorflow/models/official/utils/flags/_base.py, however, still don't know how to fix it.

By the code run in a container, which is linux 16.04.4 LTS + CUDA 9.0 + cudnn 7.0 + tensorflow 1.8.0 + python 2.7.12.

tensorflowbutler:
Thank you for your post. We noticed you have not filled out the following field in the issue template. Could you update them if they are relevant in your case, or leave them as N/A? Thanks.
What is the top-level directory of the model you are using
Have I written custom code
OS Platform and Distribution
TensorFlow installed from
TensorFlow version
Bazel version
CUDA/cuDNN version
GPU model and memory
Exact command to reproduce

Lw-Cui:
I also ran into this problem. Looks like #5145 doesn't work.

BTW, [this line](https://github.com/tensorflow/models/blob/0b0c9cfd76cd3585dffae2cd7a40f59944fab654/official/wide_deep/wide_deep_run_loop.py#L51) in `official/wide_deep/wide_deep_run_loop.py` didn't use `flags_core.help_wrap`.


* Have I written custom code: No
* OS Platform and Distribution: custom Linux
* TensorFlow installed from: Docker
* TensorFlow version: 1.7.0 (git_hash: v1.7.0-3-g024aecf414)
* Bazel version: N/A
* CUDA/cuDNN version: 9.0.176 / 7005
* GPU model and memory: Tesla P40, 23015 MB memory
* Exact command to reproduce: Just run `python movielens_main.py --help`

tensorflowbutler:
Hi There,
 We are checking to see if you still need help on this, as this seems to be considerably old issue. Please update this issue with the latest information, code snippet to reproduce your issue and error you are seeing.
 If we don't hear from you in the next 7 days, this issue will be closed automatically. If you don't need help on this issue any more, please consider closing this.

PR #7079, opened 2019-06-21
Fix help print error when stdout/stderr not use utf-8 encoding

neilhong:
Fix help print error when stdout/stderr not use utf-8 encoding, like https://github.com/tensorflow/models/issues/5637

Main reason is the help wrap did not check the stdout encoding, even the system have installed utf-8 locale package

### DIFF
--- a/official/utils/flags/_conventions.py
+++ b/official/utils/flags/_conventions.py
@@ -18,6 +18,7 @@ from __future__ import absolute_import
 from __future__ import division
 from __future__ import print_function
 
+import sys
 import codecs
 import functools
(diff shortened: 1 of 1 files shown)


tfboyd:
Hey Neil I will pick this up today.  I suspect you closed it due to the
cla.  If this is easy and works I will get the idea in asap.  Thank you for
the feedback and idea.

tfboyd:
@robieta   Can you take a quick look?  This looks fine to me, but is not my area of expertise and testing it is a bit of a pain.

PR #7081, opened 2019-06-21
some missing imports and fixed URL

NescobarAlopLop:
- URL in readme was no longer valid
- Added missing function maybe_download_and_extract, once it was part of the repo but it was removed and there is a function call to non existent func

### DIFF
--- a/tutorials/image/cifar10/README.md
+++ b/tutorials/image/cifar10/README.md
@@ -10,4 +10,4 @@ Code in this directory demonstrates how to use TensorFlow to train and evaluate 
 
 Detailed instructions on how to get started available at:
 
-http://tensorflow.org/tutorials/deep_cnn/
+https://www.tensorflow.org/tutorials/images/deep_cnn
--- a/tutorials/image/cifar10/cifar10.py
+++ b/tutorials/image/cifar10/cifar10.py
@@ -37,12 +37,14 @@ from __future__ import print_function
 
 import os
 import re
+import sys
 import tarfile
 import urllib
 import zipfile
 
-import cifar10_input
 import tensorflow as tf
+
+import cifar10_input
 
 FLAGS = tf.app.flags.FLAGS
 
@@ -149,7 +151,6 @@ def distorted_inputs():
 
 def inputs(eval_data):
   """Construct input for CIFAR evaluation using the Reader ops.
-
   Args:
     eval_data: bool, indicating if one should use the train or eval data set.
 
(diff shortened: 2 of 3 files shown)


NescobarAlopLop:
I signed it!

tutorials/image/cifar10/cifar10.py
```diff
@@ -162,6 +165,33 @@ def inputs(eval_data):
   return images, labels
 
 
+def maybe_download_and_extract():
```
tfboyd: This seems new.  There is already a download and extract script I think.  Can you explain?

tutorials/image/cifar10/README.md
```diff
@@ -10,4 +10,4 @@ Code in this directory demonstrates how to use TensorFlow to train and evaluate
 
 Detailed instructions on how to get started available at:
 
-http://tensorflow.org/tutorials/deep_cnn/
+https://www.tensorflow.org/tutorials/images/deep_cnn
```
tfboyd: Thank you.

tfboyd:
Your homer avatar is solid.

tutorials/image/cifar10/cifar10.py
NescobarAlopLop: You are correct there is a better way to get the data using tensorflow_datasets.
And there are more functions missing. I will fix it.
But currently as it is tutorials/image/cifar10/cifar10_multi_gpu_train.py wont run because of 
function call on line 269, it calls non existent `maybe_download_and_extract`
[here](https://github.com/tensorflow/models/blob/a68f65f8b06969ee314354e3692fce10be7712c7/tutorials/image/cifar10/cifar10_multi_gpu_train.py#L269)

tfboyd:
I was wrong, sorry about that.  I did not create this one.  I did the estimator in the README and then was part of the updated even better versions in /official/resnet  and official/resnet/keras.  I will try to take a look but I would rather delete this as it is not how anyone should do multi-gpu at this point.  Even the one linked in the README is by hand, which is interesting but more work than needed.  

If you confirm (I realize you kind already are confirming) that you ran this and it works I can accept it without testing myself as it looks correct to me.

NescobarAlopLop:
Hello again, 2 things:
- the code is still buggy, it might be tf version problem but there is a problem ValueError: All shapes must be fully defined: [TensorShape([Dimension(None), Dimension(24), Dimension(24), Dimension(3)]), TensorShape([Dimension(None)])]

- and there is no need to add the function, it is better to remove its call from the main.

I'll close this commit and reopen another one. But this tutorial won't work. Sorry.

tfboyd:
I would not use this tutorial is it not correct anymore.  I added the comment in the README a while back.  I would just delete it but I am not sure if it breaks some link from the website but then again a link to something that is not a good example is not great either.

PR #7084, opened 2019-06-21
readme source url and import optimization

NescobarAlopLop:
- updated URL to CNN tutorial page
- minor refactor in order of imports


tfboyd:
Thanks (very truly).  I know there is some level of demoralization with this tutorial.

PR #7086, opened 2019-06-22
Fix unit tests failures.

tfboyd:
*  Updated tests that were failing
   *  mostly to skip GPU tests when on GPU
   *  Reset flags in setup
   *  Use correct set eager if _v2.0
* What looks like needless changes to the presubmit.sh make it easy to swap in `-m unittest` which may be a hopeless dream.  I can be talked into reverting them.

I really wanted to use python3 -m unittest.  But this skips the main and requires a different approach to loading the flags which I fear would mess up the internal testing.  The benefit to -m unittest is that if a test file has a bad module/import you get a clear failure with the test method failing because the module is missing vs. the current method where test are run with python3 dir/dir/blah_test.py and the error is just like running a script with a missing module or bad import.  Maybe there is a better method but none-the-less this is better and the script update to kokoro gives a decent summary.  

We will see....  

Test pass CPU local with Python3.  I am testing GPU and then hoping to changing kokoro to use GPUs for a bigger range of tests.

### DIFF
--- a/official/mnist/mnist_eager_test.py
+++ b/official/mnist/mnist_eager_test.py
@@ -22,6 +22,7 @@ import tensorflow.contrib.eager as tfe  # pylint: disable=g-bad-import-order
 
 from official.mnist import mnist
 from official.mnist import mnist_eager
+from official.utils.misc import keras_utils
 
 
 def device():
@@ -62,6 +63,11 @@ def evaluate(defun=False):
 class MNISTTest(tf.test.TestCase):
   """Run tests for MNIST eager loop."""
 
+  def setUp(self):
+    if not keras_utils.is_v2_0():
+      tf.compat.v1.enable_v2_behavior()
+    super(MNISTTest, self).setUp()
+
   def test_train(self):
     train(defun=False)
 
@@ -76,5 +82,4 @@ class MNISTTest(tf.test.TestCase):
 
 
 if __name__ == "__main__":
-  tfe.enable_eager_execution()
   tf.test.main()
--- a/official/transformer/v2/transformer_main_test.py
+++ b/official/transformer/v2/transformer_main_test.py
@@ -20,44 +20,54 @@ from __future__ import print_function
 
 import os
 import re
+import unittest
 
 from absl import flags
+from absl.testing import flagsaver
 import tensorflow as tf
 
 from official.transformer.v2 import misc
 from official.transformer.v2 import transformer_main as tm
 
 FLAGS = flags.FLAGS
-FIXED_TIMESTAMP = "my_time_stamp"
-WEIGHT_PATTERN = re.compile(r"weights-epoch-.+\.hdf5")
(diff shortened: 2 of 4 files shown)


tfboyd:
ALL Green   :-)

official/mnist/mnist_eager_test.py
```diff
@@ -76,5 +82,4 @@ def test_evaluate_with_defun(self):
 
 
 if __name__ == "__main__":
-  tfe.enable_eager_execution()
```
guptapriya: should we also remove the tfe import? then it will work in 2.0 as well (maybe?) or are you mainly trying to get it working with 1.x?

official/transformer/v2/transformer_main_test.py
```diff
@@ -75,23 +85,26 @@ def test_train_static_batch(self):
...
-    FLAGS.distribution_strategy = "one_device"
+    FLAGS.distribution_strategy = 'one_device'
     t = tm.TransformerTask(FLAGS)
     t.train()
 
+  @unittest.skipUnless(tf.test.is_built_with_cuda(), 'requires GPU')
   def test_train_2_gpu(self):
-    FLAGS.distribution_strategy = "mirrored"
+    FLAGS.distribution_strategy = 'mirrored'
     FLAGS.num_gpus = 2
```
guptapriya: while you're here, would you mind adding this extra check so that it doesn't fail when there are <2 GPUs? (this will be useful on the internal GPU test where we have 1 GPU only. 
```
if context.num_gpus() < 2:
 self.skipTest(....)
```

official/transformer/v2/transformer_main_test.py
```diff
@@ -75,23 +85,26 @@ def test_train_static_batch(self):
     t = tm.TransformerTask(FLAGS)
     t.train()
 
+  @unittest.skipUnless(tf.test.is_built_with_cuda(), 'requires GPU')
```
guptapriya: actually i wanted this test to run even if there is no GPU. the helper method will select "one_device" with CPU if no GPUs are available. maybe we can rename the test?

official/transformer/v2/transformer_main_test.py
```diff
@@ -75,23 +85,26 @@ def test_train_static_batch(self):
     t = tm.TransformerTask(FLAGS)
```
guptapriya: does this test pass? for me it was failing with some weird reasons

tfboyd: Yes, it fails in TF 1.0 on GPU, but is not run in our test because it is CPU only.  I think because of that input problem.  I have another PR coming where all GPU tests pass on TF 1.0 where I flags this as TF 2.0 only.  And then I will have a PR where everything works on TF 2.0 CPU/GPU.  I guess is ~6 days as I now have a plan (P100s for now) to spin up 2 GPU systems for the presubmits.

PR #7101, opened 2019-06-25
Unit tests pass TF 2.0 GPU and CPU locally.

tfboyd:
Ready for Review.  

Open to revisit tests that are skipped for TF 2.0. The TF 2.0 presubmit server will be dual GPU.

### DIFF
--- a/official/benchmark/benchmark_uploader_test.py
+++ b/official/benchmark/benchmark_uploader_test.py
@@ -61,7 +61,7 @@ class BigQueryUploaderTest(tf.test.TestCase):
       json.dump({"model_name": "value"}, f)
 
   def tearDown(self):
-    tf.compat.v1.gfile.DeleteRecursively(self.get_temp_dir())
+    tf.io.gfile.rmtree(self.get_temp_dir())
 
   def test_upload_benchmark_run_json(self):
     self.benchmark_uploader.upload_benchmark_run_json(
--- a/official/benchmark/benchmark_uploader_test.py
+++ b/official/benchmark/benchmark_uploader_test.py
@@ -61,7 +61,7 @@ class BigQueryUploaderTest(tf.test.TestCase):
       json.dump({"model_name": "value"}, f)
 
   def tearDown(self):
-    tf.gfile.DeleteRecursively(self.get_temp_dir())
+    tf.compat.v1.gfile.DeleteRecursively(self.get_temp_dir())
 
   def test_upload_benchmark_run_json(self):
     self.benchmark_uploader.upload_benchmark_run_json(
--- a/official/resnet/cifar10_test.py
+++ b/official/resnet/cifar10_test.py
@@ -23,6 +23,7 @@ import numpy as np
 import tensorflow as tf  # pylint: disable=g-bad-import-order
 
 from official.resnet import cifar10_main
+from official.utils.misc import keras_utils
 from official.utils.testing import integration
 
 tf.compat.v1.logging.set_verbosity(tf.compat.v1.logging.ERROR)
@@ -42,7 +43,8 @@ class BaseTest(tf.test.TestCase):
   @classmethod
   def setUpClass(cls):  # pylint: disable=invalid-name
     super(BaseTest, cls).setUpClass()
-    tf.compat.v1.disable_eager_execution()
+    if keras_utils.is_v2_0:
+      tf.compat.v1.disable_eager_execution()
     cifar10_main.define_cifar_flags()
 
   def setUp(self):
--- a/official/resnet/cifar10_test.py
+++ b/official/resnet/cifar10_test.py
@@ -42,6 +42,7 @@ class BaseTest(tf.test.TestCase):
   @classmethod
   def setUpClass(cls):  # pylint: disable=invalid-name
     super(BaseTest, cls).setUpClass()
+    tf.compat.v1.disable_eager_execution()
     cifar10_main.define_cifar_flags()
 
   def setUp(self):
@@ -76,7 +77,7 @@ class BaseTest(tf.test.TestCase):
     self.assertAllEqual(label.shape, ())
     self.assertAllEqual(image.shape, (_HEIGHT, _WIDTH, _NUM_CHANNELS))
 
-    with self.test_session() as sess:
+    with self.session() as sess:
       image, label = sess.run([image, label])
 
       self.assertEqual(label, 7)
--- a/official/resnet/imagenet_test.py
+++ b/official/resnet/imagenet_test.py
@@ -22,6 +22,7 @@ import unittest
 import tensorflow as tf  # pylint: disable=g-bad-import-order
 
 from official.resnet import imagenet_main
+from official.utils.misc import keras_utils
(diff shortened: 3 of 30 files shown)


official/utils/logs/logger.py
```diff
@@ -391,24 +387,6 @@ def _collect_cpu_info(run_info):
         "'cpuinfo' not imported. CPU info will not be logged.")
 
 
-def _collect_gpu_info(run_info, session_config=None):
```
tfboyd: This code path is not used anymore and it is broken if using CUDA_VISIBLE_DEVICES.  Any work to fix it is not useful because we use perfzero for logging.  The entire code path can be removed at some point, just not now.

guptapriya:
Cool. let me know when I should review.

tfboyd:
@guptapriya  Ready. I did a quick review and on the surface I think this is correct.  I left a comment on the part where I removed something that seems weird without context.

official/benchmark/benchmark_uploader_test.py
```diff
@@ -61,7 +61,7 @@ def setUp(self, mock_bigquery):
       json.dump({"model_name": "value"}, f)
 
   def tearDown(self):
-    tf.gfile.DeleteRecursively(self.get_temp_dir())
+    tf.compat.v1.gfile.DeleteRecursively(self.get_temp_dir())
```
guptapriya: not 100% sure, but is this the replacement: https://www.tensorflow.org/versions/r2.0/api_docs/python/tf/io/gfile/rmtree ?

official/resnet/cifar10_test.py
```diff
@@ -42,6 +42,7 @@ class BaseTest(tf.test.TestCase):
   @classmethod
   def setUpClass(cls):  # pylint: disable=invalid-name
     super(BaseTest, cls).setUpClass()
+    tf.compat.v1.disable_eager_execution()
```
guptapriya: do you need to check "if 2.0" before this? not sure because you check in some places not others, and i wasn;t sure if that is intentional or not

official/resnet/imagenet_test.py
```diff
@@ -62,7 +63,7 @@ def reshape(shape):
 
     graph = tf.Graph()
 
-    with graph.as_default(), self.test_session(
+    with graph.as_default(), self.session(
```
guptapriya: why the test_session -> session change?

official/utils/data/file_io_test.py
```diff
@@ -28,6 +28,7 @@
 # pylint: enable=wrong-import-order
 
 from official.utils.data import file_io
+from official.utils.misc import keras_utils
```
guptapriya: totally optional but i just realized "is_v2_0" check should probably not be in a util called keras_utils :)

official/utils/testing/reference_data.py
```diff
@@ -261,12 +261,19 @@ def _evaluate_test_case(self, name, graph, ops_to_eval, correctness_function):
             .format(differences, tf_version_comparison))
 
       eval_results = [op.eval() for op in ops_to_eval]
+      print("eval_resuls:{}".format(eval_results))
```
guptapriya: do you want to leave this prints here or were they only for debugging?

official/utils/testing/reference_data.py
```diff
@@ -261,12 +261,19 @@ def _evaluate_test_case(self, name, graph, ops_to_eval, correctness_function):
...
         result_json = os.path.join(data_dir, "results.json")
+        print("results:{}".format(results))
+        print("data_dir:{}".format(data_dir))
         with tf.io.gfile.GFile(result_json, "r") as f:
           expected_results = json.load(f)
-        self.assertAllClose(results, expected_results)
+          print("expected_results:{}".format(expected_results))
+        self.assertAllClose(results,
+                            expected_results)
+                            # rtol=1e-05,
```
guptapriya: what about these commented tolerances?

official/resnet/imagenet_test.py
tfboyd: fixed

official/utils/data/file_io_test.py
tfboyd: Yeah, I will try to refactor that later.  I move it here to share.  It would be nice if TensorFlow provided this as a feature.  Not fixing in this PR.

official/utils/testing/reference_data.py
tfboyd: Removed

official/resnet/layer_test.py
```diff
@@ -166,16 +173,38 @@ def _resnet_block_ops(self, test, batch_size, bottleneck, projection,
         correctness_function=self.default_correctness_function
     )
 
+  @unittest.skipIf(tf.test.is_built_with_cuda(), "Results only match CPU.")
   def test_batch_norm(self):
+    """Tests batch norm layer correctness.
+
+    Test fails on a GTX 1080 with the last value being significantly different:
+    7.629395e-05 (expected) -> -4.159546e-02 (actual). While concerning the test
```
robieta: Yikes.

official/transformer/model/beam_search_test.py
```diff
@@ -43,22 +47,22 @@ def test_shape_list(self):
 
   def test_get_shape_keep_last_dim(self):
     y = tf.constant(4.0)
-    x = tf.ones([7, tf.to_int32(tf.sqrt(y)), 2, 5])
+    x = tf.ones([7, tf.compat.v1.to_int32(tf.sqrt(y)), 2, 5])
```
robieta: I think you can replace `tf.compat.v1.to_int32(...)` with `tf.cast(tf.sqrt(y), tf.int32)`

tfboyd: Yeah I thought about doing this right and was lazy and maybe not as sure of myself as I wanted to be.  I'll make the change, thank you for the tip.

official/utils/testing/reference_data.py
tfboyd: I removed them, it defaults to 1e-06 according to the docs and that is fine.

guptapriya:
Thanks Toby, is it ready for another review?

tfboyd:
Not yet.  I just upgraded the runner and found more tests to fix.  Monday I
will be ready.

tfboyd:
@guptapriya   I am ready and I think this is finally done.  Everything looks to be passing on GPU TF 2.0 as well as CPU as seen in the presubmit above.  The lint error is weird as the line is not over 80 characters and I can deal with it later.

official/resnet/keras/keras_common_test.py
```diff
@@ -66,7 +66,7 @@ def test_build_stats_sparse(self):
     self.assertEqual(1.9844, stats['eval_loss'])
 
   def test_time_history(self):
```
guptapriya: not a big deal but this test should probably be in keras_utils_test although i haven't checked if that exists or not

tfboyd: Yes this one could go there for sure.  Long story to even get to this point.

PR #7102, opened 2019-06-26
Perf optimization to convert gradients from sparse to dense

nnigania:


official/recommendation/ncf_keras_main.py
```diff
@@ -339,8 +339,8 @@ def step_fn(features):
           loss *= (1.0 / (batch_size*strategy.num_replicas_in_sync))
 
         grads = tape.gradient(loss, keras_model.trainable_variables)
-        optimizer.apply_gradients(list(zip(grads,
-                                           keras_model.trainable_variables)))
+        grads = neumf_model._sparse_to_dense_grads(list(zip(grads,keras_model.trainable_variables)))
```
guptapriya: couple nits:
- can we add a short comment here explaining why this helps 
- make _sparse_to_dense_grads -> sparse_to_dense_grads now that it is used outside that module
- probably need a space after `grads,` (lint should complain?)

nnigania:
@guptapriya I have added the changes you suggested.

PR #7134, opened 2019-07-02
Add StepCounterHook to hooks_helper.py

yuefengz:


official/utils/logs/hooks_helper.py
```diff
@@ -156,10 +156,17 @@ def get_logging_metric_hook(tensors_to_log=None,
       every_n_secs=every_n_secs)
 
 
+def get_step_counter_hook(**kwargs):
+  """Function to get StepCounterHook."""
+  del kwargs
+  return tf.train.StepCounterHook()
```
guptapriya: Does this work in TF 2? Where would this get used? I believe Toby is trying to get the entire model repo to work with TF 2.0 as much as possible, so let's not add something that only works in TF 1.x. Maybe use the compat symbol if you really need this.

yuefengz: Done. This is code path would not be triggered unless user specifies "stepcounterhook" in commandline flags.

PR #7135, opened 2019-07-02
Allow distibution_utils.py to work with PSStrategy or none strategy

yuefengz:
when there are multiple workers.

### DIFF
--- a/official/utils/misc/distribution_utils.py
+++ b/official/utils/misc/distribution_utils.py
@@ -114,13 +114,13 @@ def get_distribution_strategy(distribution_strategy="default",
 
   distribution_strategy = distribution_strategy.lower()
   if distribution_strategy == "off":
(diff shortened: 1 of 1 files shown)


official/utils/misc/distribution_utils.py
```diff
@@ -114,13 +114,13 @@ def get_distribution_strategy(distribution_strategy="default",
 
   distribution_strategy = distribution_strategy.lower()
   if distribution_strategy == "off":
-    if num_gpus > 1 or num_workers > 1:
+    if num_gpus > 1:
```
dubey: So what would happen if `num_workers > 1` and user specifies `off`?

yuefengz: They can use the parameter server training log built in Estimator to train their model.

PR #7138, opened 2019-07-02
add non-determinstic file reading to ImageNet input pipeline

guygurari:
When rading ImageNet data, allow files to be read in non-determinstic
order. This greatly improves performance when reading from some cloud
storage backends (e.g. Google Cloud Storage).

### DIFF
--- a/official/resnet/imagenet_main.py
+++ b/official/resnet/imagenet_main.py
@@ -212,6 +212,11 @@ def input_fn(is_training,
       cycle_length=10,
       num_parallel_calls=tf.data.experimental.AUTOTUNE)
 
+  # Allow files to be read in a non-determinstic order, improving performance
+  options = tf.data.Options()
+  options.experimental_deterministic = False
+  dataset = dataset.with_options(options)
+
   return resnet_run_loop.process_record_dataset(
       dataset=dataset,
       is_training=is_training,


official/resnet/imagenet_main.py
```diff
@@ -212,6 +212,11 @@ def input_fn(is_training,
       cycle_length=10,
       num_parallel_calls=tf.data.experimental.AUTOTUNE)
 
+  # Allow files to be read in a non-determinstic order, improving performance
+  options = tf.data.Options()
+  options.experimental_deterministic = False
```
tfboyd: This is likely going to need to be flagged rather than on all of the time.  I will ask to see what the team thinks and then respond back again.

tfboyd: @guygurari  I am thinking it might be easier to accept your change and then I can add a FLAG in a followup PR.  I am in this situation where I want to accept the change but the agreement seems to be it needs to be FLAGed and default off in the model code.  While I don't really want to do the work myself, I also don't want you to have to hunt around the flag code when I can do it a couple hours.  You did the work validating the feature helps GCS a lot.    Let me know what you think.

guygurari: That sounds fine to me, and thanks for offering to do it. I took a quick look at the way flags are organized and it does seem quite complicated, so adding a flag is not something I can do easily.

jaeyounkim:
imagenet_main.py is moved to r1 directory. Please submit a new PR, and I will approve and merge it.

PR #7174, opened 2019-07-09
Add Bounding Boxes Location

cocopambag:
well, when i use to models/research/object_detection, i usually needed bounding boxes location,
so, i suggest this merge.
thank you!


cocopambag:
> I signed it!

research/object_detection/utils/visualization_utils.py
```diff
@@ -687,6 +687,7 @@ def visualize_boxes_and_labels_on_image_array(
     classes,
     scores,
     category_index,
+    bounding_boxes_location = [],
```
pkulzc: Why don't you use the boxes above?

PR #7178, opened 2019-07-09
Improve performance for Cifar ResNet benchmarks

haoyuz:


official/resnet/keras/keras_cifar_benchmark.py
```diff
@@ -59,7 +59,7 @@ def benchmark_graph_1_gpu(self):
     self._setup()
     FLAGS.num_gpus = 1
     FLAGS.data_dir = self.data_dir
-    FLAGS.batch_size = 128
+    FLAGS.batch_size = 256
```
tfboyd: Change these all back to 128 based on our conversation.  I have too much history at 128 and I don't think 256 converges the same.  The paper is 128 unless my memory is horrible.  I spent a bunch of time getting this to converge correctly back in December.

haoyuz: Thanks for clarifying that! Reverted changes to batch size.

PR #7182, opened 2019-07-10
Transformer: Less steps for the benchmark.  

tfboyd:
Early results indicate FP16 trains with the same or maybe less steps than FP32.  No need for the 20 epochs which takes "forever" and has no added value.


tfboyd:
WRONG PR branch

PR #7183, opened 2019-07-10
Reduce transformer fp16 test to 12 iterations.

tfboyd:
Early results indicate FP16 trains with the same or maybe less steps than FP32. No need for the 20 epochs which takes "forever" and has no added value.

### DIFF
--- a/official/transformer/v2/transformer_benchmark.py
+++ b/official/transformer/v2/transformer_benchmark.py
@@ -310,7 +310,7 @@ class TransformerBigKerasAccuracy(TransformerBenchmark):
     FLAGS['bleu_ref'].value = self.bleu_ref
     FLAGS.param_set = 'big'
     FLAGS.batch_size = 3072*8
-    FLAGS.train_steps = 400000
+    FLAGS.train_steps = 20000 * 12
     FLAGS.steps_between_evals = 20000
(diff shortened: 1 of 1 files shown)


official/transformer/v2/transformer_benchmark.py
```diff
@@ -310,7 +310,7 @@ def benchmark_8_gpu_fp16(self):
     FLAGS['bleu_ref'].value = self.bleu_ref
     FLAGS.param_set = 'big'
     FLAGS.batch_size = 3072*8
-    FLAGS.train_steps = 400000
+    FLAGS.train_steps = 20000 * 12
```
reedwm: Maybe change benchmark_8_gpu_static_batch_fp16 and benchmark_xla_8_gpu_static_batch_fp16 as well

tfboyd: I will reduce them once I have full runs showing they converge.  Early tests with FP16 and static batch (I did not have FP32 results) show it only hits 28.0 not the 28.4 or even 28.1.  It also does it after more steps.  I think we will end up needing to wait for the better packing.

tfboyd:
The data is in and 12 will be fine.  I will do another PR where I detail the results in the docstring of the test.

PR #7189, opened 2019-07-11
ResNet50 and ResNet56 Add stdev to the Dense layer.

tfboyd:
**Waiting on final test result to finish: ResNet50 XLA+FP16.**

This changes ResNet50 and ResNet56.  The detailed comments below cover each situation. The change was first made to ResNet50 to aligned with TPUs and was tested on TPU to reduce variance and help push to .764 more often. I decided to apply it to ResNet56 as well as make the dense layers match. 

**ResNet56**
In a 6 run test this has make ResNet56 much more stable and should allow us to reduce the error bars for the test to better detect regressions.  We currently have to accept values as low as .924 as possibly ok.  

**Before:**  Min: .924  Max: .938  Median: .931
**After:**   Min  .929  MAX .934  Median: .932

Raw numbers:
0.9324919581
0.9309895635
0.9342948794
0.9292868376
0.9331930876
0.9320913553

**ResNet50**
I just did one run given the change was minor and I believe I have seen us set this in tf_cnn_benchmarks or maybe not.  Either way I did XLA + FP16 because it runs fast and if that works we should be good to go. 
625/625 - 244s - loss: 1.3501 - sparse_categorical_accuracy: 0.7797 - val_loss: 1.3821 - val_sparse_categorical_accuracy: **0.7638** <-- looks good.

https://sponge.corp.google.com/target?id=c09a85d7-28a1-46e0-a3c4-eebd49dbc6cc&target=tensorflow_models/perfzero/manual/v100_8_gpu/v100_8_gpu&searchFor=&show=ALL&sortBy=STATUS

### DIFF
--- a/official/resnet/keras/resnet_cifar_model.py
+++ b/official/resnet/keras/resnet_cifar_model.py
@@ -26,6 +26,7 @@ from __future__ import print_function
 import functools
 import tensorflow as tf
 from tensorflow.python.keras import backend
+from tensorflow.python.keras  import initializers
 from tensorflow.python.keras import layers
 from tensorflow.python.keras import regularizers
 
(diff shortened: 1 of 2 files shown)


tfboyd:
@haoyuz  I am 100% sure I also was part of having ResNet56 wrong.  We stopped working on it when ResNet50 started working; but I kept messing with it using my limited knowledge.  Sameer from the TPU team is the real hero; he was trying to fix the TPU and was using our code as as diff because we had proof it worked.  He then noticed if we add stdev it will reduce variance or did for TPU.  

And THANK YOU SIR  :-)

PR #7195, opened 2019-07-11
Support RunConfig overrides on command line

### LINKED ISSUE #5246: object_detection: Allow to provide RunConfig parameters on command line of model_main
### System information
- **What is the top-level directory of the model you are using**: object_detection
- **OS Platform and Distribution (e.g., Linux Ubuntu 16.04)**: Ubuntu 16.04.5 LTS (Xenial Xerus)
- **TensorFlow installed from (source or binary)**: binary
- **TensorFlow version (use command below)**: ('v1.10.1-0-g4dcfddc5d1', '1.10.1')
- **Bazel version (if compiling from source)**: -
- **CUDA/cuDNN version**: cuda-9.0
- **GPU model and memory**: Tesla K80 major: 3 minor: 7 / 11.17GiB
- **Exact command to reproduce**: - (This is a feature request)

### Describe the problem

An Estimator run is configured with a `RunConfig` object, however when starting `model_main` parameters of `RunConfig` can not be defined to overwrite defaults. That is why I propose to add command line parameters to `model_main` that allow to define `RunConfig` parameters. I see two options here:

1. Add specific command line parameters for specific `RunConfig` parameters, e.g. `--save_checkpoints_steps`, `--save_checkpoints_secs`, ...
1. Add a generic command line parameter for `RunConfig` parameters similar to the already existing `--hparams_overrides`. E.g. `--runconfig_overrides save_checkpoints_steps=100,keep_checkpoint_max=10`

Option 2 should result in less maintenance when `RunConfig` is changed. OTOH it would only allow to overwrite _primitive_ parameters, but no structured parameters like `session_config`. With Option 1 changes in `RunConfig` might need to be reflected in code. OTOH it would for example allow to define command line switches for `session_config` parameters.

I would personally vote for Option 2 and if there is interest I could also open a PR for this.

volkerstampa:
model_main is extended by a runconfig_overrides command line parameter
that allows to overwrite individual attributes of the default RunConfig
in a similar manner as hparams_overrides allows to overwrite
hyperparameters on command line.

Closes #5246

This is just a proposal. I am not sure if it makes sense to _reuse_ `parse_values` from `contrib.training`. Please advise if the implementation and the feature makes sense to you.

### DIFF
--- a/research/object_detection/model_main.py
+++ b/research/object_detection/model_main.py
@@ -24,6 +24,7 @@ import tensorflow as tf
 
 from object_detection import model_hparams
 from object_detection import model_lib
+from object_detection import model_runconfig
 
 flags.DEFINE_string(
     'model_dir', None, 'Path to output model directory '
@@ -46,6 +47,10 @@ flags.DEFINE_string(
     'represented as a string containing comma-separated '
     'hparam_name=value pairs.')
 flags.DEFINE_string(
+    'runconfig_overrides', None, 'run config overrides, '
+    'represented as a string containing comma-separated '
+    'param_name=value pairs.')
+flags.DEFINE_string(
     'checkpoint_dir', None, 'Path to directory holding a checkpoint.  If '
     '`checkpoint_dir` is provided, this binary operates in eval-only mode, '
     'writing resulting metrics to `model_dir`.')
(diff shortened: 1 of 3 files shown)


dhasl002:
Just want to mention that this worked for me. I was looking for this exact functionality. Thank you!

PR #7196, opened 2019-07-11
Record highest uncased bleu found.

tfboyd:
This records the highest bleu score found along with which iteration it occurred on.  It is possible the issues with our variance can be resolved; but for now this will give us a better indication as to if we have an issue.  

My knowledge is not complete, but the paper mentioned it took the average of the parameters of the last 5 checkpoints.  This suggest to me the result is not as stable as say ResNet.  Currently running the test and taking a reading at a predefined epoch gives results that are not conclusive or actionable.  

I also record the final bleu uncased in a different field so that is also tracked.

### DIFF
--- a/official/transformer/v2/transformer_benchmark.py
+++ b/official/transformer/v2/transformer_benchmark.py
@@ -90,7 +90,7 @@ class TransformerBenchmark(PerfZeroBenchmark):
                         'value': bleu_uncased_best[1],
                         'min_value': bleu_min,
                         'max_value': bleu_max})
-        metrics.append({'name': 'bleu_recorded_iteration',
+        metrics.append({'name': 'bleu_best_score_iteration',
                         'value': bleu_uncased_best[0]})
         metrics.append({'name': 'bleu_uncased_last',
                         'value': stats['bleu_uncased']})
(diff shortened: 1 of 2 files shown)


official/transformer/v2/transformer_benchmark.py
```diff
@@ -83,10 +83,22 @@ def _run_and_report_benchmark(self,
...
-                      'min_value': bleu_min,
-                      'max_value': bleu_max})
+      if 'bleu_uncased_history' in stats:
+        bleu_uncased_best = max(stats['bleu_uncased_history'],
+                                key=lambda x: x[1])
+        metrics.append({'name': 'bleu_uncased',
+                        'value': bleu_uncased_best[1],
+                        'min_value': bleu_min,
+                        'max_value': bleu_max})
+        metrics.append({'name': 'bleu_recorded_iteration',
```
isaprykin: From the name bleu_recorded_iteration it wasn't immediately clear to me that you meant the iteration whether the best score was recorded.  I'd suggest bleu_best_score_iteration.

tfboyd: fixed and this is one of the best part of code reviews.  More ideas on naming stuff.  Thank you.

PR #7205, opened 2019-07-12
[Transformer] Use float16 input and output for softmax in mixed-precision training

zongweiz:
Tensorflow softmax GPU internally internally performs math operations using float32 for numeric stability. So in mixed-precision training, we can keep the input and output of softmax in float16 for better performance (without manual casting and also save memory bandwidth).

### DIFF
--- a/official/transformer/v2/attention_layer.py
+++ b/official/transformer/v2/attention_layer.py
@@ -21,24 +21,6 @@ from __future__ import print_function
 import tensorflow as tf
 
 
-def _float32_softmax(logits, name=None):
-  """Computes a softmax activation in float32.
-
-  When training a model using float16, softmax is still done in float32 for
-  numeric stability.
-
-  Args:
-    logits: A tensor, with any shape accepted by `tf.nn.softmax`.
-
-  Returns:
-    A tensor with the same dtype as `logits`.
-  """
-  input_dtype = logits.dtype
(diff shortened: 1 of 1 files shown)


zongweiz:
softmax GPU kernel code:
https://github.com/tensorflow/tensorflow/blob/master/tensorflow/core/kernels/softmax_op_gpu.cu.cc#L72
accumulator_type is forced to float when input type is Eigen::half.

guptapriya:
LGTM, will let Reed approve as he has more context.

reedwm:
I think the added comment might not be entirely correct, as I think Softmax still does some operations in float16 when it's inputs/outputs are in float16. We will have to take a closer look.

In any case, I think we should try doing Softmax entirely in float32 before submitting this PR, since otherwise this might cause convergence issues. And we have to check XLA as well.

/CC @tfboyd if you have any thoughts

zongweiz:
I looked at the SoftmaxOpGPU op kernel, it seems that only input, max, output logits are in f16, other intermediate values and computation are in f32. @reedwm do you mind double check? Thanks. https://github.com/tensorflow/tensorflow/blob/master/tensorflow/core/kernels/softmax_op_gpu.cu.cc#L139

reedwm:
@zongweiz you are correct.

PR #7206, opened 2019-07-12
Ncf perf optimizations for CTL and multi GPU

nnigania:
Ncf perf changes 1)exclude metric layer from CTL train step 2) Dataset optimization to fix size of the sample_weights, preventing a costly broadcast during loss calculation for multi-gpu case

### DIFF
--- a/official/recommendation/ncf_keras_main.py
+++ b/official/recommendation/ncf_keras_main.py
@@ -224,7 +224,8 @@ def _get_keras_model(params):
       [zeros, logits],
       axis=-1)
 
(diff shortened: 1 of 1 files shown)


official/recommendation/ncf_keras_main.py
```diff
@@ -224,7 +224,8 @@ def _get_keras_model(params):
       [zeros, logits],
       axis=-1)
 
-  softmax_logits = MetricLayer(params)([softmax_logits, dup_mask_input])
+  if not params["keras_use_ctl"]:
```
guptapriya: Add a comment as to why this is not neeeded in ctl case

PR #7218, opened 2019-07-15
Initial implementation of Shakespeare character LSTM.

bfontain:


tfboyd:
Goal is to just take this, add to it and maybe replace with gnmt (or something larger).  The bar for 'staging' for now is anything we are working on and if it does not reach the level of useful we need/want we just remove it.  LGTM from me.

saberkun:
inting official/staging/shakespeare/shakespeare_main.py
Using config file /tmpfs/src/github/tensorflow_models/official/utils/testing/pylint.rcfile
************* Module shakespeare_main
C: 25, 0: third party import "import numpy as np" should be placed before "from absl import app as absl_app" (wrong-import-order)

PR #7221, opened 2019-07-15
Adds run_eagerly option for bert training utils

saberkun:
258208153  by hongkuny<hongkuny@google.com>:
    
    Adds run_eagerly option for bert.
    
--


zongweiz:
Thanks. Could you also add FLAGS.run_eagerly?

saberkun:
Hi Zongwei, 

We will do it in a different PR to call it in benchmark (maybe modifing run_classifier interface). We shall not disclose this as a flag at this moment until the performance is confident.

PR #7229, opened 2019-07-16
[NCF] Add run_eagerly for ctl.

tfboyd:


official/recommendation/ncf_keras_benchmark.py
```diff
@@ -147,6 +147,14 @@ def benchmark_1_gpu_ctl_early_stop(self):
     FLAGS.early_stopping = True
     self._run_and_report_benchmark()
 
+  def benchmark_1_gpu_ctl_default_ds_run_eagerly_early_stop(self):
+    self._setup()
+    FLAGS.distribution_strategy = 'default'
```
guptapriya: nice! did the default ds work out of the box?

tfboyd: yes and I just renamed the test.

tfboyd:
Fixed and updated the comment on the test.  PTAL

nnigania:
nice, this will allow us to run eager with CTL. Just has a question, I thought eager mode does not support distribution strategy?

official/recommendation/ncf_keras_benchmark.py
```diff
@@ -181,6 +181,13 @@ def benchmark_1_gpu_ctl_early_stop(self):
     FLAGS.early_stopping = True
     self._run_and_report_benchmark()
 
+  def benchmark_1_gpu_ctl_run_eagerly_early_stop(self):
```
guptapriya: Do we also want to do a non dist strat run eagerly?

tfboyd: Now I am confused again; and now I remember dist strat needs turned off.  I always look at this branch last when I am rushing to finish other stuff.  One more try coming up.  :-)

tfboyd: Other than removing tests this is the last test I am adding period... probably....no seriously...maybe.

nnigania:
hi,

the changes look good to me. Should be go ahead and checkin this change?

Issue #6825, opened 2019-05-20
Training Faster R-CNN from scratch: NaN loss during training


fuzzyBatman:
System: Debian 9 VM on GCP
GPU: Nvidia Tesla P100
Dataset: Custom dataset with red blood cell images (512x512)
Config: [Faster R-CNN ResNet50 COCO](https://github.com/tensorflow/models/blob/master/research/object_detection/samples/configs/faster_rcnn_resnet50_coco.config)

Changest to config:
```
num_classes: 1

min_dimension: 512
max_dimension: 512

first_stage_features_stride: 8

max_detections_per_class: 400
max_total_detections: 400
```

Train command:
```
PIPELINE_CONFIG_PATH="/home/jupyter/software/custom_detector/configs/faster_rcnn_resnet50_coco.config"
MODEL_DIR="/home/jupyter/software/custom_detector/pre_trained/faster_rcnn_resnet50_coco_2018_01_28"
SAMPLE_1_OF_N_EVAL_EXAMPLES=1
python3 ~/src/models/research/object_detection/model_main.py \
    --pipeline_config_path=${PIPELINE_CONFIG_PATH} \
    --train_dir=${MODEL_DIR} \
    --sample_1_of_n_eval_examples=$SAMPLE_1_OF_N_EVAL_EXAMPLES \
    --alsologtostderr
```

Result:
```
WARNING:tensorflow:Forced number of epochs for all eval validations to be 1.
WARNING:tensorflow:Expected number of evaluation epochs is 1, but instead encountered `eval_on_train_input_config.num_epochs` = 0. Overwriting `num_epochs` to 1.
WARNING:tensorflow:Using temporary folder as model directory: /tmp/tmpec2q810n
WARNING:tensorflow:Estimator's model_fn (<function create_model_fn.<locals>.model_fn at 0x7f082bd9b268>) includes params argument, but params are not passed to Estimator.
WARNING:tensorflow:num_readers has been reduced to 1 to match input file shards.
WARNING:tensorflow:From /home/jupyter/src/models/research/object_detection/builders/dataset_builder.py:80: parallel_interleave (from tensorflow.contrib.data.python.ops.interleave_ops) is deprecated and will be removed in a future version.
Instructions for updating:
Use `tf.data.experimental.parallel_interleave(...)`.
[...]

Issue #6536, opened 2019-04-07
object_detection: Using IoU localization loss raises "NaN loss during training"


bubbles1990:
### System information
- **What is the top-level directory of the model you are using**: object_detection
- **Have I written custom code (as opposed to using a stock example script provided in TensorFlow)**: No
- **OS Platform and Distribution (e.g., Linux Ubuntu 16.04)**: Linux Ubuntu 16.04
- **TensorFlow installed from (source or binary)**: binary
- **TensorFlow version (use command below)**: 1.13.0-dev20181228
- **Bazel version (if compiling from source)**: --
- **CUDA/cuDNN version**: 10.0/7.3.0
- **GPU model and memory**: GTX 1080 Ti 11GB
- **Exact command to reproduce**: python object_detection/model_main.py --model_dir=$MODEL_DIR --pipeline_config_path=$PIPELINE_CONFIG_PATH

### Describe the problem
When trying to use IoU localization loss, i.e. use these lines in the config file:
```
localization_loss {
    weighted_iou {
   }
}
```
After several training steps (usually 0 - 100) I always get the error:  `tensorflow.python.training.basic_session_run_hooks.NanLossDuringTrainingError: NaN loss during training.`

If I add a small constant (e.g. 1e-7) to the denominator in `matched_iou` (in [`object_detection/core/box_list_ops.py`](https://github.com/tensorflow/models/blob/master/research/object_detection/core/box_list_ops.py#L277)) like so:
```
def matched_iou(boxlist1, boxlist2, scope=None):
  """Compute intersection-over-union between corresponding boxes in boxlists.

  Args:
    boxlist1: BoxList holding N boxes
    boxlist2: BoxList holding N boxes
    scope: name scope.

  Returns:
    a tensor with shape [N] representing pairwise iou scores.
  """
  with tf.name_scope(scope, 'MatchedIOU'):
    intersections = matched_intersection(boxlist1, boxlist2)
    areas1 = area(boxlist1)
    areas2 = area(boxlist2)
    unions = areas1 + areas2 - intersections
    eps = 1e-7  # <- this is the small constant
    return tf.where(
        tf.equal(intersections, 0.0),
        tf.zeros_like(intersections), tf.truediv(intersections, unions + eps))
```
The `NaN` error is gone but loss values are really high and the training doesn't converge.

Did anyone experienced this issue or has any experience with IoU localization loss?
Thank you.

tensorflowbutler:
Hi There,
[...]

PR #7233, opened 2019-07-17
Guard the unsafe tf.exp operations to prevent possible INF or NAN values

ForeverZyh:
The inputs of `tf.exp` come from a `conv2d`, without any guard. The inputs of `tf.exp` can be large than 88 (`tf.exp(x)` with `x>88` if stored in `tf.float32` will cause INF) during training, which may cause INF values. The INF values will further influence the back propagation to cause NAN values.

Not sure if the latest issues are related. [6825](https://github.com/tensorflow/models/issues/6825) [6536](https://github.com/tensorflow/models/issues/6536)

### DIFF
--- a/research/object_detection/box_coders/faster_rcnn_box_coder.py
+++ b/research/object_detection/box_coders/faster_rcnn_box_coder.py
@@ -107,8 +107,8 @@ class FasterRcnnBoxCoder(box_coder.BoxCoder):
       tx /= self._scale_factors[1]
       th /= self._scale_factors[2]
       tw /= self._scale_factors[3]
-    w = tf.exp(tw) * wa
-    h = tf.exp(th) * ha
+    w = tf.exp(tf.minimum(tw, 80)) * wa
+    h = tf.exp(tf.minimum(th, 80)) * ha
     ycenter = ty * ha + ycenter_a
     xcenter = tx * wa + xcenter_a
     ymin = ycenter - h / 2.
--- a/research/deep_speech/deep_speech.py
+++ b/research/deep_speech/deep_speech.py
@@ -223,7 +223,7 @@ def run_deep_speech(_):
 
   # Use distribution strategy for multi-gpu training
   num_gpus = flags_core.get_num_gpus(flags_obj)
-  distribution_strategy = distribution_utils.get_distribution_strategy(num_gpus)
+  distribution_strategy = distribution_utils.get_distribution_strategy(num_gpus=num_gpus)
   run_config = tf.estimator.RunConfig(
       train_distribute=distribution_strategy)
 


research/object_detection/box_coders/faster_rcnn_box_coder.py
```diff
@@ -107,8 +107,8 @@ def _decode(self, rel_codes, anchors):
       tx /= self._scale_factors[1]
       th /= self._scale_factors[2]
       tw /= self._scale_factors[3]
-    w = tf.exp(tw) * wa
-    h = tf.exp(th) * ha
+    w = tf.exp(tf.minimum(tw, 80)) * wa
```
pkulzc: Add a comment explaining how you get the number 80 here?

ForeverZyh: Avoid INF in exp and w,h below. We set 80 as a upper_bound since tf.exp(89) yields INF.

research/object_detection/box_coders/faster_rcnn_box_coder.py
```diff
@@ -107,8 +107,10 @@ def _decode(self, rel_codes, anchors):
       tx /= self._scale_factors[1]
       th /= self._scale_factors[2]
       tw /= self._scale_factors[3]
-    w = tf.exp(tf.minimum(tw, 80)) * wa
-    h = tf.exp(tf.minimum(th, 80)) * ha
+    # Avoid INF in exp and w,h below. We set 80 as a upper_bound since tf.exp(89) yields INF.
+    UPPER_BOUND = 80
```
pkulzc: Could you please make the name more specific (e.g. UPPER_BOUND_OF_XXX) and move it to the top of this file(near line 36) ?

ForeverZyh: Sure. I've done it.

research/object_detection/core/box_list_ops.py
```diff
@@ -290,9 +291,10 @@ def matched_iou(boxlist1, boxlist2, scope=None):
     areas1 = area(boxlist1)
     areas2 = area(boxlist2)
     unions = areas1 + areas2 - intersections
+    eps = 1e-7
     return tf.where(
         tf.equal(intersections, 0.0),
-        tf.zeros_like(intersections), tf.truediv(intersections, unions))
+        tf.zeros_like(intersections), tf.truediv(intersections, tf.maximum(unions, eps)))
```
pkulzc: Can you directly use the global constant EPSILON here? Same for other eps you added.

jaeyounkim:
@pkulzc Please review this pull request again.

PR #7240, opened 2019-07-17
Update keras-cifar-main to user sparse data and loss function.

pavithrasv:
keras-cifar-main synthetic data generates sparse data, whereas the real data generates one-hot labels. The real data and model loss/metric are correct but the synthetic data is wrong with respect to this model parameters. We can convert the synthetic data to return one-hot labels but this synthetic data is shared across other models where sparse data is used. I figured it is best to convert keras-cifar-main to use sparse data as well.

### DIFF
--- a/official/resnet/keras/keras_cifar_main.py
+++ b/official/resnet/keras/keras_cifar_main.py
@@ -65,26 +65,6 @@ def learning_rate_schedule(current_epoch,
   return learning_rate
 
 
-def parse_record_keras(raw_record, is_training, dtype):
-  """Parses a record containing a training example of an image.
-
-  The input record is parsed into a label and image, and the image is passed
-  through preprocessing steps (cropping, flipping, and so on).
-
-  This method converts the label to one hot to fit the loss function.
-
-  Args:
-    raw_record: scalar Tensor tf.string containing a serialized
-      Example protocol buffer.
-    is_training: A boolean denoting whether the input is for training.
-    dtype: Data type to use for input images.
-
-  Returns:
-    Tuple with processed image tensor and one-hot-encoded label tensor.
-  """
-  return cifar_main.parse_record(raw_record, is_training, dtype)
-
-
 def run(flags_obj):
   """Run ResNet Cifar-10 training and eval loop using native Keras APIs.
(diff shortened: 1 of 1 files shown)


tfboyd:
I will check it out.  We just want it to match resnet50.

tfboyd:
I will kick off a quick benchmark as well.

pavithrasv:
Thank you Toby!

official/resnet/keras/keras_cifar_main.py
```diff
@@ -82,9 +82,7 @@ def parse_record_keras(raw_record, is_training, dtype):
   Returns:
     Tuple with processed image tensor and one-hot-encoded label tensor.
   """
-  image, label = cifar_main.parse_record(raw_record, is_training, dtype)
-  label = tf.compat.v1.sparse_to_dense(label, (cifar_main.NUM_CLASSES,), 1)
-  return image, label
+  return cifar_main.parse_record(raw_record, is_training, dtype)
```
tfboyd: While I do a test, can you remove this method all together so it is not a pass through.

pavithrasv: Done, thank you!

tfboyd:
Taking me longer than I would like.  Haoyu also said this looks ok.  I needed to finish the LSTM example first.  It may take until next week to validate performance and double check convergence.

pavithrasv:
Thank you for the update!

pavithrasv:
Hi all, just wanted to check if there is an update on this and if we can merge this change, thank you!

pavithrasv:
did the convergence tests pass? can we merge this change?

guptapriya:
@saberkun can someone from the model garden help review and validate this change?

pavithrasv:
This is blocking a bug fix change, would like to submit that by end of this week if possible. Thank you!

saberkun:
Hi, pavithrasv@
Could you please resolve the conflict first?
Please ping me internally. Note that, you are free to work on internal repo without hassle to sync to github.

pavithrasv:
Submitted this change internally.

PR #7253, opened 2019-07-19
Fix shakespeare doc

saberkun:
258871624  by hongkuny<hongkuny@google.com>:


saberkun:
This is actually testing the updated workflow to create PR from internal.

PR #7254, opened 2019-07-19
Change how TF 2 is checked

guptapriya:
The current approach checks for presence of contrib. Sometimes this is not sufficient (for e..g when testing TF 1 + enable_v2_behavior=True which is what internal tests currently do)


guptapriya:
All the tests passed! The lint error is unrelated.

PR #7256, opened 2019-07-19
NCF Keras: Fail early with TF 1.x + dist strat

guptapriya:
This combination does not yet work. Fail early with an explicit message instead of throwing error later on.


guptapriya:
All the tests pass, fixed one of the lint error, the other one is unrelated

PR #7257, opened 2019-07-19
Transformer Keras: Use add_loss instead of loss layer

guptapriya:
model.add_loss is (1) simpler (2) works in both 1.x and 2.0 with distribution strategy.

NOTE: the change was approved in https://github.com/tensorflow/models/pull/7077 previously but that PR got a bit messed up so re-doing it in a fresh client


guptapriya:
all green!

PR #7258, opened 2019-07-19
Update pylint.rcfile

saberkun:


tfboyd:
@guptapriya   This seems fine with me and I am guessing this might make it better match internal.

tfboyd:
Your change triggers a lot more lint issues.  Good/Bad  I don't know.  FYI: I am totally fine if we need to spend a couple hours powering through lint issues to line up with a new set of "rules".

guptapriya:
+1 to Toby, some description would be nice. Is this copied from somewhrre else, what rules are we  trying to follow?

saberkun:
Sorry. This is experimental and I should drop it. In fact, the previous
lintrc disabled many important things. Will try a few other improvements.

PR #7262, opened 2019-07-19
update v2 check and fix ncf v2 check error logic.

tfboyd:
Added back your change and the NCF fix.


tfboyd:
Tested on V1 this works fine, it gives the error and perfzero marks the test failed.  I will wait for all tests to pass..I hope; and the submit.

PR #7265, opened 2019-07-19
Add a new sanity check script that is able to only check incremental changes.

saberkun:
This is going to replace current lint presubmit.
After that, we can enforce correct lint rules. Now, in pylint.rcfile, there is a disable=W which disables most warning.


saberkun:
PTAL.

PR #7276, opened 2019-07-22
Single execution path tests for ResNet50, ResNet56, NCF, and Shakespeare LSTM.

tfboyd:


official/utils/flags/_performance.py
```diff
@@ -276,4 +280,11 @@ def _check_fp16_implementation(flags_dict):
         name="enable_xla", default=False,
         help="Whether to enable XLA auto jit compilation")
 
+  if force_run_distributed:
+    flags.DEFINE_boolean(
+        name="force_run_distributed", default=False,
```
haoyuz: Everything looks great except that the name of the flag looks a little confusing to me at first glance (maybe also to others without enough context). Maybe rename it `run_default_strategy` or `force_dist_strat_code_path`?

guptapriya: +1 agree. I think the original arg in model.compile is also confusing but they are planning to rename it. we should give this new flag a better name to start with. They are considering something like "run_v2". So maybe this can be something like "force_v2_in_keras_compile" ?

official/resnet/keras/keras_cifar_benchmark.py
```diff
@@ -147,38 +147,69 @@ def benchmark_1_gpu_no_dist_strat_run_eagerly(self):
...
+    FLAGS.model_dir = self._get_model_dir('benchmark_1_gpu_force_dist_strat')
     FLAGS.dtype = 'fp32'
     FLAGS.enable_eager = True
+    FLAGS.distribution_strategy = 'off'
+    FLAGS.force_run_distributed = True
     self._run_and_report_benchmark()
 
-  def benchmark_graph_2_gpu(self):
-    """Test keras based model with Keras fit and distribution strategies."""
+  def benchmark_1_gpu_force_dist_strat_run_eagerly(self):
```
guptapriya: I am not sure whether the new "force_run_distributed" has any impact when "run_eagerly=True". My understanding was that it was only useful for the default case which uses tf.function/func-graph etc. Worth confirming otherwise we may be running 2 identical benchmarks

guptapriya: Just checked with @isaprykin , it seems like it will have a different codepath, so probably fine to benchmark both versions then.

official/utils/flags/_performance.py
tfboyd: I am going to go with Pryia's because it is cryptic and if it stays in I don't want people using it.  I would rather it be confusing.

PR #7278, opened 2019-07-23
Update lint presubmit to be consistent with tensorflow

saberkun:
Only care about errors and output into an error file.


saberkun:
Current error count is also wrong. That's why we see always see errors.

PR #7299, opened 2019-07-25

saberkun:
259861112  by hongkuny<hongkuny@google.com>:
    
    Add no ds and xla perfzero tests

### DIFF
--- a/official/bert/benchmark/bert_benchmark.py
+++ b/official/bert/benchmark/bert_benchmark.py
@@ -270,7 +270,7 @@ class BertClassifyAccuracy(BertClassifyBenchmarkBase):
         max_accuracy=max_accuracy)
 
   def _setup(self):
-    super(BertClassifyAccuracy, self).__init__()
+    super(BertClassifyAccuracy, self)._setup()
     FLAGS.train_data_path = self.train_data_path
     FLAGS.eval_data_path = self.eval_data_path
     FLAGS.input_meta_data_path = self.input_meta_data_path
--- a/official/bert/benchmark/bert_benchmark.py
+++ b/official/bert/benchmark/bert_benchmark.py
@@ -33,6 +33,7 @@ from official.bert import modeling
 from official.bert import run_classifier
 from official.bert.benchmark import benchmark_utils
 from official.utils.misc import distribution_utils
+from official.utils.misc import keras_utils
 
 # pylint: disable=line-too-long
 PRETRAINED_CHECKPOINT_PATH = 'gs://cloud-tpu-checkpoints/bert/tf_20/uncased_L-24_H-1024_A-16/bert_model.ckpt'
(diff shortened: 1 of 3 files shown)


saberkun:
For testing

tfboyd:
We really really want an eager test as we do not have many CTL Eager tests and that is where eager is most likely going to be used.  I am added NCF.

official/bert/benchmark/bert_benchmark.py
```diff
@@ -70,7 +71,10 @@ def _run_bert_classifier(self, callbacks=None):
     eval_steps = int(
         math.ceil(input_meta_data['eval_data_size'] / FLAGS.eval_batch_size))
     strategy = distribution_utils.get_distribution_strategy(
-        distribution_strategy='mirrored', num_gpus=self.num_gpus)
+        distribution_strategy='mirrored' if use_ds else 'noop',
+        num_gpus=self.num_gpus)
```
tfboyd: It would be great to follow the pattern of the other models and use "off" which then triggers dist strat to be None like here:
https://github.com/tensorflow/models/blob/master/official/utils/misc/distribution_utils.py#L121

The test I want to duplicate is "off" to match the other tests.

It does mean your with scope (if used) needs to also call the util to get a scope that is not None or some other trick.  If Priya wants "off" to be tf.distribute.get_strategy() then we just updated the code.

official/bert/benchmark/bert_benchmark.py
```diff
@@ -54,7 +55,7 @@ def __init__(self, output_dir=None):
     self.num_steps_per_epoch = None
 
   @flagsaver.flagsaver
-  def _run_bert_classifier(self, callbacks=None):
+  def _run_bert_classifier(self, callbacks=None, use_ds=True, enable_xla=False):
```
tfboyd: How is a user going to turn on XLA in the actual code?  This will make it hard to debug and we really cannot have other users try this out or get the benefit.  Maybe if we added a TODO(b/) so we know this needs rolled into the model.

official/bert/benchmark/bert_benchmark.py
```diff
@@ -148,6 +155,38 @@ def benchmark_1_gpu_mrpc(self):
...
+    FLAGS.eval_data_path = self.eval_data_path
+    FLAGS.input_meta_data_path = self.input_meta_data_path
+    FLAGS.bert_config_file = self.bert_config_file
+    FLAGS.train_batch_size = 4
+    FLAGS.eval_batch_size = 4
+
+    summary_path = os.path.join(FLAGS.model_dir, 'training_summary.txt')
+    self._run_and_report_benchmark(summary_path, use_xla=True)
+
+  def benchmark_1_gpu_mrpc_no_dist_strat(self):
```
tfboyd: not dist_strat for us means "off" and returns as None.  We can change that but that is the pattern the other 4 models are following.

official/utils/misc/distribution_utils.py
```diff
@@ -120,6 +120,9 @@ def get_distribution_strategy(distribution_strategy="default",
           "flag cannot be set to 'off'.".format(num_gpus, num_workers))
     return None
 
+  if distribution_strategy == "noop":
```
tfboyd: We already have "off", this  "noop" is not a path we are testing.  We could but I would not want to call it "no dist strat".  I am also winding down testing for the RC and just want to get coverage for the matrix we have so far.  This also seems like it would be really close to the "one execution" path; which we are also testing which will be the default.

saberkun:
Not ready to review. I am using it to test perfzero as we do not want blind commit.
Will close and send another pr...

PR #7303, opened 2019-07-25
Force GPU placement for Resnet pure eager benchmark

zongweiz:


official/resnet/keras/keras_imagenet_benchmark.py
```diff
@@ -257,6 +257,7 @@ def benchmark_1_gpu_no_dist_strat_run_eagerly(self):
     FLAGS.num_gpus = 1
     FLAGS.enable_eager = True
     FLAGS.run_eagerly = True
+    FLAGS.explicit_gpu_placement = True
```
guptapriya: i believe if we are adding explicit changes in the model code that we don't expect normal users to have, we should call this a "tweaked version" of this benchmark? @tfboyd wdyt?

tfboyd: correct, let's make a new test that is "forced_placement"  or we put all the "tweaks" into a single test called "tweaked" and we tracked it with TODO(b/xxxxx) for each flag that makes it faster.  I personally think "tweaked" or "tuned" is a better option and we group them all.  If you want to track the exact impact of a given FLAG make a specific test.  This will help us drive the changes into the core product, if possible.

zongweiz: @tfboyd @guptapriya Thanks for the suggestions. I have updated the code to add the tweaked version of tests. PTAL

PR #7313, opened 2019-07-26

guptapriya:
260228553  by priyag<priyag@google.com>:
    
    Enable transformer and NCF official model tests. Also fix some minor issues so that all tests pass with TF 1 + enable_v2_behavior. 
    
--


guptapriya:
Need to sync github first

PR #7332, opened 2019-07-30
use public app runner and flags

richardbrks:
- uses `tf.app` instead of `google3.pyglib.app`
- uses `absl.flags` (already imported) instead if instead of `google3.pyglib.flags`

This removes the need for the `google3` module for the entire project, and also matches `train.py`'s use.


richardbrks:
closing to get a more substantial PR out.

PR #982, opened 2017-02-05
fixed a bug in sampled_loss(), made compatible for 0.12.0

arvind385801:
One of the main changes is swapping the arguments of the function sampled_loss(). Also made appropriate changes to package reference so it is able to run with the tensorflow's version available in the repo.

Thanks
Arvind

PR #7338, opened 2019-07-31
Replace deprecated functions  tf.rnn.nn 2 tf.contrib.rnn

Ikuyadeu:
Referred #881 and #982
This is a very easy fix for the modern style.
This change following this repositorys' most files that are using `contrib.rnn` instead of `rnn.nn`.

### DIFF
--- a/research/deep_speech/deep_speech_model.py
+++ b/research/deep_speech/deep_speech_model.py
@@ -22,9 +22,9 @@ import tensorflow as tf
 
 # Supported rnn cells.
 SUPPORTED_RNNS = {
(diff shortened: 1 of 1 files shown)


Ikuyadeu:
@googlebot I signed it!

Ikuyadeu:
@jaeyounkim Thank you for merge this!
This change is made by the [tensorflow coding rules](https://github.com/devreplay/devreplay/blob/master/src/rules/tensorflow.ts) that are available as [lint](https://devreplay.github.io/)

PR #7344, opened 2019-07-31
Change to experimental_run_tf_function.

tfboyd:


tfboyd:
TF 1.0 presubmits are pulling the wrong TensorFlow.  It should pass but I can catch the issue in the benchmark and open a bug if needed.  I do not have time to fix the presubmit to use docker like the GPU one for TF 2.0.

PR #7359, opened 2019-08-02
Remove layout off tests and related utils.

tfboyd:
Seeing if the tests will pass and putting this here so I do not forget.  Will wait for Haoyu to finish the grand refactor/move.


tfboyd:
Taking a risk and merging.  The errors look to be based on old code.

PR #7370, opened 2019-08-02
add an encoder mode for BertModel which returns all layers.

saberkun:
261393597  by hongkuny<hongkuny@google.com>:
    
    add an encoder mode for BertModel which returns all layers.
    
--


saberkun:
Hmmm, the lint is actually wrong....

PR #7383, opened 2019-08-05
[ResNet / NCF] Test force V1 path and allow V2 path as default 

tfboyd:


tfboyd:
@guptapriya  @isaprykin   Need this kind of fast so we can test.  I see some small naming issues, if you can check it out ASAP that would be great.  I hope to run some tests tonight.  I can run on the branch repo obviously as well.  Thank you if you have time.

tfboyd:
ERROR: test_train_2_gpu_fp16 

The rest of the persubmit errors are already fixed.  I am going to run some perf tests standalone first given the impact of this change.

tfboyd:
https://data.corp.google.com/sites/q1l099kp62mc/home/?f=model:in:resnet50_v1_5&f=accel_cnt_2:in:1&f=days_old:bt:0,14&f=build_label:in:custom-manual&f=label:in:prod,tobyboyd&f=dtype:in:fp32&f=xla:in:N

Reverse looks good. The force V1 path is slower as expected, a bit pointless but let it run for now.

PR #7390, opened 2019-08-06
[ResNet56] Metric fixed back to categorical_accuracy.

tfboyd:


tfboyd:
Transformer FP16 still broken. If my day had gone better I might have disabled it.  :-)

PR #7391, opened 2019-08-06
Add non-CuDNN LSTM option and PerfZero benchmarks.

robieta:


robieta:
Thanks. Looks like the test failures are related to the tf build rather than this PR, so I'm going to merge as is. (The git version of the two marshmallow test must surely be submit with the dull grey button or wait for the shiny green button. And clearly I'm in the former camp...)

PR #7402, opened 2019-08-07
Fix fp16 Transformer model.

reedwm:
Also, do Transformer inference in fp16, as well as training, when --dtype=fp16. In TF 2, layers now cannot run in multiple different dtypes, so we must use the same dtype for training and inference.

### DIFF
--- a/official/transformer/model/beam_search.py
+++ b/official/transformer/model/beam_search.py
@@ -18,11 +18,29 @@ Source implementation from Tensor2Tensor:
 https://github.com/tensorflow/tensor2tensor/blob/master/tensor2tensor/utils/beam_search.py
 """
 
+import numpy as np
 import tensorflow as tf
 from tensorflow.python.util import nest
 
-# Default value for INF
-INF = 1. * 1e7
+
+def inf(dtype):
+  """Returns a value close to infinity, but is still finite in `dtype`.
+
+  This is useful to get a very large value that is still zero when multiplied by
+  zero. The floating-point "Inf" value is NaN when multiplied by zero.
+
+  Args:
+    dtype: A dtype. The returned value will be finite when casted to this dtype.
+
+  Returns:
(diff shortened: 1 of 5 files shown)


tfboyd:
@reedwm does this fix the overall issue with Transformer and FP16.  I would like to do an end-to-end run with this PR before we accept.  I feel odd accepting this when all FP16 is broken currently for Transformer.  Open to a dialog to move this forward.

reedwm:
Yes, this fixes the overall issue with Transformer and fp16. In order to fix it, I had to do inference in fp16 as well, but it should all work.

I can start off an end-to-end run

tfboyd:
Presubmit passing is nice.  Please check lint, that check has been greatly improved.  Has a few quirks still but mostly better.

reedwm:
Not sure why I'm failing lint or presubmit. For lint, it claims some classes aren't callable when they are. Presubmit shows me failing unrelated tests.

How do I rerun the lint and presubmit?

tfboyd:
@reedwm  I was saying presubmit is finally passing with the FP16 fixes.  The TF 1.0 presubmit will be removed soon.  I already removed one of them.  You can try to trigger the lint again with `kokoro: force-run` I already added it.

tfboyd:
We may want to as @saberkun if he can turn off some of the lint checks that do not seem to be accurate.  I had a non-callable situation as well and I ended up ignoring it.  His approach is the right way; we may just need some tweaks to the rules or the environment where the checks take place.  If this happens again just make a comment, you don't need to add the disable, the lint only checks the parts you update (or is supposed to limit it to that) vs. the old lint check which made you pay for everyone else that was sloppy.

reedwm:
Running the model with this PR, the beak uncased BLEU score was 28.235626220703125 at the end of epoch 7.

PR #7416, opened 2019-08-08
[Transformer] enable static loss scaling

sgpyc:
This is a workaround to avoid tf.cond() in LossScaleOptimizer.
For fp16, single GPU step time (3072 static batch) reduces ~5ms and 8 GPU step time (3072 x 8 static batch, without XLA) reduces ~35ms.

set --loss_scale=256 and dynamic batching can yield similar Bleu scores as LossScaleOptimizer, reaching {28.45, 27.95} at 200000 steps for {uncased, cased} Bleu.

### DIFF
--- a/official/transformer/v2/transformer_main.py
+++ b/official/transformer/v2/transformer_main.py
@@ -129,6 +129,11 @@ class TransformerTask(object):
       policy = tf.keras.mixed_precision.experimental.Policy(
           "infer_float32_vars")
       tf.keras.mixed_precision.experimental.set_policy(policy)
+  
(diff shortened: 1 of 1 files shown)


sgpyc:
@tfboyd I saw some setting in utils/flags/_performance.py, but looks like changing that will change the default for all models.
Please let me know how to set it for Transformer only. Thank you very much

tfboyd:
Yeah it is messy.  I got your back.  We can get it in another PR.  I will
find the method, it is not obvious.  Sorry you used time looking around.

PR #7436, opened 2019-08-13
NCF fp16 TF1.x support

nnigania:
adding support for fp16 using the TF1.x API


official/recommendation/ncf_keras_main.py
```diff
@@ -355,8 +359,12 @@ def step_fn(features):
             softmax_logits,
             sample_weight=features[rconst.VALID_POINT_MASK])
         loss *= (1.0 / params["batch_size"])
+        if FLAGS.dtype == "fp16":
```
guptapriya: does this support fp16 only in the CTL path? 
or does wrapping the optimizer change this for the compile/fit version too?

official/recommendation/ncf_keras_benchmark.py
```diff
@@ -277,6 +278,16 @@ def benchmark_xla_1_gpu_ctl_mlperf_like(self):
     FLAGS.train_epochs = 7
     self._run_and_report_benchmark_mlperf_like()
 
+  def benchmark_xla_1_gpu_ctl_fp16_mlperf_like(self):
```
guptapriya: also a benchmark without XLA but with fp16?

official/recommendation/ncf_keras_main.py
```diff
@@ -259,6 +260,9 @@ def run_ncf(_):
           beta_1=params["beta1"],
           beta_2=params["beta2"],
           epsilon=params["epsilon"])
+      if FLAGS.dtype == "fp16":
+        optimizer = tf.compat.v1.train.experimental.enable_mixed_precision_graph_rewrite(
```
guptapriya: are we planning to add the new keras way of doing this afterwards? if yes, how will we distinguish which approach to use from the flags? what mechanism do we use in the other mdoels?

nnigania: once the tf2.0 API for fp16 is done, we will start using that. Other models use the same flag and same technique for fp16. See: official/transformer/transformer_main.py, official/r1/resnet/resnet_run_loop.py

official/recommendation/ncf_keras_benchmark.py
nnigania: sure, will add that I did not wanted to add them gradually. But I can add it now as well.

official/recommendation/ncf_keras_main.py
nnigania: This should support both (havent tested comp/fit yet). As compile/fit does not do any special gradients calculation we dont need to do any manual loss scaling that we are doing in the CTL version which uses gradient.tape()

nnigania: just tested compile/fit and it work with this API as well.

nnigania:
@guptapriya please let me know if the changes look good or I need to make more changes.

official/recommendation/ncf_keras_main.py
guptapriya: hm, i still don't understand why CTL needs the loss scaling but in compile fit we don't ?

nnigania: https://www.tensorflow.org/api_docs/python/tf/train/experimental/enable_mixed_precision_graph_rewrite
If we use "opt.minimize() or opt.compute_gradients()" to compute gradients which compile/fit does internally, then the optimizer does the loss scaling for us. But if we use Tf.GradientTape like in CTL to compute gradients, then we need to do the scaling manually.

PR #7461, opened 2019-08-16
 Adding TF2.0 support in an older commit for Wide and Deep

ashraf-bhuiyan:
Adding support of TF 2.0 to and older commit.


saberkun:
Hi, it conflicts with current master.
Could you resolve them?
Thank you for taking care of the symbol changes.

PR #7470, opened 2019-08-19

yongzhe2160:
263863588  by yongzhe:

    Fix a bug that the SetExternalContext for EdgeTPU wasn't called when initializing LSTD client.

--
263370193  by yongzhe:

    Internal change.

--

PiperOrigin-RevId: 263863588


yongzhe2160:
Reverted that change.

PR #7474, opened 2019-08-20
Use new mixed_float16 policy for resnet.

reedwm:
The old infer_float32_policies policy will be removed from TensorFlow soon.

To test convergence, I ran the Resnet50KerasAccuracy.benchmark_8_gpu_fp16 benchmark. I got an accuracy of 0.76037 and an exp_per_second of 6908.


saberkun:
Tests failed?

saberkun:
Can we disable no-name-in-module or other wrong errors inside pylint.rc?

reedwm:
I don't think the test failures are caused by this PR.

I'm not sure how to disable lint errors inside pylint.rc. I can figure it out in a future PR.

saberkun:
I don't see these tests are failing internally. Either it is the issue with docker or the changes

reedwm:
Hmmm, looks like the test failures are caused by this PR?

Any ideas how to debug this? I cannot reproduce locally, and the error does not seem related to fp16 at all.

PR #7485, opened 2019-08-21
Use new mixed_float16 policy for transformer.

reedwm:
The old infer_float32_policies policy will be removed from TensorFlow soon.

To test, I ran the TransformerBigKerasAccuracy.benchmark_8_gpu_fp16 PerfZero benchmark. The max BLEU score was 28.63 on epoch 10. Unfortunately, the run crashed during Epoch 11, but I tried doing a clean run and it also crashed, so I don't think it was caused by this PR

### DIFF
--- a/official/transformer/v2/embedding_layer.py
+++ b/official/transformer/v2/embedding_layer.py
@@ -24,24 +24,14 @@ import tensorflow as tf
 class EmbeddingSharedWeights(tf.keras.layers.Layer):
   """Calculates input embeddings and pre-softmax linear with shared weights."""
 
-  def __init__(self, vocab_size, hidden_size, dtype=None):
+  def __init__(self, vocab_size, hidden_size):
     """Specify characteristic parameters of embedding layer.
 
     Args:
       vocab_size: Number of tokens in the embedding. (Typically ~32,000)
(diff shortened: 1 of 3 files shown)


reedwm:
> tf.float32?

Done.

saberkun:
Why it is crashed? What's the error log?
is it related to the grappler bug?

reedwm:
It crashed with the error "python: can't open file '/tmpfs/kokoro_log_reader.py': [Errno 2] No such file or directory". I don't think this is caused by the PR.

I also ran on a clean client, and it crashed with the error "ERROR: Aborting VM command due to timeout of 86400 seconds". I'm not sure why it didn't crash with the same error.

saberkun:
Please get approval from toby or priya and then apply "ready to pull" tag?

tfboyd:
If you are doing long runs on kokor you need to pass --long or the VM only
runs 24 hours.  "long" is ~48 hours.  I also do not think this is related
to the code (It seems related to kokoro) and that number is acceptable.
THe run is 11 iterations (I think) and I record the best score along with
the final score.

saberkun:
Dropping this PR because of internal tests failure and copybara ongoing bug....

PR #7535, opened 2019-09-05
[TF2.0] Adding automatic mixed precision training (via graph rewrite) support to Resnet CTL mode

houtoms:
This PR adds automatic mixed precision training to Resnet in CTL mode via setting the flag combination of --dtype=fp16 --fp16_implementation=graph_rewrite.

Corresponding benchmarks are added as well.

### DIFF
--- a/official/resnet/ctl/ctl_imagenet_benchmark.py
+++ b/official/resnet/ctl/ctl_imagenet_benchmark.py
@@ -139,6 +139,24 @@ class Resnet50CtlAccuracy(CtlBenchmark):
     FLAGS.datasets_num_private_threads = 14
     self._run_and_report_benchmark()
 
+  def benchmark_8_gpu_amp(self):
+    """Test Keras model with eager, dist_strat and 8 GPUs with automatic mixed
+    precision.
+    """
+    self._setup()
+    FLAGS.num_gpus = 8
+    FLAGS.data_dir = self.data_dir
+    FLAGS.batch_size = 128 * 8
+    FLAGS.train_epochs = 90
+    FLAGS.epochs_between_evals = 10
+    FLAGS.model_dir = self._get_model_dir('benchmark_8_gpu_amp')
+    FLAGS.dtype = 'fp16'
+    FLAGS.fp16_implementation = 'graph_rewrite'
+    FLAGS.use_tf_function = True 
+    # Add some thread tunings to improve performance.
+    FLAGS.datasets_num_private_threads = 14
+    self._run_and_report_benchmark()
+
   def _run_and_report_benchmark(self):
     start_time_sec = time.time()
     stats = ctl_imagenet_main.run(flags.FLAGS)
@@ -206,6 +224,33 @@ class Resnet50CtlBenchmarkBase(CtlBenchmark):
     FLAGS.batch_size = 128
     self._run_and_report_benchmark()
 
+  def benchmark_1_gpu_amp(self):
+    """Test Keras model with 1 GPU with automatic mixed precision."""
+    self._setup()
+
+    FLAGS.num_gpus = 1
+    FLAGS.distribution_strategy = 'default'
+    FLAGS.model_dir = self._get_model_dir('benchmark_1_gpu_amp')
+    FLAGS.batch_size = 128
+    FLAGS.dtype = 'fp16'
+    FLAGS.fp16_implementation = 'graph_rewrite'
+    FLAGS.use_tf_function = True 
+    self._run_and_report_benchmark()
+
+  def benchmark_xla_1_gpu_amp(self):
+    """Test Keras model with XLA and 1 GPU with automatic mixed precision."""
+    self._setup()
+
+    FLAGS.num_gpus = 1
+    FLAGS.distribution_strategy = 'default'
+    FLAGS.model_dir = self._get_model_dir('benchmark_xla_1_gpu_amp')
+    FLAGS.batch_size = 128
+    FLAGS.dtype = 'fp16'
+    FLAGS.fp16_implementation = 'graph_rewrite'
+    FLAGS.use_tf_function = True 
+    FLAGS.enable_xla = True
+    self._run_and_report_benchmark()
+
   def benchmark_1_gpu_eager(self):
     """Test Keras model with 1 GPU in pure eager mode."""
     self._setup()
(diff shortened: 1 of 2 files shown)


official/resnet/ctl/ctl_imagenet_main.py
```diff
@@ -171,6 +171,14 @@ def run(flags_obj):
         learning_rate=common.BASE_LEARNING_RATE, momentum=0.9,
         nesterov=True)
 
+    if flags_obj.fp16_implementation == "graph_rewrite":
+      if not flags_obj.use_tf_function:
+        raise ValueError("--fp16_implementation=graph_rewrite requires "
+                         "use_tf_function to be true")
```
reedwm:
```suggestion
                         "--use_tf_function to be true")
```

official/resnet/ctl/ctl_imagenet_main.py
```diff
@@ -171,6 +171,14 @@ def run(flags_obj):
...
+      optimizer = tf.train.experimental.enable_mixed_precision_graph_rewrite(
+                      optimizer)
+      loss_scale = flags_core.get_loss_scale(flags_obj, default_for_fp16=128)
```
reedwm: You should pass this loss_scale to the second argument of `enable_mixed_precision_graph_rewrite`.

official/resnet/ctl/ctl_imagenet_main.py
```diff
@@ -203,7 +211,19 @@ def step_fn(inputs):
             loss += (l2_loss / num_replicas)
           else:
             loss += (tf.reduce_sum(model.losses) / num_replicas)
+
+          # Scale the loss
+          if flags_obj.fp16_implementation == "graph_rewrite":
+            loss = loss * tf.cast(loss_scale, loss.dtype)
```
reedwm: Better to use [`opt.get_scaled_loss`](https://www.tensorflow.org/versions/r2.0/api_docs/python/tf/keras/mixed_precision/experimental/LossScaleOptimizer#get_scaled_loss), and [`opt.get_unscaled_gradients`](https://www.tensorflow.org/versions/r2.0/api_docs/python/tf/keras/mixed_precision/experimental/LossScaleOptimizer#get_unscaled_gradients) below.

Also, I would have the condition be `if flags_obj.dtype == "fp16"` instead, because this logic will also be needed for the tf.keras mixed precision API.

official/resnet/ctl/ctl_imagenet_main.py
```diff
@@ -296,6 +316,5 @@ def main(_):
   logging.set_verbosity(logging.INFO)
   common.define_keras_flags()
   ctl_common.define_ctl_flags()
-  flags.adopt_module_key_flags(keras_common)
```
reedwm: Why remove this line?

official/resnet/ctl/ctl_imagenet_benchmark.py
```diff
@@ -139,6 +139,24 @@ def benchmark_8_gpu(self):
     FLAGS.datasets_num_private_threads = 14
     self._run_and_report_benchmark()
 
+  def benchmark_8_gpu_amp(self):
+    """Test Keras model with eager, dist_strat and 8 GPUs with automatic mixed
```
reedwm: I don't know why the lint test doesn't catch this, but the first sentence of a docstring should only span 1 line. The first line must end in a period.

You can shorten this sentence to, e.g., 

```
"""Test Keras model with eager, 8 GPUs and automatic mixed precision."""
```

official/resnet/ctl/ctl_imagenet_benchmark.py
```diff
@@ -139,6 +139,24 @@ def benchmark_8_gpu(self):
...
+    self._setup()
+    FLAGS.num_gpus = 8
+    FLAGS.data_dir = self.data_dir
+    FLAGS.batch_size = 128 * 8
+    FLAGS.train_epochs = 90
+    FLAGS.epochs_between_evals = 10
+    FLAGS.model_dir = self._get_model_dir('benchmark_8_gpu_amp')
+    FLAGS.dtype = 'fp16'
+    FLAGS.fp16_implementation = 'graph_rewrite'
+    FLAGS.use_tf_function = True
```
reedwm: This defaults to True, so no need to specify it. I don't like specifying it because then it's not clear this is exactly the same as `benchmark_8_gpu` except with AMP.

And same for the other benchmarks added.

official/resnet/ctl/ctl_imagenet_main.py
houtoms: The `keras_common` is not defined in this file. We will encounter `undefined error` if it is not removed.

houtoms:
@reedwm Thx for your review. I made some more changes. Please check.

reedwm:
Please fix lint error

houtoms:
I don't think I can see the "Details" of Lint Python Files. I got `Use your SSO username and password` when I click the details.

reedwm:
@saberkun, do you know why external users cannot access the lint page? You can reproduce by opening the link in an incognito window.

@houtoms, there is only one lint errro:

```
FAIL: Found 1 non-whitelited pylint errors:
official/resnet/ctl/ctl_imagenet_main.py:180: [C0330(bad-continuation), ] Wrong hanging indentation (remove 12 spaces).
```

saberkun:
I don't know the lint setting.

Could you run pylint locally? It also happens to the PR I merged today

houtoms:
Thanks. I fixed the bad indent issue. Please check.

official/resnet/ctl/ctl_imagenet_benchmark.py
```diff
@@ -139,6 +139,21 @@ def benchmark_8_gpu(self):
     FLAGS.datasets_num_private_threads = 14
     self._run_and_report_benchmark()
 
+  def benchmark_8_gpu_amp(self):
+    """Test Keras model with eager, 8 GPUs with automatic mixed precision."""
+    self._setup()
+    FLAGS.num_gpus = 8
+    FLAGS.data_dir = self.data_dir
+    FLAGS.batch_size = 128 * 8
```
anj-s: why not do 256 per replica?

houtoms: Thx. I will fix it.

PR #7558, opened 2019-09-15
fix the bilinear interpolation method of STN

TonyChen1995:
Hey, Google's engineers. I studied the STN (spatial transform network) for some time, and I found some following errors in your bilinear interpolation method of the STN code.
The test code is as follows:

import tensorflow as tf
from spatial_transformer import transformer
import numpy as np
U = tf.reshape(tf.range(9), [1, 3, 3, 1])
theta =  np.array([[1, 0, 0], [0, 1, 0]])
out_size = (3, 3)
y = transformer(U, theta, out_size)
sess = tf.Session()
print(sess.run(y))

It's obvious that the output y should be identical to the input U because it's a identity transform with the theta value above. However, the output y is :
[[[[0. ]
   [1.5]
   [0. ]]

  [[4.5]
   [6. ]
   [0. ]]

  [[0. ]
   [0. ]
   [0. ]]]]
Therefore there are some mistakes. I have fixed it, and please check it out.

### DIFF
--- a/research/transformer/spatial_transformer.py
+++ b/research/transformer/spatial_transformer.py
@@ -78,9 +78,9 @@ def transformer(U, theta, out_size, name='SpatialTransformer', **kwargs):
             max_y = tf.cast(tf.shape(im)[1] - 1, 'int32')
             max_x = tf.cast(tf.shape(im)[2] - 1, 'int32')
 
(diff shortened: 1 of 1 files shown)


TonyChen1995:
@googlebot I signed it!

PR #7573, opened 2019-09-17
Double batch size for fp16 benchmarks

houtoms:
For the auto mixed precision benchmarks in ResNet50 CTL, this PR doubles the batch sizes for better performance.

@reedwm


tfboyd:
We were getting out of memory issues with 16GB of memory for CTL.  If that is still the case we cannot double the batch-size. Please confirm on 16GB cards.  I have a feeling you are using 32GB cards.

If it works with AMP but not AMP that is ok just double the AMP test and that indicates and advantage and a baseline.  Again has to work with 16G cards.    Thanks.

saberkun:
Thanks. Will help with merge.

PR #7586, opened 2019-09-20
Fix path to movielens helper file

wileeam:
Just fixing the path to the helper file to download and process the Movielens dataset in the script `run.sh` after a recent move of this file within the repository


wileeam:
@googlebot I signed it

PR #7587, opened 2019-09-20
Fix bug in ptb_word_lm.py

gyojir:
If ptb_word_lm.py is executed with --num_gpus=0, initial_state and final_state are not imported. As a result, feed_dict is always empty.


jaeyounkim:
Deprecated

PR #7605, opened 2019-09-25
Remove unmaintained fork of Minigo code

brilee:
The reference implementation can be found at
https://github.com/tensorflow/minigo

This fork was originally created to experiment with performance upgrades
for MLPerf, but since MLPerf work is focused in the original repo,
this fork's existence only serves to confuse.


saberkun:
Sounds good. As it is a research model, we can just merge it.
I will merge this PR. Thanks

PR #7609, opened 2019-09-28
Added grayscale support to Slim preprocessing

petewarden:
This is needed for a person detection model running on an MCU, which only has a grayscale camera.


petewarden:
Closing, since this has been merged internally and pushed separately.

PR #7614, opened 2019-09-29
Trigger Kokoro testing, do not review.

saberkun:


saberkun:
Tests passed. tensorflow-hub installed in perfzero. LG.

PR #7617, opened 2019-10-01
update and fix the matplotlib backend as TkAgg

ujjaldas132:
TkAgg should be running in backend of Matplotlib to show the output image. if "Agg" is running in backend it wont able to do so. so it is better to fix the backend of matplotlib as TkAgg.


pkulzc:
Could you please revert the irrelevant changes in the colab?

PR #7630, opened 2019-10-07
Remove "-of" from the input path name pattern

monocongo:
Remove "-of" from the regex pattern for the input path at line 207 since the output files from the code in lines 178-192 doesn't actually produce files with "of-" in their names (as listed in lines 196-201).

### DIFF
--- a/research/object_detection/g3doc/using_your_own_dataset.md
+++ b/research/object_detection/g3doc/using_your_own_dataset.md
@@ -204,6 +204,6 @@ which can then be used in the config file as below.
 
 ```bash
 tf_record_input_reader {
-  input_path: "/path/to/train_dataset.record-?????-of-00010"
+  input_path: "/path/to/train_dataset.record-?????-00010"
 }
 ```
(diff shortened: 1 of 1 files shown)


research/object_detection/g3doc/using_your_own_dataset.md
```diff
@@ -204,6 +204,6 @@ which can then be used in the config file as below.
 
 ```bash
 tf_record_input_reader {
-  input_path: "/path/to/train_dataset.record-?????-of-00010"
```
pkulzc: I think this is actually needed. Instead you should update the file list above.

monocongo: You are correct, thanks for your sharp attention and catching my mistake.

monocongo:
I have updated the file accordingly as requested by @pkulzc, the documentation should now be accurate.

jaeyounkim:
@pkulzc Please review this pull request again.

PR #7655, opened 2019-10-11
Revert "Revert "Update usage of tf.contrib.data to tf.data.experimental""

saberkun:
Reverts tensorflow/models#7654


saberkun:
Roll forward.

PR #7678, opened 2019-10-17
Release MobileNet V3 models and SSDLite models with MobileNet V3 backbone.

pkulzc:


sguada:
Please fix Lint Python errors.

PR #7700, opened 2019-10-21
Fixes the metrics_set name

bgalvao:
So as to match that of eval_util.py, i.e.

https://github.com/bgalvao/models/blob/6766e6ddfabefbd1003d5275fa5cc32f18196c11/research/object_detection/eval_util.py#L65


bgalvao:
@googlebot I signed it!

PR #7704, opened 2019-10-22
Update Links to Google Colab and Github

raj5287:
Link has been interchanged, in place of Run in Google Colab , View source on GitHub link was given and vice-versa


raj5287:
@googlebot I signed it!

PR #7706, opened 2019-10-22
Fix BERT EmbeddingPostprocessor

Vooblin:
Change one_hot+matmul on gather. It's important for conversion to tflite because one_hot+matmul seriously change model predictions after conversion in tflite format. Gather does the same operations and doesn't have the same effect.

It is related to problem in tflite. Issue: https://github.com/tensorflow/tensorflow/issues/33613.
My fix allow to convert model to tflite before this problem will be solved


saberkun:
+ Renjie from TFLite team to understand the issue.
Thanks for the PR. I think the PR itself is good and I think it is trivial to GPU/TPU.
But I heard tflite is not happy with tf.gather?

renjie-liu:
Gather is supported in tflite

saberkun:
I see. That's great.
Let me test against internal deps.

PR #7712, opened 2019-10-24
Tensorboard summary writing for few params

madhurapande19:
Using this code on GPU, was not logging "learning_rate" and "gradient norm" etc, basically items in metric_dict. We need to use simple "tf.summary.scalar" instead of contrib one to log these params when "use_tpu" is False. Function "record_scalars" gets called on in case when "use_tpu" is False. So this change helped tensors in metric dict properly show up in Tensorboard summary. Tested on Tensorflow 1.14.


madhurapande19:
@googlebot I signed it!

PR #7713, opened 2019-10-24
Use the new website for the Open Images Dataset

rosskyl:
Change to use the new website where the Open Images dataset is hosted


rosskyl:
@googlebot I signed it!

jaeyounkim:
Fixed

PR #7724, opened 2019-10-27
Replace tf.gfile.GFile with tf.io.gfile.GFile 

r-kellerm:
to avoid depcaration warning


r-kellerm:
@googlebot I fixed it.

MarkDaoust:
Hi @r-kellerm,

Thanks for trying to make these upgrades. But I do not think this is the right approach.

Specifically, [upgrade-script guide](https://www.tensorflow.org/guide/upgrade) says:

*  [Unit Test: Ensure that the code you’re upgrading has a unit test suite with reasonable coverage [before trying to upgrade].](https://www.tensorflow.org/guide/upgrade#recommended_upgrade_process) 
* [Do not update parts of your code manually before running this script.](https://www.tensorflow.org/guide/upgrade#caveats)

I think this process needs to be undertaken on a project by project basis, by the project owners.

PR #7739, opened 2019-11-01
Update input_pipeline.py

kamalkraj:
API changed 
https://www.tensorflow.org/api_docs/python/tf/data/experimental/DistributeOptions


saberkun:
This is the new api introduced in tf-nightly. Please note that the head of this repo always follow the latest TF changes.
If you want to use a version fit your TF version, please consider release branches like:
https://github.com/tensorflow/models/tree/v2.0
Thanks!

PR #7771, opened 2019-11-07
Make BERT model more customizable

grwlf:
Hi. Let me suggest 2 patches that would make BERT model more customizable:
1. Allows users to pass custom `get_bert_model`.
2. Builds BERT's components as early as possible.

The particular problem I'm trying to solve is following: Currently Keras calls `build`s  from `__call__` which doesn't left any space for model modifications. This PR (the 2nd patch) adds `build()` calls to constructors whenever possible to offer user an opportunity to change some of the model's components before hard-wiring things with `__call__`s. This approach should reduce time/efforts required to prepare model experiments. Please check an example below.

Note that some time ago I started a related topic in TFdev mailing list:  https://groups.google.com/a/tensorflow.org/forum/#!topic/developers/B57l_ygjBPU

```py
from mytweaks import CustomTransformerBlock

def get_custom_bert_model(input_word_ids, input_mask, input_type_ids, config=None, name=None, float_type=tf.float32):
  bert_model_layer = BertModel(config=config, float_type=float_type, name=name)
  
  # Added part: update some (in this case all) layers of encoder. Wouldn't be possible wothout patch
  new_layers = []  
  for i,tb in enumerate(bert_model_layer.encoder.layers):
    self=tb
    new_layers.append(CustomTransformerBlock(
          hidden_size=self.hidden_size,
          num_attention_heads=self.num_attention_heads,
          intermediate_size=self.intermediate_size,
          intermediate_activation=self.intermediate_activation,
          hidden_dropout_prob=self.hidden_dropout_prob,
          attention_probs_dropout_prob=self.attention_probs_dropout_prob,
          initializer_range=self.initializer_range,
          backward_compatible=self.backward_compatible,
          float_type=self.float_type,
          name=("custom_layer_%d" % i)))
  bert_model_layer.encoder.layers = new_layers

  pooled_output, sequence_output = bert_model_layer(input_word_ids, input_mask,
                                                    input_type_ids)
  bert_model = tf.keras.Model(
      inputs=[input_word_ids, input_mask, input_type_ids],
      outputs=[pooled_output, sequence_output])
  return bert_model

def run():
  ...
  model,_ = classifier_model(bert_config, tf.float32, num_labels, max_seq_length,
                                              custom_bert_model=get_custom_bert_model)
  model.compile(...)
  model.fit(...)

```

### DIFF
--- a/official/nlp/bert_modeling.py
+++ b/official/nlp/bert_modeling.py
@@ -151,6 +151,7 @@ class BertModel(tf.keras.layers.Layer):
         BertConfig.from_dict(config)
         if isinstance(config, dict) else copy.deepcopy(config))
     self.float_type = float_type
+    self.build(unused_input_shapes=[])
 
   def build(self, unused_input_shapes):
     """Implements build() for the layer."""
@@ -250,6 +251,7 @@ class EmbeddingLookup(tf.keras.layers.Layer):
     self.vocab_size = vocab_size
     self.embedding_size = embedding_size
     self.initializer_range = initializer_range
+    self.build(unused_input_shapes=[])
 
   def build(self, unused_input_shapes):
     """Implements build() for the layer."""
@@ -407,6 +409,7 @@ class Attention(tf.keras.layers.Layer):
     self.attention_probs_dropout_prob = attention_probs_dropout_prob
     self.initializer_range = initializer_range
     self.backward_compatible = backward_compatible
+    self.build(unused_input_shapes=[])
 
   def build(self, unused_input_shapes):
     """Implements build() for the layer."""
@@ -728,6 +731,7 @@ class TransformerBlock(tf.keras.layers.Layer):
           "The hidden size (%d) is not a multiple of the number of attention "
           "heads (%d)" % (self.hidden_size, self.num_attention_heads))
     self.attention_head_size = int(self.hidden_size / self.num_attention_heads)
+    self.build(unused_input_shapes=[])
 
   def build(self, unused_input_shapes):
     """Implements build() for the layer."""
@@ -844,6 +848,7 @@ class Transformer(tf.keras.layers.Layer):
     self.initializer_range = initializer_range
     self.backward_compatible = backward_compatible
     self.float_type = float_type
+    self.build(unused_input_shapes=[])
 
   def build(self, unused_input_shapes):
     """Implements build() for the layer."""
--- a/official/nlp/bert_models.py
+++ b/official/nlp/bert_models.py
@@ -376,6 +376,7 @@ def squad_model(bert_config, max_seq_length, float_type, initializer=None):
       name='squad_model')
   return squad, core_model
 
+
 def classifier_model(bert_config,
                      float_type,
                      num_labels,
--- a/official/nlp/bert_models.py
+++ b/official/nlp/bert_models.py
@@ -376,13 +376,13 @@ def squad_model(bert_config, max_seq_length, float_type, initializer=None):
       name='squad_model')
   return squad, core_model
 
-
 def classifier_model(bert_config,
                      float_type,
                      num_labels,
                      max_seq_length,
                      final_layer_initializer=None,
-                     hub_module_url=None):
+                     hub_module_url=None,
+                     custom_bert_model=None):
   """BERT classifier model in functional API style.
 
   Construct a Keras model for predicting `num_labels` outputs from an input with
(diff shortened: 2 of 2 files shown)


saberkun:
grwlf@ 
Thanks for the suggestion.
We are releasing a completely new BERT libraries.
Could you hold a little bit? It will probably address your concern and we can add new components on that.
Sorry about this transition.

grwlf:
@saberkun Sure, thanks for letting now. Do you plan to publish a pre-release?

saberkun:
@grwlf I will release the library in one or two days. I will not be able to update all examples to use it but will update classifier example with new checkpoints. Please feel free to create issue to comment on that. Sorry about the commit workflow as it will directly go from internal code base without PR.

linbojin:
> @grwlf I will release the library in one or two days. I will not be able to update all examples to use it but will update classifier example with new checkpoints. Please feel free to create issue to comment on that. Sorry about the commit workflow as it will directly go from internal code base without PR.

any updates?

saberkun:
Hi Guys,
We have update our modeling library as commit: https://github.com/tensorflow/models/commit/f1d35b4ee525e4a2ef434414268eb050de81da7b
See: https://github.com/tensorflow/models/tree/master/official/nlp/modeling
Because we are moving to the new implementation, may you take a look and see if the functionality you suggests can be achieved with minimal code change?
We will finish the migration in a week or so.

Thanks!

grwlf:
> We have update our modeling library as commit: [f1d35b4](https://github.com/tensorflow/models/commit/f1d35b4ee525e4a2ef434414268eb050de81da7b)

FYI I am looking into it. Latest master doesn't work with tf2.0 anymore, so I assume you are on tf2.1 now.

saberkun:
Yes, you are right.
We are in the progress to move to the new code base and we have validated it on both pretraining/finetuning. 
It is also possible to get a "transformer scaffolding" to build customized transformer through a functional keras api. We probably will get it soon.

grwlf:
Ok, I was able to get my TF2.1 working with minor workaround. I reviewed the some parts of new design, so I agree that transition to functional API will change rules dramatically. Unfortunately, I don't see a short solution for augmenting the model anymore. 

For the reference, lets say I want to replace 2 Denses of Berts's FFN with a single Dense (I also imagine that I want to do it in only certain layers, because I know there is a python magic for global class overrides which I don't want to use). The code I came up with looked like the following originally: 
https://gist.github.com/grwlf/56ca9d220ad0b0fc8d8a10a5d437d400 .

With the new design, it becomes almost 2 times bigger (please be kind to let me know if I am missing a better approach. How can I learn more about transformer scaffolding? )
https://gist.github.com/grwlf/bba6080d950919c0a6c3bdfaf25176fe

A kind of good news - this PR is no longer important since it doesn't allow users to re-use top-level layers like `TransformerEncoder` anymore. I'll probably close it soon.

saberkun:
Hi,
I also know "replace n Denses of Berts's FFN with a single Dense" is useful. I am happy to check in this particular change. If you would like to create a PR, it would be nice.
@markomernick Hi Mark, here is a customization use case.

markomernick:
Hey all,

This is exactly what the EncoderScaffold class (coming in the next few weeks) should solve. (It's close to ready internally, but with next week being a US holiday week the external availability might take a bit longer). This allows you to pass a new hidden class that is used instead of layers.Transformer in the encoder, plus a config. For cases where you'd prefer a more complex network, like the above example where some layers are replaced and some are not, you can pass a single functional Keras model in.

grwlf:
Thanks all. Let me note that changing FFN is only one of possible customizations. One also may want to replace components with their fake-quantized versions, introduce "know-how" attentions, or just prune some blocks away. One way to do it is via whole model copypasting, but nlp models tends to be complex, so I'd like to come with a more efficient approach to keep the sources smaller.

A natural direction here it is to give users an opportunity to do certain Python assignments after the Keras model is ready but before the TF graph is created. Unfortunately, the `input_shape` math together with 'functional' trend in Keras tend to leave no gap between those two stages. That is why I'm still opposed to 'be functional' here to my own surprise. Maybe I have to continue this discussion in the TF mailing list.

So I'm looking forward at learning more about *Scaffold classes.

saberkun:
I think the direction will be: some bert/transformer layers are going to keep as it is to reproduce results. Scaffold classes will be checked in and we can make each component to be overridden.
In functional style, the layer is built when it is called. It is still possible to override attributes directly before it is called and after then instance is created.

grwlf:
Out of date now

PR #7777, opened 2019-11-10
EEG analysis

abhishek-parashar:


abhishek-parashar:
@googlebot I signed it!

abhishek-parashar:
@googlebot googlebot I signed it!

abhishek-parashar:
@saberkun please review

MarkDaoust:
Hi @abhishek-parashar,

I think the research/ directory is really meant for internal-google-researchers to use for sharing code associated with published research papers.

So I don't think this is the right place for this.

There is a tensorflow/examples repository that you could submit this to, that would be more appropriate.

PR #7780, opened 2019-11-11
Update using_your_own_dataset.md

KhanhCon:
encoded_image_data does not match encoded_cat_image_data


KhanhCon:
@googlebot I signed it!

PR #7786, opened 2019-11-12
Update for py3 and some internal changes

YknZhu:


lyc1995452-star:
This error in cityscapes train.
data_generator.py", line 170, in __init__
    raise ValueError('data split name %s not recognized' % split_name)
ValueError: data split name train not recognized
How can I solved?

PR #7794, opened 2019-11-12
Update README.md

snakch:
changed the link for kitti_archives_to_download.txt to point to the raw file rather than just the github URL.


jaeyounkim:
CLA not signed

PR #7804, opened 2019-11-14
 Fix segmentation fault on SSD tflite client destruction.

qwook:
edgetpu::EdgeTpuContext needs to be created before tflite::Interpreter.


yongzhe2160:
Are the earlier commits irrelevant?

PR #7814, opened 2019-11-15
Fixed OID tfrecord shard calculation

yaysummeriscoming:
Current shard calculation produces tfrecord files with greatly differing filesizes - 100 shard test produced files ranging form 45% to 0% of the total dataset.

To fix the problem, I copied over the shard calculation logic from create_coco_tf_record.py


yaysummeriscoming:
@derekjchow @dreamdragon could you review please?

PR #7850, opened 2019-11-21
Adding research/audioset/yamnet, a pre-trained audio event classifier

dpwe:
We are publishing a pertained audio event model based on the AudioSet data.
@plakal : I'd be grateful for an independent test before a merge.


research/audioset/yamnet/yamnet_test.py
```diff
@@ -0,0 +1,70 @@
...
+                                   (1, int(3 * params.SAMPLE_RATE))),
+        class_name='White noise')
+
+  def testSine(self):
+    self.clip_test(
+        waveform=np.reshape(
+            np.sin(2 * np.pi * 440 * np.linspace(
+                0, 3, int(3 *params.SAMPLE_RATE))),
+            [1, -1]),
+        class_name='Telephone')  # 'Sine wave' is nearly top.
```
plakal: Perhaps if we check for presence in the top N classes (N = 5 or 10), we could get away with looking for Sine wave.

PR #7863, opened 2019-11-25
add autoaugment

fsx950223:
Add autoaugment compatible with tensorflow2.


fsx950223:
@saberkun Could you review the PR?

saberkun:
Hi, we have a version sharing across all vision tasks under development. Also, we need to add tests for this feature, especially because of tf addons. Could we hold this?
@yeqingli

PR #7880, opened 2019-11-28
Replaced deprecated functions with suggestions in ImageNet example.

OMGtechy:
I wanted to have a look at an ImageNet example and came across this. However, it was full of deprecation warnings, so I've dropped in the suggested functions in their stead. There's still one warning left, however:

```
2019-11-29 00:50:28.988072: W tensorflow/core/framework/op_def_util.cc:370] Op BatchNormWithGlobalNormalization is deprecated. It will cease to work in GraphDef 
version 9. Use tf.nn.batch_normalization().
```

I wasn't as sure as to how to respond to this one, as there's no mention of BatchNormWithGlobalNormalization anywhere in the example, so I've left that one.

Regardless, this set of changes I've made is at least an improvement. Now both 2.0 and 1.15 users can use this example; beforehand only 1.15 users could.

I have signed the contributors agreement.

### DIFF
--- a/tutorials/image/imagenet/classify_image.py
+++ b/tutorials/image/imagenet/classify_image.py
@@ -76,13 +76,13 @@ class NodeLookup(object):
     Returns:
       dict from integer node ID to human-readable string.
     """
-    if not tf.gfile.Exists(uid_lookup_path):
+    if not tf.io.gfile.exists(uid_lookup_path):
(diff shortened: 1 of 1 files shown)


jaeyounkim:
I am closing this PR since the tutorials directory is deprecated.

PR #7881, opened 2019-11-29
Update vid2depth/README.md: use raw.githubcontent

ReGuess:
The wget command for the KITTI dataset in the currently downloads the html for GitHub's fancy display of the text file. This change fixes that so it instead downloads the raw txt file (no more trying to resolve host '<')


ReGuess:
@googlebot I signed it!

ReGuess:
@rezama Contributing guide said I'm supposed to contact you. I suppose an at mention suffices?

PR #7884, opened 2019-11-29
Fixed tf.compat.v1 bug

miraliahmadli:
Before that, it called __tensorflow.compat.v1.compat.v1.io.tf_record_iterator__ function, which does not exist, instead of __tensorflow.compat.v1.io.tf_record_iterator__.


miraliahmadli:
@googlebot I signed it!

saberkun:
Thanks for catching this!

PR #7898, opened 2019-12-03
Add generic object detection Dockerfile

010akv:
This commit adds a generic Dockerfile for object detection with upgraded TF version that may be helpful for non-android use cases. Further, this also reduces the image size from 8G to 4G to support various edge applications. The functionality otherwise remains same as that of android Dockerfile.


010akv:
Any interest in this change?

tqa236:
Thank you for the effort. I would love to see this PR being merged. A generic Dockerfile will speed up the setup time a lot.

tqa236:
I'm not a reviewer but here are my 2 cents: 

- Tensorflow 2.0 should be a better choice
- This Dockerfile can make use of https://www.fromlatest.io/#/  for optimization.

PR #7927, opened 2019-12-09
Transfer Learning using Keras Functional API in TensorFlow 2.0

AyeshaShafique:
Hi @all, 

This is my first ever contribution to TensorFlow repo. I have created a notebook having a way to do transfer learning using Function API in **TensorFlow 2.0** and with the TensorBorad plugin in the notebook. It will really help people who just started in TensorFlow to how to build a neural network using Transfer Learning with the latest TensorFlow Version 2.0 using Python.

Kindly merge this pull request to the actual repo if you see it can give value to the local community. 

Thanks.

### DIFF
--- a/tutorials/notebooks/transfer_learning_using_keras_functional_api_in_tensorflow_2.0.ipynb
+++ b/tutorials/notebooks/transfer_learning_using_keras_functional_api_in_tensorflow_2.0.ipynb
@@ -1,0 +1,560 @@
+{
+ "cells": [
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "# Transfer Learning using Keras Functional API in TensorFlow 2.0"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "Create the model architecture using Transfer Learning method in TensorFlow 2.0 Keras API.\n",
+    "\n",
+    "Note: We have extracted features from pretrained model mobile-net and then added custom classification layers."
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": 3,
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "from tensorflow.keras.models import Model\n",
+    "from tensorflow.keras.layers import Activation\n",
+    "from tensorflow.keras.layers import Flatten\n",
+    "from tensorflow.keras.layers import Dense\n",
+    "from tensorflow.keras.applications import MobileNet\n",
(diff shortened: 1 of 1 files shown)


AyeshaShafique:
@googlebot  I fixed it

AyeshaShafique:
Hi, @bmabey , @alextp , @moonboots  
Kindly help me. I want to merge this request. And in future i want more and more active collaboration for the TensorFlow community. 

Thanks a lot.

alextp:
@wolffg @lamberta  we should put this in the docs repo somewhere, I think.

Can you recommend where? Then we can review for content.

AyeshaShafique:
Hi Alexander,

You can put this in a tutorial section inside the repo for notebooks so
that people can learn that how to code for transfer learning using
functional api in TensorFlow 2.0

Thanks a lot.

robieta:
Drive by comment:
I agree that having a transfer learning guide would be very helpful; this is something that keras makes very easy. However I think it would benefit a lot from a simpler model and a lot deeper dive on some of the key aspects. (Shared variables across models, setting the `trainable` attribute on both Variables and Layers and what each does, how to handle moving statistics when switching datasets, etc.)

Also +1 that such a guide should live somewhere in the docs / tutorials, as it will need to be placed in the proper context to be maximally useful.

AyeshaShafique:
Great thoughts.

But I think to start with, we can go with this. Then we should improve it
more and more. I am also working on the other sides as well for transfer
learning and other stuff.

Thanks

lamberta:
Staging link: https://colab.research.google.com/github/AyeshaShafique/models/blob/master/tutorials/notebooks/transfer_learning_using_keras_functional_api_in_tensorflow_2.0.ipynb

Hi AyeshaShafique@, I believe this repo is for reference models that are *used in* tutorial docs (and not the docs themselves). So pull requests for notebooks tutorials should go in in [tensorflow/docs](https://github.com/tensorflow/docs) repo. But there's not much explanation/narrative text in this doc, so it would likely need another revision before we can bring it into the official documentation.

Alternatively, we have a number of existing transfer learning tutorials, maybe you can update one of those with useful content from this one?

* https://www.tensorflow.org/tutorials/keras/text_classification_with_hub
* https://www.tensorflow.org/tutorials/images/transfer_learning_with_hub
* https://www.tensorflow.org/tutorials/images/transfer_learning

Closing this pull request since tensorflow/models doesn't seem like the right place, but feel free to work on this doc (or others) and submit the PR to tensorflow/docs. Thanks

PR #7939, opened 2019-12-11
Swapped the github/colab link urls on the object_detection notebook

xdhmoore:
Swapped the github/colab link urls. They were reversed.


xdhmoore:
Sorry, I am not going to spend the time on a CLA for something like this.

PR #7941, opened 2019-12-12
Added terminal parameter to include masks

Ahmad-Omar-Ahsan:
This commit adds a terminal parameter so that users can decide whether they want to include masks or not. In the previous version, there was no optional argument for including masks if they want to execute the script from the terminal and the default value for include_masks was false. With this commit users can include masks in their tf records


Ahmad-Omar-Ahsan:
@googlebot I signed it!

PR #7946, opened 2019-12-14
Fix help flag error.

zachary-russell:
The flag needed is not `-h` or `--help` it's `--helpfull`.


zachary-russell:
@saberkun I'm not _super_ familiar with the flags package, but I think I added the link as you were asking.

PR #8015, opened 2020-01-08
Subtract mean list

bill2239:
fix means to take list [mean_r, mean_g, mean_b] e.g. [ 109.8 , 109.8 , 109.8 ] in subtract_channel_mean in config file for objection detection module


MuhammadAliShahzad:
@bill2239 Why did you revert your fix?

PR #8031, opened 2020-01-12
Add ALBERT model 

boronhub:
From https://github.com/google-research/ALBERT


saberkun:
Hi,

Albert models are already added: https://github.com/tensorflow/models/blob/master/official/nlp/bert/run_classifier.py#L293

Chen, could you please update readme ASAP?
@chenGitHuber

boronhub:
@saberkun are there any other NLP models that need to be added ?

jaeyounkim:
@boronhub We will let you know if we need other models to be added.

PR #8063, opened 2020-01-19
Fix bug of class DenseEinsum

ZhuBaohe:
This pr fixs bug of class DenseEinsum.

### DIFF
--- a/official/nlp/modeling/layers/dense_einsum.py
+++ b/official/nlp/modeling/layers/dense_einsum.py
@@ -148,7 +148,7 @@ class DenseEinsum(tf.keras.layers.Layer):
     input_shape = input_shape.with_rank_at_least(self._num_summed_dimensions +
                                                  1)
     for i in range(self._num_summed_dimensions):
-      if tf.dimension_value(input_shape[-1 * i]) is None:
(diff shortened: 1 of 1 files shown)


saberkun:
Thanks!
Do you know how to trigger the bug and what's the effect?

ZhuBaohe:
> Thanks!
> Do you know how to trigger the bug and what's the effect?

@saberkun 

This is a test code.
 
```
import numpy as np
from official.nlp.modeling.layers import dense_einsum

test_layer = dense_einsum.DenseEinsum(output_shape=(64,), num_summed_dimensions=2)
input_tensor = tf.constant(np.random.random_sample((10, 20, 40, 80)))
print(test_layer.compute_output_shape(input_tensor.shape))
```
Before fixed, it shows error
`AttributeError: 'DenseEinsum' object has no attribute '_units'`

After fixed, it shows the right result
`(10, 20, 64)`

saberkun:
@ZhuBaohe Thanks!
Discussing with Keras team. It seems that tf.keras does not need to implement compute_output_shape.
Will get back with conclusion.

saberkun:
Confirmed with keras team. We don't need compute_output_shape.
compute_output_shape is only used when layer.dynamic = True

PR #8090, opened 2020-01-27
Address deprecation warning in tf.while_loop

georgesterpu:
I am updating the code according to the recommendation in the deprecation message.

```python
@tf_export("while_loop", v1=[])
@deprecation.deprecated_arg_values(
    None,
    """back_prop=False is deprecated. Consider using tf.stop_gradient instead.
Instead of:
results = tf.while_loop(c, b, vars, back_prop=False)
Use:
results = tf.nest.map_structure(tf.stop_gradient, tf.while_loop(c, b, vars))""",
```


rachellj218:
@georgesterpu Thank you for the fix. Please note that I didn't update beam_search_v1.py since this is used in transformer in tf1.x, which should still use the original api.

PR #8094, opened 2020-01-29
Fixes binary write mode for Python3 and Windows

WYLDYR:
Binary write mode W might cause problems for Python 3 and Windows.


WYLDYR:
@googlebot I signed it!

PR #8110, opened 2020-02-04
added coordinate feature for each object.

parikshit223933:
With the changes made in the code, the user will now be able to:
1. Get the center-coordinates of the box, corresponding to each object in the image, which will make locating objects in the image easier fro the user.
2. User won't need to write bunch of lines for getting an array of labelled objects found in the image
3. Sequencing of the text will become easier. for example, if the user has created a model on text detection, it is often difficult for him to print the objects in a sequence. It will become much easier now.
4.  training a batch of images in a single tensorflow session, and printing the text for each image in a particular sequence will become easier.

EXAMPLE:
image
As you can see in the above image, if the first six characters have to be printed before the last four characters, then you will need to know the coordinates of each character of the image. otherwise, the characters will be printed in a shuffled order.
(In the right hand side in the command prompt, there is the string detected in a correct sequence)

### DIFF
--- a/research/object_detection/utils/visualization_utils.py
+++ b/research/object_detection/utils/visualization_utils.py
@@ -41,6 +41,14 @@ import tensorflow as tf
 from object_detection.core import standard_fields as fields
 from object_detection.utils import shape_utils
 
+label_array=[]
+# this label_array will store all the labels of detected items and percentage accuracies corresponding to them.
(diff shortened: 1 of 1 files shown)


parikshit223933:
@googlebot I signed it!

parikshit223933:
Hello sir, How long does it take to review a PR?

PR #8126, opened 2020-02-10
Updated the gast dependency

MuhammadAliShahzad:
It breaks the training on batch normalization while training object detectors.


MuhammadAliShahzad:
@googlebot I signed it!

PR #8134, opened 2020-02-13
Update research/attention_ocr model to be Python3 compatible

alexgorban:
This change contains multiple minor changes to make the Attention OCR model compatible with Python 3 as well decreases size of fsns-00000-of-00001 file from 7.6mb to 632kb.


saberkun:
Thank you!

PR #8170, opened 2020-02-21
Delete models:syntaxnet

saberkun:
delete syntax net to reduce repo size as syntax net is going to move to google-research/


candaj:
Hello,

Just saw that syntaxnet model has been remove from TF repo. But I can find out where it has been put into google-research orga or repo. Do you have a link please ? 

Regards,

jaeyounkim:
It is not yet available at the google-research GitHub repository.
You can still find it at https://github.com/tensorflow/models/tree/f2f25096d3dc6561a855dab914cf2913100728d6/research/syntaxnet

PR #8174, opened 2020-02-21
YAMNet variable sr

dpwe:
Update example notebook for YAMNet to work with wav files with any sampling rate.


dpwe:
That's right, I messed up the branching.  But if I squash and merge, does it matter?

I think my only alternative is to start over again, which is messy in itself, though I guess cleans up the timeline.

PR #8176, opened 2020-02-21
Make yamnet_visualization notebook work with any input sample rate

dpwe:
(Retry of pull request).


plakal:
I think this is redundant since I merged https://github.com/tensorflow/models/pull/8174 yesterday.

dpwe:
Redundant with #8174

Issue #8024, opened 2020-01-10
Rename Tensorflow detection model zoo to TensorFlow detection model collection


SaschaHeyer:
https://github.com/tensorflow/models/blob/master/research/object_detection/g3doc/detection_model_zoo.md

![zoo](https://user-images.githubusercontent.com/1991664/72136391-ff4c9800-3388-11ea-8165-554bf92006ab.jpg)

tensorflowbutler:
Thank you for your post. We noticed you have not filled out the following field in the issue template. Could you update them if they are relevant in your case, or leave them as N/A? Thanks.
What is the top-level directory of the model you are using
Have I written custom code
OS Platform and Distribution
TensorFlow installed from
TensorFlow version
Bazel version
CUDA/cuDNN version
GPU model and memory
Exact command to reproduce

ayushmankumar7:
Assign this to me

jaeyounkim:
We do not have a plan to rename it yet.

PR #8177, opened 2020-02-23
#8024 Update detection_model_zoo.md

ayushmankumar7:
Issue #8024 Done.


jaeyounkim:
We do not have a plan to rename it yet.

Issue #8186, opened 2020-02-24
ModuleNotFoundError: No module named 'mock'


ayushmankumar7:
**System information**
- Have I written custom code (as opposed to using a stock example script provided in TensorFlow):
- OS Platform and Distribution (e.g., Linux Ubuntu 16.04): Linux Ubuntu 18.04 LTS
- Mobile device (e.g., Pixel 4, Samsung Galaxy 10) if the issue happens on mobile device:
- TensorFlow installed from (source or binary): pip3
- TensorFlow version (use command below): 2.1.0
- Python version: 3.6
- Bazel version (if compiling from source):
- GCC/Compiler version (if compiling from source):
- CUDA/cuDNN version: 
- GPU model and memory:

**Please provide the entire URL of the model you are using?**

**Describe the current behavior**
![Screenshot from 2020-02-24 22-25-56](https://user-images.githubusercontent.com/41910134/75173550-10910e80-5755-11ea-9239-9e7f41efb333.png)

**Describe the expected behavior**
mock should be included in requirements.txt
**Code to reproduce the issue**

**Other info / logs**

saberkun:
you probably need to install mock.
See: https://github.com/tensorflow/benchmarks/blob/master/perfzero/docker/Dockerfile_ubuntu_1804_tf_v2_1#L85

saberkun:
I am fixing this using unittest.mock as users are using Python3 now

saberkun:
https://github.com/tensorflow/models/commit/4b8f80c3d31dffce153e4954b309cf52afc2c51a
Mark as fixed. 
Thanks

PR #8187, opened 2020-02-24
#8186 mock added to requirements.txt

ayushmankumar7:
Issue : #8187


saberkun:
I changed to unittest.mock. PY3 does not need to install mock

ayushmankumar7:
I was using Python 3 and i faced this problem. I had realier installed all the required librarires from the requirements.txt so, i thought that it should be present. Anyway, I respect your decision. Thanks.

PR #8221, opened 2020-03-01
Guard the unsafe tf.exp to prevent Inf cost

ForeverZyh:


research/lfads/lfads.py
```diff
@@ -792,7 +792,7 @@ def encode_data(dataset_bxtxd, enc_cell, name, forward_or_reverse,
         if hps.output_dist == 'poisson':
           log_rates_t = tf.matmul(factors[t], this_out_fac_W) + this_out_fac_b
           log_rates_t.set_shape([None, None])
-          rates[t] = dist_params[t] = tf.exp(log_rates_t) # rates feed back
+          rates[t] = dist_params[t] = tf.exp(tf.clip_by_value(log_rates_t, -hps._clip_value, hps._clip_value)) # rates feed back
```
sussillo, line 795: This logic is correct but will break the prototype example in run_lfads.py. Please add to line 580 in `run_lfads.py`
```d['_clip_value'] = xxx```, where xxx is the typical value used to protect against the exponential returning inf.

ForeverZyh: Thanks, added.

sussillo:
LGTM

PR #8225, opened 2020-03-02
Guard the unsafe tf.exp to prevent Inf cost

ForeverZyh:


jaeyounkim:
next_frame_prediction is deprecated.

PR #8238, opened 2020-03-04
Remove non existing samples folder mention

amahta:
Remove non existing samples folder mention


amahta:
@googlebot I signed it!

Issue #8246, opened 2020-03-05
ImportError: cannot import name 'profiler_v2'


ayushmankumar7:
**System information**
- Have I written custom code (as opposed to using a stock example script provided in TensorFlow):
- OS Platform and Distribution (e.g., Linux Ubuntu 16.04):
- Mobile device (e.g., Pixel 4, Samsung Galaxy 10) if the issue happens on mobile device:
- TensorFlow installed from (source or binary): pip3
- TensorFlow version (use command below): 2.1.0
- Python version: 3.6
- Bazel version (if compiling from source):
- GCC/Compiler version (if compiling from source):
- CUDA/cuDNN version:
- GPU model and memory:

**Please provide the entire URL of the model you are using?**

**Describe the current behavior**
![Screenshot from 2020-03-05 18-05-22](https://user-images.githubusercontent.com/41910134/75982396-9444bb00-5f0c-11ea-9f7d-6ba878e63723.png)

**Describe the expected behavior**
![Screenshot from 2020-03-05 18-07-15](https://user-images.githubusercontent.com/41910134/75982424-a45c9a80-5f0c-11ea-8315-ead552c513de.png)

**Code to reproduce the issue**

**Other info / logs**

saberkun:
profiler callback is removed. The profiler is integrated with TensorBoard callback. It is the right place to use it. Thanks. Closing this issue.

PR #8247, opened 2020-03-05
import name 'profiler_v2' solved

ayushmankumar7:
Issue: #8246

Current Results: 
Screenshot from 2020-03-05 18-07-15


saberkun:
Hi, the code was updated for tf-nightly/tf 2.2.
https://github.com/tensorflow/tensorflow/blob/master/tensorflow/python/profiler/profiler_v2.py
It is probably not the right way to directly import tensorflow internal modules.
Would you try tf.profiler.experimental.Profile? Also, the code should be good with TF nightly.

ayushmankumar7:
I'm currently using TF2.1.0. Maybe it's not the right way to do the thing. I just tried changing profiler_v2 to profiler and it solved the problem.

saberkun:
I see. It is good to keep this in your local change. TF 2.2 is coming and our code mostly is based on the latest TF under development. Sorry for inconvenience

ayushmankumar7:
Not a Problem. Excited for the new update.

PR #8252, opened 2020-03-06
Splitting the NCF embedding table when the accelerator is GPU.

adityaiitb:
Using a combined (bigger) embedding table followed by tensor slicing
results in an extra write and read operation, which can be completely
avoided if we maintain separate embedding tables. This results in a
significant performance improvement on the GPU.

As a sanity check, when I fixed the weights of the embedding to a
constant, and the weights were the same for all other layers before the
training, the value of the logits were the same after one step. After
the 1st step, the values will differ due to non-determinism in gradient
calculations.

For the following experiment, I used
  batch size = 99000
  synthetic data
  xla (on / off)

Without XLA
  Original: ~ 7.0 M examples/sec
  Modified: ~ 7.9 M examples/sec (12.8 % improvement)

With XLA:
  Original: ~ 9.2 M examples/sec
  Modified: ~ 9.5 M examples/sec (3.2 % improvement)

python ncf_keras_main.py \
    --model_dir /tmp/ \
    --data_dir data \
    --dataset ml-20m \
    --hooks --clean \
    --seed 3 \
    --train_epochs 3 \
    --batch_size 99000 \
    --eval_batch_size 99000 \
    --layers 256,256,128,64 \
    --num_factors 64 \
    --hr_threshold 0.635 \
    --ml_perf \
    --num_gpus 1 \
    --keras_use_ctl \
    --enable_xla \
    --use_synthetic_data

Note that since NCF is very sensitive, the number of epochs to convergence can change.

### DIFF
--- a/official/recommendation/neumf_model.py
+++ b/official/recommendation/neumf_model.py
@@ -190,37 +190,83 @@ def construct_model(user_input, item_input, params):
     x = tf.squeeze(x, [1])
     return x[:, mf_dim:]
 
-  # It turns out to be significantly more effecient to store the MF and MLP
-  # embedding portions in the same table, and then slice as needed.
-  embedding_user = tf.keras.layers.Embedding(
-      num_users,
-      mf_dim + model_layers[0] // 2,
-      embeddings_initializer=embedding_initializer,
-      embeddings_regularizer=tf.keras.regularizers.l2(mf_regularization),
-      input_length=1,
-      name="embedding_user")(
-          user_input)
+  if params["use_tpu"]:
+    # It turns out to be significantly more effecient to store the MF and MLP
+    # embedding portions in the same table, and then slice as needed.
+    embedding_user = tf.keras.layers.Embedding(
+        num_users,
+        mf_dim + model_layers[0] // 2,
+        embeddings_initializer=embedding_initializer,
+        embeddings_regularizer=tf.keras.regularizers.l2(mf_regularization),
+        input_length=1,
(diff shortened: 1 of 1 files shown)


official/recommendation/neumf_model.py
```diff
@@ -190,37 +190,83 @@ def mlp_slice_fn(x):
...
-      mf_slice_fn, name="embedding_user_mf")(embedding_user)
-  mf_item_latent = tf.keras.layers.Lambda(
-      mf_slice_fn, name="embedding_item_mf")(embedding_item)
-
-  # MLP part
-  mlp_user_latent = tf.keras.layers.Lambda(
-      mlp_slice_fn, name="embedding_user_mlp")(embedding_user)
-  mlp_item_latent = tf.keras.layers.Lambda(
-      mlp_slice_fn, name="embedding_item_mlp")(embedding_item)
+  if params["use_tpu"]:
```
hongjunChoi: let's not add this special handling.

adityaiitb: I don't think there is way out. The two architectures implement embedding differently. With one code path, one would lose performance either for the GPU or the TPU. As I mentioned in the pull request, without special handling, we lose upto 13% performance for the GPU.

jaeyounkim:
@adityaiitb Can you answer the remaining question and update your PR since there is conflict?

jaeyounkim:
We are closing this PR since there is no response from @adityaiitb. Please reopen it after resolving the conflict.

PR #8256, opened 2020-03-07
Tokenizer update

grwlf:
By this PR I propose several improvements to the tokenizer found in transformer model:

1. Unhardcode alphanumeric charset (it is still the default). This help me to use tokenizer in tasks which require a different charsets
2. Safety check before inserting EOL. The check detects the situation where users don't include default `RESERVE_TOKENS` list into their custom `reserved_tokens` argument.
3. ~~(updated)  Debug-print the status of long-runnning token counting~~

### DIFF
--- a/official/nlp/transformer/utils/tokenizer.py
+++ b/official/nlp/transformer/utils/tokenizer.py
@@ -45,11 +45,14 @@ _UNESCAPE_REGEX = re.compile(r"\\u|\\\\|\\([0-9]+);")
 
 _UNDEFINED_UNICODE = u"\u3013"
 
-# Set contains all letter and number characters.
-_ALPHANUMERIC_CHAR_SET = set(
+def alphanumeric_char_set():
+  return set(
     six.unichr(i) for i in xrange(sys.maxunicode)
     if (unicodedata.category(six.unichr(i)).startswith("L") or
         unicodedata.category(six.unichr(i)).startswith("N")))
+
+# Set contains all letter and number characters.
+_ALPHANUMERIC_CHAR_SET = alphanumeric_char_set()
 
 # min_count is the minimum number of times a subtoken must appear in the data
 # before before it is added to the vocabulary. The value is found using binary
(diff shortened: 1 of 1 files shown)


official/transformer/utils/tokenizer.py
```diff
@@ -61,11 +64,15 @@
 class Subtokenizer(object):
   """Encodes and decodes strings to/from integer IDs."""
 
-  def __init__(self, vocab_file, reserved_tokens=None):
+  def __init__(self, vocab_file, reserved_tokens=None,
+               master_char_set=None):
     """Initializes class, creating a vocab file if data_files is provided."""
```
saberkun: directly use master_char_set= _ALPHANUMERIC_CHAR_SET? Is it better

grwlf: I thought about it and finally decided to propose the `None`-style because I think it more convenient for future support:
1. It is harder to make an error when defining default arguments of downstream functions.
2. It keeps definition of `_ALPHANUMERIC_CHAR_SET` local tokenizer module. E.g. it allows hackers to tweak the global variable at runtime if they want to. AFAIK, mantioning it as a default would bind the original value of this variable to the function.
3. It seems that author of `resrved_tokens` thought the same way.

If you are sure, I am ready to change it to `_ALPHANUMERIC_CHAR_SET`.

grwlf:
> Can you apply the changes on nlp/transformer/... ? We have moved files.

Sure, will do it soon

official/nlp/transformer/utils/tokenizer.py
```diff
@@ -342,18 +357,26 @@ def _count_tokens(files, file_byte_limit=1e6, correct_strip=True):
...
   for filepath in files:
     with tf.io.gfile.GFile(filepath, mode="r") as reader:
       file_byte_budget = file_byte_limit
       counter = 0
       lines_to_skip = int(reader.size() / (file_byte_budget * 2))
-      for line in reader:
+      for i,line in enumerate(reader):
+        if i%100==0:
+          tf.compat.v1.logging.debug("Counting tokens, byte budget left for '%s:%d': %d" %
+                                     (filepath, i, file_byte_budget))
```
saberkun: i % 100 = 0

official/nlp/transformer/utils/tokenizer.py
```diff
@@ -342,18 +357,26 @@ def _count_tokens(files, file_byte_limit=1e6, correct_strip=True):
...
+
         if counter < lines_to_skip:
           counter += 1
         else:
           if file_byte_budget < 0:
+            tf.compat.v1.logging.debug("Byte budget of '%s' is over" % (filepath,))
             break
```
saberkun: overall, do we really need this debug logging?

official/nlp/transformer/utils/tokenizer.py
grwlf: removed altogether

official/nlp/transformer/utils/tokenizer.py
grwlf: Ok, it's really no that important.

Issue #8280, opened 2020-03-13
tf.Graph() can be used instead of tf.compat.v1.get_default_graph()


ayushmankumar7:
**Please provide the entire URL of the model you are using?**

https://github.com/tensorflow/models/official/vision/image_classification.common.py
**Describe the feature you request and the current behavior/state.**
Since the required task can be done with the lastest API under tf.Graph() then why use tf.compat.v1. 
In this feature request, i propose to convert tf.compat.v1.get_default_graph() to the lastest API.

**Are you willing to contribute it (Yes/No)?**
# YES
**Any other info.**

jaeyounkim:
image_classification.common.py is already deprecated. Please use our new code.

PR #8281, opened 2020-03-13
tf.compat.v1.get_default_graph()  using tf.Graph()

ayushmankumar7:
Issue : #8280


saberkun:
Thanks for the PR. We are deprecating this piece of code as it really not necessary to call graph in TF2.

ayushmankumar7:
OKay... Not a problem.

Issue #8293, opened 2020-03-14
tf.compat.v1.train.AdamOptimizer() to tf.keras.optimizers.Adam


ayushmankumar7:
**Please provide the entire URL of the model you are using?**

**Describe the feature you request and the current behavior/state.**

Few models use the old train.AdamOptimizer(). I propose to update it with the Keras API, tf.keras.optimizers.Adam().

**Are you willing to contribute it (Yes/No)?**

# YES
**Any other info.**

PR #8294, opened 2020-03-14
tf.compat.v1.train.AdamOptimizer to tf.keras.optimizers.Adam

ayushmankumar7:
Issue: #8293


saberkun:
Thanks ayushmankumar7. The models in R1 folder are frozen because they need to keep working with TF 1.15.
We recently added a note saying that we plan to remove these sample models by 2020/06.

ayushmankumar7:
@saberkun  It's done. Now it's fine, i guess.

Issue #8320, opened 2020-03-20
"ncf_test.py" failed.


ayushmankumar7:
**System information**
- Have I written custom code (as opposed to using a stock example script provided in TensorFlow):
- OS Platform and Distribution (e.g., Linux Ubuntu 16.04):
- Mobile device (e.g., Pixel 4, Samsung Galaxy 10) if the issue happens on mobile device:
- TensorFlow installed from (source or binary):pip3
- TensorFlow version (use command below): Tensorflow 2.2.0rc
- Python version: 3.6
- Bazel version (if compiling from source):
- GCC/Compiler version (if compiling from source):
- CUDA/cuDNN version:
- GPU model and memory:

**Please provide the entire URL of the model you are using?**

https://github.com/tensorflow/models/blob/master/official/recommendation/ncf_test.py
**Describe the current behavior**
![Screenshot from 2020-03-20 19-19-03](https://user-images.githubusercontent.com/41910134/77169598-c09d3180-6adf-11ea-9661-593b19760e2d.png)

**Describe the expected behavior**
![Screenshot from 2020-03-20 19-17-20](https://user-images.githubusercontent.com/41910134/77169601-c3982200-6adf-11ea-880a-de6ecc5ae72c.png)

**Code to reproduce the issue**

python3 official/recommendation/ncf_test.py
**Other info / logs**

saberkun:
Hi @ayushmankumar7 
strategy.run has been cherry picked to TF 2.2 RC1.
If you are using TF 2.1, you may revert the changes locally.

ayushmankumar7:
@saberkun  I agree with you. Since i was using tf 2.2rc0 .. i faced this problem.

saberkun:
Yeah, the master branch directly integrate with tf-nightly internally. So I would recommend to use tf-nightly with the master branch. But, yes, TF 2.2 will be released soon

saberkun:
Sorry, TF 2.2 release is a bit slow but we are sure this change will be in. Let's close this issue. Thank you for letting us know.

PR #8321, opened 2020-03-20
ncf_test.py error solved. 

ayushmankumar7:
Issue: #8320


saberkun:
Close as strategy.run is in TF2.2

PR #8344, opened 2020-03-31
Fix output shape of the position embedding layer

peakji:
The output shape of the position embedding layer should be the same as the word embeddings, which is `[batch_size, sequence_length, hidden_size]`, instead of `[1, sequence_length, hidden_size]`.

This PR explicitly calls `tf.broadcast_to()` to ensure the tf.keras.layers.Add() layer on [L133](https://github.com/tensorflow/models/blob/651677f5645831fe748a7eb7fe59a8b05c921a52/official/nlp/modeling/networks/transformer_encoder.py#L133) won't complain about merging tensors with different batch sizes.

### DIFF
--- a/official/nlp/modeling/layers/position_embedding.py
+++ b/official/nlp/modeling/layers/position_embedding.py
@@ -111,14 +111,9 @@ class PositionEmbedding(tf.keras.layers.Layer):
 
   def call(self, inputs):
     """Implements call() for the layer."""
+    input_shape = tf_utils.get_shape_list(inputs, expected_rank=3)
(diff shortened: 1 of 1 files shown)


saberkun:
Thanks for the PR. Sorry, we did not bring much tests to public. I will check if I see any issue internally.

saberkun:
Merged. Note that, we indeed have unit tests for this file open sourced. I would recommend running those before creating PR. 
The CI is in utils/testing.

peakji:
@saberkun Thanks! Will do.

PR #8382, opened 2020-04-10
Corrected tensorflow library install instruction

Noezor:
Installing tensorflow1.14 instead of 2.X

python3 -c "import delf" crashed because of call to tf.contrib.slim
contrib module has been deprecated in tf 2.0


Noezor:
@googlebot I signed it!

PR #8390, opened 2020-04-12
Updated README files of research models

jaeyounkim:
Added TensorFlow requirement
Added "No Maintenance Intended" for deprecated models


jaeyounkim:
I added a "No Maintenance Intended" badge for all research models confirmed to be archived by maintainers.

tf.compat.v1 might work for some models, yet I think it may still require some code changes. We can update a TensorFlow requirement badge again once we confirm it works with TensorFlow 2.

PR #8403, opened 2020-04-16
Make BertPretrainer to accept embedding_table explicitly

grwlf:
This PR would allow writing BERT pretraining models completely in a functional style. Currently, TrnaformerEncoder should be a custom layer, providing `get_embedding_table()` method. With this PR, we could use arbitrary `tf.keras.Model` for this.

Note, that I among TF warnings I now see the following:
```
WARNING:tensorflow:                                          
The following Variables were used a Lambda layer's call (lambda_1), but                                                   
are not present in its tracked objects:                      
  <tf.Variable 'BERT/word_embeddings/embeddings:0' shape=(30522, 768) dtype=float32>                                      
It is possible that this is intended behavior, but it is more likely                                                      
an omission. This is a strong indication that this layer should be                                                        
formulated as a subclassed Layer rather than a Lambda layer.  
```

I'm struggling to develop a style which could help to re-use as much code as possible so given the attitude towards functional style API in Keras I decided to propose this change. But I think it's OK to postone/cancel it if developers think this issue is a serious one.

### DIFF
--- a/official/nlp/bert/bert_models.py
+++ b/official/nlp/bert/bert_models.py
@@ -212,6 +212,7 @@ def pretrain_model(bert_config,
         stddev=bert_config.initializer_range)
   pretrainer_model = models.BertPretrainer(
       network=transformer_encoder,
+      embedding_table=transformer_encoder.get_embedding_table(),
       num_classes=2,  # The next sentence prediction label has two classes.
       num_token_predictions=max_predictions_per_seq,
       initializer=initializer,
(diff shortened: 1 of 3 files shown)


saberkun:
I feel embedding_table still should be a part of your source transformer encoder network in most cases?
embedding_table must be used by encoder.
But I agree assuming an encoder must have get_embedding_table() method is kind of strange.
Do you mind to make embedding_table as optional? if it is none, it accesses the source network embedding table.

saberkun:
I would still recommend you can build you own bertpretrainer model from the networks we provided. The BertPretrainer here is a canonical reference. It is not flexible for research.

grwlf:
> Do you mind to make embedding_table as optional? if it is none, it accesses the source network embedding table.

Sure, updated.

saberkun:
Thanks, Sergey. Let me pull the change to internal and review this API change.

PR #8421, opened 2020-04-21
fix index error, to support non-square kernels

bilalsal:
With kernel=[3, 3] in all entries of MOBILENETV1_CONV_DEFS, the latent index error (likely due to copy-paste) did not cause problems.
Fix is needed to ensure proper padding with non-square kernels.


bilalsal:
@googlebot I signed it!

PR #8454, opened 2020-04-30
DELF open-source library v2.0

andrefaraujo:
- Updated library to be fully compatible with TF2.1.
- New training code included.


research/delf/INSTALL_INSTRUCTIONS.md
```diff
@@ -2,15 +2,26 @@
...
+pip install 'tensorflow-gpu'
+```
+
+### TF-Slim
+
+Note: currently, we need to install the latest version from source, to avoid
+using previous versions which relied on tf.contrib (which is now deprecated).
+
+```bash
+git clone git@github.com:google-research/tf-slim.git
```
jaeyounkim, line 22: git clone https://github.com/google-research/tf-slim.git

andrefaraujo: your suggestion would be simply to use the HTTPS interface of github versus the SSH one (https://help.github.com/en/github/using-git/which-remote-url-should-i-use).

I left the default as the SSH one, but added several comments to help people troubleshoot this if needed, see https://github.com/tensorflow/models/pull/8454/commits/4879ffaa5779702274b2835bb8da4ea9efa3a5b8

PR #8477, opened 2020-05-07
Added real-time object detection demo and segmentation demo using Ope…

kd610:
I added real-time Object detection and Segmentation function using a webcam on the tutorial notebook.

What I added:

Opencv installation
run_inference function
Thank you.


kd610:
Is it pending??

Issue #8388, opened 2020-04-11
ValueError: Eval batch size 256 is not divisible by 1000


xiangshiyin:
**System information**
- Have I written custom code (as opposed to using a stock example script provided in TensorFlow):
- OS Platform and Distribution (e.g., Linux Ubuntu 16.04): MacOS Catalina Version 10.15.4 (19E287)d
- Mobile device (e.g., Pixel 4, Samsung Galaxy 10) if the issue happens on mobile device:
- TensorFlow installed from (source or binary):
- TensorFlow version (use command below): 2.2.0-dev20200411
- Python version: 3.6
- Bazel version (if compiling from source):
- GCC/Compiler version (if compiling from source):
- CUDA/cuDNN version:
- GPU model and memory:

**Please provide the entire URL of the model you are using?** 
https://github.com/tensorflow/models/blob/master/official/recommendation/ncf_keras_main.py

**Describe the current behavior**
Got the following error message when running command `python official/recommendation/ncf_keras_main.py --model_dir ../XXXX/model --data_dir ../XXXX/data`:
```
ValueError: Eval batch size 256 is not divisible by 1000
```
Full log is like below:
```sh
$ python official/recommendation/ncf_keras_main.py --model_dir ../ncf_tensorflow_sample/model --data_dir ../ncf_tensorflow_sample/data
2020-04-11 11:03:52.325168: I tensorflow/core/platform/cpu_feature_guard.cc:142] This TensorFlow binary is optimized with Intel(R) MKL-DNN to use the following CPU instructions in performance-critical operations:  AVX2 FMA
To enable them in other operations, rebuild TensorFlow with the appropriate compiler flags.
2020-04-11 11:03:52.343370: I tensorflow/compiler/xla/service/service.cc:168] XLA service 0x7f8d359bcbd0 initialized for platform Host (this does not guarantee that XLA will be used). Devices:
2020-04-11 11:03:52.343389: I tensorflow/compiler/xla/service/service.cc:176]   StreamExecutor device (0): Host, Default Version
WARNING:tensorflow:Some requested devices in `tf.distribute.Strategy` are not visible to TensorFlow: /job:localhost/replica:0/task:0/device:GPU:0
W0411 11:03:52.343887 4374371776 cross_device_ops.py:1172] Some requested devices in `tf.distribute.Strategy` are not visible to TensorFlow: /job:localhost/replica:0/task:0/device:GPU:0
[...]

PR #8509, opened 2020-05-14
Fix an example command about ncf_keras_main.py in README.md

nuka137:
# Description

Example command in README raises error, which is also mentioned at #8388.
I think an example command in README should not raise any errors because many users may execute this command as is.
This PR fixes the example command not to raise error.

## Type of change

- [x] Documentation update

## Checklist

### DIFF
--- a/official/recommendation/README.md
+++ b/official/recommendation/README.md
@@ -61,7 +61,7 @@ features in TF 2.x. Users can train the model on both GPU and TPU.
 
 To train and evaluate the model, issue the following command:
 ```
-python ncf_keras_main.py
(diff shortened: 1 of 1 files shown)


saberkun:
Really? I have update the default value of batch_size to the correct number: 9900. If eval_batch_size is None, it will use batch_size. It should not get error.

nuka137:
Maybe, it depends on the version.
When I tried v2.1.0 and v2.2.0, the error is raised.
But the issue seems to be fixed at master branch.

So, I will use master branch.
Sorry for making you confused.

PR #8604, opened 2020-06-01
Create an example for clustering mobilenet_v1 in resnet_imagenet_main.py

Ruomei:
# Description

This change adds an example test for clustering mobilenet_v1 in resnet_imagenet_main.py as well as flags changes in common.py. It can demonstrate how clustering APIs in tensorflow/model-optimization work on the last three conv2d layers of mobilenet_v1. 

## Type of change

- [ ] New feature (non-breaking change which adds functionality)

## Tests
The test is currently not dependent on any other benchmark code. To run it, we can set: 

**Test Configuration**:
Tensorflow == 2.1.0
Keras-Applications == 1.0.8
Python == 3.7.7
cudatoolkit == 10.1.243
cudnn == 7.6.5

**Test Command line example**:
python resnet_imagenet_main.py --data_dir=your_path_to_imagenet --model='mobilenet_pretrained' --train_epochs=1 --batch_size=64 --optimizer='mobilenet_fine_tune' --distribution_strategy='mirrored' --num_gpus=4 --save_files_to='./exp_folder' --clustering_method='selective_clustering'
(make sure the --data_dir flag is set to the same one needed by the previous resnet test.)

## Checklist

- [ ] I have signed the [Contributor License Agreement](https://github.com/tensorflow/models/wiki/Contributor-License-Agreements).
- [ ] I have read [guidelines for pull request](https://github.com/tensorflow/models/wiki/Submitting-a-pull-request).
- [ ] My code follows the [coding guidelines](https://github.com/tensorflow/models/wiki/Coding-guidelines).

### DIFF
--- a/official/benchmark/keras_imagenet_benchmark.py
+++ b/official/benchmark/keras_imagenet_benchmark.py
@@ -42,6 +42,7 @@ MOBILENET_V1_MAX_TOP_1_ACCURACY = 0.68
 MODEL_OPTIMIZATION_TOP_1_ACCURACY = {
     'RESNET50_FINETUNE_PRUNING': (0.76, 0.77),
     'MOBILENET_V1_FINETUNE_PRUNING': (0.67, 0.68),
+    'MOBILENET_V1_FINETUNE_CLUSTERING': (0.68, 0.70)
 }
 
 FLAGS = flags.FLAGS
@@ -1771,6 +1772,120 @@ class Resnet50KerasPruningBenchmarkReal(KerasPruningBenchmarkRealBase):
         'optimizer': 'mobilenet_default',
     }
     super(Resnet50KerasPruningBenchmarkReal, self).__init__(
+        default_flags=default_flags, **kwargs)
+
+
+class KerasClusteringAccuracyBase(keras_benchmark.KerasBenchmark):
+  """Benchmark accuracy tests for clustering method."""
+
+  def __init__(self,
+               output_dir=None,
+               root_data_dir=None,
+               default_flags=None,
+               **kwargs):
+    """An accuracy benchmark class for clustering method.
+
+    Args:
+      output_dir: directory where to output e.g. log files
+      root_data_dir: directory under which to look for dataset
+      default_flags: default flags
+      **kwargs: arbitrary named arguments. This is needed to make the
+                constructor forward compatible in case PerfZero provides more
+                named arguments before updating the constructor.
+    """
+    if default_flags is None:
+      default_flags = {}
+    default_flags['clustering_method'] = 'selective_clustering'
+    default_flags['data_dir'] = os.path.join(root_data_dir, 'imagenet')
+    default_flags['model'] = 'mobilenet_pretrained'
+    default_flags['optimizer'] = 'mobilenet_fine_tune'
+
+    flag_methods = [resnet_imagenet_main.define_imagenet_keras_flags]
+
+    super(KerasClusteringAccuracyBase, self).__init__(
+        output_dir=output_dir,
+        flag_methods=flag_methods,
+        default_flags=default_flags,
+        **kwargs)
+
+  def benchmark_8_gpu(self):
+    """Test Keras model with eager, dist_strat and 8 GPUs."""
+    self._setup()
+
+    FLAGS.num_gpus = 8
+    FLAGS.batch_size = 32 * 8
+    FLAGS.train_epochs = 1
+    FLAGS.model_dir = self._get_model_dir('benchmark_8_gpu')
+    FLAGS.dtype = 'fp32'
+    FLAGS.enable_eager = True
+    self._run_and_report_benchmark()
+
+  @benchmark_wrappers.enable_runtime_flags
+  def _run_and_report_benchmark(self,
+                                top_1_min=MODEL_OPTIMIZATION_TOP_1_ACCURACY[
+                                    'MOBILENET_V1_FINETUNE_CLUSTERING'][0],
+                                top_1_max=MODEL_OPTIMIZATION_TOP_1_ACCURACY[
+                                    'MOBILENET_V1_FINETUNE_CLUSTERING'][1]):
+    start_time_sec = time.time()
+    stats = resnet_imagenet_main.run(flags.FLAGS)
+    wall_time_sec = time.time() - start_time_sec
+
+    super(KerasClusteringAccuracyBase, self)._report_benchmark(
+        stats,
+        wall_time_sec,
+        top_1_min=top_1_min,
+        top_1_max=top_1_max,
+        total_batch_size=FLAGS.batch_size,
+        log_steps=100)
+
+
+class MobilenetV1KerasClusteringAccuracy(KerasClusteringAccuracyBase):
+  """Benchmark accuracy tests for MobilenetV1 with clustering method."""
+
+  def __init__(self, root_data_dir=None, **kwargs):
+    default_flags = {
+        'model': 'mobilenet_pretrained',
+        'optimizer': 'mobilenet_fine_tune',
+    }
+    super(MobilenetV1KerasClusteringAccuracy, self).__init__(
+        root_data_dir=root_data_dir,
+        default_flags=default_flags,
+        **kwargs)
+
+  def _run_and_report_benchmark(self):
+    super(MobilenetV1KerasClusteringAccuracy, self)._run_and_report_benchmark(
+        top_1_min=\
+        MODEL_OPTIMIZATION_TOP_1_ACCURACY['MOBILENET_V1_FINETUNE_CLUSTERING'][0],
+        top_1_max=\
+        MODEL_OPTIMIZATION_TOP_1_ACCURACY['MOBILENET_V1_FINETUNE_CLUSTERING'][1])
+
+
+class KerasClusteringBenchmarkRealBase(Resnet50KerasBenchmarkBase):
+  """Clustering method benchmarks."""
+
+  def __init__(self, root_data_dir=None, default_flags=None, **kwargs):
+    if default_flags is None:
+      default_flags = {}
+    default_flags.update({
+        'skip_eval': True,
+        'report_accuracy_metrics': False,
+        'data_dir': os.path.join(root_data_dir, 'imagenet'),
+        'train_steps': 110,
+        'log_steps': 10,
+    })
+    super(KerasClusteringBenchmarkRealBase, self).__init__(
+        default_flags=default_flags, **kwargs)
+
+
+class MobilenetV1KerasClusteringBenchmarkReal(KerasClusteringBenchmarkRealBase):
+  """Clustering method benchmarks for MobilenetV1."""
+
+  def __init__(self, **kwargs):
+    default_flags = {
+        'model': 'mobilenet_pretrained',
+        'optimizer': 'mobilenet_fine_tune',
+    }
+    super(MobilenetV1KerasClusteringBenchmarkReal, self).__init__(
         default_flags=default_flags, **kwargs)
 
 
--- a/official/benchmark/keras_imagenet_benchmark.py
+++ b/official/benchmark/keras_imagenet_benchmark.py
@@ -905,7 +905,7 @@ class Resnet50KerasBenchmarkBase(keras_benchmark.KerasBenchmark):
     FLAGS.enable_eager = True
     FLAGS.distribution_strategy = 'mirrored'
     FLAGS.model_dir = self._get_model_dir('benchmark_8_gpu')
-    FLAGS.batch_size = 128 * 8
+    FLAGS.batch_size = 128 * 8  # 8 GPUs
     self._run_and_report_benchmark()
 
   def benchmark_8_gpu_amp(self):
@@ -918,7 +918,7 @@ class Resnet50KerasBenchmarkBase(keras_benchmark.KerasBenchmark):
     FLAGS.fp16_implementation = 'graph_rewrite'
     FLAGS.distribution_strategy = 'mirrored'
     FLAGS.model_dir = self._get_model_dir('benchmark_8_gpu_amp')
-    FLAGS.batch_size = 256 * 8
+    FLAGS.batch_size = 256 * 8  # 8 GPUs
     self._run_and_report_benchmark()
 
   def benchmark_8_gpu_tweaked(self):
@@ -929,7 +929,7 @@ class Resnet50KerasBenchmarkBase(keras_benchmark.KerasBenchmark):
     FLAGS.enable_eager = True
     FLAGS.distribution_strategy = 'mirrored'
     FLAGS.model_dir = self._get_model_dir('benchmark_8_gpu_tweaked')
-    FLAGS.batch_size = 128 * 8
+    FLAGS.batch_size = 128 * 8  # 8 GPUs
     FLAGS.datasets_num_private_threads = 14
     self._run_and_report_benchmark()
 
@@ -942,7 +942,7 @@ class Resnet50KerasBenchmarkBase(keras_benchmark.KerasBenchmark):
     FLAGS.enable_xla = True
     FLAGS.distribution_strategy = 'mirrored'
     FLAGS.model_dir = self._get_model_dir('benchmark_xla_8_gpu')
-    FLAGS.batch_size = 128 * 8
+    FLAGS.batch_size = 128 * 8  # 8 GPUs
     self._run_and_report_benchmark()
 
   def benchmark_xla_8_gpu_amp(self):
@@ -956,7 +956,7 @@ class Resnet50KerasBenchmarkBase(keras_benchmark.KerasBenchmark):
     FLAGS.enable_xla = True
     FLAGS.distribution_strategy = 'mirrored'
     FLAGS.model_dir = self._get_model_dir('benchmark_xla_8_gpu_amp')
-    FLAGS.batch_size = 256 * 8
+    FLAGS.batch_size = 256 * 8  # 8 GPUs
     self._run_and_report_benchmark()
 
   def benchmark_xla_8_gpu_tweaked(self):
@@ -982,7 +982,7 @@ class Resnet50KerasBenchmarkBase(keras_benchmark.KerasBenchmark):
     FLAGS.enable_eager = True
     FLAGS.distribution_strategy = 'mirrored'
     FLAGS.model_dir = self._get_model_dir('benchmark_8_gpu_fp16')
-    FLAGS.batch_size = 256 * 8
+    FLAGS.batch_size = 256 * 8  # 8 GPUs
     self._run_and_report_benchmark()
 
   def benchmark_8_gpu_fp16_tweaked(self):
@@ -994,7 +994,7 @@ class Resnet50KerasBenchmarkBase(keras_benchmark.KerasBenchmark):
     FLAGS.enable_eager = True
     FLAGS.distribution_strategy = 'mirrored'
     FLAGS.model_dir = self._get_model_dir('benchmark_8_gpu_fp16_tweaked')
-    FLAGS.batch_size = 256 * 8
+    FLAGS.batch_size = 256 * 8  # 8 GPUs
     FLAGS.tf_gpu_thread_mode = 'gpu_private'
     FLAGS.datasets_num_private_threads = 40
     self._run_and_report_benchmark()
@@ -1009,7 +1009,7 @@ class Resnet50KerasBenchmarkBase(keras_benchmark.KerasBenchmark):
     FLAGS.distribution_strategy = 'mirrored'
     FLAGS.model_dir = self._get_model_dir(
         'benchmark_8_gpu_fp16_dynamic_tweaked')
-    FLAGS.batch_size = 256 * 8
+    FLAGS.batch_size = 256 * 8  # 8 GPUs
     FLAGS.loss_scale = 'dynamic'
     FLAGS.tf_gpu_thread_mode = 'gpu_private'
     FLAGS.datasets_num_private_threads = 40
@@ -1025,7 +1025,7 @@ class Resnet50KerasBenchmarkBase(keras_benchmark.KerasBenchmark):
     FLAGS.enable_xla = True
     FLAGS.distribution_strategy = 'mirrored'
     FLAGS.model_dir = self._get_model_dir('benchmark_xla_8_gpu_fp16')
-    FLAGS.batch_size = 256 * 8
+    FLAGS.batch_size = 256 * 8  # 8 GPUs
     self._run_and_report_benchmark()
 
   def benchmark_xla_8_gpu_fp16_tweaked(self):
@@ -1038,7 +1038,7 @@ class Resnet50KerasBenchmarkBase(keras_benchmark.KerasBenchmark):
     FLAGS.enable_xla = True
     FLAGS.distribution_strategy = 'mirrored'
     FLAGS.model_dir = self._get_model_dir('benchmark_xla_8_gpu_fp16_tweaked')
-    FLAGS.batch_size = 256 * 8
+    FLAGS.batch_size = 256 * 8  # 8 GPUs
     FLAGS.tf_gpu_thread_mode = 'gpu_private'
     FLAGS.datasets_num_private_threads = 48
     self._run_and_report_benchmark()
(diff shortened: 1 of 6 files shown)


official/benchmark/models/resnet_imagenet_main.py
```diff
@@ -275,16 +365,23 @@ def run(flags_obj):
   if not strategy and flags_obj.explicit_gpu_placement:
     no_dist_strat_device.__exit__()
 
+  if flags_obj.save_files_to:
```
TamasArm: hmm, I can't see where it is ensured that we are in clustering mode at this point

Ruomei: yes, that's a good point and I fixed in the latest commit.

Ruomei: Also, more tests were added in the latest commit ...

TamasArm: yes, sorry, I totally overlooked that

official/benchmark/keras_imagenet_benchmark.py
```diff
@@ -1720,5 +1721,119 @@ def __init__(self, **kwargs):
...
+  def __init__(self, root_data_dir=None, default_flags=None, **kwargs):
+    if default_flags is None:
+      default_flags = {}
+    default_flags.update({
+        'skip_eval': True,
+        'report_accuracy_metrics': False,
+        'data_dir': os.path.join(root_data_dir, 'imagenet'),
+        'train_steps': 110,
+        'log_steps': 10,
+    })
```
TamasArm, line 1821: I'm not sure what's happening here, could you please explain?

official/benchmark/models/resnet_imagenet_test.py
```diff
@@ -236,7 +248,10 @@ def test_end_to_end_xla_2_gpu_fp16(self, flags_key):
     extra_flags = extra_flags + self.get_extra_flags_dict(flags_key)
```
TamasArm, line 248: A larger block of code is repeated 3 times, could it be extracted into its own function?

Ruomei:
Clustering related tests should fail for the fp16 and mixed-precision training cases now. It is unclear whether we need to support fp16 before the release.

official/benchmark/keras_imagenet_benchmark.py
Ruomei: When the users pass in None, default_flags = {} is because default_flag has to be a dictionary first before we filling in any key-value pairs later, which are necessary for any clustering tests. 
I don't think it is an if-else logic here. What do you think?

official/benchmark/models/resnet_imagenet_test.py
Ruomei: hmm, will need some refactoring perhaps.

official/benchmark/models/resnet_imagenet_main.py
Ruomei: no worries.

alanchiao:
For the existing pruning integration, we didn't test fp16/mixed-precision training and it's not an explicit requirement in the guidelines from [here](https://github.com/tensorflow/models/wiki/Research-paper-code-contribution) and [here](https://github.com/tensorflow/models/wiki/Coding-guidelines)

official/benchmark/keras_imagenet_benchmark.py
TamasArm: Oh I see. What I don't understand is what happens when you do provide your own default flags in the constructor. Won't they get overridden here?

saberkun:
@alanchiao Please let us know when you feel the technical part of this PR is good.

official/vision/image_classification/resnet/common.py
```diff
@@ -331,6 +331,14 @@ def define_pruning_flags():
   flags.DEFINE_integer('pruning_frequency', 100,
                        'Frequency for pruning.')
 
+def define_clustering_flags():
+  """Define flags for clustering methods."""
+  flags.DEFINE_string('clustering_method', None,
+                      'None (no clustering) or selective_clustering.')
+  flags.DEFINE_integer('number_of_clusters', 256,
+                       'Number of clusters used in each layer.')
+  flags.DEFINE_string('save_files_to', None,
```
alanchiao, line 340: Don't think a new `save_files_to` flag should exist that's specific to clustering. There is existing export logic in the script already. The current code doesn't create a TFLite model and that seems fine.

official/vision/image_classification/resnet/common.py
```diff
@@ -331,6 +331,14 @@ def define_pruning_flags():
   flags.DEFINE_integer('pruning_frequency', 100,
                        'Frequency for pruning.')
 
+def define_clustering_flags():
+  """Define flags for clustering methods."""
+  flags.DEFINE_string('clustering_method', None,
+                      'None (no clustering) or selective_clustering.')
+  flags.DEFINE_integer('number_of_clusters', 256,
+                       'Number of clusters used in each layer.')
```
alanchiao, line 339: The description of the `number_of_clusters` flags should clarify how configurable the model really is. I'm inclined to omit this flag since it's not meaningfully configurable (it doesn't affect the whole model) and someone probably would rather directly modify the code at that point. 

For `selective_clustering` flag, I'd describe it further, like `selective_clustering, which will cluster
the last three Conv2d layers of the model`.

official/benchmark/models/resnet_imagenet_main.py
```diff
@@ -25,6 +25,7 @@
 from absl import logging
 import tensorflow as tf
 
+from tensorflow_model_optimization.python.core.clustering.keras import cluster
 import tensorflow_model_optimization as tfmot
```
alanchiao, line 29: We didn't do this previously, but should start doing it now and the long term integration:

It'd be best to import `tfmot` only when the flags indicate that tfmot will be used, so that if there were hypothetically breakages, the rest of the benchmarks could continue to run.
   

So only if flags_obj.clustering_method

official/benchmark/models/resnet_imagenet_test.py
```diff
@@ -26,12 +26,13 @@
 from official.utils.testing import integration
 from official.vision.image_classification.resnet import imagenet_preprocessing
 
-
+# TBC: joint clustering and tuning is not supported yet so only one flag should be selected
```
alanchiao, line 29: Could you clarify what this means?

official/benchmark/models/resnet_imagenet_main.py
```diff
@@ -37,6 +38,58 @@
 from official.vision.image_classification.resnet import resnet_model
 
 
+def selective_layers_to_cluster(model):
```
alanchiao, line 41: I think these four functions (`selective_layers_to_cluster` to `get_selectively_clustered_model`) have unnecessary layers of indirection.

Something like this should suffice:

```
def cluster_last_three_conv2d_layers(model):
    last_three_conv2d_layers =  [
          layer for layer in model.layers
          if isinstance(layer, tf.keras.layers.Conv2D) and
          not isinstance(layer, tf.keras.layers.DepthwiseConv2D)
     ]
     
     cluster_centroids_init = 'linear'

     def cluster_fn(layer):
        if layer not in last_three_conv2d_layers:
           return layer

        if layer == last_three_conv2d_layers[0] or
           layer == last_three_conv2d_layers[1]:
               clustered = cluster.cluster_weights(layer, clusters=256, cluster_centroids_init=cluster_centroids_init)
        else:
             clustered = cluster.cluster_weights(layer, clusters=32, cluster_centroids_init=cluster_centroids_init)

        print("Clustered {} with {} clusters".format(layer.name, clustered.number_of_clusters)

     return tf.keras.models.clone_model(model, clone_function=cluster_fn)
     
```
1. There isn't a need to have different function arguments unless they are actually configurable by an end user (which isn't needed in this current version ... only the long-term integration). 
2. See https://google.github.io/styleguide/pyguide.html#310-strings.
3. I switched to `layer` instead of `layer.name`, which I don't think there are disadvantages, and it makes the code a bit cleaner.

alanchiao:
I've reproduced the clustering results internally, so we can move forward with trying to make a TFMOT release with the clustering library.

Made the comments above, focusing on what I think is relevant, even in the longer term integration. @pengchongjin, @saberkun can comment further. 

After the release, once @pengchongjin, @saberkun have given the OK, we can bump the TFMOT version that Official Models depends on, and then try to merge this PR after addressing all comments.

alanchiao:
@pengchongjin , @saberkun : could you provide some feedback on top of what I've suggested? 

We've made the TFMOT 0.4.0 release now.

saberkun:
> @pengchongjin , @saberkun : could you provide some feedback on top of what I've suggested?
> 
> We've made the TFMOT 0.4.0 release now.

Hi Alan, if you feel it good to merge, you can just approve this PR and we can pull it to internal to let you double check then.

alanchiao:
@Ruomei: as Hongkun mentioned above, we can merge after addressing my above comments. Let's try to wrap up this work. Thanks!

official/benchmark/models/resnet_imagenet_main.py
```diff
@@ -186,6 +241,17 @@ def run(flags_obj):
           weights=None,
           classes=imagenet_preprocessing.NUM_CLASSES,
           layers=tf.keras.layers)
+    elif flags_obj.model == 'mobilenet_pretrained':
+      model = tf.keras.applications.mobilenet.MobileNet(
```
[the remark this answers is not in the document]
Ruomei: > I'd dedup the code from flags_obj.model == 'mobilenet' to reflect that.

Which one is better for the 'dedup' here? 

```
    elif flags_obj.model == 'mobilenet' or 'mobilenet_pretrained':
      # TODO(kimjaehong): Remove layers attribute when minimum TF version
      # support 2.0 layers by default.
      if flags_obj.model == 'mobilenet_pretrained':
        classes_labels = 1000
        dropout_rate = 1e-7
        initial_weights = 'imagenet'
      else:
        classes_labels = imagenet_preprocessing.NUM_CLASSES
        dropout_rate = 1e-3
        initial_weights = None
      model = tf.keras.applications.mobilenet.MobileNet(
          dropout=dropout_rate,
          weights=initial_weights,
          classes=classes_labels,
          layers=tf.keras.layers)
```
or 

```
    elif flags_obj.model == 'mobilenet':
      # TODO(kimjaehong): Remove layers attribute when minimum TF version
      # support 2.0 layers by default.
      model = tf.keras.applications.mobilenet.MobileNet(
          dropout=1e-3,
          weights=None,
          classes=imagenet_preprocessing.NUM_CLASSES,
          layers=tf.keras.layers)
    elif flags_obj.model == 'mobilenet_pretrained':
      model = tf.keras.applications.mobilenet.MobileNet(
          dropout=1e-7,
          weights='imagenet',
          classes=1000,
          layers=tf.keras.layers)    
```
Or, do you have anything else in your mind?

official/vision/image_classification/augment.py
```diff
@@ -27,7 +27,7 @@
 import tensorflow as tf
 from typing import Any, Dict, List, Optional, Text, Tuple
 
-from tensorflow.python.keras.layers.preprocessing import image_preprocessing as image_ops
```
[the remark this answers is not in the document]
Ruomei: Hi @alanchiao, I finished addressing all the review comments but this one and tested the updated commit (currently dealing with the faff cause by 2FA).

How could I revert to TensorFlow HEAD? Do you mean I should use the tf.keras APIs instead of tf.python.keras ones?
I saw so many tf.python.keras used in this repo tho. Do I need to change all the imports related to this PR?

Thanks!

official/benchmark/models/resnet_imagenet_test.py
Ruomei: It was meant to be `clustering and pruning`.
I think when the pruning and clustering flag both enabled there was a problem during testing. Perhaps we could just delete this comment?

official/benchmark/models/resnet_imagenet_main.py
Ruomei: Thanks!

official/benchmark/models/resnet_imagenet_main.py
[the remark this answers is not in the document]
alanchiao: First one is preferable. I was wondering though why the # of labels and dropout_rate need to differ. For the case of the dropout_rate, does it actually affect the resulting accuracy significantly to not use the default?

official/vision/image_classification/augment.py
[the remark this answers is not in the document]
alanchiao: Thanks. Going forward, please resolve any comments that you've resolved in the Github PR.

Just meant to revert this one-line change.

"from tensorflow.python.keras.layers import image_preprocessing as image_ops" works at tf-nightly/etc.

"from tensorflow.python.keras.layers.preprocessing import image_preprocessing as image_ops" works at older versions of TF

official/benchmark/models/resnet_imagenet_main.py
```diff
@@ -37,6 +38,40 @@
 from official.vision.image_classification.resnet import resnet_model
 
 
+def cluster_last_three_conv2d_layers(model):
+  last_three_conv2d_layers =  [
+      layer for layer in model.layers
+      if isinstance(layer, tf.keras.layers.Conv2D) and
+      not isinstance(layer, tf.keras.layers.DepthwiseConv2D)
```
alanchiao, line 45: filter for `DepthwiseConv2D` not needed anymore. weight clustering skips it now.

official/benchmark/models/resnet_imagenet_main.py
```diff
@@ -37,6 +38,40 @@
...
+    ]
+  last_three_conv2d_layers = last_three_conv2d_layers[-3:]
```
alanchiao, line 47: Would just merge this with the above line.

official/benchmark/keras_imagenet_benchmark.py
```diff
@@ -898,7 +899,7 @@ def benchmark_8_gpu_tweaked(self):
     FLAGS.enable_eager = True
     FLAGS.distribution_strategy = 'mirrored'
     FLAGS.model_dir = self._get_model_dir('benchmark_8_gpu_tweaked')
-    FLAGS.batch_size = 128 * 8  # 8 GPUs
```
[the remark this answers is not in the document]
alanchiao: Still unresolved. If there wasn't a reason, it would be good to revert.

official/benchmark/models/resnet_imagenet_main.py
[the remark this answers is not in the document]
Ruomei: yea, the dropout_rate and # of labels do affect the resulting accuracy as I recall. However, I cannot find the test results of how much of a difference they are making. I can rerun the tests if necessary?

official/benchmark/models/resnet_imagenet_main.py
alanchiao: Unresolved still.

If my comment wasn't clear, the idea is that the entire "import tensorflow_model_optimization" should be moved under an if statement and done only if clustering or pruning are enabled

official/benchmark/models/resnet_imagenet_main.py
alanchiao: That way `last_three_conv2d_layers` always is the last three layers.

official/benchmark/models/resnet_imagenet_main.py
```diff
@@ -37,6 +38,40 @@
...
+
+  cluster_weights = tfmot.clustering.keras.cluster_weights
+  CentroidInitialization = tfmot.clustering.keras.CentroidInitialization
+  clustering_params1 = {
```
alanchiao, line 51: In my other comment, I suggested just directly inlining clustering_params1 and clustering_params2 with the call to `cluster_weights`. Was there a reason not to do so?

Inlining removes a layer of indirection.

official/benchmark/models/resnet_imagenet_main.py
```diff
@@ -148,8 +182,8 @@ def run(flags_obj):
       boundaries=list(p[1] for p in common.LR_SCHEDULE[1:]),
       multipliers=list(p[0] for p in common.LR_SCHEDULE),
       compute_lr_on_cpu=True)
-  steps_per_epoch = (
```
alanchiao, line 151: If this is a change due to some linter, please check it's the same as the one that Official Models uses (if they mention it).

alanchiao: I see similar indenting changes everywhere

official/benchmark/models/resnet_imagenet_main.py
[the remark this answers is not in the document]
alanchiao: Similarly above, I see the branch in the optimizer to set a different learning rate. 

Does the experiment not work if we use the existing flags_obj.initial_learning_rate_per_sample * flags_obj.batch_size instead?

official/benchmark/models/resnet_imagenet_test.py
alanchiao: Resolving since you deleted the comment.

official/benchmark/models/resnet_imagenet_main.py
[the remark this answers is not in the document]
alanchiao: Ok. It could be that the pretrained weights themselves work with 1000 classes. I think it's fine to defer that to the future.

For dropout and the optimizer, I'd see if we can remove the branch though and run tests to see if they matter.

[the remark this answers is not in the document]
alanchiao: If they do, maybe it'll be preferable to change the dropout for pruning to match clustering. You can leave a TODO for tfmot for that.

alanchiao:
@Ruomei : you should move this to the benchmark branch of this Github.

@saberkun

official/benchmark/keras_imagenet_benchmark.py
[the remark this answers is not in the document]
Ruomei: Do you mean the inconsistency is the `# 8 GPUs` comments?
Those were in the original script. 
Should I follow the original script or just remove the comments?

official/benchmark/models/resnet_imagenet_main.py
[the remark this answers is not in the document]
Ruomei: > Does the experiment not work if we use the existing flags_obj.initial_learning_rate_per_sample * flags_obj.batch_size instead?

It works. I tested that and we can remove the optimizer branch.

For dropout, however, my coworker got different results from mine. I will double-check it next week (Haven't got time to look at this work for the past week. Sorry!).

official/benchmark/models/resnet_imagenet_main.py
```diff
@@ -275,16 +362,24 @@ def run(flags_obj):
   if not strategy and flags_obj.explicit_gpu_placement:
     no_dist_strat_device.__exit__()
 
+  if flags_obj.clustering_method:
```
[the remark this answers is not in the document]
Ruomei: Done.

official/vision/image_classification/resnet/common.py
Ruomei: yes, that makes sense. Thanks.

official/vision/image_classification/resnet/common.py
Ruomei: Done both. 
Tho I am slightly skeptical about not having the flag `number_of_clusters`, the code does look concise in your refactoring.

official/benchmark/models/resnet_imagenet_main.py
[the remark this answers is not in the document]
Ruomei: So we have run the script `resnet_imagenet_main.py` with the dropout equals to the default value several times and the accuracy was good, and the maximum variation is **0.26%**. However, when I use `tf.distribute` to test the same script, it seems that the average accuracy is consistently lower and the maximum variation is **0.73%**.

| tf.distribute  | run 1 | run 2 | run 3 | run 4 | run 5 | maximum variation
|----------------|-------|-------|-------|-------|-------|-------|
| on with 1 GPU  | 0.6907 | 0.6906 | 0.6836 | 0.6875 | 0.6909 | 0.0073|
| off with 1 GPU | 0.6939 | 0.6944 | 0.6951 | 0.6965 | 0.6964 | 0.0026 |

I think the test results are acceptable in both cases and they proved that the dropout can be set to the default value. However, the effect of `tf.distribute` on the test results can be investigated in other tasks, perhaps? @alanchiao

official/benchmark/keras_imagenet_benchmark.py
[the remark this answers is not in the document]
Ruomei: Reverted all. Thanks.

Ruomei:
Hi @alanchiao, I have just updated the PR and please let me know if anything else needs to be retested. If yes, I can set it running while addressing other comments. 
Thanks.

Ruomei:
@googlebot I fixed it.

Ruomei:
> Merge branch 'benchmark' into toupstream/clusteringexample

Nooooooo.
Tried to resolve merge conflicts with GitHub GUI conflict editor and it went wrong.

official/benchmark/keras_imagenet_benchmark.py
```diff
@@ -995,7 +996,7 @@ def benchmark_8_gpu_fp16_tweaked(self):
     FLAGS.model_dir = self._get_model_dir('benchmark_8_gpu_fp16_tweaked')
     FLAGS.batch_size = 256 * 8  # 8 GPUs
     FLAGS.tf_gpu_thread_mode = 'gpu_private'
-    FLAGS.dataset_num_private_threads = 40
+    FLAGS.datasets_num_private_threads = 40
```
alanchiao, line 999: Still not reverted.

official/vision/image_classification/augment.py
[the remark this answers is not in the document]
alanchiao: Making as unresolved - will handle when we merge the PR.

official/benchmark/models/resnet_imagenet_main.py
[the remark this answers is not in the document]
alanchiao: Fine with investigating in a separate PR.

official/benchmark/models/resnet_imagenet_main.py
alanchiao: Looks good. Resovling.

alanchiao:
@saberkun 

Ready to merge from my perspective. There are two lingering comments that can be fixed internally - my unresolved comments on "Still not reverted" and "Prior to submission, this will need to be reverted to work at TensorFlow HEAD." Happy to take care of it myself internally and if it's easiest, Ruomei can handle it instead.

Is there anything else I need to do on my end? I'm not sure also if you'd like Ruomei to squash the commits or do anything else.

@pengchongjin

Ruomei:
> @saberkun
> 
> Ready to merge from my perspective. There are two lingering comments that can be fixed internally - my unresolved comments on "Still not reverted" and "Prior to submission, this will need to be reverted to work at TensorFlow HEAD." Happy to take care of it myself internally and if it's easiest, Ruomei can handle it instead.
> 
> Is there anything else I need to do on my end? I'm not sure also if you'd like Ruomei to squash the commits or do anything else.
> 
> @pengchongjin

official/benchmark/keras_imagenet_benchmark.py
Ruomei: @alanchiao
Currently if you run the benchmark tests in `MobilenetV1KerasClusteringBenchmarkReal`, when it hits `benchmark_8_gpu_fp16_tweaked` and `benchmark_8_gpu_fp16_dynamic_tweaked`, it will throw `NotImplementedError: Clustering is currently only supported on dtype=tf.float32`. However, before I implemented this `NotImplementedError`, it throws an error `absl.flags._exceptions.UnrecognizedFlagError: Unknown command line flag 'dataset_num_private_threads'` because the flag `dataset_num_private_threads` is wrong I believe. However, I am aware of the fact that it is out of the scope of this PR. If you want this change to be reverted and deal with it later, I think it is also good.

official/vision/image_classification/augment.py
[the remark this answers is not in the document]
Ruomei: I can revert this when I do the final squash of all commits, as well as addressing comments from others (if there are any).

saberkun (approved):

official/benchmark/keras_imagenet_benchmark.py
alanchiao: Looks like you're right - I'm assuming this is a bug that wasn't caught and isn't regularly run. Good to leave as is then.

@saberkun  for visibility.

PR #5378, opened 2018-09-25
attention_ocr: A script to export SavedModel for Tensorflow Serving

hiepph:
Solution to this issue: https://github.com/tensorflow/models/issues/5330

PR #8757, opened 2020-06-30
attention_ocr: added export for SavedModel format.

xavigibert:
# Description

Add code to export model into SavedModel format. This code has the same functionality as PR #5378, but add endpoints for sequence confidence and predicted text. The code also incorporate several changes to allow running inference with batches of dynamic sizes.

## Type of change

Note: Please delete options that are not relevant.

- [x] New feature (non-breaking change which adds functionality)

## Tests

Added model_export_test.py. Ran all test to verify that nothing is broken.

**Test Configuration**:

## Checklist

### DIFF
--- a/research/attention_ocr/python/model.py
+++ b/research/attention_ocr/python/model.py
@@ -26,6 +26,7 @@ Usage example:
 import sys
 import collections
 import logging
+import numpy as np
 import tensorflow as tf
 from tensorflow.contrib import slim
 from tensorflow.contrib.slim.nets import inception
@@ -36,7 +37,7 @@ import utils
 
 OutputEndpoints = collections.namedtuple('OutputEndpoints', [
   'chars_logit', 'chars_log_prob', 'predicted_chars', 'predicted_scores',
-  'predicted_text'
+  'predicted_text', 'predicted_length', 'predicted_conf', 'normalized_seq_conf'
 ])
 
 # TODO(gorban): replace with tf.HParams when it is released.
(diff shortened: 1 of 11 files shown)


research/attention_ocr/python/model.py
```diff
@@ -380,12 +528,29 @@ def create_base(self,
         predicted_text = character_mapper.get_text(predicted_chars)
       else:
         predicted_text = tf.constant([])
+
+      text_log_prob, predicted_length = null_based_length_prediction(
+            chars_log_prob, self._params.null_code)
+      predicted_conf = lookup_indexed_value(predicted_length, text_log_prob)
+      # Convert predicted confidence from sum of logs to geometric mean
```
kyscg, line 535: I don't quite understand the reason for this, what am I missing?

xavigibert: We want to calculate the probability that the sequence of predicted characters is correct. We assume that the predictions (up to the first null character) are independent (in practice this is an approximation due to the use of the previous character in the prediction of the next one), so the joint probability is the product of probabilities. Since the confidence is the log of the probability, we sum the log probabilities.
The normalized_seq_conf is also provided, so that we can use a single threshold for the sequence confidence that does not depend on the sequence length (longer sequences have lower confidence than shorter ones).

xavigibert:
FYI. I got some presubmit errors:

https://sponge.corp.google.com/target?id=b12d2815-577d-40a7-b344-ba62a88b535b&target=tensorflow_models%2Fglint%2Fpy%2Fpresubmit&searchFor=
https://sponge.corp.google.com/target?id=eab61a32-3081-4f4a-9c2d-d52d3e923155&target=tensorflow_models%2Fubuntu%2Fpy3%2Fpresubmit_gpu_v2&searchFor=

I tried to fix them with pyformat and committed the changes. Now the presubmits don't get triggered anymore and it says "All checks have passed". However, the second commit invalidated the approval by jaeyounkim.

research/attention_ocr/python/model.py
kyscg: > in practice this is an approximation due to the use of the previous character in the prediction of the next one

Thanks for the clarification, I was thinking along these lines but wasn't quite sure. Things make sense now 😄

lynx97:
when I export, I still have to feed a fixed number of batch size. How to export serving model with batch size = -1
thanks

xavigibert:
@lynx97 thank you for reporting the issue. I believe that there is a bug on the code. Can you apply the following patch and and verify that it solves the issue?

```
diff --git a/research/attention_ocr/python/model_export.py b/research/attention_ocr/python/model_export.py
index 7c5f72e6..59b9d16d 100644
--- a/research/attention_ocr/python/model_export.py
+++ b/research/attention_ocr/python/model_export.py
@@ -117,7 +117,7 @@ def export_model(export_dir,
 
   if export_for_serving:
     images_orig = tf.placeholder(
-        tf.string, shape=[batch_size], name='tf_example')
+        tf.string, shape=[None], name='tf_example')
     images_orig_float = model_export_lib.generate_tfexample_image(
         images_orig,
         image_height,
```

lynx97:
@xavigibert thank you, it works.

PR #8797, opened 2020-07-07
How to predict future data, using previous data and Keras

vasilevskykv:
Hi! I would like to ask, does Keras Neural Networks do predictions in real time?
I see, that one of your parameters in prediction is array of data. But when you need to predict future data, you don't have an array of data - you have only previous data and time interval for prediction and nothing more.
As I undestand, you simply checked the prediction because to predict future data I need to give to a function predict a time interval in future, not a previous data trainX as necessary in this function predict

Here is my Neuro Network:
**look_back = 3
model = Sequential()
model.add(LSTM(4, input_shape=(look_back, 1)))
model.add(Dense(1))
model.compile(loss='mean_squared_error', optimizer='adam')
model.fit(trainX, trainY, epochs=20, batch_size=1, verbose=2)
trainPredict = model.predict(trainX)**

Can you show an example of predicting in real time?

### DIFF
--- a/official/bert/model_training_utils.py
+++ b/official/bert/model_training_utils.py
@@ -102,7 +102,8 @@ def run_customized_training_loop(
     metric_fn=None,
     init_checkpoint=None,
     use_remote_tpu=False,
-    custom_callbacks=None):
+    custom_callbacks=None,
+    run_eagerly=False):
   """Run BERT pretrain model training using low-level API.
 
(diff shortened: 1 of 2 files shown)


saberkun:
Could you ask this question in StackOverflow? Maybe TF or Keras github repo.
We do not response particular machine learning question here. Thank you

Issue #8793, opened 2020-07-06
Faster rcnn mobilene_v2 and mobilenet_v3 support 


anshkumar:
Since `faster_rcnn_mobilenet_v1_feature_extractor` is already provided, adding mobilenet_v2 and mobilenet_v3 should be easy. Mobilenet_v2 and v3 provides better accuracy than v1. 

I've written feature_extractor for both  [faster_rcnn_mobilenet_v2_feature_extractor.py](https://github.com/anshkumar/object_detection/blob/master/models/faster_rcnn_mobilenet_v2_feature_extractor.py) and [faster_rcnn_mobilenet_v3_feature_extractor.py](https://github.com/anshkumar/object_detection/blob/master/models/faster_rcnn_mobilenet_v3_feature_extractor.py). It can be added to the official repository.

pkulzc:
Could you create a pull request with your change? We can review it.

anshkumar:
Created one.

PR #8837, opened 2020-07-11
faster rcnn mobilene_v2 and mobilenet_v3 support

anshkumar:
Code for this [issue](https://github.com/tensorflow/models/issues/8793).


anshkumar:
@googlebot I signed it!

PR #9070, opened 2020-08-07
Consider regression model in predict mode

NivekNey:
# Description

The key `input_meta_data['num_labels']` doesn't always exist, it is defaulted to 1 when the Bert model is a regression model. In both modes `train_and_eval` and `export`, the "default to 1" setting is considered but not in `predict` mode.

Another issue this PR tries to solve is the unnecessary softmax applied to logit for regression model.

## Type of change

## Tests

Create a customized `DataProcessor` class with attribute `is_regression = True`, the output meta data will look something like follows where key `num_labels` is absent.

```
{
    "processor_type": "RegressionDataProcessor",
    "train_data_size": 19765218,
    "max_seq_length": 128,
    "task_type": "bert_regression",
    "label_type": "int",
    "eval_data_size": 1260452
}
```

This lets the finetuning process to create a regression Bert model. Finetuning finished without issue, prediction ran fine too after the proposed change, otherwise prediction will raise exception as follows.

```
KeyError: 'num_labels'
```

**Test Configuration**:

With `tf_models_nightly-2.2.0.dev20200720`

## Checklist

### DIFF
--- a/official/nlp/bert/run_classifier.py
+++ b/official/nlp/bert/run_classifier.py
@@ -262,6 +262,7 @@ def run_keras_compile_fit(model_dir,
 def get_predictions_and_labels(strategy,
                                trained_model,
                                eval_input_fn,
+                               num_labels,
                                return_probs=False):
   """Obtains predictions of trained model on evaluation data.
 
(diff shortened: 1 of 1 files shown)


NivekNey:
Actually, the issue goes deeper. For regression model, the activation https://github.com/tensorflow/models/blob/88253ce/official/nlp/bert/run_classifier.py#L291 shouldn't be applied. Should I try to include the change in this PR as well?

Issue #9057, opened 2020-08-06
Converting tf2 ssd_mobilenet-v1 object detection model to tflite model


Raphaeal19:
I am using Google Colab to run the training and conversion.

When I run the script 

!python /content/models/research/object_detection/export_tflite_ssd_graph.py \
--pipeline_config=/content/workspace/training/pipeline.config \
--trained_checkpoint=/content/workspace/training/model.ckpt \
--output_dir=/content/workspace/hands_detection \
--add_postprocessing_op=true

I get this error:

Traceback (most recent call last):
  File "/content/models/research/object_detection/export_tflite_ssd_graph.py", line 144, in <module>
    tf.app.run(main)
  File "/usr/local/lib/python3.6/dist-packages/tensorflow/python/platform/app.py", line 40, in run
    _run(main=main, argv=argv, flags_parser=_parse_flags_tolerate_undef)
  File "/usr/local/lib/python3.6/dist-packages/absl/app.py", line 299, in run
    _run_main(main, args)
  File "/usr/local/lib/python3.6/dist-packages/absl/app.py", line 250, in _run_main
    sys.exit(main(argv))
  File "/content/models/research/object_detection/export_tflite_ssd_graph.py", line 135, in main
    text_format.Merge(f.read(), pipeline_config)
  File "/usr/local/lib/python3.6/dist-packages/tensorflow/python/lib/io/file_io.py", line 116, in read
    self._preread_check()
  File "/usr/local/lib/python3.6/dist-packages/tensorflow/python/lib/io/file_io.py", line 79, in _preread_check
    self.__name, 1024 * 512)
TypeError: __init__(): incompatible constructor arguments. The following argument types are supported:
    1. tensorflow.python._pywrap_file_io.BufferedInputStream(arg0: str, arg1: int)

Invoked with: None, 524288

I traced the error to a pre_reading check for pipeline.config file. But it doesn't seem to be changed at all and all paths are correct.

ravikyram:
@Raphaeal19 

I tried in colab with TF nightly version and was able to convert  tf2 mobilenet-v1 object detection model to tflite model .Please, find the gist [here](https://colab.research.google.com/gist/ravikyram/688464077c636cdbcd71b14a8b806bcf/untitled225.ipynb).Thanks!

Raphaeal19:
This worked!! Thank you so much.

CarrierDiagnostics:
I'm getting the same error when trying to train a custom model with model_main_tf2.py

[...]

PR #9129, opened 2020-08-19
Fixing the command hint for training models

wathek:
Fixing issue #9057


wathek:
@googlebot  I signed it!

PR #9279, opened 2020-09-22
Update config_util.py  in order to fix bugs

x12901:
Fixed the BUG "UnicodeDecodeError:  'utf-8' codec can't decode byte 0xd5 in position 124: invalid continuation byte
- [ ] Bug fix ( change config_util.py)


x12901:
> @googlebot I signed it!

@googlebot I signed it!

PR #9437, opened 2020-10-30
Update 4_Neural_Style_Transfer_with_Eager_Execution.ipynb

MS-092:
# Description
> I made changes for enabling the eager execution, so this code should work for users that are using Tensorflow version 2.x, rather than using the tensorflow 1.x.

## Type of change
- [ ] TensorFlow 2 migration

1. First is that download the file or copy the code
2. Put the code or the file into google colaboratory
3. Change the runtime type into GPU used case
4. Make sure to use tensorflow version 2.x
5. Run All of the code cells

**Test Configuration**: You could use your own personal or local IDE or code editor for testing

result NST


saberkun:
Files moved to tensorflow/examples

Issue #9599, opened 2021-01-01
Installation and runtime errors with the docker setup


neilyoung:
I'm having two issues while building the docker image on MacOS Big Sur 11.1, using Docker Desktop 3.0.3

During the build I'm first getting this trace:

```
Collecting tensorflow>=2.3.0
  Downloading tensorflow-2.4.0-cp36-cp36m-manylinux2010_x86_64.whl (394.7 MB)

```

which obviously reports the download and installation of tf 2.4.0

Later this happens:

```
ERROR: pip's dependency resolver does not currently take into account all the packages that are installed. This behaviour is the source of the following dependency conflicts.
tensorflow-gpu 2.2.0 requires tensorboard<2.3.0,>=2.2.0, but you have tensorboard 2.4.0 which is incompatible.
tensorflow-gpu 2.2.0 requires tensorflow-estimator<2.3.0,>=2.2.0, but you have tensorflow-estimator 2.4.0 which is incompatible.

```

The build finishes nevertheless.

Then, at runtime, this error pops up:

```
tensorflow@b7fbf012734c:~/models/research$ python object_detection/builders/model_builder_tf2_test.py
Traceback (most recent call last):
  File "object_detection/builders/model_builder_tf2_test.py", line 21, in <module>
    import tensorflow.compat.v1 as tf
  File "/home/tensorflow/.local/lib/python3.6/site-packages/tensorflow/__init__.py", line 435, in <module>
    _ll.load_library(_main_dir)
  File "/home/tensorflow/.local/lib/python3.6/site-packages/tensorflow/python/framework/load_library.py", line 153, in load_library
    py_tf.TF_LoadLibrary(lib)
tensorflow.python.framework.errors_impl.NotFoundError: /usr/local/lib/python3.6/dist-packages/tensorflow/core/kernels/libtfkernel_sobol_op.so: undefined symbol: _ZN10tensorflow8OpKernel11TraceStringEPNS_15OpKernelContextEb
tensorflow@b7fbf012734c:~/models/research$ 


```

Any idea, what's wrong?

neilyoung:
That was easy. Created a PR. https://github.com/tensorflow/models/pull/9600

I didn't sign the CLA yet.

saberkun:
tensorflow 2.4 is just released. I guess some assumptions do not hold now.

PR #9600, opened 2021-01-01
Dockerfile update

neilyoung:
Make it TF 2.4.0 ready

Fixing #9599


neilyoung:
@googlebot I signed it!

PR #9688, opened 2021-01-29
Make YAMNet compatible with batch inputs.

MarkDaoust:
# Description

The current YAMnet implementation only allows a single waveform as input.
This change updates it to allow batches of clips. 

The tflite test was already broken when I started. It still breaks the same way.

## Type of change

- [X ] Bug fix (non-breaking change which fixes an issue)
- [X ] New feature (non-breaking change which adds functionality)

## Tests

TODO: Add more texts for the multi-clip outputs. 

## Checklist

- [ ] I have signed the [Contributor License Agreement](https://github.com/tensorflow/models/wiki/Contributor-License-Agreements).
- [ ] I have read [guidelines for pull request](https://github.com/tensorflow/models/wiki/Submitting-a-pull-request).
- [ ] My code follows the [coding guidelines](https://github.com/tensorflow/models/wiki/Coding-guidelines).

## Change notes

Basically, the keras-functional api locks in the dimensionality of the input, making it impossible to implement this with a smaller change. I had to switch this to use keras-subclassing which allows a clear separation between creating the layers and executing the model. I was careful to keep the same layers and ordering so that the keras H5 checkpoint still restores correctly.

+ export.py
  + Fixed the version check at the top.
  + Added multiple signatures with `get_concrete_function`
  + Added code to rebuild the model after testing (b/178785522)
  + Added `signatures=yamnet.single` to `tf.saved_model.save` to be sure the correct signature is set for serving.
  
+ features.py
  + All changes are to generically handle a variable number of outer dimensions. 

+ yamnet.py
  + `YAMNetBase` implements `yamnet .yamnet` and includes code to fold/unfold the batch axis.
  + `YAMNetFrames` implements `yamnet.yamnet_frames_model`.

Is this transformation to the implementation correct?
It looks like for a single clip, each frame is processed independently, so I can just fold the `batch` and `frames` axis together.
These were the shapes for an input `waveforms` of shape `(None,)`:

 
 ```
waveforms (None,)
waveform_padded (None,)
log_mel_spectrogram (None, 64)
features (None, 96, 64)
net (None, 96, 64, 1) # before reshape     shape: (frames, width, height, channels)
net (None, 96, 64, 1) # after reshape

# ... run conv layers ...

embeddings (None, 1024)
logits (None, 521)
predictions (None, 521)
```

These were the shapes for an input `waveforms` of shape `(None,)`:

```
waveforms (None, None)
waveform_padded (None, None)
log_mel_spectrogram (None, None, 64)
features (None, None, 96, 64)
net (None, None, 96, 64, 1) # before flatten      shape: (batch, frames, width, height, channels)
net (None, 96, 64, 1) # after flatten                    shape: (batchframes, width, height, channels)

# ... run conv layers ...

embeddings (None, 1024) # before fold            shape: (batchframes, embedding)
embeddings (None, None, 1024) # after fold    shape: (batch, frames, embedding)
logits (None, None, 521)
predictions (None, None, 521)
```

## TFLite is broken

Switching the assert to a `log` I get the following when testing tflite, with or without this change:

```
=====
Fri Jan 29 08:28:21 2021 | Did not find expected class Silence in top 10 predictions: [('Wail, moan', -6.907737), ('Sigh', -6.907737), ('Singing', -6.907737), ('Choir', -6.907737), ('Yodeling', -6.907737), ('Chant', -6.907737), ('Mantra', -6.907737), ('Child singing', -6.907737), ('Applause', -6.907737), ('Chatter', -6.907737)]
=====


=====
Fri Jan 29 08:28:21 2021 | Did not find expected class White noise in top 10 predictions: [('Hiccup', 2.5337985), ('Fart', 2.5639896), ('Hands', 2.6052508), ('Finger snapping', 2.6293263), ('Clapping', 2.6966517), ('Heart sounds, heartbeat', 2.7180116), ('Heart murmur', 2.767239), ('Cheering', 2.776005), ('Applause', 2.8038406), ('Chatter', 2.84695)]
=====


=====
Fri Jan 29 08:28:21 2021 | Did not find expected class Sine wave in top 10 predictions: [('Narration, monologue', -2.021668), ('Laughter', -1.6973897), ('Whispering', -1.5353699), ('Screaming', -0.5937312), ('Speech synthesizer', -0.43511826), ('Shout', 1.0447013), ('Children shouting', 1.4804856), ('Bellow', 2.7237384), ('Yell', 3.1314607), ('Whoop', 3.5263548)]
=====
```

## TFJs

TFJS seems to be working at head. 
And seems broken by this change:

```
  File "export.py", line 240, in <module>
    main(sys.argv[1:])
  File "export.py", line 237, in main
    make_tfjs_export(tflite_saved_model_dir, tfjs_export_dir)
  File "export.py", line 221, in make_tfjs_export
    tfjs_saved_model_converter.convert_tf_saved_model(
  File "/usr/local/google/home/markdaoust/venv3/lib/python3.8/site-packages/tensorflowjs/converters/tf_saved_model_conversion_v2.py", line 644, in convert_tf_saved_model
    optimize_graph(frozen_graph, signature,
  File "/usr/local/google/home/markdaoust/venv3/lib/python3.8/site-packages/tensorflowjs/converters/tf_saved_model_conversion_v2.py", line 147, in optimize_graph
    raise ValueError('Unsupported Ops in the model before optimization\n' +
ValueError: Unsupported Ops in the model before optimization
BroadcastArgs
```

### CODE REVIEW

**Comment by plakal** [comment]
Ccing @edufonseca for another set of (Keras-experienced) eyes

**Comment by plakal** [comment]
Some initial comments after a first pass through the code. I'm not doing a line-by-line review because I know that you are going to refine this. I know that Edu (cced earlier) has some comments that I will relay from him later. I also want to play with this code a little to see why the TF-JS conversion was throwing unsupported ops for you.

- The general approach seems fine and you are right that it is safe to fold the batch and frame dimensions into a single larger frame dimension because all frames are processed independently.

- I wanted to clarify: you feed in a batch of entire waveforms in and you get out the (framed) log mel spectrogram of each waveform and the embeddings and predictions for individual frames of each waveform. This is fine for visualization of the spectrograms or transfer learning using the embeddings, but if you want to do training with this model, you will also want to plumb out the logits.

- Thanks for finding and fixing the TF-Lite issues. The signature_runner() stuff is slick.

- Instead of separate signatures for single waveform and batch of waveforms, we should probably just do one signature for a batch of waveforms. Are you not generating the batch of waveforms signature in TF-Lite mode because of some unsupported ops issue?

- We could probably simplify or at least clarify the code in YAMNetBase.call() by using fixed known shapes: given that the waveform input is always of shape [batch-size, #samples], then the features tensor is always of shape [batch-size, #frames, #stft-frames, #mel-bands] and you don't have to deal with the general case of multiple batch dimensions and the tf.reshape() calls should be simpler with no calculations. 

- You inherited some confusing names from us so maybe this is a good time to fix them. The 'frames' in yamnet_frames_model and elsewhere are talking about the output of the model being frames of information, which might be confusing. Maybe we could use names like YAMNetFeaturesToFramesModel, YAMNetWaveformToFramesModel, yamnet_waveform_to_frames_model(). They're longer but hopefully clearer.

- In a pre-release version of YAMNet, I did try some in-model batching and unbatching and I remember that I had to use Lambda() to get access to the batch dimension. I'm a little confused by why you only need Lambda() for the initial expand_dims() of features but not for the subsequent reshape or the folding of batch and frame?

- In case it is useful in any way, this is what I was doing in a pre-release version of YAMNet, before DAn replaced it with what we have now.


```
# Variant of Lambda() that has access to the batch dimension and lets us process
# an entire batch at once. This is useful for whole-batch operations, such as
# creating a batch of examples from a single clip, or aggregating a batch of
# example predictions into a single clip prediction.
class BatchLambda(layers.Layer):
  def __init__(self, fn, **kwargs):
    self._fn = fn
    super(BatchLambda, self).__init__(**kwargs)
  def build(self, input_shape):
    super(BatchLambda, self).build(input_shape)
  def call(self, x):
    return self._fn(x)

def yamnet_clip_model():
  clip_waveform = layers.Input(batch_shape=(1, None), dtype='float32')
  features = BatchLambda(lambda x: features_lib.waveform_to_log_mel_examples(tf.squeeze(x, axis=0)),
                         name=params.FEATURES_LAYER_NAME)(clip_waveform)
  predictions = yamnet(features)
  clip_prediction = BatchLambda(lambda x: tf.reduce_mean(x, axis=0, keepdims=True),
                                name=params.CLIP_PREDICTION_LAYER_NAME)(predictions)
  clip_model = Model(name='yamnet_clip', inputs=clip_waveform, outputs=clip_prediction)
  return clip_model

```

**Comment by plakal** [comment]
Also, does your fix to the exporter require TF 2.5? That doesn't seem to be released yet. Can we make it work with TF 2.4? I suppose it's OK is the main code works with the current released version while the exporter might require something more recent, but it would still be nice if the entire tree worked with the current version of TF.

**Comment by plakal** [comment]
I also can't seem to get the code from your repo https://github.com/MarkDaoust/models/tree/yamnet working with either current TF 2.4.1 or nightly TF 2.5.0. Running yamnet_test.py tells me that to_keras() doesn't exist on the frame model object.

```
[  FAILED  ] setUpClass (__main__.YAMNetTest)
======================================================================
ERROR: setUpClass (__main__.YAMNetTest)
----------------------------------------------------------------------
  File "yamnet_test.py", line 34, in setUpClass
    cls._yamnet = yamnet.yamnet_frames_model(cls._params)
  File "/usr/local/google/home/plakal/local/yamnet/batched/models-yamnet/research/audioset/yamnet/yamnet.py", line 205, in yamnet_frames_model
    frames_model = YAMNetFrames(params).to_keras()
AttributeError: 'YAMNetFrames' object has no attribute 'to_keras'

----------------------------------------------------------------------
Ran 0 tests in 0.160s

FAILED (errors=1)
```

MarkDaoust:
Thanks @plakal,

So I think one blocking question right now is: 

* Does the new model need to be backwards compatible for old users? Or should we only add new signatures and not break existing ones?

That makes a big difference for this change (a lot of the added complexity is coming from the fact that keras-functional models don't support multiple signatures).

It sounds like you're leaning towards breaking existing usage. That's fine with me, I just want to make sure it's clear that that's what we're doing.

WDYT?

plakal:
@MarkDaoust sorry for not replying earlier, got swamped with stuff.

To answer your earlier question, we don't care too much about backward compatibility of the code in the repo but we do care about backward compatibility of the model checkpoint (so weights and tensor names can't change) and of the exports that we have made and uploaded to TF-Hub (so signatures can't change unless we add them as new models).

I haven't taken a look at the latest state of this PR yet but I just wanted to give you a heads-up that we are planning on making some small modifications to the YAMNe model code ourselves. We want a version of the model for some upcoming TF-Lite demos where the model takes a single frame of audio in and produces only classifier scores (in addition to some other changes that make the model friendlier to quantization).

Given this and your PR, maybe it would be better if we organized the model code into
- the core workhorse model could be one that takes a batch of audio frames and returns the various per-frame outputs as needed (features, embeddings, logits, scores). This would look a lot like the existing code.
- another model could take a batch of clips, make frames of audio from the clips, call the core model above, maybe do some unbatching or aggregation on the results. So all of your stuff could go into this model.

This gives us a lot of flexibility for various uses:
- if you want to feed a single clip for inference, you can call the clip-input model with a batch of 1 clip
- if you want to feed a single frame for inference, you can call the frame-input model with a batch of 1 frame
- if you want to feed a batch of clips for training, you can call the clip-input model with a batch of clips
- if you want to do your own framing and shuffling of clips, you can create a batch of frames and call the frame-input model with a batch of frames

How does that sound?

MarkDaoust:
Okay, so I'm back to looking at this. I've fixed the issues I was having earlier with TFLite, TFJS and tests.

So let me go through the points here one at a time.

 > [Feb 5]  if you want to do training with this model, you will also want to plumb out the logits.
 
 Training from scratch, yes.
 
 I was just interested in getting transfer leaning + fine tuning to work and for that we build our own classification head. 
 
 That would be more efficient if we include signatures that don't return the original predictions because AFAIK TensorFlow can't optimize away unused outputs across a `tf.function` call.
 
 > [Feb 5]  Thanks for finding and fixing the TF-Lite issues. The signature_runner() stuff is slick.
 
 > [Feb 5] Also, does your fix to the [TFlite] exporter require TF 2.5? That doesn't seem to be released yet. Can we make it work with TF 2.4? 
 
Yes. It requires TF2.5, that release is just around the corner now.

> [Feb 5] Are you not generating the batch of waveforms signature in TF-Lite mode because of some unsupported ops issue?

I was just taking the conservative approach, my feature doesn't involve lite so I tried to keep lite exactly as it was.
I haven't tried this. Is that something that would be useful?

> [Feb 5] You inherited some confusing names.

Added a TODO.

> [Feb 5] In a pre-release version of YAMNet, I did try some in-model batching and unbatching and I remember that I had to use Lambda() to get access to the batch dimension. I'm a little confused by why you only need Lambda() for the initial expand_dims() of features but not for the subsequent reshape or the folding of batch and frame?

First I think I have encountered "I had to use Lambda() to get access to the batch dimension" I think it was that certain ops are not available when running in the "Keras Functional" graph building mode. I don't need the lambda because I'm working in a model subclass, which is just regular tensorflow code.

For the question of "why you only need Lambda() for the initial expand_dims()". The `Lamdda` is just to maintain correspondence between the old and new code between the old and new layers. IIUC you need a 1:1 correspondence for the checkpoint to load. But since this is not a weight-bearing layer a `Lambda(lambda x:x)` would work if we need to move the reshape code.

> [Apr 8] To answer your earlier question, we don't care too much about backward compatibility of the code in the repo but we do care about backward compatibility of the model checkpoint (so weights and tensor names can't change).

Yes. This change preserves checkpoint compatibility. Your tests pass. I've added a TODO to do a before/after test with `numpy.allclose`.

> [Feb 5] Instead of separate signatures for single waveform and batch of waveforms, we should probably just do one signature for a batch of waveforms. 

> [Apr 8] and of the exports that we have made and uploaded to TF-Hub (so signatures can't change unless we add them as new models).

The two comments above seem in conflict with each other. Right now I have it exporting both signatures to hub so that we don't break compatibility with current users.

I haven't tested it but at this point the model may even work with an unknown input rank (`None` not `[None]`) 

"unless we add them as new models" - Note that TFHub models are versioned and if we make changes we can upload it as `yamnet/2`

> [Apr 8] Given this and your PR, maybe it would be better if we organized the model code into: ...

That sounds reasonable, It's really not far from what we have right now. I tink we'd just need to move the boundaries a little.  Currently:

* `export.YamNet` - Manages signatures and `tf.function` it calls `yamnet_lib.YAMNetFrames`
* `YAMNetFrames` -  Does the frame-splitting and log-mel-spectrogram extraction, and then calls `YAMNetBase`.
* `YAMNetBase` - Flattens `(batches, frames, ...)` into `(batches_and_frames, ...)` runs the model and un-flattens the result. 

As long as we don't move any weight-bearing layers that should work fine with the existing checkpoint.

I'll see what I can do.

MarkDaoust:
> another model could take a batch of clips, make frames of audio from the clips, call the core model above, maybe do...

Curently the clip-to-frames code is all hidden in the `features.waveform_to_log_mel_spectrogram_patches` function. This idea would move clip-to-frames before that.

That function converts `audio -> magnitude_spectrogram -> mel_spectrogram -> log_mel_spectrogram` and then it runs the  `tf.signal.frame(signal=log_mel_spectrogram)`. I don't know all the details abut the intermediates.

IDK if this sequence of operations commutative wrt the clip-to-frames conversion.

But with the `.npz` test I can try it and see if anything changes.

MarkDaoust:
Okay, that seems to work.

plakal:
Mark, I'll have some time early next week to take a look at this PR again in detail, if that's all right and if this is ready for another look. Thanks for all your work on this batching feature.

Also, when I last tried running the PR with TF 2.4 and a TF 2.5 release candidate, I saw some errors (see one of my previous comments). Let me know if there are any special instructions for running the code. As far as possible, we would prefer that the main code run with TF 2.4 to maximize compatibility. It's fine if the exporter uses later code since most people won't be running that.

MarkDaoust:
Yeah, I think I'm done.

The new TFLite export logic requires TF 2.5 but nothing else does.

For me `yamnet_test.py` passes now in both TF2.4 and 2.5.

PR #8846, opened 2020-07-12
Add correct MobileNet links

kyscg:
# Description

The links in the README file for Mobilenet were wrong. Fixes #8845 

## Type of change

- [x] Bug fix (non-breaking change which fixes an issue)
- [x] Documentation update

## Checklist

- [x] I have signed the [Contributor License Agreement](https://github.com/tensorflow/models/wiki/Contributor-License-Agreements).
- [x] I have read [guidelines for pull request](https://github.com/tensorflow/models/wiki/Submitting-a-pull-request).
- [x] My code follows the [coding guidelines](https://github.com/tensorflow/models/wiki/Coding-guidelines).
- [x] I have performed a self [code review](https://github.com/tensorflow/models/wiki/Code-review) of my own code.
- [x] My changes generate no new warnings.

PR #9692, opened 2021-01-31
Fix MobileNet download links

dthul:
# Description

Fix download links for the MobileNet V1 and V2 checkpoints.

## Type of change

- [X] Documentation update


kyscg:
Copy of #8846

PR #9709, opened 2021-02-06
Fix learning rate scheduler for BERT models

twilightdema:
# Description

**Root Cause**
Fix wrong learning rate scheduling for BERT models.
The root cause is that the PolynomialDecay starts the decay from first training step, where the warm up period may still be in effect. So then warm up period ends, the learning rate is at the highest value, the PolynomialDecay kicks in and the effective learning rate is lower than it should be at that step because it has alredy started decaying from the first step.

This fix is to ensure that we start the decay at the end of warm up period and end the decay correctly at the last training step with the correct target value.

## Type of change

- [ /] Bug fix (non-breaking change which fixes an issue)

## Tests

**Test Configuration**:

Testing steps:

Calling the **official.nlp.optimization.create_optimizer** with the parameters below and plot effective learning rate:
  init_lr = 0.005,
  num_train_steps = 40000,
  num_warmup_steps = 10000,
  end_lr=0.0,
  optimizer_type='adamw'

You can see the result that the effective learning rate has the discontinued value when polynomial decay starts working:
image

After this fix, the effective learning rate will be like below:
image

## Checklist

- [ /] I have signed the [Contributor License Agreement](https://github.com/tensorflow/models/wiki/Contributor-License-Agreements).
- [ /] I have read [guidelines for pull request](https://github.com/tensorflow/models/wiki/Submitting-a-pull-request).
- [ /] My code follows the [coding guidelines](https://github.com/tensorflow/models/wiki/Coding-guidelines).
- [ /] I have performed a self [code review](https://github.com/tensorflow/models/wiki/Code-review) of my own code.
- [ /] I have commented my code, particularly in hard-to-understand areas.
- [ /] I have made corresponding changes to the documentation.
- [ /] My changes generate no new warnings.
- [ /] I have added tests that prove my fix is effective or that my feature works.

### DIFF
--- a/official/nlp/optimization.py
+++ b/official/nlp/optimization.py
@@ -51,7 +51,7 @@ class WarmUp(tf.keras.optimizers.schedules.LearningRateSchedule):
       return tf.cond(
           global_step_float < warmup_steps_float,
           lambda: warmup_learning_rate,
-          lambda: self.decay_schedule_fn(step),
+          lambda: self.decay_schedule_fn(step - self.warmup_steps),
           name=name)
 
   def get_config(self):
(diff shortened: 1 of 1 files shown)


saberkun:
@twilightdema Thanks! Yes, we know this diff. The original bert has implemented the warmup in this way. We have experimented both and did not find quality difference. Note that, we have implemented warmup schedule in general inside the official/modeling/optimization package, which is combining the original decay lr schedule with the warmup, so the peak is between 0.003 and 0.004 if it follows your graph.

PR #9899, opened 2021-04-13
File movings

ykate1998:
# Description
 
Adding the CNN Image Retrieval model implementation based on the following papers:

[1] Fine-tuning CNN Image Retrieval with No Human Annotation,
Radenović F., Tolias G., Chum O., TPAMI 2018 [arXiv]
https://arxiv.org/abs/1711.02512

[2] CNN Image Retrieval Learns from BoW: Unsupervised Fine-Tuning with Hard
Examples, Radenović F., Tolias G., Chum O., ECCV 2016 [arXiv]
https://arxiv.org/abs/1604.02426

## Type of change

Note: Please delete options that are not relevant.

- [x] New feature (non-breaking change which adds functionality)
- [x] A new research paper code implementation

## Tests

Tests for the added modules are provided.

**Test Configuration**:

Same requirements as for the DELF library.


research/delf/delf/__init__.py
```diff
@@ -35,5 +35,4 @@
 from delf.python import detect_to_retrieve
 from delf.python import training
 from delf.python.training import model
-from delf.python.training import datasets
```
andrefaraujo, line 38: just wanted to double-check: after doing this change, the code continues to work, right? I thought some import like this in the top init could still be needed.

research/delf/delf/python/pooling_layers/pooling.py
```diff
@@ -19,19 +19,16 @@
 
 class MAC(tf.keras.layers.Layer):
   """Global max pooling (MAC) layer.
-
```
andrefaraujo, line 22: there are many blank-line removals in this file, can you please revert those?

andrefaraujo (changes requested):
Thanks Kate! Left some comments. Also, one question: for this PR, would it be possible/easy to remove the files "global_features_utils.py", "whiten.py" and "whiten_test.py"? They are unrelated to file movings, so this PR would be cleaner / simpler without them.

research/delf/delf/python/pooling_layers/pooling.py
ykate1998: Thank you for pointing this out, must have missed some code formatting issue. Should be fixed now

research/delf/delf/__init__.py
ykate1998: The code was running, but I now added a new updated import just to make sure it works the same way

ykate1998:
Deleted the unnecessary modules you mentioned. Thank you :)

andrefaraujo (approved):
Thanks!

PR #9902, opened 2021-04-13
Global model

ykate1998:
# Description

Adding the CNN Image Retrieval model implementation based on the following papers:

[1] Fine-tuning CNN Image Retrieval with No Human Annotation,
Radenović F., Tolias G., Chum O., TPAMI 2018 [arXiv]
https://arxiv.org/abs/1711.02512

[2] CNN Image Retrieval Learns from BoW: Unsupervised Fine-Tuning with Hard
Examples, Radenović F., Tolias G., Chum O., ECCV 2016 [arXiv]
https://arxiv.org/abs/1604.02426

## Type of change

- [x] New feature (non-breaking change which adds functionality)
- [x] A new research paper code implementation

PR #9904, opened 2021-04-13
Global model training

ykate1998:
# Description

Adding the CNN Image Retrieval model implementation based on the following papers:

[1] Fine-tuning CNN Image Retrieval with No Human Annotation,
Radenović F., Tolias G., Chum O., TPAMI 2018 [arXiv]
https://arxiv.org/abs/1711.02512

[2] CNN Image Retrieval Learns from BoW: Unsupervised Fine-Tuning with Hard
Examples, Radenović F., Tolias G., Chum O., ECCV 2016 [arXiv]
https://arxiv.org/abs/1604.02426

## Type of change

- [x] A new research paper code implementation


dan-anghel (approved):

andrefaraujo:
I will review this PR in the upcoming days, and get back to you @ykate1998 :)
In the meantime, if this PR needs to be updated due to my recent merging of #9930 , feel free to do it.

ykate1998:
Thank you, I will let you know when I fixed it

andrefaraujo:
Ok, sounds good, then I will wait for your fixes before reviewing :)

ykate1998:
@andrefaraujo, if I didn't miss anything, should be fixed now. Thank you

research/delf/delf/python/training/global_features/train.py
```diff
@@ -0,0 +1,719 @@
...
+from absl import flags
+from absl import app
+from absl import logging
+
+from delf.python.datasets import tuples_dataset
+from delf.python.datasets.revisited_op import dataset as testdataset
+from delf.python.training.losses import ranking_losses
+from delf.python.training import global_features_utils
+from delf.python.training import tensorboard_utils
+from delf.python.training.model import global_model
```
andrefaraujo, line 36: I noticed the global_model part is not already checked in. Can you point to me which PR it's in, so I can review it? Sorry if I missed here :)

research/delf/delf/python/training/global_features/train.py
```diff
@@ -0,0 +1,719 @@
...
+  global_features_utils.debug_and_log('>> Network initialized.')
+
+  global_features_utils.debug_and_log('>> Loss: {}.'.format(FLAGS.loss))
+  # Define the loss function.
+  if FLAGS.loss == 'contrastive':
+    criterion = ranking_losses.ContrastiveLoss(margin=FLAGS.loss_margin)
+  elif FLAGS.loss == 'triplet':
+    criterion = ranking_losses.TripletLoss(margin=FLAGS.loss_margin)
+  else:
+    raise RuntimeError('Loss {} not available!'.format(FLAGS.loss))
```
andrefaraujo, line 220: I guess ValueError would be the most suitable choice

research/delf/delf/python/training/global_features/train.py
```diff
@@ -0,0 +1,719 @@
...
+
+  # Defining parameters for the training.
+  start_epoch = 1
```
andrefaraujo, line 223: nit: is this variable 1-indexed? Maybe it should start at 0?

research/delf/delf/python/training/global_features/train.py
```diff
@@ -0,0 +1,719 @@
...
+    opt = tfa.optimizers.extend_with_decoupled_weight_decay(
+            tf.keras.optimizers.SGD)
+    optimizer = opt(weight_decay=FLAGS.weight_decay,
+                    learning_rate=lr_scheduler, momentum=FLAGS.momentum)
+  elif FLAGS.optimizer == 'adam':
+    opt = tfa.optimizers.extend_with_decoupled_weight_decay(
+            tf.keras.optimizers.Adam)
+    optimizer = opt(weight_decay=FLAGS.weight_decay,
+                    learning_rate=lr_scheduler)
+
```
andrefaraujo, line 244: else, raise a valueerror as well for optimizer

research/delf/delf/python/training/global_features/train.py
andrefaraujo: same for sfm120k :)

research/delf/delf/python/training/global_features/train.py
```diff
@@ -0,0 +1,719 @@
...
+      tf.random.set_seed(epoch)
+
+      # Find hard-negatives.
+      # While hard-positive examples are fixed during the whole training
+      # process and are randomly chosen from every epoch; hard-negatives
+      # depend on the current CNN parameters and are re-mined once per epoch.
+      avg_neg_distance = train_dataset.create_epoch_tuples(model,
+                                                           model_directory)
+
+      def train_gen():
```
andrefaraujo, line 338: add leading underscore

research/delf/delf/python/training/global_features/train.py
```diff
@@ -0,0 +1,719 @@
...
+              name=FLAGS.training_dataset,
+              data_root=FLAGS.data_root,
+              mode='val',
+              imsize=FLAGS.image_size,
+              nnum=FLAGS.neg_num,
+              qsize=float('Inf'),
+              poolsize=float('Inf')
+      )
+
+  output_types = [tf.float32 for i in range(2 + FLAGS.neg_num)]
```
andrefaraujo, line 311: "output_types" -- the variable naming could be improved... "output_types" of what?

research/delf/delf/python/training/global_features/train.py
```diff
@@ -0,0 +1,719 @@
...
+                                'checkpoint_epoch_{}.h5'.format(epoch))
+        model.save_weights(filename, save_format='h5')
+        global_features_utils.debug_and_log(
+                'Saved weights ({}) at {}'.format(epoch, filename))
+      except Exception as ex:
+        global_features_utils.debug_and_log(
+                'Could not save checkpoint: {}'.format(ex))
+
+
+def grad(criterion, model, input, target):
```
andrefaraujo, line 393: for this and the following functions: let's maybe move them into a separate file? This script is already really long. It could be something like "train_utils.py". I will take a look at the functions once you move them.

andrefaraujo (changes requested):
Thanks Kate! Please take a look at my initial pass of comments :)

I noticed there may be some formatting issues -- please make sure to run the pylint tool we had discussed before.

research/delf/delf/python/training/global_features/train.py
```diff
@@ -0,0 +1,719 @@
...
+  global_features_utils.debug_and_log('>> Num GPUs Available: {}'.format(
+          len(tf.config.experimental.list_physical_devices('GPU'))),
+          FLAGS.debug)
+
+  # Allow memory growth.
+  gpus = tf.config.experimental.list_physical_devices('GPU')
+  if gpus:
+    try:
+      for gpu in gpus:
+        tf.config.experimental.set_memory_growth(gpu, True)
```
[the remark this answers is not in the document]
ykate1998: It allows to see how much GPU memory is actually occupied by the code because under default conditions tensorflow takes up all of the available GPU memory.

research/delf/delf/python/training/global_features/train.py
ykate1998: @andrefaraujo, here it is:
a) Dataset related modules: https://github.com/tensorflow/models/pull/9903
b) Global model definition: https://github.com/tensorflow/models/pull/9902
c) Global model training: https://github.com/tensorflow/models/pull/9904

research/delf/delf/python/training/global_features/train.py
ykate1998: Sometimes (for example when pre-computing whitening) we run evaluation before the network training and then we set the epoch to 0. In other cases, we start from 1. Should I change that?

ykate1998:
Thank you, @andrefaraujo ,  I have fixed the issues you mentioned

research/delf/delf/python/training/global_features/train.py
```diff
@@ -0,0 +1,719 @@
...
+                         ' | '.join(_TEST_DATASET_NAMES) +
+                         ' (default: roxford5k,rparis6k).')
+flags.DEFINE_enum('precompute_whitening', default=None,
+                  enum_values=_PRECOMPUTE_WHITEN_NAMES,
+                  help='Dataset used to learn whitening: ' +
+                       ' | '.join(_PRECOMPUTE_WHITEN_NAMES) +
+                       ' (default: None).')
+flags.DEFINE_integer('test_freq', default=5,
+                     help='Run test evaluation every N epochs (default: 1).')
+flags.DEFINE_string('multiscale', default='[1.]',
```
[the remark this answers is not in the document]
andrefaraujo: I think in this case the examples you give in the `help` section of the flag would need to be changed, since now it should be input as a comma-separated string?

research/delf/delf/python/training/global_features/train.py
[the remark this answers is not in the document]
andrefaraujo: Got it, thanks for explaining! I saw you removed it, but feel free to keep in case you think it's helpful :)

research/delf/delf/python/training/global_features/train.py
andrefaraujo: Thanks! Will look at them in this order.

research/delf/delf/python/training/global_features/train.py
andrefaraujo: No, I think it should be fine. Maybe you can explain this in the comments / descriptions (and later in the README for using this training script)

research/delf/delf/python/training/global_features/train_utils.py
```diff
@@ -0,0 +1,368 @@
...
+    negatives = tf.boolean_mask(output, target == 0, axis=0)
+    negatives = tf.reshape(negatives, [tf.shape(queries)[0], neg_num,
+                                       model.meta['outputdim']])
+    # Loss calculation.
+    loss = criterion(queries, positives, negatives)
+
+  return loss, tape.gradient(loss, model.trainable_variables)
+
+
+def train_val(loader, model, criterion, optimizer, epoch, train=True,
```
andrefaraujo, line 66: reading this fn, I believe it is doing training (or validation) for one entire epoch, right? If so, maybe the name could be changed to `train_val_one_epoch`

research/delf/delf/python/training/global_features/train_utils.py
```diff
@@ -0,0 +1,368 @@
...
+        if (batch_num + 1) % update_every == 0 or (
+                batch_num + 1) == all_batch_num:
+          # Do one step for multiple batches. Accumulated gradients are
+          # used.
+          optimizer.apply_gradients(
+                  zip(accum_grads, model.trainable_variables))
+          accum_grads = [
+            tf.Variable(tf.zeros_like(tv.read_value()), trainable=False)
+            for tv in tvs]
+      except Exception as ex:
```
andrefaraujo, line 140: We shouldn't use such catch-all exceptions. From the style guide: "Never use catch-all except: statements, or catch Exception or StandardError, unless you are" (https://google.github.io/styleguide/pyguide.html#244-decision)

I am wondering if this is really needed here, or if it could be removed. When do exceptions happen in this case?

research/delf/delf/python/training/global_features/train_utils.py
```diff
@@ -0,0 +1,368 @@
...
+      input = []
+      target = []
+      try:
+        for _ in range(batch_size):
+          # Because the images are not necessarily of the same size, we can't
+          # set the batch size with .batch().
+          batch = loader.get_next()
+          input.append(batch[0:-1])
+          target.append(batch[-1])
+      except Exception as ex:
```
andrefaraujo, line 156: same exception comment as above

research/delf/delf/python/training/global_features/train_utils.py
```diff
@@ -0,0 +1,368 @@
...
+      output = tf.zeros(shape=(0, model.meta['outputdim']), dtype=tf.float32)
+      for input_tuple in input:
+        for img in input_tuple:
+          # Compute the global descriptor vector.
+          model_out = model(tf.expand_dims(img, axis=0), training=False)
+          output = tf.concat([output, model_out], 0)
+
+      # No need to reduce memory consumption (no backward pass):
+      # Compute loss for the full batch.
+      tmp_target = tf.concat(target, axis=0)
```
andrefaraujo, line 169: trying to understand, is `tmp_target` needed at all? Why can't we just use `target` directly? Sorry if I missed sthing.

research/delf/delf/python/training/global_features/train_utils.py
```diff
@@ -0,0 +1,368 @@
...
+      queries = tf.boolean_mask(output, tmp_target == -1, axis=0)
```
andrefaraujo, line 170: I think we may not need tf.boolean_mask? For example, I think we could maybe just use
output[tmp_target == -1]
?

The following seems to work -- so I think it should work here too?
a = tf.expand_dims(tf.range(5), 0)
b = tf.expand_dims(tf.range(5,10), 0)
c = tf.concat([a,b], axis=0)
c
<tf.Tensor: shape=(2, 5), dtype=int32, numpy=
   array([[0, 1, 2, 3, 4],
       [5, 6, 7, 8, 9]], dtype=int32)>
d = tf.constant([-1, 1])
c[d>0]
 <tf.Tensor: shape=(1, 5), dtype=int32, numpy=array([[5, 6, 7, 8, 9]], dtype=int32)>

research/delf/delf/python/training/global_features/train_utils.py
```diff
@@ -0,0 +1,368 @@
...
+    target: List of indexes to specify queries (-1), positives(1), negatives(0).
+    neg_num: Integer, number of negatives per a tuple.
+
+  Returns:
+    loss: Loss for the training step.
+    gradients: Computed gradients for the network trainable variables.
+  """
+  # Record gradients and loss through the network.
+  with tf.GradientTape() as tape:
+    output = tf.Variable(
```
andrefaraujo, line 48: similar to comment below, I think there is no need to make this a tf.Variable? (we could use the tf.zeros tensor directly?)

andrefaraujo, line 48: maybe rename `output` to `descriptors` would make it easier to understand. Here in particular, since the fn is computing the gradients, I was expecting output to be gradients.

research/delf/delf/python/training/global_features/train_utils.py
```diff
@@ -0,0 +1,368 @@
...
+          # Because the images are not necessarily of the same size, we can't
+          # set the batch size with .batch().
+          batch = loader.get_next()
+          input.append(batch[0:-1])
+          target.append(batch[-1])
+      except Exception as ex:
+        global_features_utils.debug_and_log(ex)
+        break
+
+      output = tf.zeros(shape=(0, model.meta['outputdim']), dtype=tf.float32)
```
andrefaraujo, line 160: similar to above, maybe let's rename `output` to `descriptors`

research/delf/delf/python/training/global_features/train_utils.py
```diff
@@ -0,0 +1,368 @@
...
+  # Record gradients and loss through the network.
+  with tf.GradientTape() as tape:
+    output = tf.Variable(
+            tf.zeros(shape=(0, model.meta['outputdim']), dtype=tf.float32))
+    for img in input:
+      # Compute descriptor vector for each image.
+      o = model(tf.expand_dims(img, axis=0), training=True)
+      output = tf.concat([output, o], 0)
+
+    queries = tf.boolean_mask(output, target == -1, axis=0)
```
andrefaraujo, line 55: see comments in the below function about boolean_mask. I am guessing that simply using output[target==-1] would work.

research/delf/delf/python/training/global_features/train_utils.py
```diff
@@ -0,0 +1,368 @@
...
+import tensorflow as tf
+
+from delf.python import whiten
+from delf.python.datasets.revisited_op import dataset as test_dataset
+from delf.python.datasets.sfm120k import sfm120k
+from delf.python.training import global_features_utils
+from delf.python.training.model import global_model
+
+
+def grad(criterion, model, input, target, neg_num=5):
```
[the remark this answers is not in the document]
andrefaraujo: Also, maybe to make more clear what it's doing, what about renaming it to _compute_loss_and_gradient? The current name seems to imply that only gradients are computed.

research/delf/delf/python/training/global_features/train_utils.py
```diff
@@ -0,0 +1,368 @@
...
+              '(Loss avg: {loss.avg:.4f})'.format(
+                      state, epoch, batch_num + 1, all_batch_num,
+                      batch_time=batch_time,
+                      data_time=data_time, loss=losses), debug=True, log=False)
+    batch_num += 1
+
+  return losses.avg
+
+
+def test(datasets, net, epoch, writer=None, model_directory=None,
```
andrefaraujo, line 203: what about a more specific name? Like, test_retrieval? This would distinguish it from the validation step above which is about loss computation (which a reader could also intuitively think would be the case here)

research/delf/delf/python/training/global_features/train_utils.py
```diff
@@ -0,0 +1,368 @@
...
+
+    # If whitening already precomputed, load it and skip the computations.
+    filename = os.path.join(
+            model_directory, 'learned_whitening_mP_{}_epoch.pkl'.format(epoch))
+    filename_layer = os.path.join(
+            model_directory,
+            'learned_whitening_layer_config_{}_epoch.pkl'.format(
+                    epoch))
+
+    if os.path.isfile(filename):
```
andrefaraujo, line 241: similar to before, please use tf.io.gfile.exists

research/delf/delf/python/training/global_features/train_utils.py
```diff
@@ -0,0 +1,368 @@
...
+
+    else:
+      start = time.time()
+      global_features_utils.debug_and_log(
+              '>> {}: Learning whitening...'.format(precompute_whitening))
+
+      # Loading db.
+      db_root = os.path.join(data_root, 'train', precompute_whitening)
+      ims_root = os.path.join(db_root, 'ims')
+      db_fn = os.path.join(db_root,
```
andrefaraujo, line 256: nit: fn -> filename

research/delf/delf/python/training/global_features/train_utils.py
```diff
@@ -0,0 +1,368 @@
...
+                           '{}-whiten.pkl'.format(precompute_whitening))
+      with tf.io.gfile.GFile(db_fn, 'rb') as f:
+        db = pickle.load(f)
+      images = [sfm120k.id2filename(db['cids'][i], ims_root) for i in
```
andrefaraujo, line 260: I see that this seems to be specific to the sfm120k datasets. So let's mention it in the docstring that only such whitening datasets are supported?

research/delf/delf/python/training/global_features/train_utils.py
```diff
@@ -0,0 +1,368 @@
...
+      precomputation required, then `None`.
+    data_root: Absolute path to the data folder.
+    multiscale: List of scales for multiscale testing.
+  """
+  global_features_utils.debug_and_log(">> Testing step:")
+  global_features_utils.debug_and_log(
+          '>> Evaluating network on test datasets...')
+
+  # For testing we use image size of max 1024.
+  image_size = 1024
```
andrefaraujo, line 228: generally it's better to avoid such hard-codings. Can we maybe move this into an argument with default 1024?

research/delf/delf/python/training/global_features/train_utils.py
```diff
@@ -0,0 +1,368 @@
...
+              dataset + ' + whiten', ranks, cfg['gnd'])
+
+      if writer:
+        tf.summary.scalar('test_accuracy_whiten_{}_E'.format(dataset),
+                          metrics[0][0], step=epoch)
+        tf.summary.scalar('test_accuracy_whiten_{}_M'.format(dataset),
+                          metrics[1][0], step=epoch)
+        tf.summary.scalar('test_accuracy_whiten_{}_H'.format(dataset),
+                          metrics[2][0], step=epoch)
+        writer.flush()
```
andrefaraujo, line 364: seems basically duplicated code from 328-341. Maybe we could create a small auxiliary function to avoid this / simplify things here.

research/delf/delf/python/training/global_features/train_utils.py
ykate1998: This is a stopping condition for loader.get_next()
We get a tensorflow.python.framework.errors_impl.OutOfRangeError: End of sequence
And then we know that we iterated through all of the pictures

ykate1998:
@andrefaraujo , I fixed the problems, thank you

research/delf/delf/python/training/global_features/train_utils.py
```diff
@@ -63,9 +63,10 @@ def grad(criterion, model, input, target, neg_num=5):
   return loss, tape.gradient(loss, model.trainable_variables)
 
 
-def train_val(loader, model, criterion, optimizer, epoch, train=True,
-              batch_size=5, query_size=2000, neg_num=5, update_every=1,
-              debug=False):
+def train_val_one_epoch(loader, model, criterion, optimizer, epoch, train=True,
+                        batch_size=5, query_size=2000, neg_num=5,
+                        update_every=1,
+                        debug=False):
```
andrefaraujo, line 69: please fix indentation here

research/delf/delf/python/training/global_features/train_utils.py
andrefaraujo: sounds good, thanks! Can you please add a comment to mention exactly this? Eg, something like "We break when we run out of range, ie, we exhausted all dataset images."

research/delf/delf/python/training/global_features/train_utils.py
andrefaraujo: just curious, did you try this?

research/delf/delf/python/training/global_features/train_utils.py
andrefaraujo: just curious, did you try this?

research/delf/delf/python/training/global_features/train_utils.py
```diff
@@ -326,43 +322,60 @@ def test(datasets, net, epoch, writer=None, model_directory=None,
...
-      if writer:
-        tf.summary.scalar('test_accuracy_whiten_{}_E'.format(dataset),
-                          metrics[0][0], step=epoch)
-        tf.summary.scalar('test_accuracy_whiten_{}_M'.format(dataset),
-                          metrics[1][0], step=epoch)
-        tf.summary.scalar('test_accuracy_whiten_{}_H'.format(dataset),
-                          metrics[2][0], step=epoch)
-        writer.flush()
+      _calculate_metrics_and_export_to_tensorboard(
+              vecs_lw, qvecs_lw, dataset, cfg, writer, epoch, whiten=True)
```
andrefaraujo, line 337: hmmm, do you need to call this with `dataset + ' + whiten'` instead of simply `dataset`?
Or, as you have done below, add the `' + whiten'` in the new function below, depending on the value of the `whiten` argument.

andrefaraujo (changes requested):
Thanks, Kate! See my small comments left, we're almost done. I'll merge the code once you address them.

research/delf/delf/python/training/global_features/train_utils.py
andrefaraujo: let me highlight this one as well, wondering if this could work. @ykate1998

research/delf/delf/python/training/global_features/train_utils.py
andrefaraujo: and here as well :) @ykate1998

research/delf/delf/python/training/global_features/train_utils.py
ykate1998: It should be working without " + whiten". The " + whiten" part in the previous code iteration was used for the naming of the tensorboard variable. Now, we overcome the need of having it by the `whiten=True` in the function, which results in choosing suitable metric names (i.e., 'test_accuracy_whiten_{}_E' instead of just 'test_accuracy_{}_E').

research/delf/delf/python/training/global_features/train_utils.py
ykate1998: Thank you for the comment, I must have overlooked it last time.

research/delf/delf/python/training/global_features/train_utils.py
ykate1998: Thanks, it works, I must have overlooked it last time.

ykate1998:
@andrefaraujo, thank you! If I didn't overlook anything, I think I fixed the problems in this PR

research/delf/delf/python/training/global_features/train_utils.py
andrefaraujo: Right, but I was talking about the line `metrics = global_features_utils.compute_metrics_and_print(dataset, ranks,
                                                            cfg['gnd'])`

Before, you had `dataset + ' + whiten'` instead of `dataset` in the above line, when you were logging the whitening case.

Anyway, it's not a big issue at all, it will I guess just log things slightly differently. Feel free to ignore this, or change the code to follow the same behavior as before -- as you prefer. Feel free to resolve this comment as well.

andrefaraujo (approved):
Thanks Kate!

andrefaraujo:
@ykate1998 Can you please fix the conflict? I'll merge the PR once you fix it. Thanks :)

ykate1998:
@andrefaraujo, fixed the conflict. Thank you

andrefaraujo:
Thanks Kate. However, shouldn't you keep using `from delf.python.datasets.revisited_op import dataset as revisited_dataset`? Seems like you removed it, but I guess it would be necessary, otherwise it could conflict with the variable `dataset`. Similarly, you had renamed the `dataset` variable to `dataset_name`, which was better, but it seems like you are undoing this now.

So overall I am thinking that you should not undo these changes? But please let me know in case I am missing something.

This comment applies to both this and the other open PR (#9902).

andrefaraujo (approved):
Thanks Kate!

PR #9914, opened 2021-04-15
Added instructions to convert the frozen graph into a TFLite model

uzair-ali10:
Added instructions to convert the frozen graph into a TFLite model


uzair-ali10:
@jch1 @pkulzc @tombstone 
Kindly review this PR Its been on hold for a long time ;/

Issue #9880, opened 2021-04-08
Bug in centernet object detection


ApproximateIdentity:
I'm running into a problem when going through this file:

https://github.com/sglvladi/TensorFlowObjectDetectionTutorial/blob/master/docs/source/auto_examples/plot_object_detection_checkpoint.rst

The plain English summary of the problem is the following: Going through that example fails when predicting on images. It throws an error about the tensor shapes being incorrect. I tracked down the commit in the tensorflow/models repo which causes this issue (see below), but I don't know how to adapt this example to the new repository. So basically I see at least two possibilities:

1. Is this a bug with tensorflow/models? In that case, it should be fixed here in this repo.
2. Is it something that should be changed in the tutorial example above? In that case, what needs changing? I'm new to tensorflow so I'm a bit at a loss...

## General description and system information

While going through this tutorial I ran into an issue:

https://tensorflow-object-detection-api-tutorial.readthedocs.io/en/latest/auto_examples/plot_object_detection_checkpoint.html

I am not sure if it is a bug in this repository or not (though it must be a bug in either this repository or the one with the tutorials). I have stripped down a simplified python script to exhibit the issue. I am also providing a script to build the necessary environment (at least on a system similar to my debian 11/testing/bullseye system). I have no GPU. My python was downloaded from the official site and compiled by myself. Here is some system information:

```
$ lsb_release -a
No LSB modules are available.
Distributor ID:	Debian
Description:	Debian GNU/Linux bullseye/sid
Release:	testing
Codename:	bullseye
$ python3.8 -V
Python 3.8.8
```

## Reproducing the necessary environment

Running the following bash script should rebuild the environment necessary to reproduce the bug:

`source_me.sh`

```bash
mkdir root && cd root

python3.8 -m venv venv
source venv/bin/activate

# Install these packages first so the installation doesn't fall down later
pip3 install -U pip wheel
pip3 install cython 

# Install the tensorflow models packages
git clone https://github.com/tensorflow/models.git
cd models/research
protoc object_detection/protos/*.proto --python_out=.
cp object_detection/packages/tf2/setup.py .
pip3 install .
cd ../..

# Download the model
wget download.tensorflow.org/models/object_detection/tf2/20200711/centernet_hg104_1024x1024_coco17_tpu-32.tar.gz
tar xf centernet_hg104_1024x1024_coco17_tpu-32.tar.gz
```

Next save the following python script inside the prior created `root/` directory:

`test.py`

```python
import tensorflow as tf

[...]

PR #9941, opened 2021-04-23
Fix center net parameter parsing

ApproximateIdentity:
This commit https://github.com/tensorflow/models/commit/6e5c5a1bd50e07aab485fddf7bab958b80d95bee seems to have introduced a slight bug in the parameter parsing. This issue is in the following line:

https://github.com/tensorflow/models/blob/6e5c5a1bd50e07aab485fddf7bab958b80d95bee/research/object_detection/builders/model_builder.py#L950

That evaluates to `True` even if the pipeline config does not have the `center_head_params` set to anything. A fix is to change the line to something like (which is what my commit does):

```
if oc_config.center_head_params.num_filters and oc_config.center_head_params.kernel_sizes:
```

That restores the original default behavior and seems to match the author's intent. Issue https://github.com/tensorflow/models/issues/9880 contains complete instructions as to how to reproduce the environment and to test the code.

### DIFF
--- a/research/object_detection/builders/model_builder.py
+++ b/research/object_detection/builders/model_builder.py
@@ -946,7 +946,8 @@ def object_center_proto_to_params(oc_config):
   if oc_config.keypoint_weights_for_center:
     keypoint_weights_for_center = list(oc_config.keypoint_weights_for_center)
 
-  if oc_config.center_head_params:
+  if (oc_config.center_head_params.num_filters and
+      oc_config.center_head_params.kernel_sizes):
     center_head_num_filters = list(oc_config.center_head_params.num_filters)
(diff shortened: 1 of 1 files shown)


ApproximateIdentity:
I have no idea if I am able to sign a CLA. I'll have to look into that later when I have time. This change is quite clearly devoid of any sort of creativity and only a slight technical modification of the work made by the previous committer of this code so I don't think any court would hold this commit as copyrightable at all (at least in the US), but I'm unaware of Google's policies on this matter. In any case, it fixes the issue and seems to quite obviously be what was originally meant when the code was introduced.

vighneshbirodkar:
Fixed via 
https://github.com/tensorflow/models/commit/d7a784e6a85528292abf960be3a7cc643fa2b02c

PR #9987, opened 2021-05-11
mask evaluation from record (issues: 9957, 9980)

anshkumar:
This PR allows maskrcnn evaluation to be done directly from record file. This will be useful in deepmac_maskrcnn.


arashwan (dismissed):

saberkun:
np.generic inclue np.btypes. The code does not work. I am changing to np.array only for the internal process now.

saberkun:
@arashwan there are multiple failures according to the internal tests. Please check.

anshkumar:
> @arashwan there are multiple failures according to the internal tests. Please check.

What kind of failures ?

anshkumar:
@saberkun I've added mask resizing in `maskrcnn_input.py` which should fix errors. I'm using `np.generic` just to check whether `groundtruths['masks'][i][j, k]` is an instance of numpy or not. I've tested using coco validation record. Please let me know if it's working fine.

arashwan:
This PR causes OOM due to saving the masks into the memory. We have supported eval from record by generating the annotation.json if annotation file isn't provided. https://github.com/tensorflow/models/commit/e0ad9ff2daf3c7771161dcd6f8b76ccf5d4181c4

PR #9995, opened 2021-05-13
(doc) Add training log note to README.md

bileschi:
# Description

Adds a documentation note about including references to training logs.


saberkun:
I think we need to be more clear on the message. Are you OK with creating an internal PR as the source of truth in the internal repo? A few vision models might have tensorboard links offered but others are not.

bileschi:
Yes, please reach out to me at my internal address and let's work together to build a message that works for all.

jaeyounkim (approved):

Issue #10062, opened 2021-06-11
Same padding in causal convolution provides wrong results


Atze00:
# Prerequisites

Please answer the following questions for yourself before submitting an issue.

- [x ] I am using the latest TensorFlow Model Garden release and TensorFlow 2.
- [x ] I am reporting the issue to the correct repository. (Model Garden official or research directory)
- [x ] I checked to make sure that this issue has not been filed already.

## 1. The entire URL of the file you are using
https://github.com/tensorflow/models/blob/927e31aa1de2d23fd62b7b2644b67b29d658b944/official/vision/beta/modeling/layers/nn_layers.py

## 2. Describe the bug
1. Same padding in causal convolutions can leave some of the input unattended, look at the section 3 for more details.
2. <s> Same padding can be achieved padding either left or right, bottom or up. it appears that the implementation is different in the case of causal convolution with respect to standard convolution. The standard implementation starts from right, bottom, the causal convolution starts from left, top.</s> In some scenarios does both padding left and right, bottom and top, when just padding bottom right is the right choice.

## 3. Steps to reproduce
Same padiding leaves some of the input unattended:
```python
from official.vision.beta.modeling.layers.nn_layers import Conv3D 
kernel_size=(3,3,3)
stride=(1,3,3)
tf_conv = Conv3D( 1, kernel_size,kernel_initializer='ones', strides=stride, padding='causal',
                                                        use_bias=False)

a = tf.ones([1, 1, 3, 3, 1])
tf_conv(a)
```

Same padiding behaves differently<s>, padding left and top</s>, padding left right, bottom top:
```python
kernel_size=(3,3,3)
stride=(1,2,2)
tf_conv = Conv3D( 1, kernel_size,kernel_initializer='ones', strides=stride, padding='causal',
                                                        use_bias=False)

a = tf.ones([1, 1, 4, 4, 1])
tf_conv(a)
```

## 4. Expected behavior
First output should have been :
```<tf.Tensor: shape=(1, 1, 1, 1, 1), dtype=float32, numpy=array([[[[[9.]]]]], dtype=float32)>```
instead the output is: (this leaves the bottom right part of the input unattended)
```<tf.Tensor: shape=(1, 1, 1, 1, 1), dtype=float32, numpy=array([[[[[4.]]]]], dtype=float32)> ```

Second output should have been :
```<tf.Tensor: shape=(1, 1, 2, 2, 1), dtype=float32, numpy=
 array([[[[[9.],
           [6.]],
 
[...]

PR #10081, opened 2021-06-19
fixed same padding in causal convolutions

Atze00:
# Description

> * Fixed this issue #10062

## Type of change

- [x] Breaking change (Models trained with 3D causal convolution won't work as expected)

## Tests

> * Tested against standard convolution implementation, with different kernels, strides and input shape. This kind of test requires time to run, so it is not included in the pull request.
> * The tests are the same as the one provided originally, the results have been corrected.

## Checklist

### DIFF
--- a/official/vision/beta/modeling/layers/nn_layers.py
+++ b/official/vision/beta/modeling/layers/nn_layers.py
@@ -706,7 +706,7 @@ class CausalConvMixin:
     self._use_buffered_input = variable
 
   def _compute_buffered_causal_padding(self,
-                                       inputs: Optional[tf.Tensor] = None,
+                                       inputs: tf.Tensor,
                                        use_buffered_input: bool = False,
                                        time_axis: int = 1) -> List[List[int]]:
     """Calculates padding for 'causal' option for conv layers.
@@ -720,7 +720,7 @@ class CausalConvMixin:
     Returns:
       A list of paddings for `tf.pad`.
     """
-    shape_in = inputs.shape[1:-1]
+    input_shape = tf.shape(inputs)[1:-1]
     del inputs
 
     if tf.keras.backend.image_data_format() == 'channels_first':
@@ -731,11 +731,10 @@ class CausalConvMixin:
          (self.kernel_size[i] - 1) * (self.dilation_rate[i] - 1))
         for i in range(self.rank)
     ]
-    pad_total = [max(kernel_size_effective[i] - (self.strides[i]), 0)
-                 if (shape_in[i]%self.strides[i]) == 0 else
-                 max(kernel_size_effective[i] - 
-                     (shape_in[i]%self.strides[i]), 0)
-                 for i in range(self.rank)]
+    pad_total = [kernel_size_effective[0] - 1]
+    for i in range(1, self.rank):
+      overlap = (input_shape[i] - 1) % self.strides[i] + 1
+      pad_total.append(tf.maximum(kernel_size_effective[i] - overlap, 0))
     pad_beg = [pad_total[i] // 2 for i in range(self.rank)]
     pad_end = [pad_total[i] - pad_beg[i] for i in range(self.rank)]
     padding = [[pad_beg[i], pad_end[i]] for i in range(self.rank)]
@@ -768,7 +767,8 @@ class CausalConvMixin:
     # across time should be the input shape minus any padding, assuming
     # the stride across time is 1.
     if self._use_buffered_input and spatial_output_shape[0] is not None:
-      padding = self._compute_buffered_causal_padding(use_buffered_input=False)
+      padding = self._compute_buffered_causal_padding(
+          tf.zeros([1] + spatial_output_shape + [1]), use_buffered_input=False)
       spatial_output_shape[0] -= sum(padding[1])
     return spatial_output_shape
 
--- a/official/vision/beta/modeling/layers/nn_layers.py
+++ b/official/vision/beta/modeling/layers/nn_layers.py
@@ -721,7 +721,6 @@ class CausalConvMixin:
       A list of paddings for `tf.pad`.
     """
     input_shape = tf.shape(inputs)[1:-1]
-    del inputs
 
     if tf.keras.backend.image_data_format() == 'channels_first':
       raise ValueError('"channels_first" mode is unsupported.')
--- a/official/vision/beta/modeling/layers/nn_layers.py
+++ b/official/vision/beta/modeling/layers/nn_layers.py
@@ -720,6 +720,7 @@ class CausalConvMixin:
     Returns:
       A list of paddings for `tf.pad`.
     """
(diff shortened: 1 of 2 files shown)


official/vision/beta/modeling/layers/nn_layers.py
```diff
@@ -720,6 +720,7 @@ def _compute_buffered_causal_padding(self,
     Returns:
       A list of paddings for `tf.pad`.
     """
+    shape_in = inputs.shape[1:-1]
```
Hyperparticle, line 723: Let's rename to `input_shape`. Also, `inputs.shape` does not work in graph mode, so we should use `tf.shape(inputs)`:

```suggestion
    input_shape = tf.shape(inputs)[1:-1]
```

official/vision/beta/modeling/layers/nn_layers.py
```diff
@@ -730,7 +731,11 @@ def _compute_buffered_causal_padding(self,
          (self.kernel_size[i] - 1) * (self.dilation_rate[i] - 1))
         for i in range(self.rank)
     ]
-    pad_total = [kernel_size_effective[i] - 1 for i in range(self.rank)]
+    pad_total = [max(kernel_size_effective[i] - (self.strides[i]), 0)
```
Hyperparticle, line 734: This should only be applied after the first temporal dimension, otherwise the causal padding won't match with `tf.nn.conv1d(padding='causal')`.

```suggestion
    pad_total = [kernel_size_effective[0] - 1]
    for i in range(1, self.rank):
      overlap = (input_shape[i] - 1) % self.strides[i] + 1
      pad_total.append(max(kernel_size_effective[i] - overlap, 0))
```

official/vision/beta/modeling/layers/nn_layers_test.py
```diff
@@ -311,14 +311,14 @@ def test_depthwise_conv3d_causal(self):
...
             [6.0, 6.0, 6.0]],
            [[6.0, 6.0, 6.0],
-            [9.0, 9.0, 9.0]]],
-          [[[8.0, 8.0, 8.0],
+            [4.0, 4.0, 4.0]]],
+          [[[18.0, 18.0, 18.0],
             [12., 12., 12.]],
            [[12., 12., 12.],
-            [18., 18., 18.]]]]])
+            [8., 8., 8.]]]]])
```
Hyperparticle, line 321: Can you also add these tests?

```python
  def test_conv3d_causal_padding_2d(self):
    """Test to ensure causal padding works like standard padding."""
    conv3d = nn_layers.Conv3D(
        filters=1,
        kernel_size=(1, 3, 3),
        strides=(1, 2, 2),
        padding='causal',
        use_buffered_input=False,
        kernel_initializer='ones',
        use_bias=False,
    )

    keras_conv3d = tf.keras.layers.Conv3D(
        filters=1,
        kernel_size=(1, 3, 3),
        strides=(1, 2, 2),
        padding='same',
        kernel_initializer='ones',
        use_bias=False,
    )

    inputs = tf.ones([1, 1, 4, 4, 1])

    predicted = conv3d(inputs)
    expected = keras_conv3d(inputs)

    self.assertEqual(predicted.shape, expected.shape)
    self.assertAllClose(predicted, expected)

    self.assertAllClose(predicted,
                        [[[[[9.],
                            [6.]],
                           [[6.],
                            [4.]]]]])

  def test_conv3d_causal_padding_1d(self):
    """Test to ensure causal padding works like standard padding."""
    conv3d = nn_layers.Conv3D(
        filters=1,
        kernel_size=(3, 1, 1),
        strides=(2, 1, 1),
        padding='causal',
        use_buffered_input=False,
        kernel_initializer='ones',
        use_bias=False,
    )

    keras_conv1d = tf.keras.layers.Conv1D(
        filters=1,
        kernel_size=3,
        strides=2,
        padding='causal',
        kernel_initializer='ones',
        use_bias=False,
    )

    inputs = tf.ones([1, 4, 1, 1, 1])

    predicted = conv3d(inputs)
    expected = keras_conv1d(tf.squeeze(inputs, axis=[2, 3]))
    expected = tf.reshape(expected, [1, 2, 1, 1, 1])

    self.assertEqual(predicted.shape, expected.shape)
    self.assertAllClose(predicted, expected)

    self.assertAllClose(predicted,
                        [[[[[1.]]],
                          [[[3.]]]]])
```

Hyperparticle (changes requested):
Thanks for the fix! I left some small changes.

official/vision/beta/modeling/layers/nn_layers_test.py
Hyperparticle: Right below this line, we should also test `_spatial_output_shape`, as it is used in the training loop. Add these lines:

```python
    output_shape = conv3d._spatial_output_shape([4, 4, 4])
    self.assertAllClose(output_shape, [2, 2, 2])
```

official/vision/beta/modeling/layers/nn_layers.py
Hyperparticle: The change to requiring the input shape will affect the `_buffered_spatial_output_shape` method. Can we change it to:

```python
  def _buffered_spatial_output_shape(self, spatial_output_shape: List[int]):
    """Computes the spatial output shape from the input shape."""
    # When buffer padding, use 'valid' padding across time. The output shape
    # across time should be the input shape minus any padding, assuming
    # the stride across time is 1.
    if self._use_buffered_input and spatial_output_shape[0] is not None:
      padding = self._compute_buffered_causal_padding(
          tf.zeros([1] + spatial_output_shape + [1]), use_buffered_input=False)
      spatial_output_shape[0] -= sum(padding[1])
    return spatial_output_shape
```

Hyperparticle (changes requested):
I noticed that these changes will break `_compute_causal_padding`, so I added some additional suggestions.

official/vision/beta/modeling/layers/nn_layers.py
Atze00: couldn't this cause to not attend to some of the input also for the standard implementation?

official/vision/beta/modeling/layers/nn_layers.py
```diff
@@ -720,7 +720,7 @@ def _compute_buffered_causal_padding(self,
     Returns:
       A list of paddings for `tf.pad`.
     """
-    shape_in = inputs.shape[1:-1]
+    input_shape = tf.shape(inputs)[1:-1]
     del inputs
```
Hyperparticle, line 724: This line isn't necessary, you can remove it.

Hyperparticle (changes requested):

Hyperparticle (approved):

Hyperparticle:
We might need to add a flag and disable this behavior by default while we wait for new pretrained models.

yeqingli (approved):

jaeyounkim (approved):

PR #10117, opened 2021-07-07
move sub layer creation into init in RandomFeatureGaussianProcess

Ghostvv:
# Description

when sub layers are created in `self.build` with subsequent `build` call, the names of the weights of sublayers get the name of the main layer which prevents keras saving in `h5` format. Fix it by moving layer creation into init, then they will be built when they are called.

## Type of change

## Tests

no functionality was added or changed, so old test should be enough

**Test Configuration**:

## Checklist

### DIFF
--- a/official/nlp/modeling/layers/gaussian_process.py
+++ b/official/nlp/modeling/layers/gaussian_process.py
@@ -156,18 +156,15 @@ class RandomFeatureGaussianProcess(tf.keras.layers.Layer):
       if self.custom_random_features_activation is None:
         self.custom_random_features_activation = tf.math.cos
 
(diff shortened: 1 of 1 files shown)


saberkun:
Hi, I think there were other issues if we don't build layers. @jereliu 

Could we use the savedmodel format to save the keras model? It is the default and recommended path to save.

Ghostvv:
i cannot do that, since `crf` from tf addons don't support savedmodel format

saberkun:
Keras team is developing a new format and the savedmodel is the current standard. This change will make checkpoint incompatible.

PR #10249, opened 2021-09-08
added dice loss file

gamenerd457:
# In this commit i have added dice loss to losses folder

> * dice loss can be used as a loss function for training segmentation . models
>

## Tests

tested the file using unittest


saberkun:
keras_cv is ramping down in this repo now. We do not accept changes to this folder. Thanks

Issue #10222, opened 2021-08-24
Bert2Bert model call fails when setting padded_decode to True


LoicDagnas:
## 1. The entire URL of the file you are using

https://github.com/tensorflow/models/blob/master/official/nlp/nhnet/models.py

## 2. Describe the bug

While using a Bert2Bert model instance setting the `padded_decode` parameter to True (e.g. for TPU usage), I am forced to specify the batch size in the input when calling the model. 

## 3. Steps to reproduce

You can simply run the following code:
```python
import tensorflow as tf
from official.nlp.nhnet.configs import UNITTEST_CONFIG, BERT2BERTConfig
from official.nlp.nhnet.models import Bert2Bert, get_bert2bert_layers

bert2bert_config_dict = UNITTEST_CONFIG.copy()
bert2bert_config_dict["len_title"] = 32
bert2bert_config_dict["max_position_embeddings"] = 200
bert2bert_config_dict["padded_decode"] = True

bert2bert_config = BERT2BERTConfig.from_args(**bert2bert_config_dict)
bert_layer, decoder_layer = get_bert2bert_layers(params=bert2bert_config)

bert2bert = Bert2Bert(bert2bert_config, bert_layer, decoder_layer)

inputs = {
    "input_ids": tf.keras.layers.Input((200,), dtype=tf.int32, name="input_ids"),
    "input_mask": tf.keras.layers.Input((200,), dtype=tf.int32, name="input_mask"),
    "segment_ids": tf.keras.layers.Input((200,), dtype=tf.int32, name="segment_ids"),
    "target_ids": tf.keras.layers.Input((32,), dtype=tf.int32, name="target_ids")
}

output = bert2bert(inputs, mode='predict')
```

you'll get the following stack:
```
[...]
C:\dev\ml\OnnxConversionLab\venv\lib\site-packages\official\nlp\nhnet\models.py:168 predict_decode  *
        decoded_ids, scores = beam_search.sequence_beam_search(
    C:\dev\ml\OnnxConversionLab\venv\lib\site-packages\official\nlp\modeling\ops\beam_search.py:622 sequence_beam_search  *
        return sbs.search(initial_ids, initial_cache)
    C:\dev\ml\OnnxConversionLab\venv\lib\site-packages\official\nlp\modeling\ops\beam_search.py:158 search  *
        state, state_shapes = self._create_initial_state(initial_ids, initial_cache,
[...]

PR #10372, opened 2021-11-16
Fix the Bert2Bert call when padded_decode is set to True

LoicDagnas:
# Description

As mentioned in the following [issue](https://github.com/tensorflow/models/issues/10222), it was impossible to call a Bert2Bert model with the configuration option `padded_decode` set to True.

It comes from how the beam search modules deals with batch size. The special case implemented when `padded_decode` is True doesn't seem useful and caused the following error:

```
TypeError: Failed to convert object of type <class 'list'> to Tensor. Contents: [None, 1]. Consider casting elements to a supported type.
```

## Type of change

## Tests

I have added a naive unittest which failed before my modification and passes now.

## Checklist

### DIFF
--- a/official/nlp/modeling/ops/beam_search.py
+++ b/official/nlp/modeling/ops/beam_search.py
@@ -152,9 +152,7 @@ class SequenceBeamSearch(tf.Module):
     Returns:
       finished_seq and finished_scores.
     """
-    batch_size = (
-        initial_ids.shape.as_list()[0]
-        if self.padded_decode else tf.shape(initial_ids)[0])
+    batch_size = tf.shape(initial_ids)[0]
     state, state_shapes = self._create_initial_state(initial_ids, initial_cache,
                                                      batch_size)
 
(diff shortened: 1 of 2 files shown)


LoicDagnas:
@saberkun or anyone else if you have a bit of time for this one, it will be very helpful. It is quite critical to make my model serveable in ONNX run time or in TPU vms.

If I have to modify the code or add unittest, just tell me :) 

Thanks in advance !

official/nlp/modeling/ops/beam_search.py
```diff
@@ -455,19 +453,19 @@ def _create_initial_state(self, initial_ids, initial_cache, batch_size):
               tf.TensorShape([]),
           _StateKeys.ALIVE_SEQ:
               tf.TensorShape(
-                  [batch_size, self.beam_size, self.max_decode_length + 1]),
```
saberkun, line 458: The batch size need to be "known" at least for TPUs.

LoicDagnas: yes that's true, but so far it don't work either on classical hardware.

saberkun: I am pretty sure this works on TPU/GPU savedmodel serving. The beam search code is used in multiple projects. If this change is useful to your ONNX path, you can keep in your fork.

Issue #10509, opened 2022-02-19
Default learning rate in TF2 SSD MobileNet V2 config file is way too high. Is it a typo?


EdjeElectronics:
The learning rate set in the [TF2 SSD MobileNet V2 config file](https://github.com/tensorflow/models/blob/master/research/object_detection/configs/tf2/ssd_mobilenet_v2_320x320_coco17_tpu-8.config) is 10x higher than that of the other SSD MobileNet models. This causes loss during training to get extremely high. Is it a typo? 

The default `ssd_mobilenet_v2_320x320_coco17_tpu-8.config` configuration has this for the learning rate:
```
  optimizer {
    momentum_optimizer: {
      learning_rate: {
        cosine_decay_learning_rate {
          learning_rate_base: .8
          total_steps: 50000
          warmup_learning_rate: 0.13333
          warmup_steps: 2000
        }
      }
      momentum_optimizer_value: 0.9
```

Meanwhile, the FPNLite version ( `ssd_mobilenet_v2_fpnlite_320x320_coco17_tpu-8.config` ) has this:
```
  optimizer {
    momentum_optimizer: {
      learning_rate: {
        cosine_decay_learning_rate {
          learning_rate_base: .08
          total_steps: 50000
          warmup_learning_rate: .026666
          warmup_steps: 1000
        }
      }
      momentum_optimizer_value: 0.9
```

When I train with the default values in the `ssd_mobilenet_v2_320x320_coco17_tpu-8.config` file, the huge learning rate throws training way off. When I change the values from `.8` to `.08` and `.13333` to `.013333`, training works much better. I think whoever wrote the config file missed a decimal point.

Loss graph BEFORE changing learning rate values (the loss is way higher)
![image](https://user-images.githubusercontent.com/26154245/154808561-9fed6aac-b023-4a05-831f-bdbb4e01b2c5.png)

Loss graph AFTER changing learning rate values
![image](https://user-images.githubusercontent.com/26154245/154777226-27681388-059e-4d53-8c7f-ca6fcaaa2cb9.png)

Annieliaquat:
> The learning rate set in the [TF2 SSD MobileNet V2 config file](https://github.com/tensorflow/models/blob/master/research/object_detection/configs/tf2/ssd_mobilenet_v2_320x320_coco17_tpu-8.config) is 10x higher than that of the other SSD MobileNet models. This causes loss during training to get extremely high. Is it a typo?
> 
[...]

PR #10531, opened 2022-03-09
Updated learning rate and warmup_learning_rate

jvishnuvardhan:
Looks like this is a typo. All the other models uses small learning rate excepti this specific model. As the user mentioned below, the loss spikes with this learning rate.

Updated learning rate and warmup_learning_rate based on the following GH issue

Fixes https://github.com/tensorflow/models/issues/10509


pkulzc:
Are you training using a TPU v2-8 or v3-8? This config is tuned for that setup and the large learning rate is desired.

PR #11187, opened 2024-04-04
loading a2 rather than a0

MustafaLotfi:
In the documentation that is provided for MoViNet tutorial, it says we will download MoViNet-A0-Stream. But In the code MoViNet-A2 is downloaded and used.

### DIFF
--- a/official/projects/movinet/movinet_tutorial.ipynb
+++ b/official/projects/movinet/movinet_tutorial.ipynb
@@ -439,7 +439,7 @@
       "source": [
         "## Run Streaming Model Inference with TensorFlow Hub and Plot Predictions\n",
         "\n",
-        "We will load MoViNet-A0-Stream from TensorFlow Hub as part of the [MoViNet collection](https://tfhub.dev/google/collections/movinet/).\n",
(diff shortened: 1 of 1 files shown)


laxmareddyp:
Hi @Chen-Zhe @chaoyan1037 ,

PR having typo changes , Please approve in your free time.Thanks

LakshmiKalaKadali:
Hi @Chen-Zhe, @chaoyan1037,

Could you please review the PR.

Thank You

chaoyan1037 (changes requested):
Instead of change comments, maybe update "load_movinet_from_hub('a2', 'stream', hub_version=3)" to be "load_movinet_from_hub('a0', 'stream', hub_version=3)" since the following section Export a Streaming Model to TensorFlow Lite for Mobile also use 'a0' as example.

MustafaLotfi:
@chaoyan1037 
In the previous cells, MoViNet-A2-base is loaded and used, Wouldn't it be a better approach to use a same version (A2) for stream model too?

chaoyan1037:
I see.

chaoyan1037 (approved):

PR #11244, opened 2024-07-31
Fix bibtex citation in README.md

khatchad:
I am getting an error from `biblatex` that the entry has too many commas.


LakshmiKalaKadali:
Hi @yeqingli,

Could you please review this PR. Thank You

PR #13509, opened 2024-12-17
Updated `parse_policy_info` function in augment.py

LakshmiKalaKadali:
This PR is raised to add more flexible control over the randomness in the augmentation process by changing `level += tf.random.normal([], dtype=tf.float32)` to `level += level_std * tf.random.normal([], dtype=tf.float32)` to the function `parse_policy_info` instead of standard deviation always 1.

### DIFF
--- a/official/vision/ops/augment.py
+++ b/official/vision/ops/augment.py
@@ -1869,7 +1869,7 @@ def _parse_policy_info(name: str,
   func = NAME_TO_FUNC[name]
 
   if level_std > 0:
-    level += tf.random.normal([], dtype=tf.float32)
+    level += level_std*tf.random.normal([], dtype=tf.float32)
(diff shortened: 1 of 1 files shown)


official/vision/ops/augment.py
```diff
@@ -1869,7 +1869,7 @@ def _parse_policy_info(name: str,
   func = NAME_TO_FUNC[name]
 
   if level_std > 0:
-    level += tf.random.normal([], dtype=tf.float32)
+    level += level_std*tf.random.normal([], dtype=tf.float32)
```
luotigerlsx, line 1872: This will change the behavior completely. I am worrying about the effects of this. Have you done any tests verifying it won't break existing results using this augmentation ?

luotigerlsx: It's OK to bring more flexibility, but the default behavior should keep backward compatibility.

